PR #47, opened 2022-04-22
feat: fencing limitation removed

### LINKED ISSUE #44: Enhance fencing mechanism
Currently, fencing one single instance in a cluster disables any failover/switchover operation in the Postgres cluster.

We should remove this limitation so that, for example, when one of the two replica instances is fenced, failover might still be triggered on the other standby.

Additionally, in case after an investigation a fenced instance is deleted, the cluster should resume by recreating the instance.

In case of fencing of a primary, once the fence is lifted, the primary should start up again (without issuing a failover like we currently do).

phisco:
With this patch we push down the fencing condition, so that fencing a
single instance will not anymore imply the whole cluster being fenced.
This way if a follower instace is fenced, in case of falure of the
primary instance, failovers will still be performed. In case of
deletion of a fenced instance, the pod will be recreated, relabelled
according to the previous role and if it was a primary instance, the
operator will not perform a failover as before.

Closes #44

### DIFF
--- a/controllers/cluster_controller.go
+++ b/controllers/cluster_controller.go
@@ -423,7 +423,7 @@ func (r *ClusterReconciler) reconcileResources(
 	}
 
 	// If we still need more instances, we need to wait before setting healthy status
-	if cluster.Status.ReadyInstances != cluster.Spec.Instances-instancesStatus.InstancesReportingMightBeUnavailable() {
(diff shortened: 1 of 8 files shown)


phisco:
Closes #44

gbartolini (approved):

mnencia (approved):

PR #74, opened 2022-04-29
fix: ensures that operator pods always have the latest certificates

### LINKED ISSUE #55: PKI setup should only run on the leader instance of the controller
We've noticed that PKI initialization runs unconditionally at the start of an instance, so it will run multiple times if the controller has multiple instances. To be correct, it should run only one time after the leader election.

We could move the code inside a dedicated LeaderElectionRunnable to handle the CA initialization and the refresh of the certificates. We also need to watch the certificate secrets to react to their changes in the non-leader instances.

armru:
This patch mounts the secret containing the certificate in the operator deployment and marks it as optional to allow the first boot. The kubelet will now make sure to update the mounted files each time the secrets get updated.

This patch also reorganizes the cert package by unexporting some methods and by improving its logic encapsulation.
 
Closes #55


pkg/certs/k8s.go
```diff
@@ -72,16 +74,98 @@ type PublicKeyInfrastructure struct {
...
+}
+
+// Setup ensures that we have the required PKI infrastructure to make the operator and the clusters working
+func (pki *PublicKeyInfrastructure) Setup(
+	ctx context.Context,
+	clientSet *kubernetes.Clientset,
+	apiClientSet *apiextensionsclientset.Clientset,
+) error {
+	err := retry.OnError(retry.DefaultRetry, func(err error) bool {
+		return apierrors.IsNotFound(err) || apierrors.IsAlreadyExists(err)
```
armru, line 128: we now also retry on isAlreadyExists error

litaocdl (approved):

PR #76, opened 2022-04-29
feat: provide application database configuration for restore

### LINKED ISSUE #73: Need a way to specify the app password when bootstrap a cluster from recovery or pg_basebackup
When boostrapping a new cluster from a recovery or pg_basebackup, we need a way to specify the app password secret together with Owner and Database in the BootstrapRecovery configuration and BootstrapPgBaseBackup configuration

litaocdl:
This patch provides the function to configure the application database 
credentials when Bootstrap a cluster using recovery or pg_basebackup. 
like initDB options, this patch allows to specific the database name, owner
 and credential secret while restore a cluster using recovery or pg_basebackup.  

Closes #73  

Signed-off-by: Tao Li [tao.li@enterprisedb.com](mailto:tao.li@enterprisedb.com)


api/v1/cluster_types.go
```diff
@@ -665,24 +665,30 @@ type CertificatesStatus struct {
 	Expirations map[string]string `json:"expirations,omitempty"`
 }
 
-// BootstrapInitDB is the configuration of the bootstrap process when
-// initdb is used
-// Refer to the Bootstrap page of the documentation for more information.
-type BootstrapInitDB struct {
+// DatabaseInfo contains the information to connect to a user database
+type DatabaseInfo struct {
```
litaocdl, line 669: Is this not compatible with old `initDB` section ? existed e2e need be refactored if we support new databaseInfo section

api/v1/cluster_types.go
```diff
@@ -1317,79 +1299,48 @@ func (cluster *Cluster) GetEnableSuperuserAccess() bool {
...
+// GetDatabaseInfoSecretName will return the name or the secret if it exists or the default name
+func GetDatabaseInfoSecretName(info *DatabaseInfo, clusterName string) string {
+	if info != nil && info.Secret != nil {
+		return info.Secret.Name
+	}
+	return fmt.Sprintf("%v%v", clusterName, ApplicationUserSecretSuffix)
+}
+
+// GetDatabaseInfoName if a name is set returns that, otherwise the default name
+func GetDatabaseInfoName(info *DatabaseInfo, clusterName string) string {
```
litaocdl, line 1311: the default database name is app

api/v1/cluster_types.go
```diff
@@ -1562,89 +1513,36 @@ func (cluster *Cluster) ShouldResizeInUseVolumes() bool {
...
+	case cluster.Spec.Bootstrap.InitDB != nil:
+		if CheckApplicationDatabase(cluster.Spec.Bootstrap.InitDB.DatabaseInfo) {
+			return true
+		}
+	case cluster.Spec.Bootstrap.Recovery != nil:
+		if CheckApplicationDatabase(cluster.Spec.Bootstrap.Recovery.DatabaseInfo) {
+			return true
+		}
+	case cluster.Spec.Bootstrap.PgBaseBackup != nil:
+		if CheckApplicationDatabase(cluster.Spec.Bootstrap.Recovery.DatabaseInfo) {
```
litaocdl, line 1541: is this a typo,  should be `cluster.Spec.Bootstrap.PgbaseBackup.DatabaseInfo` ?

api/v1/cluster_webhook.go
```diff
@@ -213,33 +215,21 @@ func (r *Cluster) defaultInitDB() {
 
 // defaultRecovery enriches the recovery with defaults if not all the required arguments were passed
 func (r *Cluster) defaultRecovery() {
```
litaocdl, line 217: The benifit of introducing in databaseInfo is when this value is not specified, which means user does not want to setup the application database after recovery.

api/v1/cluster_types_test.go
```diff
@@ -137,21 +137,105 @@ var _ = Describe("Node maintenance window", func() {
...
+				},
+			},
+		}
+
+		Expect(cluster.ShouldPgBaseBackupCreateApplicationDatabase()).To(BeTrue())
+		Expect(cluster.GetApplicationDatabaseName()).To(Equal("appDB"))
+		Expect(cluster.GetPgBaseBackupApplicationSecretName()).To(Equal("appSecret"))
+	})
+
+	It("will get default application secrets name if not specific", func() {
```
hh24k, line 231: typo: specific --> specified

api/v1/cluster_webhook.go
```diff
@@ -49,6 +49,10 @@ const (
 	// DefaultMonitoringSecretName is the name of the target secret with the default monitoring queries,
 	// if configured
 	DefaultMonitoringSecretName = DefaultMonitoringConfigMapName
+	// DefaultApplicationDatabaseName is the name of application database if not specific
```
hh24k, line 52: typo: specific --> specified

api/v1/cluster_webhook.go
```diff
@@ -49,6 +49,10 @@ const (
...
+	DefaultApplicationDatabaseName = "app"
+	// DefaultApplicationUserName is the name of application database owner if not specific
```
hh24k, line 54: typo: specific --> specified

pkg/management/postgres/restore.go
```diff
@@ -566,7 +568,21 @@ func (info InitInfo) ConfigureInstanceAfterRestore(env []string) error {
 		}
 	}
 
-	return nil
+	// if the application database or user are not specific
```
hh24k, line 571: typo: specific --> specified

hh24k (changes requested):

hh24k (approved):

armru (approved):

docs/src/bootstrap.md
```diff
@@ -520,6 +520,40 @@ spec:
...
+#### Update application database password (recovery)
+
+By default, the user credentials for the application database are preserved. You can update 
+the application user password with additional configuration. You can generate your passwords, 
+store them as secrets, and update the database use your own secrets after recovery. Or you can 
+also let the operator generate a secret with randomly secure password for use.
+Please reference the  ["Bootstrap an empty cluster"](#bootstrap-an-empty-cluster-initdb) section
+for more information about secrets.
+
+The following example update the application user password with supplied secret `app-secret`
```
leonardoce, line 532: I think we should explain here how this will affect bootstrap of replica clusters. The same consideration applies to the `pg_basebackup` bootstrap method.

If we change something after having restored a database, the cluster won't be able to reattach to the master.

leonardoce:
I just added my signoff to the partial code review I did

docs/src/bootstrap.md
mnencia: Maybe we should disable the synchronization with PostgreSQL content if the `replica` mode is enabled.

gbartolini: This is a good call. When `replica` is active, this throws an error. Also, while in most cases when replica mode is enabled we don't want to have this code running (as it might have been performed on the primary), in some cases this could be desired (see when you import a database from outside Kubernetes in CloudNativePG). So, question:

- can we skip for now if in replica mode - and properly document it?
- can we open a new ticket to add a test for this case
- can we open a new ticket to manage reconciliation of the app database and user when the instance is promoted?

Thanks,
Gabriele

leonardoce: yes, I'd like to have both: blocking it from the webhook and disable the synchronization on replica mode

mnencia (approved):

Issue #131, opened 2022-05-10
Support for Azure AD Workload Identity


valeriodelsarto:
When using [AzureCredentials](https://github.com/cloudnative-pg/cloudnative-pg/blob/4e5f5bab3f4f237da2ff04045ea661e8115daa96/api/v1/cluster_types.go#L1048-L1055), allow to inherit the Azure credentials needed to interact with the Storage Account from the Service Account, using [Azure AD Workload Identity](https://azure.github.io/azure-workload-identity/docs/introduction.html).

valeriodelsarto:
Created https://github.com/EnterpriseDB/barman/issues/565 to address a warning that is always displayed while using this configuration

PR #132, opened 2022-05-10
feat: adding support for Azure AD Workload Identity

valeriodelsarto:
Closes https://github.com/cloudnative-pg/cloudnative-pg/issues/131

Signed-off-by: Valerio Del Sarto [valerio.delsarto@enterprisedb.com](mailto:valerio.delsarto@enterprisedb.com)

### DIFF
--- a/api/v1/cluster_types.go
+++ b/api/v1/cluster_types.go
@@ -1052,6 +1052,8 @@ type S3Credentials struct {
 //
 // - storageKey
 // - storageSasToken
+//
+// - inheriting the credentials from the pod environment by setting inheritFromAzureAD to true
 type AzureCredentials struct {
 	// The connection string to be used
 	ConnectionString *SecretKeySelector `json:"connectionString,omitempty"`
@@ -1066,6 +1068,10 @@ type AzureCredentials struct {
 	// A shared-access-signature to be used in conjunction with
 	// the storage account name
 	StorageSasToken *SecretKeySelector `json:"storageSasToken,omitempty"`
+
+	// Use the Azure AD based authentication without providing explicitly the keys.
+	// +optional
+	InheritFromAzureAD bool `json:"inheritFromAzureAD"`
 }
 
 // GoogleCredentials is the type for the Google Cloud Storage credentials.
--- a/pkg/management/barman/commandbuilder.go
+++ b/pkg/management/barman/commandbuilder.go
@@ -30,10 +30,17 @@ func AppendCloudProviderOptionsFromConfiguration(
 	options []string,
 	barmanConfiguration *v1.BarmanObjectStoreConfiguration,
 ) ([]string, error) {
+	azureInheritFromAzureAD := false
+	if barmanConfiguration.AzureCredentials != nil {
+		if barmanConfiguration.AzureCredentials.InheritFromAzureAD {
+			azureInheritFromAzureAD = true
+		}
+	}
 	return appendCloudProviderOptions(options,
 		barmanConfiguration.S3Credentials != nil,
 		barmanConfiguration.AzureCredentials != nil,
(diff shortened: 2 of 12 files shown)


api/v1/cluster_types.go
```diff
@@ -1066,6 +1068,10 @@ type AzureCredentials struct {
 	// A shared-access-signature to be used in conjunction with
 	// the storage account name
 	StorageSasToken *SecretKeySelector `json:"storageSasToken,omitempty"`
+
+	// Use the Azure AD based authentication without providing explicitly the keys.
+	// +optional
+	InheritFromAzureAD bool `json:"inheritFromAzureAD"`
```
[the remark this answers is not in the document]
valeriodelsarto: I agree, can be also `EnableAzureADWorkLoadIdentity`. I've initially used used `InheritFromAzureAD` which is similar to `InheritFromIAMRole` that we have for AWS, but I agree that we can probably use a more meaningful name that refers to the related Azure feature.

valeriodelsarto:
Barman was already supporting the feature needed to avoid logging the warning message, as I said here https://github.com/EnterpriseDB/barman/issues/565#issuecomment-1122543081. I've added a new commit to support this parameter here https://github.com/cloudnative-pg/cloudnative-pg/pull/132/commits/c04fe69628791e4257fa21f985318f2fb265d515

sxd:
@valeriodelsarto this patch it's only lacking documentation, it is possible to add a few lines probably here https://github.com/cloudnative-pg/cloudnative-pg/blob/main/docs/src/backup_recovery.md#azure-blob-storage ?

valeriodelsarto:
> @valeriodelsarto this patch it's only lacking documentation, it is possible to add a few lines probably here https://github.com/cloudnative-pg/cloudnative-pg/blob/main/docs/src/backup_recovery.md#azure-blob-storage ?

@sxd I've added that part, thanks for reminding me.

sxd (approved):

api/v1/cluster_types.go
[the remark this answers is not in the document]
dhilipkumars: agreed when the config field maps to the underlying cloud capability then its lot clearer to the user without us needing additional documentation, otherwise we need to explicitly state that `InHeritFromAzureAD` uses `workloadIdentity` under the hood.

pkg/management/barman/commandbuilder.go
```diff
@@ -74,6 +89,12 @@ func appendCloudProviderOptions(
 				options,
 				"--cloud-provider",
 				"azure-blob-storage")
+			if azureInheritFromAzureAD {
```
dhilipkumars, line 92: @valeriodelsarto shouldn't we be checking for barman version `2.18` and above as it was only introduced only @ 2.18? 
https://github.com/EnterpriseDB/barman/releases

dhilipkumars (commented):
tiny comments.

pkg/management/barman/commandbuilder.go
valeriodelsarto: You're right @dhilipkumars, I haven't check for that but here we need to be sure that we are using an image which have Barman version 2.18 or above. We need to add this check in this feature branch.

valeriodelsarto: I've added the check, waiting for a review. Thanks!

dhilipkumars (approved):

pkg/management/barman/commandbuilder.go
valeriodelsarto: I've also added `//nolint: nestif` because the linter was complaining about the complexity of the nested if statements here, feel free to refactor it removing this

armru (approved):

PR #133, opened 2022-05-10
fix: enable CSV logpipe inside WithActiveInstance function

### LINKED ISSUE #130: CSV log pipe is not started by jobs
When PostgreSQL is started by the [WithActiveInstance function](https://github.com/cloudnative-pg/cloudnative-pg/blob/476f5a1185260a35e5080fbab226f371bb398c7f/pkg/management/postgres/instance.go#L445), the log pipe that is handling the logging collector CSV output is not started.

As a result, any problem during Postgres execution is not logged to stdout and collected. This can happen, for example, with a failure during a bootstrap recovery job.

armru:
This patch ensures that we can collect the Postgres log files when we start the instance with the WithActiveInstance function.

It does that by starting a CSV logpipe process with the instance and terminating it before terminating the WithActiveInstance function.

Closes #130

### DIFF
--- a/pkg/management/postgres/instance.go
+++ b/pkg/management/postgres/instance.go
@@ -17,6 +17,7 @@ limitations under the License.
 package postgres
 
 import (
+	"context"
 	"database/sql"
 	"errors"
 	"fmt"
@@ -37,6 +38,7 @@ import (
 	"github.com/cloudnative-pg/cloudnative-pg/pkg/fileutils"
 	"github.com/cloudnative-pg/cloudnative-pg/pkg/management/execlog"
 	"github.com/cloudnative-pg/cloudnative-pg/pkg/management/log"
+	"github.com/cloudnative-pg/cloudnative-pg/pkg/management/postgres/logpipe"
 	"github.com/cloudnative-pg/cloudnative-pg/pkg/management/postgres/pool"
 	"github.com/cloudnative-pg/cloudnative-pg/pkg/postgres"
 	"github.com/cloudnative-pg/cloudnative-pg/pkg/specs"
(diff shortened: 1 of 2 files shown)


mnencia (approved):

pkg/management/postgres/instance.go
```diff
@@ -447,10 +449,22 @@ func (instance Instance) WithActiveInstance(inner func() error) error {
 	if err != nil {
 		return fmt.Errorf("while activating instance: %w", err)
 	}
+
+	ctx, ctxCancel := context.WithCancel(context.Background())
+	csvPipe := logpipe.NewLogPipe()
```
gbartolini, line 454:
```suggestion
	# Start the CSV logpipe to redirect log to stdout
	ctx, ctxCancel := context.WithCancel(context.Background())
	csvPipe := logpipe.NewLogPipe()
```

pkg/management/postgres/instance.go
```diff
@@ -447,10 +449,22 @@ func (instance Instance) WithActiveInstance(inner func() error) error {
...
+
+	go func() {
+		if err := csvPipe.Start(ctx); err != nil {
```
phisco, line 458: this will always fail when WithActiveInstance is called from CompleteCrashRecovery, but that's not a real issue.

phisco (dismissed):

pkg/management/postgres/instance.go
phisco: Actually this could be an issue as log pipes don't know whether they are already running or not, so multiple goroutines could be spawned and attached to the same file. Given the current usages of WithActiveInstance this should never happen afaict, but it could happen in the future if the code is changed.

phisco (approved):

PR #135, opened 2022-05-10
fix: updating resizing status if resize is pending

### LINKED ISSUE #134: Update status if pvc resize is pending
Currently we update the status of the cluster if a resize is happening but there are scenarios where a resize could be pending for a non trivial amount of time. In these instances we should update the Cluster status to reflect that a resize is scheduled to happen but has yet to happen.

jlong49:
Closes #134


mnencia (approved):

sxd (approved):

gbartolini (approved):

gbartolini:
While writing the release notes, I could not figure out why we are doing this. Are we sure we want to ship this in the next patch release? This is our last chance to revert it.

jlong49:
We thought this change was needed but no longer do. This can be reverted.

gbartolini:
Also we don't have any E2E tests written for this. It seems a bit lame to me to proceed. I'd rather postpone this when we have a clearer idea about why we need this and what we need to do to test it works. Do you agree?

leonardoce:
Yes, this can be reverted. Unfortunately the underlying context is not clear here too.

PR #188, opened 2022-05-25
feat: add replication slots for HA

### LINKED ISSUE #8: Replication slots for HA
Cloud Native Postgres currently relies only on the archive log to synchronize those standby servers that have fallen out of sync. However, for some high workload scenarios, this can be inefficient. We need to introduce a smart way for the operator to manage replication slots on the primary and designated primary instances, so that the operator itself regularly moves forward the WAL location based on the most delayed standby. The operator must also remove the replication slot from the former primary and add it to the new one following a failover or switchover.

sxd:
Closes #8


mnencia:
superseded by #671

Issue #220, opened 2022-05-31
E2E to verify that recovery works with custom `serverName` on the source yaml


jsilvela:
Issue https://github.com/cloudnative-pg/cloudnative-pg/issues/209 fixes a bug that could prevent recovery when a source cluster used a custom `serverName` in the backup spec.
We should add an E2E test to cover this case.

hh24k:
Removed the defer func since "dumping failure report" has been done in the JustAfterEach setup node.

jsilvela:
@hh24k the `defer` was not redundant. There is a new cluster created with this E2E on top of the "normal" cluster used in the remainder of the test. Without this defer, a failure of the test will not trigger saving the cluster environment.
OTOH perhaps I should not have used a shadowed name for the cluster

PR #221, opened 2022-05-31
test: add e2e test for recovery with custom serverName

jsilvela:
adds E2E test to verify the patch fixing the recovery from a cluster
with a custom `serverName` in the backup configuration

Closes https://github.com/cloudnative-pg/cloudnative-pg/issues/220

### DIFF
--- a/tests/e2e/backup_restore_test.go
+++ b/tests/e2e/backup_restore_test.go
@@ -286,7 +286,7 @@ var _ = Describe("Backup and restore", Label(tests.LabelBackupRestore), func() {
 		It("backs up and restore a cluster with PITR MinIO", func() {
 			restoredClusterName := "restore-cluster-pitr-minio"
 
-			prepareClusterForPITROnMinio(namespace, clusterName, backupFile, 2, currentTimestamp)
+			prepareClusterForPITROnMinio(namespace, clusterName, backupFile, 3, currentTimestamp)
 
 			err := testUtils.CreateClusterFromBackupUsingPITR(namespace, restoredClusterName, backupFile, *currentTimestamp, env)
 			Expect(err).NotTo(HaveOccurred())
--- a/tests/e2e/backup_restore_test.go
+++ b/tests/e2e/backup_restore_test.go
@@ -210,6 +210,62 @@ var _ = Describe("Backup and restore", Label(tests.LabelBackupRestore), func() {
 			})
 		})
 
+		// We backup and restore a cluster, and verify some expected data to
+		// be there
+		It("backs up and restores a cluster with custom backup serverName", func() {
+			const (
+				targetDBOne                      = "test"
+				targetDBTwo                      = "test1"
+				targetDBSecret                   = "secret_test"
+				testTableName                    = "test_table"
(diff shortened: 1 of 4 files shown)


hh24k (approved):

jsilvela:
@hh24k the `defer` was not redundant. There is a new cluster created with this E2E on top of the "normal" cluster used in the remainder of the test. Without this defer, a failure of the test will not trigger saving the cluster environment.
OTOH perhaps I should not have used a shadowed name for the cluster

jsilvela:
Maybe it's better if we create a top-level variable for the `custom serverName` cluster, from before the JustAfter code is declared, and there we could dump the environment of the second cluster, if necessary.
Let's sync on this.

tests/e2e/backup_restore_test.go
```diff
@@ -210,6 +210,56 @@ var _ = Describe("Backup and restore", Label(tests.LabelBackupRestore), func() {
...
+				targetDBTwo              = "test1"
+				targetDBSecret           = "secret_test"
+				testTableName            = "test_table"
+				clusterRestoreSampleFile = fixturesDir + "/backup/cluster-from-restore-custom.yaml"
+				// clusterWithMinioCustomSampleFile has metadata.name != backup.barmanObjectStore.serverName
+				clusterWithMinioCustomSampleFile = fixturesDir + "/backup/minio/cluster-with-backup-minio-custom-servername.yaml"
+				backupFileCustom                 = fixturesDir + "/backup/minio/backup-minio-custom-servername.yaml"
+			)
+
+			clusterName, err := env.GetResourceNameFromYAML(clusterWithMinioCustomSampleFile)
```
hh24k, line 227: use "clusterName, err :=" here means initializing a new variable `clusterName` instead of assigning another value to the global variable `clusterName` so the failure of the test will not trigger saving the cluster environment

hh24k, line 227: use "clusterName, err :=" here means initializing a new variable `clusterName` instead of assigning another value to the global variable `clusterName` so the failure of the test will not trigger saving the cluster environment

hh24k:
> @hh24k the `defer` was not redundant. There is a new cluster created with this E2E on top of the "normal" cluster used in the remainder of the test. Without this defer, a failure of the test will not trigger saving the cluster environment. OTOH perhaps I should not have used a shadowed name for the cluster

@jsilvela I checked it again ... yes, the issue when a failure of the test will not trigger saving the cluster environment does exist, it's because this https://github.com/cloudnative-pg/cloudnative-pg/pull/221/files#r887387630 ... also I did some tests, for some reason, the defer func doesn't really work as expect there, it only gets executed when the spec succeeds.

so I think, the defer func still needs to be removed but with one more minor change added (like your other comment mentioned).

leonardoce (approved):

hh24k:
@leonardoce Hi Leo ... I just added a new commit based on the discussion with @jsilvela last Thursday, the defer func I previously removed is not really redundant, it just needs a few more adjustments ... May I bother you to have a more check with this?

leonardoce:
No problem @hh24k I'll have a look at it

PR #224, opened 2022-05-31
feat: adds monolithic and microservice database import for bootstrap

### LINKED ISSUE #9: Offline data import and major upgrades for Postgres
We need to provide a declarative way to import existing Postgres databases in CloudNativePG, starting from offline migrations - which also cover major upgrades of Postgres databases. 

**IMPORTANT:** Offline means that with this method applications must stop their write operations to the source until the database is migrated to the new cluster. Offline migrations are in contrast to online migrations with 0 cutover time - which are implemented using native logical replication (see “Logical replica clusters” for online migrations).

Offline migrations extend the “initdb” bootstrap method of CloudNativePG to enable users to create a new PostgreSQL cluster using a snapshot of the data available in another PostgreSQL database - which can be accessed via the network through a superuser connection.

Import is from any supported version of Postgres.

The logical import relies on pg_dump and pg_restore executed from the primary of the new cluster, for all the databases that are part of the operation and, if requested, for the roles.

There are two types to import a database:

- microservice: select the single database to import in the new cluster as the main application database
- monolith: select one or more databases to import (including “*” wildcard), as well as roles (superuser privileges are removed)

The microservice approach is the preferred method to import databases in CloudNativePG, while the monolith approach is more aligned with a shared PostgreSQL instance having multiple databases that DBAs are familiar with.

armru:
Closes #9


armru:
something we should keep in mind:

we need to check if we handle the dump files once we finish the restore (example: delete them, etc...)

sxd (approved):

docs/src/database_import.md
```diff
@@ -0,0 +1,217 @@
...
+  folder is automatically deleted by the operator.
+- At least one database to be specified in the `initdb.import.databases` array
+- Any role that is required by the imported databases must be specified inside
+  `initdb.import.roles`, with the limitations below:
+    - The following roles, if present, are not imported: `postgres`, `streaming_replica`, `cnp_pooler_pgbouncer`
+    - The `SUPERUSER` option is removed from any imported role
+- Wildcard `"*"` can be used as the only element in the `databases` and/or
+  `roles` arrays to import every object of the kind; When matching databases
+  the wildcard will ignore databases not allowing connections, template databases,
+  and `postgres` database
```
gbartolini, line 214:
```suggestion
  the wildcard will ignore the `postgres` database, template databases,
  and those databases not allowing connections
```

gbartolini (approved):

mnencia (approved):

leonardoce (approved):

PR #241, opened 2022-06-08
fix: add support to dump logs of  multiple clusters in e2e.

### LINKED ISSUE #227: E2E most tests don't dump the affected clusters on error
In the ginkgo tests in our E2E suite we are relying on a JustAfterEach node to dump the cluster environment.
Like so:

``` go
	JustAfterEach(func() {
		if CurrentSpecReport().Failed() {
			env.DumpClusterEnv(namespace, clusterName,
				"out/"+CurrentSpecReport().LeafNodeText+".log")
		}
	})
```

But in many cases, for example on replication tests, a particular spec may create new clusters. On error, the DumpClusterEnv will not be applied to those clusters, and we will lose valuable context.

We should ensure that the affected clusters in a failed spec have their environment dumped.
This ticket should be for the development of a strategy.
The application of said strategy throughout the code base could be done on extra tickets, since it will involve a scan of the whole suite.

danishedb:
test: adding support to dump logs of  multiple clusters in case e2e fails.

closes #227 

Signed-off-by: Danish Khan <danish.khan@enterprisedb.com>

### DIFF
--- a/tests/e2e/apparmor_test.go
+++ b/tests/e2e/apparmor_test.go
@@ -50,8 +50,7 @@ var _ = Describe("AppArmor support", Serial, Label(tests.LabelNoOpenshift), func
 
 	JustAfterEach(func() {
 		if CurrentSpecReport().Failed() {
-			env.DumpClusterEnv(namespace, clusterName,
-				"out/"+CurrentSpecReport().LeafNodeText+".log")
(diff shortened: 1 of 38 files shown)


jsilvela (approved):

hh24k (approved):

hh24k:
CI finished with a few unrelated failures ... PR ready to be merged

sxd (approved):

PR #257, opened 2022-06-13
ci: remove automatically e2e test run for push and pull request

### LINKED ISSUE #256: saving e2e test resource by reducing the running times
1. remove automatically e2e run for non-main and non-autobot branch
2. reduce automatically e2e run for pull request, only run e2e automatically when pull request is created

litaocdl:
This fix reduce the chance of e2e test automatically run. for push event, 
e2e test will only automatically run on main and dependabot/** branch. 
For pull request event, e2e test case will only automatically run when 
event type is opened. 

Closes #256 

### DIFF
--- a/.github/workflows/continuous-delivery.yml
+++ b/.github/workflows/continuous-delivery.yml
@@ -3,8 +3,10 @@ name: continuous-delivery
 on:
   push:
     branches:
-      - '**'
+      - main
+      - 'dependabot/**'
   pull_request:
+    types: [ opened ]
   workflow_dispatch:
     inputs:
       depth:


.github/workflows/continuous-delivery.yml
```diff
@@ -3,8 +3,10 @@ name: continuous-delivery
 on:
   push:
     branches:
-      - '**'
+      - main
+      - 'dependabot/**'
   pull_request:
+    types: [ opened ]
```
jsilvela, line 9: i think this should be `action`, not `types` right?
https://docs.github.com/en/developers/webhooks-and-events/webhooks/webhook-events-and-payloads#pull_request

jsilvela: I can answer my own question:
From the GH docs.
>By default, a workflow only runs when a pull_request event's activity type is opened, synchronize, or reopened. To trigger workflows by different activity types, use the types keyword.

PR #267, opened 2022-06-16
test: fix backup restore e2e test failure

### LINKED ISSUE #266: e2e test failed for backup and restore
backup and restore case failure when using minio as storage provider. this ticket is opened to deliver the fix in opensource code.

litaocdl:
fix the end to end test case failure 

Closes #266 

### DIFF
--- a/tests/utils/minio.go
+++ b/tests/utils/minio.go
@@ -35,6 +35,11 @@ import (
 	"github.com/cloudnative-pg/cloudnative-pg/pkg/postgres"
 )
 
+const (
+	minioImage       = "minio/minio:RELEASE.2020-04-23T00-58-49Z"
+	minioClientImage = "minio/mc:RELEASE.2022-06-11T21-10-36Z"
+)
+
 // MinioSetup contains the resources needed for a working minio server deployment:
 // a PersistentVolumeClaim, a Deployment and a Service
 type MinioSetup struct {
@@ -135,7 +140,7 @@ func MinioDefaultDeployment(namespace string, minioPVC corev1.PersistentVolumeCl
 						{
 							Name: "minio",
 							// Latest Apache License release
-							Image:   "minio/minio:RELEASE.2020-04-23T00-58-49Z",
+							Image:   minioImage,
 							Command: nil,
 							Args:    []string{"server", "data"},
 							Ports: []corev1.ContainerPort{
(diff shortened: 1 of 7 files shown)


litaocdl:
for latest run, the failed case are about set up env. will have separate ticket to follow up that. 
https://github.com/cloudnative-pg/cloudnative-pg/runs/6913081701?check_suite_focus=true

jsilvela (approved):

mnencia:
I've checked with @gbartolini and it is fine to use the latest minio.

tests/utils/minio.go
```diff
@@ -321,7 +321,7 @@ func MinioDefaultClient(namespace string) corev1.Pod {
 			Containers: []corev1.Container{
 				{
 					Name:  "mc",
-					Image: "minio/mc:RELEASE.2021-04-22T17-40-00Z",
+					Image: "minio/mc:RELEASE.2022-06-11T21-10-36Z",
```
mnencia, line 324: What about putting this value in a constant on top of the source file? It would be more evident.

litaocdl: yeah, I can do that.

tests/utils/minio.go
```diff
@@ -35,6 +35,11 @@ import (
 	"github.com/cloudnative-pg/cloudnative-pg/pkg/postgres"
 )
 
+const (
+	minioImage       = "minio/minio:RELEASE.2020-04-23T00-58-49Z"
+	minioClientImage = "minio/mc:RELEASE.2022-06-11T21-10-36Z"
```
mnencia, line 40: Is there a reason why we are not also upgrading minioImage to a recent version?

litaocdl: @mnencia during the testing, I only found that sometimes mc client leaves in hang status. let me upgrade both and run e2e again.

litaocdl: upgraded and rerun a test case.

litaocdl:
I found one random case, when I execute 
```
"psql -U postgres app -tAc 'CHECKPOINT; SELECT pg_walfile_name(pg_switch_wal())'"
```

it supports to return the latest war name, but from the log, it returns
CHECKPOINT
<latest war name>


error log:
```
STEP: verify the existence of WAL */pg-backup-minio/*/*/CHECKPOINT
    000000010000000000000003.gz in minio 06/23/22 02:50:48.421
<< End Captured GinkgoWriter Output
Timed out after 60.001s.
  Expected
      <int>: 0
  to be equivalent to
      <int>: 1
In [It] at: /home/runner/work/cloudnative-pg/cloudnative-pg/tests/e2e/asserts_test.go:605
```

I will update this utils too

mnencia (approved):

PR #270, opened 2022-06-16
fix: use backup ID in Point in Time Recovery

### LINKED ISSUE #233: Point in time recovery with an external cluster only works with targetTime
Currently, CloudNativePG does not correctly handle point in time recovery from a recovery object store via the external cluster definition (using `source`) when any of these options is specified:

- targetXID
- targetName
- targetLSN
- targetImmediate

The reason is that the operator at the moment only selects the latest available base backup.

**IMPORTANT:** this does not affect the primary method to perform PITR, which is via time stamp. However, if for example you want to recover to a restore point that you previously created and therefore specify `targetName`, you have no way to instruct from which base backup to start the recovery.

As a result, it makes sense to add another option called `backupId` in the `recoveryTarget` section to be used only when `source` is specified. By default `backupId` is empty, meaning that the operator will automatically detect the backup from which to start the recovery in case `targetTime` is specified, or use the latest available. If non empty, the operator will not perform the automated detection algorithm and always use that backup ID as start of the recovery process - entirely leaving the responsibility to the end user.

jbattiato:
This fix will enable user to specify a backup ID
when performing a PITR using one of the following
recovery targets:

* TargetName (created with pg_create_restore_point)
* TargetXID
* TargetImmediate

Restore will fail if backupID is not defined when restoring
with one of the above targets.

Closes #233

Signed-off-by: Jonathan Battiato <jonathan.battiato@enterprisedb.com>


jbattiato:
It misses E2E test for the PITR using the backupID.

api/v1/cluster_webhook.go
```diff
@@ -867,6 +843,17 @@ func (r *Cluster) validateRecoveryTarget() field.ErrorList {
 		}
 	}
 
+	// validate BackupID is defined when TargetName or TargetXID or TargetImmediate are set
+	if recoveryTarget.TargetName != "" ||
+		recoveryTarget.TargetXID != "" ||
+		recoveryTarget.TargetImmediate != nil {
+		if recoveryTarget.BackupID == "" {
```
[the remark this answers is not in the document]
phisco:
```suggestion
	// validate BackupID is defined when TargetName or TargetXID or TargetImmediate are set
	if recoveryTarget.BackupID == "" &&
		recoveryTarget.TargetName == "" &&
		recoveryTarget.TargetXID == "" &&
		recoveryTarget.TargetImmediate == nil {
```

sxd (approved):

jlong49 (approved):

api/v1/cluster_webhook.go
```diff
@@ -764,7 +764,7 @@ func (r *Cluster) validateBootstrapRecoverySource() field.ErrorList {
 	var result field.ErrorList
 
 	// This validation is only applicable for recovery based bootstrap
-	if r.Spec.Bootstrap == nil || r.Spec.Bootstrap.Recovery == nil || r.Spec.Bootstrap.Recovery.Source == "" {
+	if r.Spec.Bootstrap == nil || r.Spec.Bootstrap.Recovery == nil || r.Spec.Bootstrap.Recovery.Backup != nil {
```
gbartolini, line 767: Why has this been changed? I don't see how it relates to the patch. Can you please comment on this @jbattiato?

config/crd/bases/postgresql.cnpg.io_clusters.yaml
```diff
@@ -1220,6 +1220,11 @@ spec:
                           (i.e., timestamp, transaction Id, LSN, ...). More info:
                           https://www.postgresql.org/docs/current/runtime-config-wal.html#RUNTIME-CONFIG-WAL-RECOVERY-TARGET'
                         properties:
+                          backupID:
+                            description: 'The backup ID to restore when one of the
+                              following targets is specified: targetName, targetXID
+                              and targetImmediate.'
```
gbartolini, line 1226: Mhhh ... Why not target time too? what if I want to restore using a previous backup?

api/v1/cluster_webhook.go
jbattiato: I don't know, @jlong49 did it.

config/crd/bases/postgresql.cnpg.io_clusters.yaml
jbattiato: Because it does not make sense to me. Could you please elaborate more with a use case when this could be necessary?

gbartolini: Because the closest backup might be corrupted and you want to restore from the previous one. Rare case, but still I would not prevent the choice from the start. It's actually simpler to explain. If you provide the Backup ID we always take that.

api/v1/cluster_webhook.go
mnencia: I think this change needs to be reverted.

config/crd/bases/postgresql.cnpg.io_clusters.yaml
gbartolini: This was my initial input:

> As a result, it makes sense to add another option called backupId in the recoveryTarget section to be used only when source is specified. By default backupId is empty, meaning that the operator will automatically detect the backup from which to start the recovery in case targetTime is specified, or use the latest available. If non empty, the operator will not perform the automated detection algorithm and always use that backup ID as start of the recovery process - entirely leaving the responsibility to the end user.

See #233

gbartolini:
I have made some changes to the logic. Basically BackupID is now processed as first option, then we proceed as before.

I made significant changes to the doc too.

gbartolini (approved):

mnencia (approved):

PR #285, opened 2022-06-20
refactor: leverage Sort on array types

jsilvela:


pkg/postgres/status_test.go
```diff
@@ -94,108 +90,105 @@ var _ = Describe("PostgreSQL status", func() {
...
 })
 
 var _ = Describe("PostgreSQL status real", func() {
 	f, err := os.Open("testdata/lsn_overflow.json")
+	Expect(err).ToNot(HaveOccurred())
 	defer func() {
 		_ = f.Close()
 	}()
-	Expect(err).ToNot(HaveOccurred())
+	defer GinkgoRecover()
```
mnencia, line 172: Am I missing something, or should this not be needed as we are not in a goroutine?

jsilvela: Pretty surprised myself. Seems to be necessary because of `json.NewDecoder(f).Decode(&list)`
I wonder if, under the covers, that executes as a goroutine.
Anyway, plan to remove this or explain it better.

pkg/management/catalog/catalog.go
```diff
@@ -94,62 +91,62 @@ func (catalog *Catalog) FindClosestBackupInfo(recoveryTarget *v1.RecoveryTarget)
...
 		if !barmanBackup.isBackupDone() {
 			continue
 		}
 		if (strconv.Itoa(barmanBackup.TimeLine) == targetTLI ||
 			// if targetTLI is not an integer, it will be ignored actually
 			currentTLIRegex.MatchString(targetTLI)) &&
-			barmanBackup.EndTime.Before(targetTime) {
-			return &catalog.List[i], nil
+			barmanBackup.BeginTime.Before(targetTime) {
+			return &catalog[i], nil
```
phisco, line 134: this is reverting https://github.com/cloudnative-pg/cloudnative-pg/commit/216cc54a5a215b70d63725ea1d1ef637410ff223, is this intentional?

jsilvela: eeeek. No! Rebase artifact.. Thanks for catching @phisco

pkg/management/barman/backuplist.go
```diff
@@ -133,5 +139,5 @@ func GetBackupList(
 		return nil, err
 	}
 
-	return backupList, nil
+	return &backupList, nil
```
phisco, line 142: no need for the pointer to be returned here now, the function should return a catalog.Catalog directly

phisco (approved):

Issue #250, opened 2022-06-09
PostInitApplicationSQLRefs


Anhui-tqhuang:
It might be good to involve a way to read sql files from configmap or secrets to init db
```yaml
postInitApplicationSQLRefs:
  secretRefs: []
  configmapRefs:
  - name: configmapName1
    key: file.sql
  - name: configmapName2
    key: file.sql
```
related:
- https://github.com/cloudnative-pg/cloudnative-pg/issues/242

the implementation to these references is:
- secrets -> configmaps
- inside secrets or configmaps, the order is same as the order of each array

MarkCupitt:
This would be very useful for us as well ..

gbartolini:
This issue was not assigned to any given milestone. As a result, we did not look at it for 1.16.

Are we still on time to add it to 1.16 or shall we postpone to 1.17?

Anhui-tqhuang:
@gbartolini thx, added it to 1.17 is fine to me

sxd:
@Anhui-tqhuang we're waiting for #426 to be completed so we can test this PR with the E2E test on your side and our side

Anhui-tqhuang:
@sxd thx for the info!

gbartolini:
@sxd @Anhui-tqhuang any news here? #426 has been merged in the meantime.

Anhui-tqhuang:
@gbartolini hey i am waiting for a new pass of review

PR #287, opened 2022-06-21
feat: add postInitApplicationSQLRefs

Anhui-tqhuang:
Add postInitApplicationSQLRefs, as a result users could refer to post init application SQLs from 
files inside configmap or secrets.

Closes https://github.com/cloudnative-pg/cloudnative-pg/issues/250 

Signed-off-by: TQ [tianqiu.huang@enterprisedb.com](tianqiu.huang@enterprisedb.com)


internal/cmd/manager/instance/initdb/cmd.go
```diff
@@ -40,10 +53,20 @@ func NewCmd() *cobra.Command {
...
 		Use: "init [options]",
 		RunE: func(cmd *cobra.Command, args []string) error {
+			_, err := client.New(ctrl.GetConfigOrDie(), client.Options{
+				Scheme: scheme,
+			})
+			if err != nil {
+				log.Error(err, "Error while getting ")
+				return err
+			}
+
```
phisco, line 69: Given that these secrets and configmaps are needed only in the init job we could think of mounting them directly in the pod, maybe in a single directory with an index (e.g. "00-....sql", 01-... etc) so that we can keep track of the ordering. This way we would avoid requiring permissions to access the configmaps and the secrets, a client and we could pass a flag specifying the folder only, instead of the whole list of secrets with keys. What do you think, @Anhui-tqhuang ?

Anhui-tqhuang: yeah, that's a good point

pkg/specs/volumes.go
```diff
@@ -75,9 +77,68 @@ func createPostgresVolumes(cluster apiv1.Cluster, podName string) []corev1.Volum
...
+			cluster.Spec.Bootstrap.InitDB.PostInitApplicationSQLRefs,
+		)...)
+	}
+
 	return result
 }
 
+func createPostgresVolumesForPostInitApplicationSQLRefs(refs *apiv1.PostInitApplicationSQLRefs) []corev1.Volume {
+	length := len(refs.ConfigMapRefs) + len(refs.SecretRefs)
+	res := make([]corev1.Volume, 0, length)
```
sxd, line 91: magic numbers, the user needs to provide the volume for this, if the user wants and will know the size of the volume, other wise, we can be creating a volume that the user will know nothing about

pkg/specs/volumes.go
```diff
@@ -116,5 +177,44 @@ func createPostgresVolumeMounts(cluster apiv1.Cluster) []corev1.VolumeMount {
 		)
 	}
 
+	if cluster.ShouldInitDBRunPostInitApplicationSQLRefs() {
```
sxd, line 180: again, the user doesn't know about this magic volume

api/v1/cluster_types.go
```diff
@@ -758,6 +758,23 @@ type BootstrapInitDB struct {
 	// after the cluster has been created - to be used with extreme care
 	// (by default empty)
 	PostInitTemplateSQL []string `json:"postInitTemplateSQL,omitempty"`
+
+	// list of references to SQL files which are executed as a superuser in the
+	// application database right after is created - to be used with extreme care
```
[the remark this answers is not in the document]
Anhui-tqhuang: reference to secrets or configmaps

api/v1/cluster_webhook.go
```diff
@@ -414,6 +414,30 @@ func (r *Cluster) validateInitDB() field.ErrorList {
 				"WAL segment size must be a power of 2"))
 	}
 
+	if initDBOptions.PostInitApplicationSQLRefs != nil {
+		for _, item := range initDBOptions.PostInitApplicationSQLRefs.SecretRefs {
+			if item.Name == "" || item.Key == "" {
+				result = append(
```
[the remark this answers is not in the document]
Anhui-tqhuang: if we don't have the secrets or configmaps ready, the init job will not be abled to be completed

api/v1/cluster_webhook.go
```diff
@@ -414,6 +414,30 @@ func (r *Cluster) validateInitDB() field.ErrorList {
...
+					result,
+					field.Invalid(
+						field.NewPath("spec", "bootstrap", "initdb", "postInitApplicationSQLRefs", "secretRefs"),
+						item,
+						"key and name must be specified"))
+			}
+		}
+
+		for _, item := range initDBOptions.PostInitApplicationSQLRefs.ConfigMapRefs {
+			if item.Name == "" || item.Key == "" {
```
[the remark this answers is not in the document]
Anhui-tqhuang: if we don't have the secrets or configmaps ready, the init job will not be abled to be completed

pkg/management/postgres/initdb.go
```diff
@@ -224,6 +231,29 @@ func (info InitInfo) ConfigureNewInstance(instance *Instance) error {
 		return fmt.Errorf("could not execute init Application queries: %w", err)
 	}
 
+	err = filepath.Walk(
```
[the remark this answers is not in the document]
Anhui-tqhuang: gotcha, thanks!

pkg/specs/volumes.go
Anhui-tqhuang: user doesn't provider volume, they provider refs to secrets or configmaps only, and we attach volumes to init job spec by the controller.

i assume user cannot change the spec of init job directly

pkg/specs/jobs.go
```diff
@@ -70,6 +70,11 @@ func CreatePrimaryJobViaInitdb(cluster apiv1.Cluster, nodeSerial int) *batchv1.J
 			"--app-user", cluster.Spec.Bootstrap.InitDB.Owner)
 	}
 
+	if cluster.ShouldInitDBRunPostInitApplicationSQLRefs() {
```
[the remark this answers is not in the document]
Anhui-tqhuang: i follow the strategy used by other stuffs like `postInitApplicationSQL`

pkg/specs/volumes.go
Anhui-tqhuang: also i want to keep the magic number here to make sure the order of implementation to sql files

sxd:
@Anhui-tqhuang if this is a work in progress please set it as draft so we look after it's finished :)

Anhui-tqhuang:
@sxd converted

phisco:
🤩 awesome patch @Anhui-tqhuang! I will review and test it later more in detail, but at a quick glance it looks really well done! Thanks for your contribution!

Anhui-tqhuang:
@sxd @phisco hello, the PR is ready, could i get a review from you plz

api/v1/cluster_webhook.go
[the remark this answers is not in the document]
sxd: Then the init job should fail saying that the secret it's missing

pkg/specs/jobs.go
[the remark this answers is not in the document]
sxd: that doesn't mean things can be improved...

pkg/specs/volumes.go
sxd: nop, it's not acceptable to keep magic numbers, in that case a directory inside $PGDATA it's a better solution

api/v1/cluster_types.go
```diff
@@ -758,6 +758,25 @@ type BootstrapInitDB struct {
...
+// PostInitApplicationSQLRefs points references to configmaps or secrets which
+// contain SQL files, the general implementation order to these references is
+// from all secrets to all configmaps, and inside secrets or configmaps,
+// the implementation order is same as the order of each array.
+type PostInitApplicationSQLRefs struct {
+	// SecretRefs holds a list of references to secrets.
+	SecretRefs []corev1.SecretKeySelector `json:"secretRefs,omitempty"`
+
+	// ConfigMapRefs holds a list of references to configmaps.
+	ConfigMapRefs []corev1.ConfigMapKeySelector `json:"configMapRefs,omitempty"`
```
phisco, line 779: please use our own SecretKeySelector and ConfigMapKeySelector, see https://github.com/Anhui-tqhuang/cloudnative-pg/blob/main/api/v1/base_types.go#L32.

pkg/specs/volumes.go
```diff
@@ -75,9 +77,68 @@ func createPostgresVolumes(cluster apiv1.Cluster, podName string) []corev1.Volum
...
+
+	count := 0
+	for {
+		if length < 10 {
+			count++
+			break
+		}
+		length /= 10
+		count++
+	}
```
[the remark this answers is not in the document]
Anhui-tqhuang: good point, will improve

pkg/specs/volumes.go
```diff
@@ -116,5 +177,48 @@ func createPostgresVolumeMounts(cluster apiv1.Cluster) []corev1.VolumeMount {
...
+
+	for i := range refs.SecretRefs {
+		res = append(res, corev1.VolumeMount{
+			Name:      fmt.Sprintf("%0*d-post-init-application-sql", count, i),
+			MountPath: fmt.Sprintf("/etc/post-init-application-sql-refs/%0*d.sql", count, i),
+			SubPath:   fmt.Sprintf("%0*d.sql", count, i),
+			ReadOnly:  true,
```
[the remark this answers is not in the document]
Anhui-tqhuang: you are right

Anhui-tqhuang:
@phisco hey i addressed all comments, and the PR is ready for another PASS

phisco:
@Anhui-tqhuang can you run a `make checks` and push any changed file to solve https://github.com/cloudnative-pg/cloudnative-pg/runs/7017512429?check_suite_focus=true?

phisco:
As I was saying the patch looks good, the only thing I wanted to improve was to find a way to let users know the initdb job is pending because of a missing configmap/secret we'd have to mount. I have a rough implementation at the moment, but i think we could address this more broadly, maybe we could discuss it with the rest of the team to get also their feedback and decide whether to move on this PR as is and address that later or use this pr for that too.

gbartolini:
What's the status of this patch? I see it is not assigned to any milestone, but I see that 2 maintainers have already reviewed it. Why did not it land in 1.16?

gbartolini:
I chatted with @phisco and he suggested we move this to 1.17 for now.

mnencia:
This PR needs to be rebased on top of the offline database import patch, which also introduces the `PostImportApplicationSQL` field that is a good candidate for a `Refs` pair.

Anhui-tqhuang:
> This PR needs to be rebased on top of the [offline data import patch](https://github.com/cloudnative-pg/cloudnative-pg/commit/1b63d6a3be2efe14e62d32ef7e38df268ef9fc41), which also introduces the PostImportApplicationSQL field that is a good candidate for a Refs pair.

@mnencia good point, i would like to add this in another PR to avoid making this PR so big

Anhui-tqhuang:
@phisco hello, i rebased the branch and resolved the conflicts, could we move forward on this PR?

phisco (approved):

phisco:
@Anhui-tqhuang :wave: we have discussed this patch in the latest community call, could you update the documentation around [here](https://github.com/cloudnative-pg/cloudnative-pg/blob/adf57a2340ec03315f35deba052eab4d89436cab/docs/src/bootstrap.md#bootstrap-an-empty-cluster-initdb) and maybe add an example in the examples folder similar to [this](https://github.com/cloudnative-pg/cloudnative-pg/blob/main/docs/src/samples/cluster-example-initdb.yaml). We agreed it's not t going to get into 1.16, but will be definitely merged for 1.17, and about what @mnencia was mentioning [here](https://github.com/cloudnative-pg/cloudnative-pg/pull/287#issuecomment-1173920449), we agreed it can be handled in a different patch, let us know if you want to address that too!

Anhui-tqhuang:
@phisco hey, i added the doc and example in this [commit](https://github.com/cloudnative-pg/cloudnative-pg/pull/287/commits/8eafa229aff0002b27cdef146c5442f52d273fd3)

> and about what @mnencia was mentioning https://github.com/cloudnative-pg/cloudnative-pg/pull/287#issuecomment-1173920449, we agreed it can be handled in a different patch, let us know if you want to address that too!

yeah, i would like to address that too!

phisco (approved):

docs/src/bootstrap.md
```diff
@@ -275,6 +275,38 @@ spec:
     as queries are run as a superuser and can disrupt the entire cluster.
     An error in any of those queries interrupts the bootstrap phase, leaving the cluster incomplete.
 
+Moreover you can specify a list of secrets and/or configmaps which contains sql files that will be executed after the database is created and configured. These sql files will be executed as the *superuser* (`postgres`), connected to the database specified in the initdb:
```
sxd, line 278: the proper words should be `Secrets` and  `ConfigMaps`

docs/src/bootstrap.md
```diff
@@ -275,6 +275,38 @@ spec:
...
+          key: secret.sql
+  storage:
+    size: 1Gi
+```
+
+!!! Note
+    SQL files in secrets will be executed before SQL files in configmaps, meanwhile SQL files in secrets will be executed in the order of the `secretRefs`, same as the `configMapRefs`.
+
+!!! Warning
+    Please make sure the existence of sql files inside configmaps or secrets specified in the `postInitApplicationSQLRefs`, otherwise the bootstrap will be interrupted.
```
sxd, line 307: the proper words should be `Secrets` and `ConfigMaps`

sxd:
@Anhui-tqhuang can you please add the sample file add into the samples.md file with the proper prerequisites too please

docs/src/bootstrap.md
Anhui-tqhuang: done

docs/src/bootstrap.md
Anhui-tqhuang: done

Anhui-tqhuang:
@sxd thx for the review
> @Anhui-tqhuang can you please add the sample file add into the samples.md file with the proper prerequisites too please

i will add it to samples.md without the section `prerequisites` as my sample file already contains the definitions to Secrets and ConfigMaps

pkg/utils/digits.go
```diff
@@ -0,0 +1,31 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package utils
+
+// IterativeDigitsCount counts number of digits in an integer.
+func IterativeDigitsCount(number int) int {
```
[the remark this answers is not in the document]
Anhui-tqhuang: > return len(fmt.Sprintf("%d", n))

this one works, i will use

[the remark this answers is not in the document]
Anhui-tqhuang: > Also the name isn't too descriptive, probably a better name will be "PaddingZero" or sometlhing like that that actually say what this function it's doing

i assume it's a bit different

this operation could call `PaddingZero`
```go
fmt.Sprintf("%0*d", len(fmt.Sprintf("%d", number)), i)
```

this operation could only call `IterativeDigitsCount`
```go
return len(fmt.Sprintf("%d", number))
```

internal/cmd/manager/instance/initdb/cmd.go
```diff
@@ -109,6 +113,9 @@ func NewCmd() *cobra.Command {
 		"executed inside application database right after the database is created")
 	cmd.Flags().StringVar(&postInitTemplateSQLStr, "post-init-template-sql", "", "The list of SQL queries to be "+
 		"executed inside template1 database to configure the new instance")
+	cmd.Flags().StringVar(&postInitApplicationSQLRefsFolder, "post-init-application-sql-refs-folder",
```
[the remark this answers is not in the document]
Anhui-tqhuang: does it has access to secrets or configmaps, if so, it would easy, we even don't need to keep them in a folder, but if we want to keep them in a folder, anyway we need to know the folder name, except hardcoding the path, i cannot think of other ideas.

moreover the value of this option could be used to control the the logic
```go
func (info InitInfo) executePostInitApplicationSQLRefs(sqlUser *sql.DB) error {
	if info.PostInitApplicationSQLRefsFolder == "" {
		return nil
	}
        # ....
}
```

pkg/utils/digits.go
```diff
@@ -0,0 +1,24 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package utils
+
+import "fmt"
+
+// CountDigits counts number of digits in an integer.
+func CountDigits(number int) int {
```
sxd, line 22: Now doesn't make sense to have this function in one file and one file to test this function, since it's an inliner it can be used directly when you need the value

pkg/specs/volumes.go
```diff
@@ -78,6 +81,67 @@ func createPostgresVolumes(cluster apiv1.Cluster, podName string) []corev1.Volum
 	return result
 }
 
+func createVolumesAndVolumeMountsForPostInitApplicationSQLRefs(
+	refs *apiv1.PostInitApplicationSQLRefs,
+) ([]corev1.Volume, []corev1.VolumeMount) {
+	length := len(refs.ConfigMapRefs) + len(refs.SecretRefs)
+	digitsCount := utils.CountDigits(length)
```
sxd, line 88: here, we should use the inline code to get the count and will be easy to know how you get that number

pkg/utils/digits.go
Anhui-tqhuang: good point

docs/src/bootstrap.md
```diff
@@ -275,6 +275,38 @@ spec:
...
+        - name: my-configmap
+          key: configmap.sql
+        secretRefs:
+        - name: my-secret
+          key: secret.sql
+  storage:
+    size: 1Gi
+```
+
+!!! Note
```
leonardoce, line 303: I think we should say that we execute every SQL statement in a single exec on the server, and they will be executed according to the PostgreSQL semantics (described here https://www.postgresql.org/docs/current/protocol-flow.html#PROTOCOL-FLOW-MULTI-STATEMENT).

This means, i.e., that the user can add comments to the script but cannot use any `psql` internal command.

leonardoce (changes requested):

leonardoce:
@Anhui-tqhuang this is awesome! Thank you!

pkg/specs/volumes.go
```diff
@@ -78,6 +80,67 @@ func createPostgresVolumes(cluster apiv1.Cluster, podName string) []corev1.Volum
...
+							Path: fmt.Sprintf("%0*d.sql", digitsCount, i),
+						},
+					},
+				},
+			},
+		})
+
+		volumeMounts = append(volumeMounts, corev1.VolumeMount{
+			Name:      fmt.Sprintf("%0*d-post-init-application-sql", digitsCount, i),
+			MountPath: fmt.Sprintf("/etc/post-init-application-sql-refs/%0*d.sql", digitsCount, i),
```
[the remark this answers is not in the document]
Anhui-tqhuang: good point

.gitignore
```diff
@@ -29,3 +29,5 @@ tests/e2e/out/
 
 # Spellcheck dictionary
 dictionary.dic
+
+vendor
```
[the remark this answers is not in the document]
Anhui-tqhuang: i have to vendor locally, otherwise GOLang server wouldn't work well

docs/src/bootstrap.md
```diff
@@ -275,6 +275,38 @@ spec:
...
+  name: cluster-example-initdb
+spec:
+  instances: 3
+
+  bootstrap:
+    initdb:
+      database: app
+      owner: app
+      postInitApplicationSQLRefs:
+        configMapRefs:
```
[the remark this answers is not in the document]
Anhui-tqhuang: understand, good point

Anhui-tqhuang:
@leonardoce, thanks for the review i addressed comments, might be ready for another pass

docs/src/bootstrap.md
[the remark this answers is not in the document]
leonardoce: 👍 could you please update the example inside `docs/src/samples/cluster-example-initdb-sql-refs.yaml` too? Thank you! 🙏

pkg/specs/jobs.go
```diff
@@ -29,6 +29,12 @@ import (
 	"github.com/cloudnative-pg/cloudnative-pg/pkg/utils"
 )
 
+const (
+	// PostInitApplicationSQLRefsFolder points to the folder of
+	// postInitApplicationSQL files in the primary job with initdb.
+	PostInitApplicationSQLRefsFolder = "/etc/post-init-application-sql-refs"
```
leonardoce, line 35: I think this can be private, as we use it just inside the `spec` package. I'm sorry to be so picky...

leonardoce (changes requested):

.gitignore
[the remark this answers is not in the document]
leonardoce: You could have added your rule inside `.git/info/exclude` instead of changing the versioned `.gitignore`.

pkg/specs/jobs.go
sxd: should this go in `/etc/` ? probably we can put it in a directory inside `/controller`, I haven't check everything from scratch again so I may be forgetting sometlhing

leonardoce: We already mount secrets in `/etc/superuser-secret` and `/etc/app-secret` so I don't see any downside on using `/etc` name here. Maybe we could devise a better name for that directory... I think `post-init-application-sql` it's better than `post-init-application-sql-refs`. There are no refs here but just contents.

sxd: Yes, but those are configuration or secrets require for configuration, that's why `/etc` it's ok, but as you said this is content, that's why I think `/etc/` it's no the proper place for content file, by convention `/etc` has always contain configuration and files needed by configuration, on the other hand `/controller` contains files with all the needed by the instance manager, even file with content

Anhui-tqhuang: fixed with the following action:
- make the const be private
- replace `post-init-application-sql-refs` with `post-init-application-sql  `

leonardoce (approved):

docs/src/bootstrap.md
```diff
@@ -275,6 +275,39 @@ spec:
...
+          key: secret.sql
+        configMapRefs:
+        - name: my-configmap
+          key: configmap.sql
+  storage:
+    size: 1Gi
+```
+
+!!! Note
+    The SQL scripts referenced in `secretRefs` will be executed before the ones referenced in `configMapRefs`. For both sections the SQL scripts will be executed respecting the order in the list.
```
armru, line 304: Is this still true?
now we do sort.Strings(files)

`	
if err := fileutils.EnsureDirectoryExist(info.PostInitApplicationSQLRefsFolder); err != nil {
		return fmt.Errorf("could not find directory: %s, err: %w", info.PostInitApplicationSQLRefsFolder, err)
	}

	files, err := fileutils.GetDirectoryContent(info.PostInitApplicationSQLRefsFolder)
	if err != nil {
		return fmt.Errorf("could not get directory content from: %s, err: %w",
			info.PostInitApplicationSQLRefsFolder, err)
	}

	sort.Strings(files)

	for _, file := range files {
		sql, ioErr := fileutils.ReadFile(path.Join(info.PostInitApplicationSQLRefsFolder, file))
		if ioErr != nil {
			return fmt.Errorf("could not read file: %s, err; %w", file, err)
		}

		if err = info.executeQueries(sqlUser, []string{string(sql)}); err != nil {
			return fmt.Errorf("could not execute queries: %w", err)
		}
	}
`

armru (approved):

leonardoce:
We run the E2e test manually on EDB's infrastructure as we'd like this work to reach main soon.
Everything was green.

PR #299, opened 2022-06-22
fix: retry issue in `Prepare the enviroment` step

### LINKED ISSUE #298: E2E tests fail in prepareEnv step
As @mnencia analyzed, the issue here is that when the timeout occurs in `retry`, it tries to kill the subprocess, but the subprocess runs as root due to the `sudo` invocation.

We have several options: 

- Search for another retry action that supports having subprocesses running as root
- Make the timeout long enough to never occur

hh24k:
Closes #298 


jlong49 (approved):

jsilvela (approved):

jsilvela:
Tests looking good, no failures out of Postgres 1

PR #305, opened 2022-06-23
chore: adds webhook validation for syncReplicaElectionConstraints

### LINKED ISSUE #304: add a validation webook for syncReplicaElectionConstraint
Right now this definition would pass validation:

```
    syncReplicaElectionConstraint:
      enabled: true
      nodeLabelsAntiAffinity: [  ]
```

This is wrong because it would result in a disabled sync replicas.

We should ensure in our webhook validators that if syncReplicaElectionConstraint is enabled it should also have some labels passed for evaluation

armru:
Closes #304 

### DIFF
--- a/api/v1/cluster_webhook.go
+++ b/api/v1/cluster_webhook.go
@@ -773,6 +773,12 @@ func (r *Cluster) validateConfiguration() field.ErrorList {
 		}
 	}
 
+	if err := validateSyncReplicaElectionConstraint(
+		r.Spec.PostgresConfiguration.SyncReplicaElectionConstraint,
+	); err != nil {
+		result = append(result, err)
+	}
+
 	return result
 }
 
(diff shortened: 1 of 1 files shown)


jlong49 (approved):

api/v1/cluster_webhook.go
```diff
@@ -769,9 +775,32 @@ func (r *Cluster) validateConfigurationChange(old *Cluster) field.ErrorList {
 		}
 	}
 
+	if err := validateSyncReplicaElectionConstraint(
+		r.Spec.PostgresConfiguration.SyncReplicaElectionConstraint,
+	); err != nil {
+		result = append(result, err)
+	}
+
```
phisco, line 783: This should not be needed, we are already checking the same thing in the Validate method, which is called before ValidateChange on updates.

phisco:
```suggestion
```

phisco (approved):

PR #330, opened 2022-06-29
fix: version upgrade logic in patch #198

### LINKED ISSUE #329: Wrong logic to upgrade the conditions to the new format during the upgrade
The part of pull request #198 regarding the upgrade from a version setting reasons containing space is wrong because it is impossible to update the status in a defaulting webhook or with a simple Update call. Moreover, the setDefault function misses the difference because it only compares the content of the spec section.

In most cases, the current code works anyway because the Cluster reconciliation loop must update the status after an operator upgrade. However, in rare cases, it could lead to a cluster that is impossible to upgrade.

mnencia:
The part of pull request #198 regarding the upgrade from a version
setting reasons containing space is wrong because it is impossible to
update the status in a defaulting webhook or with a simple Update
call. Moreover, the setDefault function misses the difference because
it only compares the content of the spec section.

In most cases, the previous code works anyway because the Cluster
reconciliation loop must update the status after an operator
upgrade. However, in rare cases, it could lead to a cluster that is
impossible to upgrade.

This patch fixes the issue by moving the upgrade code from the webhook
to the main controller reconciliation loop.

Closes #329

### DIFF
--- a/controllers/cluster_status.go
+++ b/controllers/cluster_status.go
@@ -31,7 +31,7 @@ import (
 	batchv1 "k8s.io/api/batch/v1"
 	corev1 "k8s.io/api/core/v1"
 	apierrs "k8s.io/apimachinery/pkg/api/errors"
-	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/util/retry"
 	"k8s.io/utils/strings/slices"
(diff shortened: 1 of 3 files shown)


phisco (approved):

armru (approved):

controllers/cluster_status.go
```diff
@@ -365,6 +365,24 @@ func (r *ClusterReconciler) updateResourceStatus(
...
+	conditions := make([]metav1.Condition, 0, len(cluster.Status.Conditions))
+	for _, entry := range cluster.Status.Conditions {
+		if utils.IsConditionReasonValid(entry.Reason) {
+			conditions = append(conditions, entry)
+		}
+	}
+
+	if !reflect.DeepEqual(cluster.Status.Conditions, conditions) {
+		cluster.Status.Conditions = conditions
+		return r.Status().Update(ctx, cluster)
```
Anhui-tqhuang, line 380: if we call this `r.Status.Update(ctx, cluster)` without finishing this reconcile, this will cause
- a new worker will pick up the update event in the work queue and proceed the cluster
- this reconcile will continue to proceed the cluster

and when they call update, one of them will fail due to the conflict
```sh
the object has been modified; please apply your changes to the latest version and try again
```

PR #336, opened 2022-06-30
fix: avoid two patch in one reconcile loop

### LINKED ISSUE #335: avoid two patch in one reconcile loop
This is regrading to the cluster condition migration code, trying to avoid spec and status patch in one reconcile loop

litaocdl:
Avoid race condition if two patch is applied in one reconcile loop

Closes #335 


mnencia (approved):

phisco (approved):

phisco (changes requested):
All the E2E are failing

litaocdl:
@phisco yes,yes, you are right. if the original cluster has no condition. the 
```
if !reflect.DeepEqual(cluster.Status.Conditions, conditions)
```
will still be true and update status will happen but seems not trigger new reconcile (because status is not changed ?).  but we stop the current reconcile loop. I added one condition to check if original cluster condition is zero or not.

litaocdl:
e2e pass.

mnencia (approved):

phisco (approved):

PR #357, opened 2022-07-04
fix: support application database monitoring for replica clusters

### LINKED ISSUE #345: CNPG doesn't report cluster dbName in Replica cluster
Inside the codebase, we obtain the dbname this way:

https://github.com/cloudnative-pg/cloudnative-pg/blob/bf1546052f6b2f4b759c890a206b36052b19823a/api/v1/cluster_types.go#L1482

But inside the function:

https://github.com/cloudnative-pg/cloudnative-pg/blob/bf1546052f6b2f4b759c890a206b36052b19823a/api/v1/cluster_types.go#L1718

we return false if it is a replica cluster.

This affects the monitoring:

https://github.com/cloudnative-pg/cloudnative-pg/blob/bf1546052f6b2f4b759c890a206b36052b19823a/internal/management/controller/instance_controller.go#L691

We should proceed to fix this issue.

We should also investigate if it affects the user GetApplicationDatabaseOwner functions.

leonardoce:
This patch enables CNPG to execute monitoring queries against the
application database for replica clusters, bootstrapped with the
`pgbasebackup` and the `recovery` bootstrap method.

We are fixing the behavior of `GetApplicationDatabaseName` too, directly
getting the database name (and owner) regardless of whenever the
application database need to be created or not.

Closes #345

### DIFF
--- a/api/v1/cluster_types.go
+++ b/api/v1/cluster_types.go
@@ -1463,6 +1463,9 @@ func (cluster *Cluster) GetApplicationSecretName() string {
 // GetApplicationDatabaseName get the name of the application database for a specific bootstrap
 func (cluster *Cluster) GetApplicationDatabaseName() string {
 	bootstrap := cluster.Spec.Bootstrap
+	if bootstrap == nil {
+		return ""
+	}
 
 	if bootstrap.Recovery != nil && bootstrap.Recovery.Database != "" {
 		return bootstrap.Recovery.Database
@@ -1482,6 +1485,9 @@ func (cluster *Cluster) GetApplicationDatabaseName() string {
 // GetApplicationDatabaseOwner get the owner user of the application database for a specific bootstrap
 func (cluster *Cluster) GetApplicationDatabaseOwner() string {
 	bootstrap := cluster.Spec.Bootstrap
+	if bootstrap == nil {
+		return ""
+	}
 
 	if bootstrap.Recovery != nil && bootstrap.Recovery.Owner != "" {
 		return bootstrap.Recovery.Owner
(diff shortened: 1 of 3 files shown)


api/v1/cluster_types.go
```diff
@@ -1462,30 +1462,40 @@ func (cluster *Cluster) GetApplicationSecretName() string {
...
-	switch {
-	case cluster.ShouldRecoveryCreateApplicationDatabase():
-		return cluster.Spec.Bootstrap.Recovery.Database
-	case cluster.ShouldPgBaseBackupCreateApplicationDatabase():
-		return cluster.Spec.Bootstrap.PgBaseBackup.Database
-	case cluster.ShouldInitDBCreateApplicationDatabase():
-		return cluster.Spec.Bootstrap.InitDB.Database
-	default:
-		return ""
+	bootstrap := cluster.Spec.Bootstrap
```
mnencia, line 1465: Bootstrap could be `nil` here. It is unlikely, but it is worth adding an if for protection.

api/v1/cluster_types.go
```diff
@@ -1462,30 +1462,40 @@ func (cluster *Cluster) GetApplicationSecretName() string {
...
-	switch {
-	case cluster.ShouldRecoveryCreateApplicationDatabase():
-		return cluster.Spec.Bootstrap.Recovery.Owner
-	case cluster.ShouldPgBaseBackupCreateApplicationDatabase():
-		return cluster.Spec.Bootstrap.PgBaseBackup.Owner
-	case cluster.ShouldInitDBCreateApplicationDatabase():
-		return cluster.Spec.Bootstrap.InitDB.Owner
-	default:
-		return ""
+	bootstrap := cluster.Spec.Bootstrap
```
mnencia, line 1484: See previous comment

gbartolini (approved):

leonardoce:
I tested it manually and it works fine

PR #362, opened 2022-07-04
refactor: reorganize addBarmanEndpointCAToJob function

### LINKED ISSUE #361: refactor addBarmanEndpointCAToJob function
Right now the function:

https://github.com/cloudnative-pg/cloudnative-pg/blob/1a322fa7a703c232b401b4fa7e50db676c3a1d9f/pkg/specs/jobs.go#L135

Is extremely hard to maintain given to the approach used. Adding new providers would be extremely error prone

We should refactor and make it more maintainable.

armru:
This patch reorganizes in a more structured way the fields related to the barman credentials and how they are used to build the environment variables for the instance.

Closes #361 


pkg/specs/pods.go
```diff
@@ -329,3 +339,61 @@ func PodWithExistingStorage(cluster apiv1.Cluster, nodeSerial int) *corev1.Pod {
...
+
+// BarmanCredentials an object containing the potential credentials for each cloud provider
+type BarmanCredentials struct {
+	Google *apiv1.GoogleCredentials
+	AWS    *apiv1.S3Credentials
+	Azure  *apiv1.AzureCredentials
+}
+
+// AddBarmanEndpointCAToPodSpec adds the required volumes and env variables needed by barman to work correctly
+func AddBarmanEndpointCAToPodSpec(
```
armru, line 351: This is intentionally exported

jlong49:
> @jlong49, here's the link to the E2E on CNPG workflow run: https://github.com/cloudnative-pg/cloudnative-pg/actions/runs/2699817277

One test failed in this run but it was just a timeout and I re-ran the test locally and it passed

jlong49 (approved):

leonardoce:
I added a review commit that is really a refactoring to avoid something like #453 to happen again.

From the serialization PoV nothing changes but people using this codebase as a dependency will need to change a bit how they are creating clusters. We may decide to keep my review commit or remove it, anyway I'd like to have your feedback guys!

mnencia (approved):

mnencia:
The PR is OK for me. I added the "do not merge" label because I want @armru to review it before the final merge. @armru, once satisfied, please remove the label and merge it.

PR #371, opened 2022-07-06
feat: adds logs during switchover, failover and demotion

### LINKED ISSUE #57: Report switchover and failover time in the log
### Problem

At the moment it is not trivial to know how long it took for a switchover to be completed (meaning the primary being promoted, not the former primary to re-attach). Also, by calculating the difference between timestamps of target and current primary, you only know the last switchover.

### Solution

Please report in the log the following:

* **switchover**: the new primary starts up - measure the difference between target and current timestamp and print that in the log
* **switchover**: the former primary is reattached - measure the difference between the start of the standby and the target timestamp, then print it
* **failover**: the new primary starts up - measure the difference between the current timestamp and the most relevant timestamp we have to detect the moment of failure.

### Additional problem

How can we push the above duration metrics somehow to Prometheus so that they can be clearly shown in Grafana?

### Validation

Check the log contains those timestamps via automated tests

armru:
This patch adds a new method for `cluster` that logs timestamp information to stdout, and uses it to log this information during the switchover, failover and demotion of a primary instance

Partially Closes #57

E2E tests will be handled by a different PR

### DIFF
--- a/internal/management/controller/instance_controller.go
+++ b/internal/management/controller/instance_controller.go
@@ -1033,7 +1033,8 @@ func (r *InstanceReconciler) reconcilePrimary(ctx context.Context, cluster *apiv
 
 	// if the currentPrimary doesn't match the PodName we set the correct value.
 	if cluster.Status.CurrentPrimary != r.instance.PodName {
-		// we log the timestamp information before overriding them
+		// we log the timestamp information before overriding them. This is useful to know how much time it took.
+		contextLogger.Info("printing timestamp info before setting myself as the current primary")
 		cluster.LogTimestamps(ctx)
 		contextLogger.Info("Setting myself as the current primary")
 		cluster.Status.CurrentPrimary = r.instance.PodName
--- a/internal/management/controller/instance_controller.go
+++ b/internal/management/controller/instance_controller.go
@@ -1022,6 +1022,7 @@ func (r *InstanceReconciler) reconcilePrimary(ctx context.Context, cluster *apiv
 
 	// If I'm not the primary, let's promote myself
 	if !isPrimary {
+		cluster.LogTimestampsWithMessage(ctx, "setting myself as primary")
 		if err := r.promoteAndWait(ctx, cluster); err != nil {
 			return false, err
 		}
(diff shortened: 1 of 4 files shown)


gbartolini:
A few comments:

- time differences should be expressed in milliseconds
- calculate the difference between current timestamp and timestamp:
    - 0 = initial cluster
    - > 0 = last downtime
    - < 0 = elapsed downtime (failover/switchover in progress)

internal/management/controller/instance_controller.go
```diff
@@ -1020,23 +1023,26 @@ func (r *InstanceReconciler) reconcilePrimary(ctx context.Context, cluster *apiv
...
 
-	// If it is already the current primary, everything is ok
+	// if the currentPrimary doesn't match the PodName we set the correct value.
 	if cluster.Status.CurrentPrimary != r.instance.PodName {
+		// we log the timestamp information before overriding them. This is useful to know how much time it took.
+		contextLogger.Info("printing timestamp info before setting myself as the current primary")
+		cluster.LogTimestamps(ctx)
+		contextLogger.Info("Setting myself as the current primary")
 		cluster.Status.CurrentPrimary = r.instance.PodName
 		cluster.Status.CurrentPrimaryTimestamp = pkgUtils.GetCurrentTimestamp()
```
armru, line 1041: we could potentially log the timestamp after we patch the resource on the cluster, but that would make the code a bit more dirty

internal/management/controller/instance_controller.go
```diff
@@ -1020,23 +1023,26 @@ func (r *InstanceReconciler) reconcilePrimary(ctx context.Context, cluster *apiv
...
 		}
 		restarted = true
 	}
 
-	// If it is already the current primary, everything is ok
+	// if the currentPrimary doesn't match the PodName we set the correct value.
 	if cluster.Status.CurrentPrimary != r.instance.PodName {
+		// we log the timestamp information before overriding them. This is useful to know how much time it took.
+		contextLogger.Info("printing timestamp info before setting myself as the current primary")
+		cluster.LogTimestamps(ctx)
```
armru, line 1038: we could potentially log the timestamp after we patch the resource on the cluster, but that would make the code a bit more dirty

gbartolini (approved):

sxd (approved):

PR #396, opened 2022-07-11
fix: collector panic when query rows returns a non nil error

### LINKED ISSUE #395: Bug in error handling causes panic in metrics collector and crashes the instance controller
The code path here https://github.com/cloudnative-pg/cloudnative-pg/blob/main/pkg/management/postgres/metrics/collector.go#L342
triggers a panic when tx.Rows returns a non nil error, the panic causes the instance controller to crash and the container to restart.

https://github.com/cloudnative-pg/cloudnative-pg/pull/396

francoispqt:
If `rows.Err()` returns a non nil error, the log tries to call `err.Error()` on the wrong `err` that might be nil causing a nil pointer dereference and crashing the whole controller.

Closes #395

### DIFF
--- a/pkg/management/postgres/metrics/collector.go
+++ b/pkg/management/postgres/metrics/collector.go
@@ -339,7 +339,7 @@ func (c QueryCollector) collect(conn *sql.DB, ch chan<- prometheus.Metric) error
 			log.Warning("Error while closing metrics extraction",
 				"err", err.Error())
 		}
-		if rows.Err() != nil {
+		if err := rows.Err(); err != nil {
 			log.Warning("Error while loading metrics",
 				"err", err.Error())
 		}


armru:
Hello @francoispqt , AFAIK rows.Err() should have a built-in nil handler

armru (approved):

armru:
sounds good!

francoispqt:
I'm considering adding unit tests using [go-sqlmock](https://github.com/DATA-DOG/go-sqlmock), maybe in another PR though.

phisco (approved):

francoispqt:
Thanks for your quick reviews! let me know if I need to do anything else :pray:

sxd:
@francoispqt I just changed the title to make it more readable and the comment in the PR to link it with the issue, using the "Closes" word

PR #398, opened 2022-07-11
ci: early warning on bad DefaultImage

### LINKED ISSUE #397: Early detection that the `versions.DefaultImage` is invalid
Recently we had a bunch of E2E failing, because the `versions.DefaultImage` was not
pullable (this happened in the closed source with the image in quay.io).
We should have a very clear warning about having a non-pullable DefaultImage.

jsilvela:
When the default image cannot be pulled, this should result
in a clear alert.

Closes #397 

### DIFF
--- a/pkg/versions/versions.go
+++ b/pkg/versions/versions.go
@@ -23,7 +23,7 @@ const (
 	Version = "1.15.1"
 
 	// DefaultImageName is the default image used by the operator to create pods
-	DefaultImageName = "ghcr.io/cloudnative-pg/postgresql:14.6"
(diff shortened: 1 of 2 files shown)


leonardoce (approved):

leonardoce:
This patch is introducing an external dependency on the `docker` command.
This is not a major issue because we are already using it from the Makefile, but calls for an evaluation of a different solution.

We evaluated using a simple HTTPS request against the registry endpoint but it turned out to be complex since some registries require token authentication. In the future we may want to use a Docker registry client to do this.

PR #429, opened 2022-07-18
feat: add ready-to-test command and independent container registry

### LINKED ISSUE #426: Allow running E2E test from forked repos
Currently, users are getting a "403 Forbidden" error while running E2E test, even if we allow those test to run, this is due external organization user doesn't have write permissions on the packages.

sxd:
We introduce a new behaviour in this commit:

1.- The commits now can run in the forked repo, using the external
    user registry to push the image and the run the E2E.
2.- A new command `ready-to-test` that use the `pull_request_target`
    event to run the test and run the E2E test on the PR, this will
    allow the forked repo to run test from the PR when a user allows it.

Previously we were not able to run the E2E on PR from forked repo
due to the fixed use of the cloudnative-pg repo and since the
`pull_request` event doesn't provide a GitHub token with write permissions
for packages.

We still allow running E2E test from internal branch only when the `test`
command is used, since it will trigger the event dispatch with the client
payload and is used to verify that comes from the internal repo.

Closes #426 

### DIFF
--- a/.github/workflows/continuous-delivery.yml
+++ b/.github/workflows/continuous-delivery.yml
@@ -110,6 +110,8 @@ jobs:
       -
         name: Checkout code
         uses: actions/checkout@v3
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
       -
         name: Install Go
         uses: actions/setup-go@v3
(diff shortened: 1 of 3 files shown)


.github/workflows/continuous-delivery.yml
```diff
@@ -324,7 +324,7 @@ jobs:
           tags=''
           labels=''
           commit_sha=${{ github.event.pull_request.head.sha || github.sha }}
-          commit_date=$(git log -1 --pretty=format:'%ad' --date short "${commit_sha}")
+          commit_date=$(git log -1 --pretty=format:'%ad' --date short "${commit_sha}" || : )
```
sxd, line 327: isn't this a bit out of the scope of this issue?

litaocdl (approved):

jsilvela (approved):

mnencia (approved):

PR #461, opened 2022-07-22
feat: set cluster condition ready when the phase is healthy

### LINKED ISSUE #84: Expose conditions to provide a simple and reliable way to wait for a CNPO Cluster to be ready
Currently (re)deployment of a CNPO cluster seems to be a bit tricky to wait for completion of in a kubectl-friendly way.
For pods, you might

`1. kubectl wait --for=status=Ready -n the-cnp-cluster-namespace pods `

but this is very race-prone. It tends to report premature success if the pods aren't created yet, or if the old pods haven't started termination. Or it can hang indefinitely if some of the old pods still exist but are terminating, because kubectl wait will wait forever for a pod that went away. And it might return premature success if only some of the pods are created yet, but others haven't started creation.
CNPO deletes and creates pods after a new or updated `Cluster` resource is applied to the k8s config. So the new configured `Cluster` resource is visible but the pods haven't been terminated or launched yet. That's fine and normal for an operator, but poses issues if you want to wait for readiness.
AFAICS there is no `Condition` exposed by the `Cluster` CRD, so something like this will not work:

`1. kubectl wait -n app-db --for=condition=Ready clusters/app-db`

and `kubectl` doesn't know how to wait for arbitrary fields like `State=Cluster in healthy state`. It doesn’t support waiting for `Phase` which is considered deprecated, nor for generic wait conditions ([kubectl wait on arbitrary jsonpath · Issue #83094 · kubernetes/kubernetes](https://github.com/kubernetes/kubernetes/issues/83094) ) .
`kubectl`'s `--field-selector` doesn't seem to help either as it's a pretty limited feature that doesn't seem to work well if at all with CRDs.
I'd like to be able to wait for a `Condition` that becomes true when `readyInstances == pvcCount` i.e. `phase: Cluster in healthy state`
In the mean time all I can see to do is to poll until `Cluster in healthy state`, as shown in the `Status` or `phase` fields. But is that i18n-safe? Is that a stable key I can rely on not changing?

wadlejitendra:
This PR has introduces cluster conditions and adding cluster is ready or not in cluster status like:

```
status:

  conditions:
    - lastTransitionTime: "2022-07-22T10:03:02Z"
      reason: ClusterIsReady
      status: "True"
      type: Ready
      message: Cluster is Ready
  
  - lastTransitionTime: "2022-07-22T10:02:02Z"
      reason: ClusterIsNotReady
      status: "False"
      type: Ready
      message: Cluster Is Not Ready
```

Closes #84  

Signed-off-by: Jitendra Wadle <jitendra.wadle@enterprisedb.com>

### DIFF
--- a/controllers/cluster_status.go
+++ b/controllers/cluster_status.go
@@ -360,9 +360,6 @@ func (r *ClusterReconciler) updateResourceStatus(
 		return err
 	}
 
-	// set cluster conditions in cluster status
-	r.SetClusterConditionsInStatus(ctx, cluster)
-
 	if !reflect.DeepEqual(existingClusterStatus, cluster.Status) {
 		return r.Status().Update(ctx, cluster)
 	}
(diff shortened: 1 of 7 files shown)


controllers/cluster_status.go
```diff
@@ -877,3 +881,35 @@ func getPodsTopology(
...
+	ctx context.Context,
+	cluster *apiv1.Cluster,
+) {
+	contextLogger := log.FromContext(ctx)
+	podList, err := r.getManagedPods(ctx, cluster)
+	if err != nil {
+		contextLogger.Error(err, "pods are not found, skipping to set cluster conditions")
+	}
+	var conditions metav1.Condition
+	if cluster.Spec.Instances == utils.CountReadyPods(podList.Items) {
```
phisco, line 899: could be already checking the status as it's already being set. https://github.com/cloudnative-pg/cloudnative-pg/pull/461/files#diff-5cd470cc9a2a125dc3413b4f557332272a815309dd2dbb25d182fcb7b80763ecR276

wadlejitendra: @phisco Could you please provide more details.
I am not getting your point.

jsilvela (approved):

leonardoce:
I think we're missing the documentation here. @wadlejitendra can you please add it?

docs/src/troubleshooting.md
```diff
@@ -398,6 +398,7 @@ event to occur instead of relying on the overall cluster health state. Available
 
 - LastBackupSucceeded
 - ContinuousArchiving
+- Ready
```
leonardoce, line 401: Perhaps we could explain what we mean by "Ready". What do you think about it?

leonardoce (approved):

armru (approved):

PR #462, opened 2022-07-22
feat: add instances timeline id in cluster status

### LINKED ISSUE #449: Add the current timeline in the cluster status
For monitoring purposes, it would be handy if the cluster status reported the current timeline in the Postgres cluster.

armru:
This patch adds a new field in the cluster status called 'InstanceReportedState' it contains several fields reported by the instance manager endpoint, in detail:
- isPrimary: reports if the instance considers itself the primary
- timelineId: reports on which timeline the instance is

It also adds the cluster TimelineId at the root of the status object

Closes #449 


armru:
the failing tests should be tied to the postgres15 issue

jlong49 (approved):

litaocdl (approved):

mnencia (approved):

PR #467, opened 2022-07-22
fix: get latest WAL correctly for Postgres 15 in upgrade E2E test

### LINKED ISSUE #420: E2E - consistent failure of upgrade test for the 15 beta
The only consistent failure in the nightly builds is the upgrade test in the Postgres15beta branches.
See https://github.com/cloudnative-pg/cloudnative-pg/runs/7313093278?check_suite_focus=true

The upgrade tests seem to be the only ones failing in the 15beta branches.

jsilvela:
In the upgrade E2E tests, we get the latest WAL with two combined
commands: `CHECKPOINT; SELECT pg_walfile_name(pg_switch_wal())`
In Postgres15, by default all query command outputs are shown.
See https://github.com/postgres/postgres/commit/3a5130672296ed4e682403a77a9a3ad3d21cef75

When we then create a CLI string to check the last WAL file, the output
contains more than just the WAL number, and this breaks the construction
of the `minio` path for the test.

We fix this  by using the setting `SHOW_ALL_RESULTS=off`

Closes #420 

### DIFF
--- a/tests/e2e/upgrade_test.go
+++ b/tests/e2e/upgrade_test.go
@@ -381,7 +381,7 @@ var _ = Describe("Upgrade", Label(tests.LabelUpgrade, tests.LabelNoOpenshift), O
 				"kubectl exec -n %v %v -- %v",
 				upgradeNamespace,
 				primary,
-				"psql -U postgres appdb -tAc 'CHECKPOINT; SELECT pg_walfile_name(pg_switch_wal())'"))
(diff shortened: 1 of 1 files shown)


mnencia (approved):

sxd (approved):

jsilvela:
The Postgres 15 branches still fail, but now at a different place. The `rolling_update` test.
Will investigate that.

jsilvela:
Failure on rolling update tracked in a different ticket. This one should be merged.

PR #478, opened 2022-07-25
chore: introduces metav1.Microtime for status timestamps

### LINKED ISSUE #446: Convert current/target primary timestamps to Kubernetes Time objects
In b603233a2dbda824afbe9c276d74691d1987f6b0 we introduced two fields: `CurrentPrimaryTimestamp` and `TargetPrimaryTimestamp`. They are part of the status, in string format.

We should convert them to the Kubernetes `Time` type, and make sure that they are then displayed with the nanosecond precision (RFC3339Nano).

Clean the code by removing the unnecessary methods that we might have created in the meantime and that can be replaced by the Kubernetes library.

armru:
This patch changes the type of the timestamps inside our status from the generic `string` to the more upstream type of `metav1.MicroTime`

Closes #446 


gbartolini (approved):

gbartolini:
This looks good to me Armando. Well done, thanks.

armru:
we will not proceed with this issue with the current release, we will make the switch after we have switched to microtime and  a some time have passed to avoid disruptive changes

PR #486, opened 2022-07-26
Single replica breaking OCP upgrade

aaronmcohen:
When running a single replica  POC the PDB will not allow eviction. This means OCP cannot upgrade the cluster unless the it is in a maintenance window. The problem with this logic is OCP administration and Postgres administration may not be in sync.  The proposal is to separate the single replica logic  from the maintenance window logic. 

Signed-off-by: Aaron M. Cohen <amcohen@us.ibm.com>

### DIFF
--- a/controllers/cluster_create.go
+++ b/controllers/cluster_create.go
@@ -103,19 +103,19 @@ func (r *ClusterReconciler) createPostgresClusterObjects(ctx context.Context, cl
 }
 
 func (r *ClusterReconciler) reconcilePodDisruptionBudget(ctx context.Context, cluster *apiv1.Cluster) error {
+
+	if cluster.Spec.Instances == 1 {
+		// If this a single-instance cluster, we need to delete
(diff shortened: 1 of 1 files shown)


sxd:
Hi @aaronmcohen 

A couple of things to notice:
* Please create an issue so we can talk an discuss it there and assign a responsible for this PR
* Fix the linter issue
* Please run the E2E test on this on your repo 

On the other hand, this will basically delete the PDB on every cluster that we create, but we may not want that since this
protects the node from being drained and this needs to be done with during maintenance window and knowing that we
can reuse the PVC, otherwise, this may turn everything into a possible mess, do you have some examples that you can
mention in the issue?

Regards

aaronmcohen:
Created https://github.com/cloudnative-pg/cloudnative-pg/issues/498 to continue the discussion.

sxd:
This is happening in a discussion and this PR it's not valid anymore since looks like we can take another path or there was a solution already for this

PR #502, opened 2022-07-29
tests(e2e) : use client go instead of kubectl for cluster creation

### LINKED ISSUE #452: E2E - use Kubernetes client for all cluster creation
We still use the `kubectl` command within the E2E test.
in `AssertCreateCluster`, as well as several one-off invocations just to appply a cluster YAML.

1. rewrite AssertCreateCluster to use the Kubernetes client
2. remove all remaining `kubectl create/apply -f` invocations that create clusters.

litaocdl:
Rewrites Testing to use K8s Go client for cluster and resource creation,
rather than the `kubectl` command. This way, error handling can be
done in Go, rather than having to parse stdout / stderr.
Additionally, eliminates the `envsubst` logic from the shell scripts,
and handles variable substitutions into YAML template files in the
Go codebase. This will make it much easier to develop tests
iteratively in dev machines.

Closes #452 


jsilvela (approved):

tests/utils/yaml.go
```diff
@@ -0,0 +1,91 @@
...
+		if string(bytes.TrimSpace(section)) == "\n" || len(bytes.TrimSpace(section)) == 0 {
+			continue
+		}
+
+		obj, _, err := decoder.Decode(section, nil, nil)
+		if err != nil {
+			log.Printf("ERROR CreateObjectsFromYAML decoding: %s", err)
+			return nil, err
+		}
+		switch o := obj.(type) {
```
leonardoce, line 46: Perhaps we could avoid this type switch here by using a `MetadataAccessor` (https://pkg.go.dev/k8s.io/apimachinery/pkg/api/meta#MetadataAccessor) or doing something similar to what we did here: https://github.com/cloudnative-pg/cloudnative-pg/blob/4eb7d0a79e291f97e4d1c9bc7504b4848e0ef50f/tests/utils/environment.go#L257 ?

tests/utils/yaml.go
```diff
@@ -0,0 +1,91 @@
...
+// CreateObjectsFromYAML creates kubernetes objects defined in a YAML file, in the namespace
+// NOTE: this function is meant to be used in Tests. It only recognizes a subset of
+//  kubernetes and CNPG object types, throwing errors for any other object types
+func CreateObjectsFromYAML(yamlPath, namespace string) ([]client.Object, error) {
+	data, err := ioutil.ReadFile(filepath.Clean(yamlPath))
+	if err != nil {
+		return nil, err
+	}
+
+	sections := bytes.Split(data, []byte("---"))
```
leonardoce, line 29: If I remember correctly `kubectl apply` is not handling the YAML stream splitting manually but relying on the YAML parser to do that. Perhaps we could explore a bit what's happening here:

https://github.com/kubernetes/kubernetes/blob/aea9f9887d08922378b9e7da0e81a46e733f2995/staging/src/k8s.io/kubectl/pkg/cmd/apply/apply.go#L401

to see if we can get some ideas?

tests/e2e/asserts_test.go
```diff
@@ -2234,14 +2236,23 @@ func collectAndAssertDefaultMetricsPresentOnEachPod(namespace, clusterName, curl
...
 func CreateResourceFromFile(namespace, sampleFilePath string) {
-	Eventually(func() error {
-		_, _, err := testsUtils.RunUnchecked("kubectl apply -n " + namespace + " -f " + sampleFilePath)
-		if err != nil {
-			return err
+	objects, err := testsUtils.CreateObjectsFromYAML(sampleFilePath, namespace)
+	Expect(err).ToNot(HaveOccurred())
+	for _, obj := range objects {
+		if cluster, ok := obj.(*apiv1.Cluster); ok {
+			storageClassName := os.Getenv("E2E_DEFAULT_STORAGE_CLASS")
```
leonardoce, line 2245: Perhaps we could replace the environment variables such as `E2E_DEFAULT_STORAGE_CLASS` before parsing the YAML. Doing that every resource would be handled in the same way.

tests/utils/yaml.go
leonardoce: This patch is already too big! Let's do this later. In the end this works

leonardoce (approved):

litaocdl:
all green, going to merge it!

PR #504, opened 2022-07-29
chore: set timestamps to RFC3339Micro

### LINKED ISSUE #503: Use timestamps compatible with MicroTime
While developing the PR for #478 we noticed that the current timestamp we set inside our status is not compatible with the MicroTime object which leads to upgrade test failures.

We should proceed to migrate to a compatible timestamp and at a later date switch to the MicroTime object

armru:
This patch switches RFC3339Nano to the same RFC3339Time given that is used by metav1.MicroTime

Closes #503 

### DIFF
--- a/pkg/utils/time.go
+++ b/pkg/utils/time.go
@@ -20,23 +20,25 @@ import (
 	"time"
 
 	"github.com/lib/pq"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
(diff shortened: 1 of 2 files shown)


armru:
Please take extra care that is compatible with branch #478.

I've updated the #478 to correctly point to this branch for the test

jsilvela (approved):

sxd (approved):

PR #513, opened 2022-08-02
feat: introduce WAL archive volume

### LINKED ISSUE #10: Creation of a new cluster with a separate volume for WAL files (`pg_wal`)
For vertical scalability, introduce a separate and dedicated volume for storing the WAL archive (`pg_wal`) to parallelize I/O operations between transaction logs and database pages.

This also improves business continuity as we can limit the amount of space used by the archive and dedicate the PGDATA volume for Postgres pages.

In this initial implementation, we only enable this feature at cluster creation time and disable changing this setting on existing clusters.

The second implementation (which will be in a separate ticket) should implement adding the separate volume to an existing cluster that's without it. We should also enable resizing, as well as removing the separate volume from an existing cluster.

armru:
Closes #10


pkg/specs/volumes.go
```diff
@@ -75,6 +74,18 @@ func createPostgresVolumes(cluster apiv1.Cluster, podName string) []corev1.Volum
 		)
 	}
 
+	if cluster.ShouldCreateWalArchiveVolume() {
+		result = append(result,
+			corev1.Volume{
+				Name: "wal-archive",
```
gbartolini, line 80:
```suggestion
				Name: "pg_wal",
```

pkg/specs/volumes.go
```diff
@@ -116,5 +127,14 @@ func createPostgresVolumeMounts(cluster apiv1.Cluster) []corev1.VolumeMount {
 		)
 	}
 
+	if cluster.ShouldCreateWalArchiveVolume() {
+		volumeMounts = append(volumeMounts,
+			corev1.VolumeMount{
+				Name:      "wal-archive",
```
gbartolini, line 133:
```suggestion
				Name:      "pg_wal",
```

pkg/specs/volumes.go
```diff
@@ -116,5 +127,14 @@ func createPostgresVolumeMounts(cluster apiv1.Cluster) []corev1.VolumeMount {
 		)
 	}
 
+	if cluster.ShouldCreateWalArchiveVolume() {
+		volumeMounts = append(volumeMounts,
+			corev1.VolumeMount{
+				Name:      "pg_wal",
+				MountPath: "/var/lib/postgresql/data/pg_wal",
```
gbartolini, line 134:
```suggestion
				MountPath: "/var/lib/postgresql/pg_wal",
```

gbartolini:
I have made some minor changes in the preferred location. I suggest that the "pg_wal" volume is a sibling of the PGDATA one. What is missing now is the invocation of the `initdb` parameter with `--waldir` option (see https://www.postgresql.org/docs/current/app-initdb.html)

api/v1/cluster_webhook.go
```diff
@@ -274,13 +275,15 @@ func (r *Cluster) Validate() (allErrs field.ErrorList) {
 		r.validateCerts,
 		r.validateBootstrapMethod,
 		r.validateStorageConfiguration,
+		r.validateWalStorageConfiguration,
```
leonardoce, line 278: I opened https://github.com/cloudnative-pg/cloudnative-pg/issues/514 to refactor and fix the two different error messages issue over the storage size.

leonardoce:
I think we need to improve the `DetectPVCs` function too, because we can't reattach a PVC of a data directory if we don't have the corresponding WAL volume.

controllers/cluster_create.go
```diff
@@ -1248,3 +1238,37 @@ func (r *ClusterReconciler) removeDanglingPVCs(ctx context.Context, cluster *api
 
 	return nil
 }
+
+func (r *ClusterReconciler) createPVC(
+	ctx context.Context,
+	ownedBy *apiv1.Cluster,
+	storageConfiguration apiv1.StorageConfiguration,
+	name string,
```
leonardoce, line 1246: instead of passing the name of the PVC, perhaps we could pass the suffix to be used for this PVC. This would be empty for the data directory, and `wal` for the WAL archive.

leonardoce:
```
$ k get pvc -w
NAME                       STATUS   VOLUME                                     CAPACITY   ACCESS MODES   STORAGECLASS   AGE
cluster-example-1          Bound    pvc-a419cd79-d644-4234-82fa-fd3f6dacdd14   1Gi        RWO            standard       4s
pg-wal-cluster-example-1   Bound    pvc-4e07f615-f3ac-4d08-bff1-d935e09c9cec   1Gi        RWO            standard       4s
cluster-example-1          Bound    pvc-a419cd79-d644-4234-82fa-fd3f6dacdd14   1Gi        RWO            standard       11s
```

I think our users are used to have the resources created by `Cluster` having the same prefix. This is what happens usually in Kubernetes.

Perhaps we could have `cluster-example-1-wal` as a name, instead of `pg-wal-cluster-example-1`

pkg/specs/jobs.go
```diff
@@ -68,6 +68,10 @@ func CreatePrimaryJobViaInitdb(cluster apiv1.Cluster, nodeSerial int) *batchv1.J
 			"--app-user", cluster.Spec.Bootstrap.InitDB.Owner)
 	}
 
+	if cluster.ShouldCreateWalArchiveVolume() {
+		initCommand = append(initCommand, "--pg-wal", pgWalVolumePath+"/pg_wal")
```
leonardoce, line 72: I think we could use `path.Join` here

leonardoce:
We should think about the meaning of the `DanglingPVC`, `ResizingPVC`, `InitializingPVC`, `HealthyPVC` fields too.
In this moment `ResizingPVC` is considering the WAL PVCs too, while the other ones are not including it.

pkg/specs/pvcs.go
```diff
@@ -131,54 +145,92 @@ func DetectPVCs(
 			continue
 		}
 
+		// TODO optimize
+		// we save all the instance we found so far
+		instanceName := pvc.Labels[utils.InstanceLabelName]
```
leonardoce, line 150: We need to make it backward compatible and, if we don't have that label name, look at a Pod with the name of the PVC. Or, maybe, we could retag the PVC before...

pkg/specs/pvcs.go
```diff
@@ -131,54 +145,92 @@ func DetectPVCs(
...
-		if podFound {
-			// We found a Pod using this PVC so this
-			// PVC is not dangling
-			result.Healthy = append(result.Healthy, pvc.Name)
-			continue
-		}
-
-		jobFound := false
 		for idx := range jobList {
 			if IsJobOperatingOnPVC(jobList[idx], pvc) {
```
leonardoce, line 178: I think we need to improve the definition of `IsJobOperatingOnPVC` to look at the Job definition, like what we do in the previous loop. Maybe we could have a function doing exactly.

api/v1/cluster_types.go
```diff
@@ -353,16 +365,16 @@ type ClusterStatus struct {
 
 	// List of all the PVCs created by this cluster and still available
 	// which are not attached to a Pod
-	DanglingPVC []string `json:"danglingPVC,omitempty"`
+	DanglingPVC []InstancePVC `json:"danglingPVC,omitempty"`
```
leonardoce, line 368: I think we need to backward compatible here

docs/src/storage.md
```diff
@@ -303,6 +326,6 @@ $ kubectl get pods
 NAME                           READY   STATUS      RESTARTS   AGE
 cluster-example-1              1/1     Running     0          5m58s
 cluster-example-2              1/1     Running     0          5m43s
-cluster-example-4-join-v2bfg   0/1     Completed   0          17s
+cluster-example-4-join-v2      0/1     Completed   0          17s
```
armru, line 329: maek checks throws error for the 'bfg' keyword

controllers/cluster_create.go
```diff
@@ -1156,9 +1162,9 @@ func (r *ClusterReconciler) reconcilePVCs(
 	pod := specs.PodWithExistingStorage(*cluster, nodeSerial)
 
 	if configuration.Current.EnableAzurePVCUpdates {
-		for _, resizingPVC := range cluster.Status.ResizingPVC {
+		for _, pvcName := range cluster.Status.ResizingPVC {
 			// This code works on the assumption that the PVC have the same name as the pod using it.
-			if resizingPVC == pvc.Name {
+			if pvcName == pvc.Name {
```
armru, line 1167: need to change

internal/cmd/manager/instance/restore/cmd.go
```diff
@@ -32,6 +32,8 @@ func NewCmd() *cobra.Command {
 	var clusterName string
 	var namespace string
 	var pgData string
+	// TODO: add support
+	var pgWal string
```
armru, line 36: do we need to do anything special for the restore cmd?

controllers/cluster_controller.go
```diff
@@ -665,8 +665,7 @@ func (r *ClusterReconciler) ensureHealthyPVCsAnnotation(
 	for _, pvcName := range cluster.Status.HealthyPVC {
 		pvc := resources.getPVC(pvcName)
 		if pvc == nil {
-			contextLogger.Warning("unable to find pvc to annotate it as ready", "pvc", pvc.Name)
-			continue
+			return fmt.Errorf("pvc not found")
```
leonardoce, line 668: I'm not fully aware of the context behind this, but I agree with you, this seems a programmatic error.
We should raise an error here and perhaps improve the error message including the name of the PVC that we don't found.

controllers/cluster_controller.go
```diff
@@ -1071,22 +1070,22 @@ func (r *ClusterReconciler) markPVCReadyForCompletedJobs(
...
 				break
 			}
 		}
+
 		if pvcName == "" {
 			continue
 		}
 
 		// finding the PVC having the same name as pod
+		// TODO: we need to do a null check
```
leonardoce, line 1084: Yep. Totally agree with it. Let's make this function return an error and improve the caller too. From what I can see it's called just one time.

controllers/cluster_create.go
```diff
@@ -1137,6 +1139,10 @@ func (r *ClusterReconciler) reconcilePVCs(
 	}
 
 	pvc := resources.getPVC(pvcToReattach)
+	if pvc == nil {
+		return ctrl.Result{}, fmt.Errorf(
+			"a pvc to reattach was nominated but it could not be found from the pvc list")
```
leonardoce, line 1144: Let's improve this error message too, including the name of the PVC we didn't found.

controllers/cluster_create.go
```diff
@@ -1213,9 +1219,9 @@ func electPvcToReattach(cluster *apiv1.Cluster) string {
 		return ""
 	}
 
-	for _, name := range pvcs {
-		if name == cluster.Status.TargetPrimary {
-			return name
+	for _, pvc := range pvcs {
+		if specs.DoesBelongToInstance(cluster.Status.TargetPrimary, pvc) {
+			return pvc
```
leonardoce, line 1224: I think the function `electPvcToReattach` is not suitable anymore, as we may want to reattach two PVCs: the pgdata one and the WALs one. Maybe we could change its name as `electInstanceToReattach` and make sure we always return the name of the instance, which corresponds to the name of the pgdata Pod.

controllers/cluster_upgrade.go
```diff
@@ -239,10 +239,10 @@ func IsPodNeedingRollout(status postgres.PostgresqlStatus, cluster *apiv1.Cluste
 	}
 
 	if configuration.Current.EnableAzurePVCUpdates {
-		for _, pvc := range cluster.Status.ResizingPVC {
-			// This code works on the assumption that the PVC have the same name as the pod using it.
-			if status.Pod.Name == pvc {
-				return true, false, fmt.Sprintf("rebooting pod to complete resizing %s", pvc)
+		for _, pvcName := range cluster.Status.ResizingPVC {
+			// This code works on the assumption that the PVC beings with the name of the pod using it.
```
leonardoce, line 243:
```suggestion
			// This code works on the assumption that the PVC begins with the name of the pod using it.
```

controllers/cluster_upgrade.go
```diff
@@ -239,10 +239,10 @@ func IsPodNeedingRollout(status postgres.PostgresqlStatus, cluster *apiv1.Cluste
...
+			if specs.DoesBelongToInstance(status.Pod.Name, pvcName) {
```
leonardoce, line 244: We should verify here that we are rebooting the instance just one time after we have resized all the PVC that need resizing for a certain instance.

internal/cmd/manager/instance/restore/cmd.go
leonardoce: IIRC barman-cloud-restore will not restore the WAL files but create an empty folder. Anyway we should certainly override this folder with a symlink pointing to the correct location.

pkg/specs/pvcs.go
```diff
@@ -176,12 +173,56 @@ func DetectPVCs(
...
+		expectedPVCs := getExpectedInstancePVCNames(cluster, instance)
+		var indexOfFoundPVCs []int
+
+		for idx, pvcName := range result.Healthy {
+			if slices.Contains(expectedPVCs, pvcName) {
+				indexOfFoundPVCs = append(indexOfFoundPVCs, idx)
+			}
+		}
+
+		if len(expectedPVCs) == len(indexOfFoundPVCs) {
```
leonardoce, line 190: We could transform this in a `switch` statement, I bet this will be more readable.

pkg/management/postgres/restore.go
```diff
@@ -398,6 +398,12 @@ func (info InitInfo) writeRestoreWalConfig(backup *apiv1.Backup, cluster *apiv1.
 			0o600)
 	}
 
+	if info.PgWal != "" {
+		if err := os.Symlink(info.PgWal, path.Join(info.PgData, "pg_wal")); err != nil {
```
leonardoce, line 402: I think this will fail because Barman is going to create the `pg_wal` directory and `os.Symlink` is not going to overwrite it. Perhaps it's better to have a generic function that migrate the existing content of `pg_wal` inside  the volume, removes recursively the `pg_wal` directory, and create the symlink.

We could use this function when migrating an existing cluster which doesn't have a PVC for WALS to one having it.

controllers/cluster_upgrade.go
leonardoce: Talking with @armru we decided that in this moment we don't have enough information to correctly implement this. The worst case here is when the user changes the storage configuration of the pgdata and the storage configuration of the WALs at the same time. In that case the operator may restart that Pod twice.

Let's address this in a separate ticket as this behavior may be desired or not depending on the situation.

@armru can you create a ticket for that?

docs/src/storage.md
```diff
@@ -295,6 +359,17 @@ As an example, to recreate the storage for `cluster-example-3` you can:
 $ kubectl delete pvc/cluster-example-3 pod/cluster-example-3
 ```
 
+!!! Important
+    In case you have instanciated a dedicated WAL volume, both PCVs will have to be deleted during this process.
```
gbartolini, line 363:
```suggestion
    In case you have instantiated a dedicated WAL volume, both PCVs will have to be deleted during this process.
```

docs/src/failure_modes.md
```diff
@@ -29,11 +31,19 @@ following command:
 kubectl delete -n [namespace] pvc/[cluster-name]-[serial] pod/[cluster-name]-[serial]
 ```
 
+!!! Note
+    In case you have instanciated a dedicated WAL volume it will also have to be deleted during this process.
```
gbartolini, line 35:
```suggestion
    In case you have instantiated a dedicated WAL volume it will also have to be deleted during this process.
```

gbartolini (approved):

docs/src/failure_modes.md
NiccoloFei: Fixed

docs/src/storage.md
NiccoloFei: Fixed

controllers/cluster_controller.go
```diff
@@ -1060,36 +1064,35 @@ func (r *ClusterReconciler) mapNodeToClusters(ctx context.Context) handler.MapFu
...
+		return nil
 	}
 
 	for _, job := range completeJobs {
-		var pvcName string
 		for _, pvc := range resources.pvcs.Items {
-			if specs.IsJobOperatingOnPVC(job, pvc) {
-				pvcName = pvc.Name
-				break
+			pvc := pvc
```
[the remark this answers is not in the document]
jsilvela: I suggest to rename this to `pvcCopy` with a comment to explain the need

docs/src/failure_modes.md
```diff
@@ -29,11 +31,19 @@ following command:
 kubectl delete -n [namespace] pvc/[cluster-name]-[serial] pod/[cluster-name]-[serial]
 ```
 
+!!! Note
+    In case you have instantiated a dedicated WAL volume it will also have to be deleted during this process.
```
jsilvela, line 35:
```suggestion
    If you have specified a dedicated WAL volume, it will also have to be deleted during this process.
```

docs/src/operator_capability_levels.md
```diff
@@ -91,6 +91,9 @@ workload requirements, based on what the underlying Kubernetes environment
 can offer. This implies choosing a particular storage class in
 a public cloud environment or fine-tuning the generated PVC through a
 PVC template in the CR's `storage` parameter.
+For better performance and finer control, you can also decide to host your
```
jsilvela, line 94:
```suggestion
For better performance and finer control, you can also choose to host your
```

docs/src/storage.md
```diff
@@ -128,6 +132,66 @@ spec:
       volumeMode: Filesystem
 ```
 
+## Volume for WAL
+
+By default, PostgreSQL stores all its data in the so called `PGDATA`.
```
jsilvela, line 137:
```suggestion
By default, PostgreSQL stores all its data in the so-called `PGDATA` (a directory).
```

docs/src/storage.md
```diff
@@ -128,6 +132,66 @@ spec:
...
+One of the core directories inside `PGDATA` is called `pg_wal` (historically
```
jsilvela, line 138:
```suggestion
One of the core directories inside `PGDATA` is `pg_wal` (historically
```

docs/src/storage.md
```diff
@@ -128,6 +132,66 @@ spec:
...
+known as `pg_xlog` in PostgreSQL), which contains the log of transactional
+changes occurred in the database, in the form of segment files.
+
+!!! Info
+    Normally, each segment is 16 MB in size, but the size is can be changed
```
jsilvela, line 143:
```suggestion
    Normally, each segment is 16 MB in size, but the size can be configured
```

docs/src/storage.md
```diff
@@ -128,6 +132,66 @@ spec:
...
+    at cluster initialization time through the `walSegmentSize` option, as
```
jsilvela, line 144:
```suggestion
    through the `walSegmentSize` option, applied at cluster initialization time, as
```

docs/src/storage.md
```diff
@@ -128,6 +132,66 @@ spec:
...
+    described in ["Bootstrap an empty cluster"](bootstrap.md#bootstrap-an-empty-cluster-initdb).
+
+While in most of the cases having `pg_wal` on the same volume where `PGDATA`
```
jsilvela, line 147:
```suggestion
While in most cases, having `pg_wal` on the same volume where `PGDATA`
```

docs/src/storage.md
```diff
@@ -295,6 +359,17 @@ As an example, to recreate the storage for `cluster-example-3` you can:
 $ kubectl delete pvc/cluster-example-3 pod/cluster-example-3
 ```
 
+!!! Important
+    In case you have instantiated a dedicated WAL volume, both PCVs will have to be deleted during this process.
```
jsilvela, line 363:
```suggestion
    In case you have created a dedicated WAL volume, both PVCs will have to be deleted during this process.
```

pkg/specs/pvcs.go
```diff
@@ -176,12 +177,55 @@ func DetectPVCs(
...
+
+		for idx, pvcName := range result.Healthy {
+			if slices.Contains(expectedPVCs, pvcName) {
+				indexOfFoundPVCs = append(indexOfFoundPVCs, idx)
+			}
+		}
+
+		switch {
+		case len(expectedPVCs) == len(indexOfFoundPVCs):
+			continue
```
jsilvela, line 196: hmm, we continue in all the cases, no? Then this can be expressed more explicitly

jsilvela (commented):
I am making a few changes (comment and doc corrections), but others I am putting as comments, as I want to clarify before making any corrections.

pkg/specs/pvcs.go
armru: Not sure that I got the question correctly, but it can happen that one PVC is gone and then we need to null the other too, see the  case `len(expectedPVCs) > len(indexOfFoundPVCs):`

Once we have more than one volume (not just the PGDATA) all of them are needed to make the pod start correctly

jsilvela (approved):

leonardoce (approved):

PR #531, opened 2022-08-05
fix: timeouts for restarts and switchovers

### LINKED ISSUE #530: switchoverDelay ignored in switchovers
Commit ce948a03052f5149695a086e424907bc53019ac5 introduced a bug in how switchoverDelay is used. This bug was introduced in version **1.14**.

Specifically, from that moment, switchover was honoring `stopDelay`, while restarts were using `switchoverDelay`. See the excerpt from internal/cmd/manager/instance/run/lifecycle/lifecycle.go:

```go
        case postgres.RestartSmartFast:
                return true, tryShuttingDownSmartFast(i.instance.MaxSwitchoverDelay, i.instance)
        case postgres.ShutDownFastImmediate:
                if err := tryShuttingDownFastImmediate(i.instance.MaxStopDelay, i.instance); err != nil {
                        log.Error(err, "error shutting down instance, proceeding")
                }
```

This meant that if you issued a promote, the primary would have been fast shutdown before `stopDelay` seconds (default 30).

If you instead required a restart (for example a configuration change), the restart would have been performed in smart mode first (waiting for active clients to disconnect) for `switchoverDelay`/2, then fast. The issue is that, by default, `switchoverDelay` is set to a very high value (close to one year).

leonardoce:
This patch fix the timeouts that we were using during a restart and
during switchovers to use the right fields in the Cluster specification.

Closes #530

### DIFF
--- a/internal/cmd/manager/instance/run/lifecycle/lifecycle.go
+++ b/internal/cmd/manager/instance/run/lifecycle/lifecycle.go
@@ -182,9 +182,9 @@ func (i *PostgresLifecycle) handleInstanceCommandRequests(
 		}
 		return false, err
 	case postgres.RestartSmartFast:
-		return true, tryShuttingDownSmartFast(i.instance.MaxSwitchoverDelay, i.instance)
+		return true, tryShuttingDownSmartFast(i.instance.MaxStopDelay, i.instance)
 	case postgres.ShutDownFastImmediate:
-		if err := tryShuttingDownFastImmediate(i.instance.MaxStopDelay, i.instance); err != nil {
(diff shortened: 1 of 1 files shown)


gbartolini (approved):

armru (approved):

gbartolini:
This patch needs to be backported

gbartolini:
What about adding some unit tests that make sure that inside "tryShuttingDownSmartFast" the default delay that is passed matches stopDelay's value and that in "tryShuttingDownFastImmediate" it is switchoverDelay's value? @leonardoce @armru

@mnencia would you please look at this patch as well and approve it?

sxd (approved):

PR #551, opened 2022-08-08
chore: properly recognize `instances` pods

### LINKED ISSUE #550: rename getManagedPods in getManagedInstances
While developing #10  and talking with @leonardoce, we came to the conclusion that it would be better to rename this function in getManagedInstances.

Right now this function returns only the instances pod and all the code base expects this to be true.
In future, this could change and lead to subtle bugs or failures so it is better to: a) make the query more explicit to be sure that we get only the instances, b) make the behavior more explicit in the codebase.

https://github.com/cloudnative-pg/cloudnative-pg/blob/f989a97359569595f5d69a0507ef27913d75d08e/controllers/cluster_status.go#L125

armru:
Closes #550

Descriptions of what is done are contained in each commit. Do not squash.

### DIFF
--- a/controllers/cluster_status.go
+++ b/controllers/cluster_status.go
@@ -63,10 +63,10 @@ var StatusRequestRetry = wait.Backoff{
 // and need to be managed by the controller
 type managedResources struct {
 	// nodes this is a map composed of [nodeName]corev1.Node
-	nodes map[string]corev1.Node
-	pods  corev1.PodList
-	pvcs  corev1.PersistentVolumeClaimList
-	jobs  batchv1.JobList
+	nodes     map[string]corev1.Node
+	instances corev1.PodList
+	pvcs      corev1.PersistentVolumeClaimList
+	jobs      batchv1.JobList
 }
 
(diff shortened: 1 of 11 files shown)


jlong49 (approved):

controllers/cluster_status.go
```diff
@@ -165,7 +165,7 @@ func (r *ClusterReconciler) getNodes(ctx context.Context) (map[string]corev1.Nod
 	return data, nil
 }
 
-func (r *ClusterReconciler) getManagedPods(
+func (r *ClusterReconciler) getManagedInstances(
```
armru, line 168: we cannot scope the List query to pick up only the instance pods because we don't keep track in an unambiguous way of the instance at the pod creation.

The instance pods are identified only when the label role is applied but this happens after the creation.

This is why we are adding the new label on the instance pods, but it will make time until we can safely use it in our code due to backward compatibility. 

We should also think of a function that reconciles the pod labels with the new value that we are adding

leonardoce (approved):

PR #560, opened 2022-08-09
refactor: allows the override of the default external secret directory

### LINKED ISSUE #559: Allow the possibility to override `externalSecretsPath`
While developing a fix for #392, we noticed that we could not write unit tests for the behavior because we are blocked by:

https://github.com/cloudnative-pg/cloudnative-pg/blob/4e5f5bab3f4f237da2ff04045ea661e8115daa96/pkg/management/external/utils.go#L32

We should add a way to override the default path so we can write unit tests

armru:
This patch allows the developer to specify the directory where the external cluster secrets will be saved.
This is needed to allow the unit testing of certain external cluster methods.

Closes #559

As discussed with @jsilvela  @gbartolini  and @jlong49  this is a temporary fix until we properly refactor the whole chain call of functions

### DIFF
--- a/pkg/management/external/utils.go
+++ b/pkg/management/external/utils.go
@@ -27,18 +27,18 @@ import (
 )
 
 const (
-	// externalSecretsPath is the path where the cryptographic material
+	// defaultExternalSecretsPath is the default path where the cryptographic material
 	// needed to connect to an external cluster will be dumped
 	defaultExternalSecretsPath = "/controller/external" // #nosec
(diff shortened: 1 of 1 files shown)


jsilvela (approved):

leonardoce:
I see we already had a full green E2e suite on this, and after that we just changed commits. That's good, we don need to `/test` here.

leonardoce (approved):

PR #565, opened 2022-08-09
chore: reduce the amount of tests per push and pull_request

### LINKED ISSUE #564: Improving test starting from a bug
### Discussed in https://github.com/cloudnative-pg/cloudnative-pg/discussions/525

<div type='discussions-op-text'>

<sup>Originally posted by **sxd** August  4, 2022</sup>
Currently, since we implemented the `/ready-to-test` line which use `pull_request_target` instead of `pull_request` to run the test, we're expecting the label `ready to test :elephant:` and we cannot see this when running the test on `pull_request` or even when running the `/test` command.

The proposal is:
- Only run on `pull_request_target` when label is equal to `ready to test :elephant:` and not on any other situation, even when labels are removed.
- The `/test` or `/ready-to-test` command should be used to run all the E2E test, unit test can stay for now to run on every situation.
- Even if members of the project create a PR this will not run until `/test` or `/ready-to-test` command are trigger.

The idea with this approach is to:
* Do not run test every time we create a PR or we push
* Avoid the excessive consumption of runners, since we don't have so much runners
* Run test in PR from forked repo

What do you think about this @phisco @leonardoce @gbartolini ?</div>

sxd:
Reduce the amount of test that will run when pushing and creating
pull requests by using only on pull_request_target and those triggering
the tests using /test or /ready-to-test commands

Closes #564 


.github/workflows/continuous-delivery.yml
```diff
@@ -160,7 +159,13 @@ jobs:
         (
            github.event_name == 'pull_request_target' &&
            github.event.action == 'labeled' &&
-           contains(github.event.pull_request.labels.*.name, 'ready to test :elephant:')
+           (
+              contains(github.event.pull_request.labels.*.name, 'ready to test :elephant:')
+           )
+        ) ||
+        (
+          github.head_ref == 'main' ||
```
gbartolini, line 167: This looks suspicious @sxd

hh24k (approved):

litaocdl (approved):

mnencia (approved):

litaocdl:
@mnencia @sxd Could help force merge the pr, there are some `required` status can not get update here and I can not have merge button enabled. I am OK with the changes and there are part of function related with main branch I need this get merged then do the further check again.

PR #569, opened 2022-08-10
fix(externalcluster): sslmode is set based on the passed parameters

### LINKED ISSUE #392: Monolith database import is requiring a TLS connection
We're currently requiring a TLS connection between the CNP cluster to be created and the PostgreSQL instance we're downloading data from but just from the **monilith** import approach, but we don't to that for the **microservice** approach.

The reason behind this is that the microservice approach is going to create a connection from the instance manager to the source PostgreSQL instance using **lib/pq** which is by setting `sslmode` to `required` by default.
Unfortunately TLS is not commonly configured for development databases; see [this article](https://errorsingo.com/github.com-lib-pq-err-ssl-not-supported/) for more information.

We should check whether it's best to document this behavior or change our implementation to make the monolith and the microservice approach have the same requirements.

### LINKED ISSUE #571: we should ensure that the dbname is removed when creating the connection pooler for the imports
In our code, we create but don't use the modified `externalCluster` struct with the `dbname` `connectionParameter` removed.

We should fix this, as the connection string without dbname is needed to avoid crashes during the creation of the various import connection strings

https://github.com/cloudnative-pg/cloudnative-pg/blob/bc15477ae116d9c923edf614b78a2645cac04dfa/pkg/management/postgres/initdb.go#L386

armru:
This PR contains two fixes. For the description see the relative commits

Closes #392
Closes #571

### DIFF
--- a/docs/src/bootstrap.md
+++ b/docs/src/bootstrap.md
@@ -951,3 +951,21 @@ This will open up two main use cases:
 - replication over different Kubernetes clusters in CloudNativePG
 - *0 cutover time* migrations to CloudNativePG with the `pg_basebackup`
   bootstrap method
+
+## About `sslmode` in `externalCluster`
+
+We recommend that you explicitly specify `sslmode` in the
+`externalClusters.connectionParameters` section when connecting to the source
+PostgreSQL server, always.
+
+!!! Important
+    Please make sure you go through the available options from the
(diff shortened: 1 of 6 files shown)


armru:
We should document the sslmode election behavior

leonardoce (approved):

armru:
Do not squash the commits

leonardoce:
@armru can you please add a paragraph to the documentation explaining the behavior we implement? Thank you!

gbartolini (approved):

Issue #628, opened 2022-08-22
e2e failure due to can not find "cluster-storage-class.yaml" 


litaocdl:
After ticket https://github.com/cloudnative-pg/cloudnative-pg/issues/452 merged, we parse template file directly. 
1. There is some missing places which reference cluster-storage-class.yaml instead of template, will correct them.
2. Some fixtures without .template is added back from other pr, going to add back the gitignore and remote those files.

PR #629, opened 2022-08-22
test: use fixtrues as e2e test yaml file directory

litaocdl:
This patch fix the e2e test case to use the template files under fixtures directory
instead of samples directory
Closes: #628 


jlong49 (approved):

wadlejitendra (approved):
Done.

mnencia (approved):

PR #632, opened 2022-08-23
chore(pvc): add instance role on the pvc labels

### LINKED ISSUE #631: Propagate the role label to the underlying PVCs
We should propagate the `role` label also in the PVCs. When the operator manages the `role` label on a Pod,  we should also set this value on the underlying PVCs. This will help us to keep track of the PVC status.

armru:
This patch adds support for the `role` label on the PVC.

The PVC labeling is executed in a new method called `updateClusterRoleLabelsOnPVCs`, this ensures that the PVC has the correct role label.

Closes #631


armru:
a lot of weird failures, will rerun tests

leonardoce (approved):

armru:
Please rerun /test before merging

leonardoce (approved):

mnencia (approved):

PR #642, opened 2022-08-24
Load additional libraries when initializing database

redbaron:
This change makes `postInit*SQL` run on a instance with loaded additional libraries, which comes handy for `CREATE EXTENSION`  commands which can use features provided by additional libs.

Fixes: https://github.com/cloudnative-pg/cloudnative-pg/issues/649

### DIFF
--- a/pkg/management/postgres/initdb.go
+++ b/pkg/management/postgres/initdb.go
@@ -27,7 +27,6 @@ import (
 	"os/exec"
 	"path"
 	"sort"
-	"strings"
 
 	"github.com/lib/pq"
 	ctrl "sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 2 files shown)


redbaron:
@sxd, CI fails on container push with 403 Forbidden. Is it because I am submitting PR from a fork?

leonardoce:
@redbaron I slightly changed your patch to take care of the `shared_preload_libraries` that are automatically added by the operator, such at `pgaudit` when it's configured. 

Think about the following example:

```yaml
kind: Cluster
metadata:
  name: cluster-example
spec:
  instances: 3

  postgresql:
    parameters:
      "pgaudit.log": "all, -misc"
      "pgaudit.log_catalog": "off"
      "pgaudit.log_parameter": "on"
      "pgaudit.log_relation": "on"

  storage:
    size: 1Gi
```

What do you think about that? Is that fine with you?

redbaron:
@leonardoce, I am fine either way, feel free to take over this PR or even create your own one, since CI doesn't seem to like PR from forks.

sxd:
@redbaron @leonardoce the test passed except for this one https://github.com/cloudnative-pg/cloudnative-pg/runs/8019245065?check_suite_focus=true which is not related, I think we can merge.

I'm rebasing so the next person available can merge it! I'll try to do it before the end of my day

sxd (approved):

leonardoce (approved):

PR #655, opened 2022-08-26
Atomically check whether inplace instance upgrade is in progress

### LINKED ISSUE #654: instance manager's /update insufficient check to prevent parallel upgrades
HTTP handler for `/update` checks and sets `InstanceManagerIsUpgrading` non atomically, which makes it possible for more than one parallel upgrade processes to proceed

redbaron:
Fixes #654

### DIFF
--- a/pkg/management/postgres/webserver/remote.go
+++ b/pkg/management/postgres/webserver/remote.go
@@ -157,6 +157,8 @@ func (ws *remoteWebserverEndpoints) updateInstanceManager(
 		// Unfortunately this point, if everything is right, will not be reached.
 		// At this stage we are running the new version of the instance manager
 		// and not the old one.
+		// This is also the reason why we don't need to reset the
+		// `ws.instance.InstanceManagerIsUpgrading` variable.
 		_, _ = fmt.Fprint(w, "OK")
 	}
 }
--- a/pkg/management/postgres/webserver/remote.go
+++ b/pkg/management/postgres/webserver/remote.go
@@ -142,7 +142,7 @@ func (ws *remoteWebserverEndpoints) updateInstanceManager(
 		}
 
 		// No need to do anything if we are already upgrading
-		if ws.instance.InstanceManagerIsUpgrading {
+		if !ws.instance.InstanceManagerIsUpgrading.CompareAndSwap(false, true) {
 			http.Error(w, "instance manager is already upgrading", http.StatusTeapot)
 			return
 		}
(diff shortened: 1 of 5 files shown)


pkg/management/upgrade/upgrade.go
```diff
@@ -53,12 +53,10 @@ func FromReader(
 	instance *postgres.Instance,
 	r io.Reader,
 ) error {
-	// Mark this instance manager as upgrading
-	instance.InstanceManagerIsUpgrading = true
 	defer func() {
 		// Something didn't work if we reach this point, as this process
 		// should have already been replaced by a new one with the new binary
-		instance.InstanceManagerIsUpgrading = false
+		instance.InstanceManagerIsUpgrading.Store(false)
```
phisco, line 59: Given that now we are setting this to true outside this method, I'd move the reset too. WDYT?

mnencia:
I prefer to add a comment explaining why we do not need to store `false` if there is no error during the upgrade (that function never returns on success).

leonardoce:
Well, we could improve the old comment:

```
		// Unfortunately this point, if everything is right, will not be reached.
		// At this stage we are running the new version of the instance manager
		// and not the old one.
```

pkg/management/postgres/webserver/remote.go
```diff
@@ -142,14 +142,15 @@ func (ws *remoteWebserverEndpoints) updateInstanceManager(
 		}
 
 		// No need to do anything if we are already upgrading
-		if ws.instance.InstanceManagerIsUpgrading {
+		if !ws.instance.InstanceManagerIsUpgrading.CompareAndSwap(false, true) {
 			http.Error(w, "instance manager is already upgrading", http.StatusTeapot)
```
redbaron, line 146: I just noticed `StatusTeapot` :) Maybe 409 Conflict is more appropriate?

leonardoce:
I removed the code implementing the temporary file name, and put it in #659 to make review easier.

pkg/management/postgres/webserver/remote.go
leonardoce: Yeah, I don't have strong opinions on that!

leonardoce (approved):

jsilvela (approved):

jsilvela:
The logic for unsetting the `InstanceManagerIsUpgrading` was a bit hidden and magical.
I have made an explicit `defer` for the unsetting.

litaocdl:
The latest e2e triggered is https://github.com/cloudnative-pg/cloudnative-pg/actions/runs/2968471359, not sure why the unit test statuses could not be updated.

PR #660, opened 2022-08-26
fix: use a temporary file to host the new instance manager binary

### LINKED ISSUE #659: Use a temporary file to host the new instance manager binary
In this moment we are using a file with an hardcoded name to host the file where the operator will upload the new instance manager binary. We should use a temporary file with a generated name to avoid a possible race condition.

leonardoce:
Avoid using an hardcoded file name to store the newly uploaded
instance manager, preventing a possible race condition to happen.

Closes #659


leonardoce:
This code is based on #655, and should be merged after it and rebased over the latest main.

armru (approved):

jsilvela (approved):

mnencia (approved):

PR #664, opened 2022-08-29
feat: make webserver timeouts configurable

### LINKED ISSUE #630: Configurable HTTP server timeout parameters
In https://github.com/cloudnative-pg/cloudnative-pg/commit/197b3f23daab610d5796dcea8e3a8f551ef32e95  we introduced timeouts on the http.Server as a security measure against the SlowLoris type of attack.

While this may be totally correct, we are using the HTTP server to upload the new instance manager executable from the operator Pod to the PostgreSQL Pod and the timeouts we have now, which is hardcoded to 30 seconds, may not be enough depending on the network configuration of that Kubernetes cluster.

This parameters should be configurable at least at operator level.

hh24k:
Closes #630 


sxd:
The issue was abandoned and we should do the same with the PR waiting for a future conversation

PR #668, opened 2022-08-30
chore: replaces connection driver `postgres` with `pgx`

### LINKED ISSUE #667: Replace libpq with pgx
Following https://github.com/cloudnative-pg/cloudnative-pg/pull/569 limitations,  we should replace libpq with pgx as it is better maintained.

armru:
This patch replaces the connection driver `postgres` with `pgx` to avoid some of the limitations in the future that we encountered while developing #569.

Most of the code still relies on the simple protocol connection, further work will be needed to use the advanced protocol.

Closes #667


sxd (approved):

docs/src/database_import.md
```diff
@@ -219,8 +219,6 @@ There are a few things you need to be aware of when using the `monolith` type:
   ["The `externalClusters` section"](bootstrap.md#the-externalclusters-section))
 - Traffic must be allowed between the Kubernetes cluster and the
   `externalCluster` during the operation
-- You need to specify `sslmode: disable` in the `connectionParameters` section
```
armru, line 222: we are using pgx but  still using the simple protocol, I think the same limitations still apply

litaocdl (approved):

leonardoce (approved):

PR #674, opened 2022-08-31
tests: Add E2E for replication slots

NiccoloFei:
First implementation of the E2E for replication slots.

E2E tests for #8.

Signed-off-by: Niccolò Fei <niccolo.fei@enterprisedb.com>

### DIFF
--- a/tests/e2e/asserts_test.go
+++ b/tests/e2e/asserts_test.go
@@ -2476,18 +2476,10 @@ func AssertPvcHasLabels(
 // and that only the required slots exist in the target pod.
 func AssertRepSlotsOnPod(
 	namespace,
-	clusterName,
-	podName string,
+	clusterName string,
+	pod corev1.Pod,
 ) {
-	namespacedName := types.NamespacedName{
-		Namespace: namespace,
-		Name:      podName,
-	}
-	targetPod := &corev1.Pod{}
-	err := env.Client.Get(env.Ctx, namespacedName, targetPod)
-	Expect(err).ToNot(HaveOccurred())
-
-	expectedSlots, err := testsUtils.GetExpectedRepSlotsOnPod(namespace, clusterName, podName, env)
+	expectedSlots, err := testsUtils.GetExpectedRepSlotsOnPod(namespace, clusterName, pod.GetName(), env)
 	Expect(err).ToNot(HaveOccurred())
 
 	for _, slot := range expectedSlots {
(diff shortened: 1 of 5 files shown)


tests/utils/replication_slots.go
```diff
@@ -0,0 +1,89 @@
...
+		if lsn != lsnList[0] {
+			return false
+		}
+	}
+	return true
+}
+
+// DiscoverRepSlotsOnPod returns a list of replication slot names which should be present
+// in a given pod
+func DiscoverRepSlotsOnPod(namespace, clusterName, podName string, env *TestingEnvironment) ([]string, error) {
```
litaocdl, line 41: This function returns the expected replication slot names in a pod through name converting, should we use getExpectedRepSlotOnPod.

tests/e2e/asserts_test.go
```diff
@@ -2426,3 +2426,59 @@ func AssertPvcHasLabels(
...
+		Name:      podName,
+	}
+	targetPod := &corev1.Pod{}
+	err := env.Client.Get(env.Ctx, namespacedName, targetPod)
+	Expect(err).ToNot(HaveOccurred())
+
+	slots, err := testsUtils.DiscoverRepSlotsOnPod(namespace, clusterName, podName, env)
+	Expect(err).ToNot(HaveOccurred())
+
+	for _, slot := range slots {
```
litaocdl, line 2448: Already verified all expected replication slot is in this pod, vice versa, going to add logic to verify all existing replication slot is as expected, no redundant. this could be useful. when scale down or switch over, check redundant replication shot is deleted.

tests/e2e/replication_slot_test.go
```diff
@@ -0,0 +1,105 @@
...
+		if CurrentSpecReport().Failed() {
+			env.DumpNamespaceObjects(namespace, "out/"+CurrentSpecReport().LeafNodeText+".log")
+		}
+	})
+	AfterEach(func() {
+		err := env.DeleteNamespace(namespace)
+		Expect(err).ToNot(HaveOccurred())
+	})
+
+	It("manage Replication slots for HA", func() {
```
litaocdl, line 50: Going to discuss and cover more scenario, with both highAvailability.enable=true and false

tests/e2e/asserts_test.go
NiccoloFei: > going to add logic to verify all existing replication slot is as expected, no redundant. this could be useful. when scale down or switch over, check redundant replication shot is deleted.

We already have a list of expected rep_slots here, so we could gather the current list of rep_slots on the targetPod and expect the 2 lists to eventually be equivalent (that would cover also scale down/switchover where "old" slots needs to be removed by the operator)

tests/utils/replication_slots.go
NiccoloFei: Done.

tests/e2e/asserts_test.go
NiccoloFei: Done.

PR #680, opened 2022-08-31
chore: ensure that `cnpg.io/instanceName` and `cnpg.io/podRole` are correctly reconciled on existing resources

### LINKED ISSUE #679: reconcile missing labels on resources created on older versions
Right now we lack a way to reconcile some of the labels on the resources  that were created on older versions, we should proceed to add  a logic that ensures that the expected labels are always present

armru:
This patch ensures that the labels:
- `"cnpg.io/instanceName` is present on both PVCs and instance pods
- `"cnpg.io/podRole"` on the instance pod

are correctly reconciled during the `reconcileResources` phase.

This will make sure that:
- those labels will be eventually present on  the pods and pvcs created on previous versions
- in case of accidental deletion they will be readded by the operator

Closes #679

### DIFF
--- a/controllers/cluster_controller.go
+++ b/controllers/cluster_controller.go
@@ -380,12 +380,21 @@ func (r *ClusterReconciler) reconcileResources(
 ) (ctrl.Result, error) {
 	contextLogger, ctx := log.SetupLogger(ctx)
 
+	// TODO: refactor how we handle label and annotation reconciliation.
+	// TODO: We should generate a fake pod containing the expected labels and annotations and compare it to the living pod
+
 	// Update the labels for the -rw service to work correctly
 	if err := r.updateRoleLabelsOnPods(ctx, cluster, resources.instances); err != nil {
 		return ctrl.Result{}, fmt.Errorf("cannot update role labels on pods: %w", err)
 	}
 
-	if err := r.updateClusterRoleLabelsOnPVCs(ctx, resources.instances, resources.pvcs); err != nil {
+	// updated any labels that are coming from the operator
+	if err := r.updateOperatorLabelsOnInstances(ctx, resources.instances); err != nil {
+		return ctrl.Result{}, fmt.Errorf("cannot update instance labels on pods: %w", err)
+	}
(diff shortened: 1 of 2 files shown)


controllers/replicas.go
```diff
@@ -552,30 +552,67 @@ func (r *ClusterReconciler) updateRoleLabelsOnPods(
...
 	ctx context.Context,
-	pods corev1.PodList,
-	pvcs corev1.PersistentVolumeClaimList,
+	instances corev1.PodList,
 ) error {
-	for _, pod := range pods.Items {
-		podRole, podHasRole := pod.ObjectMeta.Labels[specs.ClusterRoleLabelName]
-		if !podHasRole {
+	for _, instance := range instances.Items {
+		instance := instance
```
jlong49, line 561: why are we creating a new variable here when we could just use the index?

instances.Items[i] would be a more efficient way of writing this

controllers/cluster_controller.go
```diff
@@ -385,7 +385,15 @@ func (r *ClusterReconciler) reconcileResources(
 		return ctrl.Result{}, fmt.Errorf("cannot update role labels on pods: %w", err)
 	}
 
-	if err := r.updateClusterRoleLabelsOnPVCs(ctx, resources.instances, resources.pvcs); err != nil {
+	// updated any labels that are coming from the operator
+	if err := r.updateOperatorLabelsOnInstances(ctx, resources.instances); err != nil {
+		return ctrl.Result{}, fmt.Errorf("cannot update instance labels on pods: %w", err)
+	}
+
+	// TODO: refactor how we handle label and annotation reconciliation. We should handle them in a more programmatic way
```
armru, line 393: we should move to a design where we compare the expected pod labels and annotations with the current living one and reconcile it.

litaocdl (approved):

leonardoce (approved):

leonardoce:
The e2e tests have been executed on the EDB infrastructure to spare the ones from the community.

PR #682, opened 2022-09-01
feat: replication slots for HA: manage slots in the primary

### LINKED ISSUE #671: Replication slots for HA: manage slots in the primary
Following #669, we need to instruct the current primary in the cluster to manage the **primary HA slots** if `replicationSlots.highAvailability.enabled` is true.

In CloudNativePG terminology, a primary HA slot is a physical replication slot whose lifecycle is entirely managed by the current primary of the cluster, and whose purpose is to map to a specific standby in streaming replication. Such slots live on the primary only.

The idea is to make sure that, based on the expected standbys, the primary verifies that there's a corresponding replication slot with the name `_cnpg_<REPLICA_POD_NAME>` (for simplicity and better readability we are using `_cnpg_` here, but it should actually be set based on the value of `slotPrefix`.

The instance manager in the primary will create the missing replication slots, without initializing them.

Write unit tests for the logic that builds the list of expected replication slot names.

mnencia:
Closes #671

Co-authored-by: Marco Nenciarini <marco.nenciarini@enterprisedb.com>


mnencia (commented):
This is not unused. It's part of the interface.

armru (approved):

litaocdl (approved):

PR #684, opened 2022-09-01
fix: avoid creating Pods for an incomplete group of PVCs

### LINKED ISSUE #683: Avoid creating Pods for an incomplete group of PVCs
When a cluster is configured with separate storage for WAL files,
if you delete a Pod and only one of their PVCs, the operator tries to recreate the Pod
even if one of its PVCs is not present anymore.

Instead, we should ignore the unusable PVC and provision a new instance to reach the correct number of replicas.

mnencia:
Closes #683


armru (approved):

pkg/specs/pvcs.go
```diff
@@ -178,84 +184,123 @@ pvcLoop:
...
+				result.Unusable = append(result.Unusable, pvcNames...)
+				continue instancesLoop
 			}
 		}
 
-		if pvc.Annotations[PVCStatusAnnotationName] != PVCStatusReady {
-			// This PVC has not a Job nor a Pod using it, but it is not marked as PVCStatusReady
-			// we need to ignore it here
-			continue
+		// If we have more PVC than expected, the extra PVCs are unusable
```
leonardoce, line 224: If I understand correctly, what happens here is that we don't have "**more PVC than expected**" (as in count) but we found a PVC which we don't expect (as in subsets definition).

If I'm correct, perhaps we could fix this comment.

pkg/specs/pvcs.go
```diff
@@ -178,84 +184,123 @@ pvcLoop:
...
-			// we lost some pvc null them all
-			for _, index := range indexOfFoundPVCs {
-				result.Dangling = append(result.Dangling, result.Healthy[index])
-				result.Healthy = removeElementByIndex(result.Healthy, index)
+		// At this point the PVCs are dangling, so we need to check if some of them has
+		// a PVC status annotation marking it as non-ready.
+		// In that case, all the instance PVCs are unusable
+		for _, pvc := range pvcs {
+			// We ignore any PVC that is not expected
+			if !slices.Contains(expectedPVCs, pvc.Name) {
```
leonardoce, line 267: Is this really possible? We already have https://github.com/cloudnative-pg/cloudnative-pg/pull/684/files#diff-fd3956fd8d840a5772b8b430eca9c8eb8b36d261a1c91e240052ae46dd2e4964R226

leonardoce (approved):

pkg/specs/pvcs.go
mnencia: Yes, you are right

pkg/specs/pvcs.go
mnencia: That check doesn't remove the extra PVC from the list, so we need to filter them in this check, otherwise they could invalidate the whole instance.

mnencia: Here we are iterating over PVCs not PVC names because we need to check the annotations.

pkg/specs/pvcs.go
armru: I agree

PR #690, opened 2022-09-05
chore: raises the default read timeout values of the instance manager

### LINKED ISSUE #630: Configurable HTTP server timeout parameters

armru:
This patch raises the webserver constants `DefaultReadTimeout` and `DefaultReadHeaderTimeout` to avoid network issues on slower Kubernetes clusters.

This is a temporary fix until further work is done to dynamically configure these parameters.

Closes #630

### DIFF
--- a/pkg/management/pgbouncer/metricsserver/metricsserver.go
+++ b/pkg/management/pgbouncer/metricsserver/metricsserver.go
@@ -65,8 +65,8 @@ func ListenAndServe() error {
 	server = &http.Server{
 		Addr:              fmt.Sprintf(":%d", url.PgBouncerMetricsPort),
 		Handler:           serveMux,
-		ReadTimeout:       webserver.DefaultReadTimeout,
-		ReadHeaderTimeout: webserver.DefaultReadHeaderTimeout,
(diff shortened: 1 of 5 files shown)


armru:
if we decide to merge also the second commit we should not squash

hh24k (approved):

armru:
remember to run `/test` before we merge just to be extra sure

hh24k (approved):

armru:
We decided to not proceed with this implementation, the default values already in place are good enough for most cases.
What we will work on is the possibility of overriding the default values

PR #693, opened 2022-09-05
feat: Replication slots for HA: manage slots in the standby servers

### LINKED ISSUE #689: Replication slots for HA: manage slots in the standby servers
Following #671, we need to instruct each standby in the cluster to manage the standby HA slots if `replicationSlots.highAvailability.enabled` is `true`.

In CloudNativePG terminology, a standby HA slot is a physical replication slot whose lifecycle is entirely managed by another standby in the cluster, based on the content of the `pg_replication_slots` view in the primary - updated at regular intervals. Its purpose is to map to a specific standby in streaming replication. Such slots live on the standbys only.

Replicas will use the `<SLOT_PREFIX><REPLICA_POD_NAME>` replication slot when connecting to the primary for streaming replication purposes.

Each replica will also maintain a local copy of the replication slots of all the other replicas (excluding the primary and themselves) by querying the `pg_stat_replication_slots` view on the primary at regular intervals, and advance the position of the local slot using pg_replication_slot_advance() - available from PostgreSQL 11. The query should be similar to: `SELECT slot_name FROM pg_replication_slots WHERE NOT temporary AND restart_lsn IS NOT NULL AND  slot_name ~ ‘^<SLOT_PREFIX>’ AND slot_name != ‘<MY_SLOT_NAME>’` - where `restart_lsn IS NOT NULL` denotes an initialized slot. These slots are called standby HA slots.

`<SLOT_PREFIX>` is the value of the `replicationSlots.highAvailability.slotPrefix` setting.

This is a basic idea, please explore and expand during prototyping.

mnencia:
Closes #689


internal/management/controller/slots/slots.go
```diff
@@ -104,7 +104,7 @@ func (sm PostgresManager) List(
 
 	rows, err := db.QueryContext(
 		ctx,
-		`SELECT slot_name, slot_type, active, restart_lsn FROM pg_replication_slots
+		`SELECT slot_name, slot_type, active, coalesce(restart_lsn::text, '') as restart_lsn FROM pg_replication_slots
```
mnencia: That is unnecessary as a null value is an empty string when scanned as a string value.

armru (approved):

internal/management/controller/slots/infrastructure/replicationslot.go
```diff
@@ -0,0 +1,51 @@
...
+// ReplicationSlot represents a single replication slot
+type ReplicationSlot struct {
+	SlotName   string   `json:"slotName,omitempty"`
+	Type       SlotType `json:"type,omitempty"`
+	Active     bool     `json:"active"`
+	RestartLSN string   `json:"restartLSN,omitempty"`
+}
+
+// ReplicationSlotList contains a list of replication slots
+type ReplicationSlotList struct {
```
jsilvela, line 34: The more I look at this, the more I think it should be a map[string]ReplicationSlot.
I.e. we should not allow, nor ever have, two slots with the same name.

jsilvela (approved):
👏

jsilvela:
The E2E test for replication slots is failing consistently for Postgres 10 images.
Makes sense. Will update the test to expect the validation hook rejection on PG10.

tests/e2e/replication_slot_test.go
```diff
@@ -48,6 +50,19 @@ var _ = Describe("Replication Slot", func() {
 	})
 
 	It("can manage Replication slots for HA", func() {
+		// check if current postgres version is pg 11
+		if strings.Contains(os.Getenv("POSTGRES_IMG"), ":10") {
+			AssertCreateNamespace(namespace, env)
```
mnencia, line 55: This should be factored together with the creation on line 67

tests/e2e/fixtures/replication_slot/cluster-pg-replication-slot.yaml.template
```diff
@@ -4,6 +4,7 @@ metadata:
   name: cluster-pg-replication-slot
 spec:
   instances: 3
+  imageName: "${POSTGRES_IMG}"
```
mnencia, line 7: I think this should not be required. If the default image is postgres <= 10 it should fail even if not explicitly stated. Otherwise, it's a bug.

jsilvela: For manual testing this was not working as expected.

jsilvela: I mean for running locally

tests/e2e/replication_slot_test.go
jsilvela: Hm. This is slightly different. The AssertCreateCluster already includes waiting for the namespace.
However, even if there is some functional redundancy, no harm done, will do as you suggest.
In a future ticket I'd like to refactor the AssertCreateCluster to assume the namespace is already there.

mnencia: I would just use AssertCreateCluster also in the other place. It doesn't harm to wait until the object is visible.

tests/e2e/fixtures/replication_slot/cluster-pg-replication-slot.yaml.template
mnencia: I would remove it anyway because it is not needed for the test.

tests/e2e/replication_slot_test.go
jsilvela: Yep

jsilvela: Better yet, do it in the beginning of the test. Either pg < 11 or >= 11 need it created.

Issue #413, opened 2022-07-13
use policyv1 for PodDisruptionBudget for Kubernetes 1.25+


armru:
In Kubernetes 1.25+ `PodDisruptionBudget` beta1 is getting dropped and this is relevant because we currently use it in our codebase.

To solve this issue we can choose between:
- Create betav1policyv1 if kubernetes versions is 1.19 or 1.20. This perhaps will require introducing in our codebase a Kubernetes version getter and parser just like we do for Postgres.
- Drop the support for 1.19 and 1.20 and update to policyv1  

Feel free to suggest any other option available to us

mdallaire:
Is there any news on this? I just upgraded to 1.25 and now my cnpg-operator is crashing :)

PR #717, opened 2022-09-09
Use non-deprecated PodDisruptionBudget version

alexpearce:
PodDisruptionBudget version policyv1beta1 was dropped in Kubernetes v1.25.

Addresses #413 via option 2 presented there. I don't know how important 1.19/1.20 support is for CNPG, but if "not very" then this one possible fix.

### DIFF
--- a/controllers/cluster_controller.go
+++ b/controllers/cluster_controller.go
@@ -29,7 +29,7 @@ import (
 
 	batchv1 "k8s.io/api/batch/v1"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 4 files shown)


mnencia:
Unit tests on 1.20 are failing. This PR should update the list of tested k8s versions and the documentation about supported versions.

mnencia:
fixed in a64eda701984bff6736d75b361890895f43df4e0

alexpearce:
Thanks very much @mnencia!

PR #720, opened 2022-09-09
fix: ensure that the conditions contained in the cluster status are made compatible with metav1.Conditions struct

### LINKED ISSUE #719: ensure that clusters created by the 1.15 version contains conditions that are compatible with the successive releases
In version 1.15 of the operator, we use conditions that are not compatible with the metav1.Conditions structs.
Given that we started to use metav1.Conditions struct in 1.16 we should be sure that in the 1.15 version we have conditions that can be converted to the new struct

armru:
This patch cherry-picks the PR containing the logic needed to sanitize the cluster conditions

Closes #719 

### DIFF
--- a/api/v1/cluster_types.go
+++ b/api/v1/cluster_types.go
@@ -402,6 +402,28 @@ const (
 	ConditionUnknown ConditionStatus = "Unknown"
 )
 
+// ConditionReason defines the reason why a certain
+// condition changed
+type ConditionReason string
+
+const (
+	// ConditionReasonLastBackupSucceeded means that the condition changed because the last backup
+	// has been taken successfully
+	ConditionReasonLastBackupSucceeded ConditionReason = "LastBackupSucceeded"
+
+	// ConditionReasonLastBackupFailed means that the condition changed because the last backup
+	// failed
+	ConditionReasonLastBackupFailed ConditionReason = "LastBackupFailed"
+
+	// ConditionReasonContinuousArchivingSuccess means that the condition changed because the
+	// WAL archiving was working correctly
+	ConditionReasonContinuousArchivingSuccess ConditionReason = "ContinuousArchivingSuccess"
+
+	// ConditionReasonContinuousArchivingFailing means that the condition has changed because
+	// the WAL archiving is not working correctly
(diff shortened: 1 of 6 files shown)


jlong49 (approved):

mnencia (approved):

danishedb (approved):

mnencia:
I just discovered that the condition with the wrong reason is still in the code. We need to fix it before merging.

mnencia (changes requested):
I just discovered that the condition with the wrong reason is still in the code. We need to fix it before merging.

```
$ git grep -n Reason internal/cmd/manager/walarchive/cmd.go pkg/management/postgres/backup.go
internal/cmd/manager/walarchive/cmd.go:348:                     Reason:  "Continuous Archiving is Failing",
internal/cmd/manager/walarchive/cmd.go:355:             Reason:  "Continuous Archiving is Working",
pkg/management/postgres/backup.go:484:                  Reason:  "Last backup failed",
pkg/management/postgres/backup.go:491:          Reason:  "Last backup succeeded",
```

we need to pick from c2fa5d9fb4e33e0bab9319a54b2018cdec1dfa45 the constants `apiv1.ConditionReasonContinuousArchivingFailing`, `apiv1.ConditionReasonContinuousArchivingSuccess`, `apiv1.ConditionBackupStarted`, `apiv1.ConditionReasonLastBackupFailed`, and `apiv1.ConditionReasonLastBackupSucceeded` and use them to fill the content of the Reason field.

armru:
The patch should remove the incompatible version set i allowing the user to upgrade to the newer CNPG versions.
I think that modifying the existing conditions on 1.15 could be a different PR

mnencia:
> The patch should remove the incompatible condition allowing the user to upgrade to the newer CNPG versions. I think that modifying the existing conditions on 1.15 should be in a different PR

The problem is not fixing the existing condition. We continue to add the wrong reasons in the conditions even in this version, then remove the condition immediately after. It is a regression because, after this patch, we will not report any condition in our clusters.

mnencia (approved):

PR #737, opened 2022-09-14
fix: ensures that the promoted standby active connections are dropped after promotion

### LINKED ISSUE #736: ensure we drop all active connections on the standby that is promoted
We should ensure that we drop all the read-only connections after we promote a standby.

armru:
This patch drops all the connections made before the standby promotion are dropped.

Closes #736

### DIFF
--- a/internal/management/controller/instance_controller.go
+++ b/internal/management/controller/instance_controller.go
@@ -1034,10 +1034,15 @@ func (r *InstanceReconciler) reconcilePrimary(ctx context.Context, cluster *apiv
 	if cluster.Status.CurrentPrimary != r.instance.PodName {
 		cluster.Status.CurrentPrimary = r.instance.PodName
 		cluster.Status.CurrentPrimaryTimestamp = pkgUtils.GetCurrentTimestamp()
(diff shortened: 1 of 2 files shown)


jlong49 (approved):

pkg/utils/time.go
```diff
@@ -27,11 +27,15 @@ import (
...
+	if parsedTimestamp, err := time.Parse(metav1.RFC3339Micro, timestamp); err == nil {
+		return parsedTimestamp.Format("2006-01-02 15:04:05.000000Z07:00")
+	}
+
+	if parsedTimestamp, err := time.Parse(time.RFC3339, timestamp); err == nil {
+		return parsedTimestamp.Format("2006-01-02 15:04:05.000000Z07:00")
 	}
-	return parsedTimestamp.Format("2006-01-02 15:04:05.000000Z07:00")
+
+	return timestamp
```
mnencia, line 38: This is a good change, but I do not think it belongs to this patch.

mnencia (approved):

jsilvela (approved):

PR #740, opened 2022-09-15
Replication slots for HA

### LINKED ISSUE #8: Replication slots for HA

mnencia:
Closes #8


armru (approved):

internal/cmd/manager/instance/run/cmd.go
```diff
@@ -178,6 +179,12 @@ func runSubCommand(ctx context.Context, instance *postgres.Instance) error {
 		return err
 	}
 
+	slotReplicator := runner.NewReplicator(instance)
+	if err = mgr.Add(slotReplicator); err != nil {
+		setupLog.Error(err, "unable to create slot slotReplicator")
```
phisco, line 184:
```suggestion
		setupLog.Error(err, "unable to create slot Replicator")
```

mnencia: @leonardoce fixed it in his review.

gbartolini (approved):

leonardoce (approved):

PR #782, opened 2022-09-26
feat: introduce hibernate on command

### LINKED ISSUE #781: Add "hibernate on" command to kubectl plugin
Add the following command to the `cnpg` plugin for `kubectl`:

```console
kubectl cnpg hibernate on [-n NAMESPACE] <CLUSTER_NAME>
```

As a result, the plugin will be:

- Fencing the cluster
- Destroying one standby at a time, including their own PVCs
- Destroying the primary by keeping the PVC group, by removing the ownership metadata
- Adding the necessary annotations (to be determined) to all PVCs, in particular the cluster information, excluding the status

armru:
Closes #781


armru:
I'm retaining the cluster status in the PVC annotation to facilitate the debugging if a PVC won't restart correctly

internal/cmd/plugin/hibernate/on.go
```diff
@@ -0,0 +1,209 @@
...
+		specs.PostgresContainerName,
+		&timeout,
+		"pg_ctl", "status")
+
+	// todo: do not rely on strings
+	// pg_ctl returns exit code 3 if no postgres is running
+	if err != nil && !strings.Contains(err.Error(), "command terminated with exit code 3") {
+		return false, fmt.Errorf("isInstanceRunning command returned the following error: %s", err)
+	}
+	return stdout == "pg_ctl: no server running", nil
```
jsilvela, line 208: isn't this the opposite of what the function name implies?

jsilvela:
A comment here. I was doing a test locally, and the hibernation command did not recognize that it had stopped the instances (which it had). After the retries, the command failed, and the instances remained stopped and the PVCs were not annotated.
Should not there be some transactionality here?

internal/cmd/plugin/hibernate/on.go
```diff
@@ -0,0 +1,209 @@
...
+		&timeout,
+		"pg_controldata")
+	if err != nil {
+		return "", err
+	}
+
+	return stdout, nil
+}
+
+func isInstanceStopped(
```
jsilvela, line 188: wondering if we could leverage `func (instance *Instance) isStatusRunning()` from the instance manager.
Going to experiment a bit...

jsilvela (approved):

leonardoce (approved):

internal/cmd/plugin/hibernate/on.go
```diff
@@ -0,0 +1,415 @@
...
+	fencedInstances, err := utils.GetFencedInstances(on.cluster.Annotations)
+	if err != nil {
+		return fmt.Errorf("could not check if cluster is fenced: %v", err)
+	}
+
+	if fencedInstances.Len() > 0 {
+		if on.force {
+			contextLogger.Warning("Continuing hibernation procedure even there are fenced instances")
+		} else {
+			return fmt.Errorf("cannot hibernate a cluster with has fenced instances")
```
sxd, line 170: Isn't this something like: `cannot hibernate a cluster that has fenced instances` ?

PR #800, opened 2022-10-01
chore: persist manifest file for dev branch

### LINKED ISSUE #771: Facilitate testing of the latest snapshot
At the moment, there is no straightforward way - as an end user - to test the operator's latest snapshot. While we build the images, we are not building the YAML manifest associated to that specific release and that automatically uses that operator image.

Would it be possible to add the YAML file to the artifacts for that release?

Once this is done, we should add instructions in the CONTRIBUTING section or even in the documentation itself.

hh24k:
Closes #771 


NiccoloFei (approved):

.github/workflows/continuous-delivery.yml
```diff
@@ -487,6 +498,71 @@ jobs:
...
           image=$(sed -n '1{s/,.*//; p}' <<< "$TAGS")
           echo "::set-output name=image::${image}"
+      -
+        name: Generate manifest for operator deployment
+        id: generate-manifest
+        run: |
+          make generate-manifest
+      -
+        name: Upload the operator manifest as artifact in workflow
+        if: always()
```
mnencia, line 508: Why do we use `if: always()` here? What step could fail before this without compromising the manifest generation?

.github/workflows/continuous-delivery.yml
```diff
@@ -487,6 +498,71 @@ jobs:
...
+          name: operator-manifest.yaml
+          path: dist/operator-manifest.yaml
+          retention-days: 7
+      -
+        name: Replicate the artifact or not (to cloudnative-pg/artifacts repo)
+        id: replicate-or-not
+        run: |
+          branch_name=${{ steps.build-meta.outputs.branch_name }}
+          # TODO after review/testing, the `|| ${branch_name} == 'dev/771'` should be removed
+          if [[ ${branch_name} == 'main' || ${branch_name} =~ 'release/' || ${branch_name} == 'dev/771' ]]; then
```
mnencia, line 520: We need to remove `dev/711` from here before merging.

.github/workflows/continuous-delivery.yml
```diff
@@ -487,6 +498,71 @@ jobs:
...
+      -
+        name: Upload the operator manifest as artifact in workflow
+        if: always()
+        uses: actions/upload-artifact@v3
+        with:
+          name: operator-manifest.yaml
+          path: dist/operator-manifest.yaml
+          retention-days: 7
+      -
+        name: Replicate the artifact or not (to cloudnative-pg/artifacts repo)
```
mnencia, line 515: I do not like calling this step "replicate". It is publishing. I would change the step title and all the related variables.

mnencia (commented):
I do not like to have all the publishing work at the end of the build. I would upload the generated manifest as an artifact of this build and then publish it in the artifacts repo in a separate Job.

.github/workflows/continuous-delivery.yml
hh24k: If publishing the manifest in the artifacts repo in a separate Job, we will need to download the manifest as an artifact before publishing it since different jobs don't share files directly.

mnencia (approved):

PR #807, opened 2022-10-04
feat: add support for http proxies

fyannk:
One of my cloud providers expects me to manage proxies to go outside of the Kubernetes cluster, especially for S3 storage.

Here is a proposal to take common proxies environment variables at operator install and push them to child clusters at creation time.

Example:

Environment added at operator install:

```yaml
    - name: HTTP_PROXY
      value: http://someproxy:8080
    - name: HTTPS_PROXY
      value: http://someproxy:8080
    - name: NO_PROXY
      value: .cluster.local,.svc,10.128.0.0/14,127.0.0.0/8,127.0.0.1,localhost
```

Without patch, environment for a pod of a Cluster:

```yaml
    env:
    - name: PGDATA
      value: /var/lib/postgresql/data/pgdata
    - name: POD_NAME
      value: pgcluster-1
    - name: NAMESPACE
      value: somenamespace
    - name: CLUSTER_NAME
      value: pgcluster
    - name: PGPORT
      value: "5432"
    - name: PGHOST
      value: /controller/run
```

With patch, environment for a pod of a Cluster:

```yaml
    env:
    - name: PGDATA
      value: /var/lib/postgresql/data/pgdata
    - name: POD_NAME
      value: pgcluster-1
    - name: NAMESPACE
      value: somenamespace
    - name: CLUSTER_NAME
      value: pgcluster
    - name: PGPORT
      value: "5432"
    - name: PGHOST
      value: /controller/run
    - name: HTTP_PROXY
      value: http://someproxy:8080
    - name: HTTPS_PROXY
      value: http://someproxy:8080
    - name: NO_PROXY
      value: .cluster.local,.svc,10.128.0.0/14,127.0.0.0/8,127.0.0.1,localhost
```

### DIFF
--- a/internal/configuration/configuration.go
+++ b/internal/configuration/configuration.go
@@ -48,9 +48,9 @@ type Data struct {
 	EnablePodDebugging bool `json:"enablePodDebugging" env:"POD_DEBUG"`
 
 	// Proxies variable
-	EnvHttpProxy  string `env:"HTTP_PROXY"`
-	EnvHttpsProxy string `env:"HTTPS_PROXY"`
-	EnvNoProxy    string `env:"NO_PROXY"`
+	EnvHttpProxy  string `json:"envHttpProxy" env:"HTTP_PROXY"`
+	EnvHttpsProxy string `json:"envHttpsProxy" env:"HTTPS_PROXY"`
+	EnvNoProxy    string `json:"envNoProxy" env:"NO_PROXY"`
 
 	// OperatorNamespace is the namespace where the operator is installed
 	OperatorNamespace string `json:"operatorNamespace" env:"OPERATOR_NAMESPACE"`
--- a/internal/configuration/configuration.go
+++ b/internal/configuration/configuration.go
@@ -47,6 +47,11 @@ type Data struct {
 	// EnablePodDebugging enable debugging mode in new generated pods
 	EnablePodDebugging bool `json:"enablePodDebugging" env:"POD_DEBUG"`
 
+	// Proxies variable
+	EnvHttpProxy  string `env:"HTTP_PROXY"`
+	EnvHttpsProxy string `env:"HTTPS_PROXY"`
+	EnvNoProxy    string `env:"NO_PROXY"`
+
 	// OperatorNamespace is the namespace where the operator is installed
 	OperatorNamespace string `json:"operatorNamespace" env:"OPERATOR_NAMESPACE"`
 
--- a/api/v1/cluster_types.go
+++ b/api/v1/cluster_types.go
@@ -1527,6 +1527,30 @@ func (cluster *Cluster) GetImageName() string {
 	return configuration.Current.PostgresImageName
 }
 
+// SetEnvProxies to pass proxies defined for operator to clusters
+func (cluster *Cluster) SetEnvProxies(envvar []corev1.EnvVar) []corev1.EnvVar {
+
+	if configuration.Current.EnvHttpProxy != "" {
+		envvar = append(envvar, corev1.EnvVar{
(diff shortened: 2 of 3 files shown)


internal/configuration/configuration.go
```diff
@@ -47,6 +47,11 @@ type Data struct {
 	// EnablePodDebugging enable debugging mode in new generated pods
 	EnablePodDebugging bool `json:"enablePodDebugging" env:"POD_DEBUG"`
 
+	// Proxies variable
+	EnvHttpProxy  string `env:"HTTP_PROXY"`
+	EnvHttpsProxy string `env:"HTTPS_PROXY"`
+	EnvNoProxy    string `env:"NO_PROXY"`
```
phisco, line 53: Please add also the json tag, in camel case, as the other fields in the struct

phisco:
Thanks for your contribution! The only doubt I have is whether it's worth introducing directly a map and allowing setting any required environment variable or we just want to keep it only for proxy configs. WDYT @mnencia @sxd @leonardoce @armru?

netthier:
+1, the only way of doing this at the moment is modifying the underlying container image.

sxd:
@fyannk we talked about this in the community meeting, you can see the agenda here were we put our thoughts https://docs.google.com/document/d/1Bmf2AZG5WLKAyESJbYk7MbsfiuD3jgdIDQrDkNuKT9w/ you will able to watch the recording later today, but basically we will want to have a more generic way to inject env variables into the pods, what do you think about it?

fyannk:
@sxd thank you for your reply

As I saw several questions during your meeting, I confirm this is for a in-premises k8s cluster (with a strange architecture), and my problem is for egress for backups, as without proxies pgBarman cannot contact S3 storage that is outside of the cluster.

For now as it is not possible to put env vars with stock cnpg operator inside pg pods, I'll be happy even if we have to put it at pgcluster levels.

My main problem here is more philosophical, as those values are for egress on this specific k8s cluster.
I especially wanted to avoid end users that creates pgclusters to have some cloud-specific parameters to manage, this burden should be only at the cloud admin level as he is the one with the specific architecture.

For instance, if CNPG is installed through Helm, it is by cloud admin by definition, so he puts proxies at the installation and, with this patch, everything is propagated silently.

If it is installed with OLM, at least on Openshift (i should test it on stock k8s to see what happens), cloud admin should have done a proxy configuration at cluster level, and we'll have also the propagation of values to the operators.

There shouldn't be any reconciliation on that (in my opinion), as env vars modification implicates a creation of new pods (those values aren't editable for pods, only for deployments, replica sets, ....... that create new pods).
So in case of a value change by cloud admins (unlikely, this'll impact everything that's doing an egress, so not a very light operation on a cloud), I think it should be at the pgcluster admin level to decide when to recreate their pods, as this is a delicate operation (much like an upgrade of postgres).

References:
- [OLM Proxy in GO](https://sdk.operatorframework.io/docs/building-operators/golang/references/proxy-vars/)
- [Proxy support for OLM in OpenShift](https://docs.openshift.com/container-platform/4.9/operators/admin/olm-configuring-proxy-support.html)

fyannk:
up, don't forget me :)

also I've digged a little more on OLM, and on stock Kubernetes as far as I can say we cannot have the same behaviour as on Openshift.

Hence a feature request opened on OLM side:

https://github.com/operator-framework/operator-lifecycle-manager/issues/2877

sxd:
@fyannk following the discussion in the last community meeting, we decided that this it's not the way we want to go and we prefer to have a more generic way to add environment variables since there's more environment variables that we may want to add, like LANG or LC.

PR #819, opened 2022-10-05
chore: improve archive WAL check

### LINKED ISSUE #818: improve the logic to check if an object storage already contains a WAL
Right now our logic for checking if any existing WAL exists in the storage is very error prone because we rely on the WAL numeration, we should improve it by:

- We stop looking at the WAL file name to trigger the check for an empty archive

- At the end of the init job add a flag file in the data directory. I.e., $PGDATA/.cnp-wal-never-archived

- If an instance manager sees the ConditionContinuousArchiving true, remove the flag file if exists

- At the start of the walarchive command, if the flag file exists, run the check for an empty archive

- If the walarchive command succeeds remove the flag file if exists

improvement proposal originally made by @mnencia  and @gbartolini

armru:
Closes #818

### DIFF
--- a/pkg/management/barman/archiver/archiver.go
+++ b/pkg/management/barman/archiver/archiver.go
@@ -23,15 +23,22 @@ import (
 	"math"
 	"os/exec"
 	"path"
+	"path/filepath"
 	"sync"
 	"time"
 
 	apiv1 "github.com/cloudnative-pg/cloudnative-pg/api/v1"
+	"github.com/cloudnative-pg/cloudnative-pg/pkg/fileutils"
 	"github.com/cloudnative-pg/cloudnative-pg/pkg/management/barman"
 	barmanCapabilities "github.com/cloudnative-pg/cloudnative-pg/pkg/management/barman/capabilities"
 	"github.com/cloudnative-pg/cloudnative-pg/pkg/management/barman/spool"
 	"github.com/cloudnative-pg/cloudnative-pg/pkg/management/execlog"
 	"github.com/cloudnative-pg/cloudnative-pg/pkg/management/log"
+)
+
+const (
+	// CheckWalArchiveFile the flag file that activates the wal archiver checks
+	CheckWalArchiveFile = ".check-wal-archive"
 )
 
 // WALArchiver is a structure containing every info need to archive a set of WAL files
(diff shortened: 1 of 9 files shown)


pkg/management/postgres/restore.go
```diff
@@ -654,7 +654,7 @@ func (info *InitInfo) checkBackupDestination(
 
 	// Instantiate the WALArchiver to get the proper configuration
 	var walArchiver *archiver.WALArchiver
-	walArchiver, err = archiver.New(ctx, cluster, env, walarchive.SpoolDirectory)
+	walArchiver, err = archiver.New(ctx, cluster, env, walarchive.SpoolDirectory, os.Getenv("PGDATA"))
```
mnencia, line 657: Pgdata is already in the info struct

pkg/management/barman/archiver/archiver.go
```diff
@@ -23,17 +23,24 @@ import (
...
 	"github.com/cloudnative-pg/cloudnative-pg/pkg/management/barman"
 	barmanCapabilities "github.com/cloudnative-pg/cloudnative-pg/pkg/management/barman/capabilities"
 	"github.com/cloudnative-pg/cloudnative-pg/pkg/management/barman/spool"
 	"github.com/cloudnative-pg/cloudnative-pg/pkg/management/execlog"
 	"github.com/cloudnative-pg/cloudnative-pg/pkg/management/log"
 )
 
+const (
+	// CheckWalArchiveFile the flag file that activates the wal archiver checks
+	CheckWalArchiveFile = ".check-wal-archive"
```
gbartolini, line 41:
```suggestion
	CheckWalArchiveFile = ".check-empty-wal-archive"
```

armru: done

sxd (approved):

pkg/management/postgres/restore.go
```diff
@@ -209,6 +210,14 @@ func (info InitInfo) restoreDataDir(backup *apiv1.Backup, env []string) error {
 		return err
 	}
 	log.Info("Restore completed")
+
+	filePath := filepath.Join(info.PgData, archiver.CheckEmptyWalArchiveFile)
+	// We create the check empty wal archive file to tell that we should check if the
+	// destination path it is empty
+	if err := fileutils.CreateEmptyFile(filepath.Clean(filePath)); err != nil {
```
armru, line 217: I think ConfigureNewInstance is invoked also inside the restore process, so maybe this is not needed here

sxd: fixed

gbartolini (approved):

mnencia (approved):

PR #828, opened 2022-10-07
feat: adds skipEmptyWalArchiveCheck annotation

### LINKED ISSUE #823: Add safety check for WAL archiving during the recovery phase
Extend the work done in #818 to the recovery phase. Ideally, we should provide a way to disable the safety check, either at configuration level (preferred) or through the plug-in?

armru:
Closes #823 

### DIFF
--- a/internal/cmd/manager/walarchive/cmd.go
+++ b/internal/cmd/manager/walarchive/cmd.go
@@ -165,8 +165,10 @@ func run(ctx context.Context, podName, pgData string, args []string, client clie
 	walFilesList := gatherWALFilesToArchive(ctx, walName, maxParallel)
 
 	// Step 4: Check if the archive location is safe to perform archiving
-	if err := checkWalArchive(ctx, cluster, walArchiver, client, pgData); err != nil {
-		return err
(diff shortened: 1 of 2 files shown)


armru:
if we decide to go with this approach we should add the relative documentation

internal/cmd/manager/walarchive/cmd.go
```diff
@@ -164,9 +164,11 @@ func run(ctx context.Context, podName, pgData string, args []string, client clie
 	// Step 3: gather the WAL files names to archive
 	walFilesList := gatherWALFilesToArchive(ctx, walName, maxParallel)
 
-	// Step 4: Check if the archive location is safe to perform archiving
-	if err := checkWalArchive(ctx, cluster, walArchiver, client, pgData); err != nil {
-		return err
+	if utils.IsEmptyWalArchiveCheckEnabled(&cluster.ObjectMeta) {
+		// Step 4: Check if the archive location is safe to perform archiving
```
mnencia, line 168: I would swap these 2 lines.

mnencia (approved):

jsilvela (approved):

PR #883, opened 2022-10-19
Clarify ScheduledBackup.spec.schedule format

alexpearce:
I found the documentation confusing as it claimed the `ScheduledBackup.spec.schedule` value should use "the same format used in Kubernetes CronJobs", which is the [unix-style five-field syntax](https://kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/#cron-schedule-syntax), however the implementation actually uses the [six-field syntax from the robfig/cron package](https://pkg.go.dev/github.com/robfig/cron#hdr-CRON_Expression_Format) where the additional fields denotes seconds. This PR clarifies the format.

(I discovered my `0 1 * * *` format was resulting in a backup every hour at 1 minute past the hour, rather than my intention of one backup each day at 1 am 😄)


sxd:
@alexpearce please run `make apidoc` and `make checks` on your branch and push the changes, I wasn't able to do it myself and push the changes to your branch

alexpearce:
> please run make apidoc and make checks on your branch and push the changes

Done! Thanks for the hint @sxd.

PR #930, opened 2022-10-25
chore: avoid setting seccompProfile when not supported

### LINKED ISSUE #928: Detect kubernetes version to decide if we should use the seccomprofile

sxd:
In Kubernetes version 1.24 and above we have SeccompProfile which it's one of the new security meassures we can use, but for older versions this it's not supported and we should be aware of.

Closes #928


pkg/specs/containers.go
```diff
@@ -59,6 +60,11 @@ func addManagerLoggingOptions(cluster apiv1.Cluster, container *corev1.Container
 func CreateContainerSecurityContext() *corev1.SecurityContext {
 	trueValue := true
 	falseValue := false
+	seccompProfile := &corev1.SeccompProfile{Type: corev1.SeccompProfileTypeRuntimeDefault}
+
+	if !utils.HaveSeccompSupport() {
```
jsilvela, line 65: nice, future-looking to put Seccomp active by "default"

jsilvela (approved):

NiccoloFei (approved):

armru (approved):

PR #931, opened 2022-10-26
feat: defaulting to jsonlog for pg15

### LINKED ISSUE #792: Adopt PostgreSQL 15 native jsonlog capability
With [this patch by Michael Paquier](https://git.postgresql.org/gitweb/?p=postgresql.git;a=commitdiff;h=dc686681e0799b12c40f44f85fc5bfd7fed4e57f), PostgreSQL 15 introduces direct support for JSON log.

CloudNativePG should set the default behaviour to match the capability introduced in version 15 and directly use in the configuration:

```ini
log_destination = jsonlog
```

For older versions, fall back to the current implementation which makes usage of CSV and parses the content.

This should be transparent for the end user.

phisco:
PostgreSQL 15 intruduced a new possible value for `log_destination`, `jsonlog`. This patch sets it by default for PG >= 15 and then introduces a new log pipe to parse logs from json instead of csv, producing the same output log format as before (pgaudit included).

Closes #792 

Signed-off-by: Philippe Scorsolini <p.scorsolini@gmail.com>

### DIFF
--- a/internal/cmd/manager/instance/run/cmd.go
+++ b/internal/cmd/manager/instance/run/cmd.go
@@ -139,12 +139,20 @@ func runSubCommand(ctx context.Context, instance *postgres.Instance) error {
 	postgresStartConditions = append(postgresStartConditions, reconciler.GetExecutedCondition())
 
 	// postgres CSV logs handler (PGAudit too)
-	postgresLogPipe := logpipe.NewLogPipe()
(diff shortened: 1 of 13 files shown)


phisco:
I'll have to take a look at the failing e2es more in depth

phisco:
We agreed this doesn't add anything w.r.t. the operator's current behavior, parsing JSON instead of CSV would only worsen the log pipeline performances without any pros.

PR #935, opened 2022-10-26
fix(hibernate): evaluate `pg_ctl status` exit code to determine if an instance is running

### LINKED ISSUE #934: improve how we determine if an instance is stopped during hibernate on
Right now during hibernation, we determine if an instance is stopped by doing a string match on pg_ctl: no server running but this could be not always true given that is dependant on the Postgres locale set.

We should improve the logic by relying on exit codes instead.

armru:
This patch ensures that we rely only on the exit code to determine if an instance is stopped. 

This removes the previous behavior of checking the pg_ctl message which generated errors with languages different than English

Closes #934

### DIFF
--- a/internal/plugin/resources/instance.go
+++ b/internal/plugin/resources/instance.go
@@ -19,21 +19,32 @@ package resources
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"time"
 
 	v1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/api/errors"
+	apierrs "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
+	"k8s.io/client-go/util/exec"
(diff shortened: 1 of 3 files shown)


internal/plugin/resources/instance.go
```diff
@@ -163,3 +173,35 @@ func getPVC(ctx context.Context, name string) (*v1.PersistentVolumeClaim, error)
...
+	timeout := time.Second * 2
+	clientInterface := kubernetes.NewForConfigOrDie(plugin.Config)
+	stdout, _, err := utils.ExecCommand(
+		ctx,
+		clientInterface,
+		plugin.Config,
+		pod,
+		specs.PostgresContainerName,
+		&timeout,
+		"/bin/bash", "-c", "pg_ctl status &>/dev/null; echo $?")
```
mnencia, line 191:
```suggestion
		"/bin/sh", "-c", "pg_ctl status >/dev/null 2>&1; echo $?")
```

mnencia: better, we are ignoring stderr anyway, so just remove the stderr redirection...

armru:
take extra care in not squashing

leonardoce (approved):

PR #944, opened 2022-10-28
feat: adds plugin install command

### LINKED ISSUE #938: Implement the `install` command in the `cnpg` plugin for `kubectl`
Right now, installation is primarily done in two methods:

- static release YAML manifest
- helm chart

We could leverage the `cnpg` plugin for `kubectl`  with the introduction of another command called `install`. In this first implementation, the plugin should simply provide a way to install the latest patch release for the latest minor release.

For example:

```
kubectl cnpg install
```

should return the content of the `https://raw.githubusercontent.com/cloudnative-pg/artifacts/RELEASE_BRANCH/manifests/operator-manifest.yaml` file at the time of the release of the plugin, where release is the latest patch release available for the latest minor.

Users can select another minor release with:

```
kubectl cnpg install --release-version 1.17
```

Additionally, we could specify:

- the namespace we would like the operator to be installed
- possibly which namespaces to control
- the number of instances for the operator deployment
- others to be determined in the analysis which might or not be included in this ticket, depending on the difficulty

armru:
Closes #938


internal/cmd/plugin/install/generate.go
```diff
@@ -0,0 +1,313 @@
...
+	for _, ir := range supportedResources {
+		ir := ir
+		err := machineryYaml.UnmarshalStrict(document, ir.obj)
+		if err != nil {
+			continue
+		}
+		return ir, nil
+	}
+
+	return installationResource{}, fmt.Errorf("could not parse the yaml document: \n %s", string(document))
```
armru, line 199: reminder: move the document on the logger

litaocdl (approved):

jlong49 (approved):

gbartolini (approved):

Issue #946, opened 2022-10-31
Seccomp detection is failing on EKS 1.21


leonardoce:
Sometimes the reported minor version info is not really an integer:

```
❯ kubectl version
Client Version: version.Info{Major:"1", Minor:"21", GitVersion:"v1.21.13", GitCommit:"80ec6572b15ee0ed2e6efa97a4dcd30f57e68224", GitTreeState:"clean", BuildDate:"2022-05-24T12:40:44Z", GoVersion:"go1.16.15", Compiler:"gc", Platform:"darwin/arm64"}
Server Version: version.Info{Major:"1", Minor:"21+", GitVersion:"v1.21.13-eks-84b4fe6", GitCommit:"e1318dce57b3e319a2e3fecf343677d1c4d4aa75", GitTreeState:"clean", BuildDate:"2022-06-09T18:22:07Z", GoVersion:"go1.16.15", Compiler:"gc", Platform:"linux/amd64"}
```

and this will prevent the operator from starting with the following error:

```
could not detect SeccompProfile support: strconv.Atoi: parsing "21+": invalid syntax
```

PR #947, opened 2022-10-31
fix: seccomp detection is failing when minor version is not an integer

leonardoce:
When the reported minor version is not an integer (i.e. `21+`) the seccomp detection is failing and will prevent the operator from starting correctly.

Closes: #946

This patch makes that function more resilient.

### DIFF
--- a/pkg/utils/discovery.go
+++ b/pkg/utils/discovery.go
@@ -33,11 +33,11 @@ var haveSCC bool
 // This variable specifies whether we should set the SeccompProfile or not in the pods
 var supportSeccomp bool
 
-// minorVersionRegexp is used to extract the minor version from
+// `minorVersionRegexp` is used to extract the minor version from
 // the Kubernetes API server version. Some providers, like AWS,
-// append a "+" to the Kubernetes minor version indicated that
-// there's some maintenance backport patch over the standard
-// release beyond its end-of-life.
+// append a "+" to the Kubernetes minor version to presumably
+// indicate that some maintenance patches have been back-ported
+// beyond the standard end-of-life of the release.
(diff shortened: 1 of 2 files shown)


jsilvela (approved):

jlong49 (approved):

pkg/utils/discovery.go
```diff
@@ -30,6 +33,13 @@ var haveSCC bool
 // This variable specifies whether we should set the SeccompProfile or not in the pods
 var supportSeccomp bool
 
+// minorVersionRegexp is used to extract the minor version from
+// the Kubernetes API server version. Some providers, like AWS,
+// append a "+" to the Kubernetes minor version indicated that
+// there's some maintenance backport patch over the standard
+// release beyond its end-of-life.
```
gbartolini, line 40:
```suggestion
// `minorVersionRegexp` is used to extract the minor version from
// the Kubernetes API server version. Some providers, like AWS,
// append a "+" to the Kubernetes minor version to presumably
// indicate that some maintenance patches have been back-ported
// beyond the standard end-of-life of the release.
```

pkg/utils/discovery.go
```diff
@@ -98,6 +108,18 @@ func HaveSeccompSupport() bool {
 	return supportSeccomp
 }
 
+// extractK8sMinorVersion extracts and parse the Kubernetes minor version from
+// the version info as detected by discovery client
```
gbartolini, line 112:
```suggestion
// extractK8sMinorVersion extracts and parses the Kubernetes minor version from
// the version info that's been  detected by discovery client
```

gbartolini (approved):

PR #967, opened 2022-11-08
Better support for Istio

### LINKED ISSUE #801: Better support for sidecar containers in batch jobs (aka Istio support)
CloudNativePG uses Kubernetes jobs to initialize the PGDATA content on the persistent volumes.

However, when used in conjunction with Istio, Kubernetes jobs do not terminate, as the injected sidecar running the Envoy based `pilot-agent` runs indefinitely. For details: https://medium.com/redbox-techblog/handling-istio-sidecars-in-kubernetes-jobs-c392661c4af7

Although there are a few options, including [annotations](https://istio.io/latest/docs/reference/config/annotations/), the most reasonable way forward is to introduce support for the `/quitquitquit` endpoint of the [Envoy server](https://www.envoyproxy.io/docs/envoy/latest/operations/admin).

Such support was introduced in [Istio 1.3](https://istio.io/latest/news/releases/1.3.x/announcing-1.3/change-notes/) - see https://github.com/istio/istio/issues/15041 for details.

This is also the recommended approach in this discussion: https://discuss.istio.io/t/best-practices-for-jobs/4968

hh24k:
1. Addressing the issue when the primary container requests the K8S API Server, while the Istio (Evoy) proxy is not ready yet, the API call will be refused and fail... it will cause the pod to restart.
2. Addressing the issue when the job's primary container is completed, while the pod which the Istio proxy sidecar has injected in, won't be marked as completed since the sidecar keeps running.

Closes #801 


.gitignore
```diff
@@ -1,4 +1,4 @@
-
+/vendor
```
mnencia, line 1: I do not think this change should be in this pull request.

hh24k: yeah, that was made for my local dev's convenience ... just removed it

sxd (approved):

leonardoce (changes requested):

leonardoce (approved):

leonardoce (approved):

armru (approved):

armru (approved):

leonardoce (approved):

Issue #975, opened 2022-11-09
Implement `list` command in `kubectl-cnpg`


till:
I found a `list` command to be very useful and I have some code that'll push to a PR next. This is basically a "simpler" way to do `kubectl get clusters.postgresql.cnpg.io`.

PR #976, opened 2022-11-09
Update: kubectl-cnpg list command

till:
For: #975
Signed-off-by: till <till@php.net>

### DIFF
--- a/internal/cmd/plugin/list/cmd.go
+++ b/internal/cmd/plugin/list/cmd.go
@@ -45,7 +45,7 @@ func NewCmd() *cobra.Command {
 	listCmd.PersistentFlags().BoolVarP(&allNamespaces,
 		"all-namespaces", "A", false, "Apply operation to all clusters in all namespaces")
 	listCmd.PersistentFlags().StringP(
-		"selector", "", "", "Label selector: label=value,label in (foo, bar)")
+		"selector", "l", "", "Label selector: label=value,label in (foo, bar)")
 	listCmd.Flags().StringP(
 		"output", "o", "text", "Output format. One of text|json")
 
--- a/internal/cmd/plugin/list/cmd.go
+++ b/internal/cmd/plugin/list/cmd.go
@@ -31,10 +31,9 @@ func NewCmd() *cobra.Command {
 	listCmd := &cobra.Command{
 		Use:   "list",
 		Short: "List all created PostgreSQL cluster",
-		// Args:  cobra.ExactArgs(1),
+		Args:  cobra.ExactArgs(0),
 		RunE: func(cmd *cobra.Command, args []string) error {
 			ctx := context.Background()
-			// clusterName := args[0]
 
 			output, _ := cmd.Flags().GetString("output")
 
--- a/internal/cmd/plugin/list/cmd.go
+++ b/internal/cmd/plugin/list/cmd.go
@@ -36,13 +36,16 @@ func NewCmd() *cobra.Command {
 			ctx := context.Background()
 
 			output, _ := cmd.Flags().GetString("output")
+			selector, _ := cmd.Flags().GetString("selector")
 
-			return List(ctx, allNamespaces, plugin.OutputFormat(output))
+			return List(ctx, allNamespaces, selector, plugin.OutputFormat(output))
 		},
 	}
 
 	listCmd.PersistentFlags().BoolVarP(&allNamespaces,
 		"all-namespaces", "A", false, "Apply operation to all clusters in all namespaces")
+	listCmd.PersistentFlags().StringP(
+		"selector", "", "", "Label selector: label=value,label in (foo, bar)")
 	listCmd.Flags().StringP(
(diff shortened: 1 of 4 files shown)


internal/cmd/plugin/list/cmd.go
```diff
@@ -0,0 +1,51 @@
...
+)
+
+// NewCmd create the new "status" subcommand
+func NewCmd() *cobra.Command {
+	var allNamespaces bool
+
+	listCmd := &cobra.Command{
+		Use:   "list",
+		Short: "List all created PostgreSQL cluster",
+		// Args:  cobra.ExactArgs(1),
```
phisco, line 34: Maybe set exactArgs(0) here

internal/cmd/plugin/list/cmd.go
```diff
@@ -0,0 +1,51 @@
...
+			ctx := context.Background()
+			// clusterName := args[0]
+
+			output, _ := cmd.Flags().GetString("output")
+
+			return List(ctx, allNamespaces, plugin.OutputFormat(output))
+		},
+	}
+
+	listCmd.PersistentFlags().BoolVarP(&allNamespaces,
```
phisco, line 45: What about a label selector too?

phisco (dismissed):

internal/cmd/plugin/list/cmd.go
till: I hadn't thought of that. But I can add that if you like.

till: I just had a look, so something like `kubectl-cnpg list --selector=app=tas-core`? Not sure how to do this with Cobra. :D I tried to look for an example, but don't see anything in this repo yet.

till: I think done. :D

till:
@phisco would you mind running the tests/linters? LMK what you think.

gbartolini:
Don't forget the docs @till ! ;)

till:
> Don't forget the docs @till ! ;)

Happy to, where do I add that?

till:
@gbartolini  Think I found it, hope that works.

till:
Is it possible to include this in the next release?

internal/cmd/plugin/list/cmd.go
```diff
@@ -0,0 +1,53 @@
...
+			selector, _ := cmd.Flags().GetString("selector")
+
+			return List(ctx, allNamespaces, selector, plugin.OutputFormat(output))
+		},
+	}
+
+	listCmd.PersistentFlags().BoolVarP(&allNamespaces,
+		"all-namespaces", "A", false, "Apply operation to all clusters in all namespaces")
+	listCmd.PersistentFlags().StringP(
+		"selector", "", "", "Label selector: label=value,label in (foo, bar)")
```
phisco, line 48:
```suggestion
		"selector", "l", "", "Label selector: label=value,label in (foo, bar)")
```
To be compliant with kubectl and other tools

phisco (changes requested):
Other

internal/cmd/plugin/list/list.go
```diff
@@ -0,0 +1,91 @@
...
+	clusters.AddHeader(
+		"Namespace",
+		"Cluster Name",
+		"Labels",
+	)
+
+	for _, item := range clusterList.Items {
+		var labelOut []string
+		for k, v := range item.Labels {
+			labelOut = append(labelOut, fmt.Sprintf("%s: %s", k, v))
```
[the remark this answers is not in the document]
till: For some reason it will not work, tried that before. It's like it doesn't honor the columns when I put a `\n` in there.

phisco:
Any more information we could add to the clusters's list? Maybe ready instances (x/z) and cluster phase? What do you think @gbartolini?

till:
I added phase and instances. Instances makes sense to me, I used the cluster status, phase, not entirely sure.

```
Available Clusters:
Namespace  Cluster Name  Instances  Phase               Labels
---------  ------------  ---------  -----               ------
db-test    db-test       1/1        Setting up primary  app: db-foobar-tralala
```

I just read up on the phases, and it looks like this is the "ready" state? I mean, I can use it, but the "phase" suggests it's doing something, but that's probably another ticket problem.

till:
@phisco @gbartolini Anything I can do to move this along? I understand that everyone is busy, but can you share next steps or something so I can manage on my end?

phisco:
Hi, sorry for the late reply, I wanted to discuss it with the rest of the team during today's bi-weekly meeting.
So, the general agreement is that this subcommand is not adding much with respect to what you can already get with a simple `kubectl get clusters`, therefore we wanted to understand your point of view here, @till, can you expand on what additional information you wanted to be able to see?

till:
I guess the only difference are the labels, but I am not invested enough right now to want them. You can still get them through `kubectl get clusters -A -o json` and a little jq. I think that works for me also.

Issue #986, opened 2022-11-10
Remove PostgreSQL 10 from supported versions


gbartolini:
As of Nov 10, 2022 Postgres 10 is no longer supported. See https://www.postgresql.org/about/news/postgresql-151-146-139-1213-1118-and-1023-released-2543/

Please remove it from the matrix of tests.

PR #1009, opened 2022-11-14
ci: remove PostgreSQL 10 from supported versions.

wadlejitendra:
closes : #986 

Signed-off-by: Jitendra Wadle <jitendra.wadle@enterprisedb.com>


docs/src/replication.md
```diff
@@ -59,9 +59,6 @@ CREATE USER streaming_replica WITH REPLICATION;
    -- NOSUPERUSER INHERIT NOCREATEROLE NOCREATEDB NOBYPASSRLS
 ```
 
-!!! Note
```
gbartolini, line 62: @wadlejitendra why don't we then remove the superuser code?

gbartolini: Given that the scope of this ticket is to remove support from version 1.19, please make sure that we remove the setting of superuser in streaming_replica which was required in Postgres 10.

gbartolini:
@wadlejitendra with this patch we officially remove support for PostgreSQL 10 from CloudNativePG version 1.19. So please remove anything that introduces exceptions in the code to work with Postgres 10.

armru (approved):

sxd (approved):

PR #1022, opened 2022-11-16
chore: replace `postgresql` with `cnpg.io/cluster` as service selector

### LINKED ISSUE #1021: remove deprecated selector inside specs/service.go file
Currently inside specs/service.go file we still use `"postgresql": cluster.Name` as a selector that is deprecated. 
We should change it and use the new instance selector `"cnpg.io/cluster"`

We should also consolidate the corev1.Port generation inside a small function instead of repeating the code

armru:
Closes #1021


armru:
do not squash

armru:
do not merge, this change could have unintended consequences.

We need to talk about this further

sxd (approved):

PR #1105, opened 2022-11-28
feat: customize labels and annotations for the service account

### LINKED ISSUE #1083: Customize labels and annotations for the service account
Based on the use case triggered in #756, we could solve the issue on a more general case - rather than having to document it. This would allow us to customize workload identities in public cloud providers such as EKS, AKS, and GKE.

We could add a new section in the cluster's `spec` to enable definition of custom labels and annotations for the service account that controls the cluster. This enables us to associate the service account to an IAM role or a workload identity. For example:

In Azure:

```
...
  serviceAccount:
    labels:
      azure.workload.identity/use: true
    annotations:
      azure.workload.identity/client-id: XXXXXXXXXXXX
```

or, in EKS:

```
...
  serviceAccount:
    annotations:
      eks.amazonaws.com/role-arn: arn:aws:iam::111122223333:role/my-role
```

or, in GKE:

```
...
  serviceAccount:
    annotations:
      iam.gke.io/gcp-service-account: GSA_NAME@GSA_PROJECT.iam.gserviceaccount.com
```

Please update the documentation accordingly, especially for the backup part.

gabriele-wolfox:
Closes #1083


gbartolini:
Shouldn't this be split into two separate commits/PRs @gabriele-wolfox ? You are refactoring how Metadata is passed to existing resources like the Pooler, IIUC.

leonardoce:
> Shouldn't this be split into two separate commits/PRs @gabriele-wolfox ? You are refactoring how Metadata is passed to existing resources like the Pooler, IIUC.

I think this was introduced by the review that @armru did on his work. We could definitely split it, I don't have strong opinions on that.
In my opinion we could even have just one commit: we didn't change the API of the `Pooler` resource.

docs/src/backup_recovery.md
```diff
@@ -89,6 +98,26 @@ The destination path can be any URL pointing to a folder where
 the instance can upload the WAL files, e.g.
 `s3://BUCKET_NAME/path/to/folder`.
 
+#### IAM Role for Service Account (IRSA)
+
+In order to use IRSA we need an `annotation` in the cluster's ServiceAccount.
```
gbartolini, line 103:
```suggestion
In order to use IRSA you need to set an `annotation` in the `ServiceAccount` of the Postgres cluster.
```

gbartolini (approved):

armru (approved):

gbartolini (approved):

Issue #1133, opened 2022-12-05
schedule e2e test is failing 


litaocdl:
I’m seeing an issue with the last two nightly builds. Both aborted after ~20 min, and failed on the E2E job generation step.
Run python .github/e2e-matrix-generator.py \
usage: e2e-matrix-generator.py [-h]
                               [-m {issue_comment,workflow_dispatch,schedule,pull_request,main,pull_request_target,push}]
                               [-l LIMIT]
e2e-matrix-generator.py: error: argument -m/--mode: invalid choice: ‘’ (choose from ‘issue_comment’, ‘workflow_dispatch’, ‘schedule’, ‘pull_request’, ‘main’, ‘pull_request_target’, ‘push’)
{‘KIND’: {‘min’: ‘1.22’, ‘max’: ‘’}}
Error: Process completed with exit code 2.

PR #1134, opened 2022-12-05
fix: Give default value for schedule workflow

litaocdl:
Give default value for schedule workflow

Closes: #1133 

### DIFF
--- a/.github/workflows/continuous-delivery.yml
+++ b/.github/workflows/continuous-delivery.yml
@@ -47,7 +47,9 @@ defaults:
 jobs:
   check_commenter:
     if: |
-      github.event_name == 'issue_comment' && startsWith(github.event.comment.body, '/test')
+      github.event_name == 'issue_comment' &&
+      github.event.issue.pull_request &&
+      startsWith(github.event.comment.body, '/test')
     name: Retrieve command
     runs-on: ubuntu-20.04
     outputs:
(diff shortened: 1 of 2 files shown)


.github/workflows/continuous-delivery.yml
```diff
@@ -498,7 +516,7 @@ jobs:
     runs-on: ubuntu-20.04
     env:
       # 'TEST_DEPTH' determines the matrix of E2E test cases that will be run.
-      TEST_DEPTH: ${{ needs.evaluate_options.outputs.test_level }}
+      TEST_DEPTH: ${{ needs.evaluate_options.outputs.depth }}
```
sxd, line 519: this it's test_level not depth, yeah I know that that confusing :P

litaocdl: Aha, OK, I see, we use different in python, Let me change back

litaocdl: seems from here, got it
```
// testDepthEnvVarName is the environment variable we expect the user to set
// to change the default test depth level
const testDepthEnvVarName = "TEST_DEPTH"
```

NiccoloFei: Fixed

NiccoloFei (approved):

hh24k (approved):

sxd (approved):

Issue #625, opened 2022-08-19
'FATAL:  data directory "/var/lib/postgresql/data/pgdata" has invalid permissions' when bootstrapping cluster from backup


9numbernine9:
Hello! 👋 

I've recently started experiment with Cloudnative-pg (installed via Helm) in my homelab setup, a three-node K3S cluster running Kubernetes `v1.24.3+k3s1` and using Longhorn `1.3.1` as my CSI . After destroying a cluster, I'm attempting to restore the cluster from an S3 backup but I've noticed this message in the logs of the `full-recovery` container:
```
{"level":"info","ts":1660872541.8728054,"msg":"Starting up instance","logging_pod":"mycluster-1","pgdata":"/var/lib/postgresql/data/pgdata"}
{"level":"info","ts":1660872541.8913906,"logger":"pg_ctl","msg":"waiting for server to start....2022-08-19 01:29:01.891 UTC [33] FATAL:  data directory \"/var/lib/postgresql/data/pgdata\" has invalid permissions","pipe":"stdout","logging_pod":"mycluster-1"}
{"level":"info","ts":1660872541.8915691,"logger":"pg_ctl","msg":"2022-08-19 01:29:01.891 UTC [33] DETAIL:  Permissions should be u=rwx (0700) or u=rwx,g=rx (0750).","pipe":"stdout","logging_pod":"mycluster-1"}
{"level":"info","ts":1660872541.9828606,"logger":"pg_ctl","msg":" stopped waiting","pipe":"stdout","logging_pod":"mycluster-1"}
{"level":"info","ts":1660872541.9829462,"logger":"pg_ctl","msg":"pg_ctl: could not start server","pipe":"stderr","logging_pod":"mycluster-1"}mycluster
```
It seems that the permissions of the `/var/lib/postgresql/data/pgdata` directory aren't being set correctly for some reason? The `bootstrap-controller` initContainer seems to work fine:
```
{"level":"info","ts":1660872533.6525264,"msg":"Installing the manager executable","destination":"/controller/manager","version":"1.16.1","build":{"Version":"1.16.1","Commit":"14600bab","Date":"2022-08-12"}}
{"level":"info","ts":1660872534.7739167,"msg":"Setting 0755 permissions"}
{"level":"info","ts":1660872534.7739878,"msg":"Bootstrap completed"}
```
But then when the `full-recovery` container starts is when things go awry.

The YAML specs for my cluster are as follows:
```
[...]

PR #1164, opened 2022-12-11
fix: permissions error on bootstrap

onedr0p:
Fixes: https://github.com/cloudnative-pg/cloudnative-pg/issues/625

I was able to confirm this fix is indeed working by forcing `0750` permissions on my PVC right after the bootstrap process started

Signed-off-by: Devin Buhl <onedr0p@users.noreply.github.com>

### DIFF
--- a/internal/cmd/manager/bootstrap/cmd.go
+++ b/internal/cmd/manager/bootstrap/cmd.go
@@ -44,7 +44,7 @@ func NewCmd() *cobra.Command {
 				panic(err)
 			}
 
-			log.Info("Setting 0755 permissions")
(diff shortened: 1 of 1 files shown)


internal/cmd/manager/bootstrap/cmd.go
```diff
@@ -45,7 +45,7 @@ func NewCmd() *cobra.Command {
 			}
 
 			log.Info("Setting 0755 permissions")
```
gbartolini, line 47:
```suggestion
			log.Info("Setting 0750 permissions")
```

gbartolini (approved):

PR #1205, opened 2022-12-15
fix: use correct size to update WAL PVC

### LINKED ISSUE #1190: WAL volume is resized without a user request
The operator automatically resizes the WAL volume on a new cluster (in the end it matches data volume size).

  storage:
    storageClass: ebs-gp3
    size: 16Gi
  walStorage:
    storageClass: ebs-gp3
    size: 8Gi

phisco:
Closes #1190 

Previously we were using the size of the main PVC to update the one dedicated to WAL storage. This led to "-wal" PVCs being created with the correct size at first, but then, in case the main PVC had a different size, to be resized to the main PVC's size.

Signed-off-by: Philippe Scorsolini <p.scorsolini@gmail.com>

### DIFF
--- a/api/v1/cluster_types.go
+++ b/api/v1/cluster_types.go
@@ -82,6 +82,10 @@ const (
 	// get the name of the pull secret
 	ClusterSecretSuffix = "-pull-secret"
 
+	// WalArchiveVolumeSuffix is the suffix appended to the instance name to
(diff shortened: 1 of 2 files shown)


jsilvela (approved):
👍

mnencia (approved):

PR #1213, opened 2022-12-16
fix: don't look for PodMonitor when not needed

### LINKED ISSUE #1212: Don't look for podmonitor if it's not enabled in the cluster
Currently we're looking for the PodMonitor resource even if it wasn't requested by the user, we should just not do it since it's not needed, this will avoid a lot of log messages and cpu cycles

sxd:
Since we know that the cluster doesn't have the option `enablePodMonitor` set to `true` there's no need to continue and ask every time to the Kubernetes API for this type of resource.

Closes #1212 

### DIFF
--- a/controllers/cluster_create.go
+++ b/controllers/cluster_create.go
@@ -771,12 +771,10 @@ func (r *ClusterReconciler) createOrPatchPodMonitor(ctx context.Context, cluster
 		return err
 	}
 
-	// If the PodMonitor CRD does not exist, and the cluster has monitoring enabled,
+	// If the PodMonitor CRD does not exist, but the cluster has monitoring enabled,
 	// the controller cannot do anything until the CRD is installed
-	if !havePodMonitorCRD {
-		if cluster.IsPodMonitorEnabled() {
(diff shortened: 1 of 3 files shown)


gbartolini (approved):

jsilvela (approved):

danishedb (approved):

danishedb (approved):

controllers/cluster_create.go
```diff
@@ -767,11 +767,20 @@ func (r *ClusterReconciler) createOrPatchPodMonitor(ctx context.Context, cluster
...
 	}
 
+	// If the PodMonitor does not exist then check in cluster if it is present, log the  warning.
+	// If it is not enabled in the cluster then no need to continue.
+	if !havePodMonitor {
+		if cluster.IsPodMonitorEnabled() {
+			contextLogger.Warning("Kind PodMonitor not detected, " +
+				"Please create after enabling PodMonitor in cluster")
+		}
+		return nil
```
jsilvela, line 781: this logic is completely wrong: if we don't have pod monitor, and it is enabled, log and do nothing?!!!

PR #1242, opened 2022-12-23
fix: ensure ExecCommand obeys timeout

jsilvela:
fixes the problem in the `utils.ExecCommand` where the timeout argument
was not being honored.
This patch uses the provided timeout to enrich the rest.Config, and to create
a timed context to pass to the newly available `StreamWithContext`.


pkg/utils/exec.go
```diff
@@ -86,6 +86,9 @@ func ExecCommand(
 		Stderr:    true,
 	}, scheme.ParameterCodec)
 
+	if timeout != nil {
+		config.Timeout = *timeout
+	}
```
mnencia, line 91: I think this should be merged with the if statement at https://github.com/cloudnative-pg/cloudnative-pg/pull/1242/files#diff-ce145a4bb4fe260559e08df65c268267e5b81e737b1fc4c46937dbf0930901adR78-R80

Moreover, I do not think blindly setting config.Timeout here is wise because it could affect the caller (the config is passed as a pointer)

NiccoloFei (approved):

sxd (approved):

PR #1247, opened 2022-12-26
test: enhance E2E test's observability

### LINKED ISSUE #1246: Adjust Ginkgo polling progress timings & cap the number of operator logs lines getting dumped
--poll-progress-after and --poll-progress-interval generate a lot of output lines during the E2E runs which makes skimming through output a lot harder. That output may be useful for finding specific corner cases. But, we need to find a compromise between readability and debug logs (the same applies to the Operator Logs we dump whenever a test fails):

- Set --poll-progress-after to a very high timeout that’s unlikely we hit unless there are some serious issues in a test (e.g 1200 seconds seems sensible)

- Increase --poll-progress-interval, it’s useless to cycle through the same error message every 30 seconds (e.g 150 seconds seems sensible)

- Change saveOperatorLogs() to print a limited number of operator log lines on failure (in the Ginkgo Output). The full logs will still be available in the file attached as an artifact to the workflow.

hh24k:
Closes #1246 

### DIFF
--- a/tests/e2e/suite_test.go
+++ b/tests/e2e/suite_test.go
@@ -107,7 +107,7 @@ var _ = SynchronizedAfterSuite(func() {
 //
 // along the way it parses the timestamps for convenience, BUT the lines
 // of output are not legal JSON
-func saveOperatorLogs(buf bytes.Buffer, specName string, output io.Writer) {
+func saveOperatorLogs(buf bytes.Buffer, specName string, output io.Writer, capLines int) {
 	scanner := bufio.NewScanner(&buf)
 	filename := "out/operator_logs_" + specName + ".log"
 	f, err := os.Create(filepath.Clean(filename))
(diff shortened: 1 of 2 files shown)


tests/e2e/suite_test.go
```diff
@@ -125,8 +125,11 @@ func saveOperatorLogs(buf bytes.Buffer, specName string, output io.Writer) {
 			fmt.Fprintln(output, "ERROR while closing file:", err)
 		}
 	}()
-	for scanner.Scan() {
+
+	lineIdx := 0
+	for scanner.Scan() && lineIdx < capLines {
```
mnencia, line 130: We should display the last lines, not the initial part

hh24k: Fixed ...

jsilvela (approved):

sxd (approved):

PR #1255, opened 2022-12-27
make cloudnative-ng compatible with k8s 1.18 to 1.20 version

ns-jhu:


sxd:
@ns-jhu you have conflicts on this patch and you're downgrading a lot of stuff including breaking the support of Kubernetes 1.25, have you actually tried this on Kubernetes 1.25 ?

ns-jhu:
unfortunately, this is the only way I can do to get it working in k8s version 1.18 to 1.20 in our production cluster.

PR #1307, opened 2023-01-06
chore: check for PostgreSQL version and block PostgreSQL 10

### LINKED ISSUE #1306: Block PostgreSQL 10 in version 1.19
At webhook level we have to stop accepting PostgreSQL 10

sxd:
Since version 1.19 we will not support PostgreSQL 10, and this will be blocked at webhook level.

Closes #1306


api/v1/cluster_webhook.go
```diff
@@ -864,6 +864,13 @@ func (r *Cluster) validateConfiguration() field.ErrorList {
 		// validateImageName function
 		return result
 	}
+	if psqlVersion <= 100000 {
+		result = append(result,
+			field.Invalid(
+				field.NewPath("spec", "imageName"),
+				r.Spec.ImageName,
+				"Invalid postgresql version. Versions 11 or above are supported"))
```
hh24k, line 872: How about using "Unsupported postgresql version" instead of "Invalid postgresql version" @sxd

hh24k (approved):

api/v1/cluster_webhook.go
```diff
@@ -864,6 +864,13 @@ func (r *Cluster) validateConfiguration() field.ErrorList {
 		// validateImageName function
 		return result
 	}
+	if psqlVersion <= 100000 {
```
jsilvela, line 867: there is a problem here. E.g. version 10.23 will be read as 100023
So we really need to check against 110000

jsilvela (approved):

gbartolini (approved):

PR #1310, opened 2023-01-06
chore: add two new fields to the plugin install generation process

### LINKED ISSUE #1309: Add new operator arguments to operator
Following this PR https://github.com/cloudnative-pg/cloudnative-pg/pull/843 we should add those new options to the generation of the install yaml file of the operator in the plugin

sxd:
We added to new fields to the command line:
* --log-field-level
* --log-field-timestamp

Both used to change the respective values in the operator deployment.

Closes #1309

### DIFF
--- a/internal/cmd/plugin/install/generate.go
+++ b/internal/cmd/plugin/install/generate.go
@@ -102,7 +102,7 @@ func newGenerateCmd() *cobra.Command {
 	cmd.Flags().Int32Var(
 		&replicas,
 		"replicas",
-		1,
+		0,
 		"Number of replicas in the deployment. Default is zero, meaning that no override is applied on the "+
 			"installation manifest (normally it is a single replica deployment)",
 	)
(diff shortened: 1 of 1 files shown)


internal/cmd/plugin/install/generate.go
```diff
@@ -98,7 +102,7 @@ func newGenerateCmd() *cobra.Command {
 	cmd.Flags().Int32Var(
 		&replicas,
 		"replicas",
-		0,
+		1,
```
jsilvela, line 105: hm, why are we doing this?

sxd: The default is 1, why not make it the default also in the options?

jsilvela (approved):

internal/cmd/plugin/install/generate.go
gbartolini: I think that's because we rely on the source manifest, which could come with a different value.

gbartolini, line 105:
```suggestion
		0,
```

I would not change this in this patch - assuming it makes sense to change it.

PR #1340, opened 2023-01-19
refactor: add declarative logic to reconcile PersistentVolumeClaim

### LINKED ISSUE #1321: add a declarative way  inside the codebase to reconcile PersistentVolumeClaims
Right now in our codebase, we have an imperative way to handle PVC reconciliation and this makes it hard to:
- add support for new PVCs
- reconcile the desired state without adding ad-hoc logic for every situation
- having a single point of responsibility that handles the PVC logic

We should start moving to a more declarative approach.

armru:
Closes #1321


api/v1/cluster_webhook.go
```diff
@@ -1190,6 +1190,10 @@ func (r *Cluster) validateWalStorageChange(old *Cluster) field.ErrorList {
 		return nil
 	}
 
+	if old.Spec.WalStorage == nil && r.Spec.WalStorage != nil {
+		return nil
+	}
+
```
mnencia, line 1196: This is caught by the previous if.

mnencia (approved):

gbartolini (approved):

Issue #677, opened 2022-08-31
Verify coherence of `min_wal_size` and `max_wal_size` with the size of `walStorage`


gbartolini:

gbartolini:
If the user has specified a separate volume for `walStorage`, verify the settings for `min_wal_size` and `max_wal_size` are coherent. See https://www.postgresql.org/docs/current/runtime-config-wal.html

Sonali-Rajput:
Can you please further explain where the changes should be made?

PR #1378, opened 2023-01-31
fix: verify max/min wal size setting with wal storageSize

litaocdl:
This patch verify the postgres settings max_wal_size and min_wal_size with 
wal storage volume size. make sure. 
 min_wal_size < max_wal_size < wal volume size
when min_wal_size and max_wal_size is not specified, default value will be considered

Close: #677  

### DIFF
--- a/api/v1/cluster_webhook.go
+++ b/api/v1/cluster_webhook.go
@@ -921,9 +921,10 @@ func validateWalSizeConfiguration(
 
 	var result field.ErrorList
 
-	minWalSize := postgresConfig.Parameters[minWalSizeKey]
+	minWalSize, hasMinWalSize := postgresConfig.Parameters[minWalSizeKey]
 	if minWalSize == "" {
 		minWalSize = minWalSizeDefault
+		hasMinWalSize = false
 	}
 	minWalSizeValue, err := parseWalSettingValue(minWalSize)
 	if err != nil {
@@ -935,9 +936,10 @@ func validateWalSizeConfiguration(
 				fmt.Sprintf("Invalid value for configuration parameter %s", minWalSizeKey)))
 	}
 
-	maxWalSize := postgresConfig.Parameters[maxWalSizeKey]
+	maxWalSize, hasMaxWalSize := postgresConfig.Parameters[maxWalSizeKey]
 	if maxWalSize == "" {
 		maxWalSize = maxWalSizeDefault
+		hasMaxWalSize = false
 	}
 	maxWalSizeValue, err := parseWalSettingValue(maxWalSize)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


jsilvela (approved):

mnencia:
I've changed several things:
1) We do not compare against the data directory anymore, which would be misleading because the purpose of the primary storage is to hold data, not to store WAL files.
2) I've fixed the parser to accept all the values accepted by Postgres
3) I've updated the tests accordingly

mnencia (approved):

mnencia:
I did more changes:
* Ignore the volume size validation if neither min_wal_size nor max_wal_size is specified
* Fix parsing of the 'kB' unit
* Added specific unit test for the `parseWalSettingValue` function

api/v1/cluster_webhook.go
```diff
@@ -904,6 +908,122 @@ func (r *Cluster) validateConfiguration() field.ErrorList {
...
+			field.Invalid(
+				field.NewPath("spec", "postgresql", "parameters", minWalSizeKey),
+				minWalSize,
+				fmt.Sprintf("Invalid configuration. Parameter %s (default %s) should be smaller than WAL volume size",
+					minWalSizeKey, minWalSizeDefault)))
+	}
+
+	if hasMaxWalSize &&
+		!maxWalSizeValue.IsZero() &&
+		maxWalSizeValue.Cmp(*walVolumeSize) >= 0 {
```
litaocdl, line 983: should remove = in compare `maxWalSizeValue.Cmp(*walVolumeSize) > 0`, so the default storageSize 1Gi is not broken ?

api/v1/cluster_webhook.go
```diff
@@ -895,6 +895,10 @@ func (r *Cluster) validateConfiguration() field.ErrorList {
 		}
 	}
 
+	// verify the postgres setting min_wal_size < max_wal_size < volume size
+	result = append(result, validateWalSizeConfiguration(
+		r.Spec.PostgresConfiguration, r.Spec.WalStorage.GetSizeOrNil())...)
+
```
litaocdl, line 901: if user did not specific walStorage, do we need to consider using the size of storage like

	if r.ShouldCreateWalArchiveVolume() {
		walVolumeSize = r.Spec.WalStorage.GetSizeOrNil()
	} else {
		walVolumeSize = r.Spec.StorageConfiguration.GetSizeOrNil()
	}

api/v1/cluster_webhook.go
mnencia: I considered this, but that will give the user a false sense of correctness. Given that it is not a hard limit in Postgres space usage, setting the maximum equal to the volume size is not correct.

api/v1/cluster_webhook.go
mnencia: For the same reason as the previous comment, that would give the user a false sense of correctness. The storage area is meant to keep the data, not the WAL files. We can only validate the user input if a walStorage section is present.

armru (approved):

PR #1408, opened 2023-02-07
fix: do not run CHECKPOINT as streaming_replica user

### LINKED ISSUE #1407: Creating a new `Backup` with `immediate` `true` fails because the `streaming_replica` user doesn't have rights to do `CHECKPOINT`
If I create a new Backup with `immediate` `true` it fails because the `streaming_replica` user doesn't have rights to do `CHECKPOINT`, see the following logs:

```code
{"level":"info","ts":"2023-02-07T11:11:03Z","msg":"WAL archiving is not working","logging_pod":"cluster-name-1","err":"error while requiring a checkpoint: ERROR: must be superuser or have privileges of pg_checkpoint to do CHECKPOINT (SQLSTATE 42501)"}
```

mnencia:
Closes #1407

### DIFF
--- a/pkg/management/postgres/wal.go
+++ b/pkg/management/postgres/wal.go
@@ -34,6 +34,26 @@ type walArchiveAnalyzer struct {
 	dbFactory func() (*sql.DB, error)
 }
 
+// ensureWalArchiveIsWorking behave slightly differently when executed on primary or a standby.
+// On primary, it could run very early, when the first WAL has never completed. For this reason it
+// could require a WAL switch, to quicken the check.
+// On standby, the mere existence of the standby guarantee that a WAL file has already been generated
+// by the pg_basebakup used to prime the standby data directory, so we check only if the WAL
+// archive process is not failing.
+func ensureWalArchiveIsWorking(instance *Instance) error {
+	isPrimary, err := instance.IsPrimary()
+	if err != nil {
+		return err
+	}
+
+	if isPrimary {
+		return newWalArchiveBootstrapperForPrimary().ensureFirstWalArchived(retryUntilWalArchiveWorking)
+	}
+
+	return newWalArchiveAnalyzerForReplicaInstance(instance.GetPrimaryConnInfo()).
+		mustHaveFirstWalArchivedWithBackoff(retryUntilWalArchiveWorking)
+}
+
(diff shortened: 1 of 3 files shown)


armru (approved):

pkg/management/postgres/wal.go
```diff
@@ -28,63 +28,41 @@ import (
...
+func (w *walArchiveAnalyzer) execute() error {
+	return retry.OnError(w.backoff, resources.RetryAlways, func() error {
+		db, err := w.dbFactory()
+		if err != nil {
+			return err
 		}
-		return db, nil
+		return w.mustHaveFirstWalArchived(db)
 	})
 }
```
phisco, line 63: this connection is never closed, is that intentional? @armru

armru: yes this was intentional in the old implementation, but I've streamlined the logic further

pkg/management/postgres/wal.go
```diff
@@ -97,20 +98,74 @@ func (w *walArchiveBootstrapper) tryBootstrapWal() error {
...
-	if w.firstWalArchiveTriggered {
-		log.Info("Waiting for the first WAL file to be archived")
-		return errors.New("waiting for first wal-archive")
+	return errNoWalArchivePresent
+}
+
+type walArchiveBootstrapper struct {
+	walArchiveAnalyzer
+	createdFirstWal bool
+}
```
jsilvela, line 116: I think this could use some documentation

pkg/management/postgres/wal.go
```diff
@@ -97,20 +98,74 @@ func (w *walArchiveBootstrapper) tryBootstrapWal() error {
...
+				log.Debug("Error while closing connection", "err", closeErr.Error())
+			}
+		}()
+
+		err = w.mustHaveFirstWalArchived(db)
+		if !errors.Is(err, errNoWalArchivePresent) {
+			return err
+		}
+
+		if w.createdFirstWal {
```
jsilvela, line 157: I think this is very wrong...
And, I don't see where it gets set (other than in test)

jsilvela: I think that in the shipWalFile method, we should set `w.createdFirstWal` to true, no?

jsilvela (approved):
Looks good, but have found the code difficult to parse initially.
We should open a new ticket to simplify.
And, just a personal thing, but `dbFactory` unsettles me.

PR #1432, opened 2023-02-12
doc: extend the architecture page covering the Kubernetes layer

### LINKED ISSUE #1431: Integrate the "Architecture" page covering the Kubernetes cluster
The architecture page implies that the Kubernetes cluster extends of multiple availability zones. While this is the case in the cloud, we cannot take it for granted in the self-managed space. I personally see cases where customers install two K8s clusters in the same building - which makes me think of a lift-and-shift operation from the traditional world, forcing our database users to solely rely on the replica cluster feature, losing all the capabilities of our operator such as self-healing, failover, rolling updates, and so on.

Briefly describe the problem, and state our position that - as PostgreSQL experts - we highly recommend planning for a 3 data centre spanning Kubernetes cluster in the first place. This is more for posterity, given that for some who have already made this decision, this might come too late.

gbartolini:
Closes #1431

Signed-off-by: Gabriele Bartolini <gabriele.bartolini@enterprisedb.com>


leonardoce (approved):

armru (approved):

docs/src/architecture.md
```diff
@@ -1,20 +1,138 @@
...
+can only work within a single Kubernetes cluster.
+
+## PostgreSQL architecture
+
 CloudNativePG supports clusters based on asynchronous and synchronous
 streaming replication to manage multiple hot standby replicas within the same
 Kubernetes cluster, with the following specifications:
 
-* One primary, with optional multiple hot standby replicas for High Availability
+* One primary, with optional multiple hot standby replicas for HA
```
sxd, line 135: Why not keep both terms? Some people may not understand HA as high availability

docs/src/architecture.md
```diff
@@ -85,18 +203,13 @@ In a distributed PostgreSQL cluster there can only be a single PostgreSQL
...
-    If you are interested in a PostgreSQL architecture where all instances accept writes, 
-    please take a look at  [EDB Distributed Postgres by EDB](https://www.enterprisedb.com/docs/pgd/latest/). 
-    For Kubernetes, EDB Distributed Postgres will have its own Operator, expected later in 2022.
-
 However, for business continuity objectives it is fundamental to:
 
 - reduce global **recovery point objectives** (RPO) by storing PostgreSQL backup data
   in multiple locations, regions and possibly using different providers
-  (**Disaster Recovery**)
+  (DR)
```
sxd, line 210: Using DR has always been a bit strange for me that I'm not used to this terminology, i think it's better explain since the operator it's going to be used for people that may not be used to this term

docs/src/architecture.md
```diff
@@ -1,30 +1,166 @@
...
+
+PostgreSQL is a database management system and, as such, it needs to be treated
+as a **stateful workload** in Kubernetes. While, for High Availability (HA) and
+Disaster Recovery (DR) purposes, stateless applications primarily take care of
+traffic redirection, in the case of a database its state must be replicated in
+multiple locations, preferably in a continuous and instantaneous way, by adopting
+either of the following two strategies:
+
+* *storage-level replication*, normally persistent volumes
+* *application-level replication*, in this specific case PostgreSQL
```
jsilvela, line 21: here I want to check with you what we mean to express.
As it stands, seems to me a tiny overview of replication in Kubernetes for stateful vs stateless.

docs/src/architecture.md
```diff
@@ -1,30 +1,166 @@
...
+
+The multi-availability zone Kubernetes architecture, having three (3) or more
+zones, is the one that we recommend for PostgreSQL usage.
+This scenario is typical of Kubernetes services managed by Cloud Providers.
+
+![Kubernetes cluster spanning over 3 independent data centers](./images/k8s-architecture-3-az.png)
+
+Such an architecture enables the CloudNativePG operator to control the full
+lifecycle of a `Cluster` resource within a single Kubernetes cluster, by
+treating all the availability zones as active: this includes, among the others,
```
jsilvela, line 72: want to double-check what we want to convey with "zones as active"

docs/src/architecture.md
```diff
@@ -1,30 +1,166 @@
...
+even though failovers and promotions in this case must be done manually.
+
+![Example of a multiple Kubernetes cluster architecture distributed over 3 regions each with 3 independent data centers](./images/k8s-architecture-multi.png)
+
+### Single availability zone Kubernetes clusters
+
+In case your Kubernetes cluster has only one availability zone, CloudNativePG
+still provides you with a lot of features to improve HA and DR results for your
+PostgreSQL databases, by pushing out the single point of failure (SPoF) to the
+level of the zone as much as possible.
```
jsilvela, line 94: not sure if this is a bit technical for what we want to express

docs/src/architecture.md
```diff
@@ -1,30 +1,166 @@
...
+section below for details on how you can design your PostgreSQL clusters within
+your single availability zone Kubernetes cluster through shared-nothing
+deployments at the storage and worker node levels only. For HA, in such a
+scenario it becomes even more important that the PostgreSQL instances are
+located on different worker nodes and do not share the same storage.
+
+For DR, you can push the SPoF out by taking advantage of additional
+[Kubernetes clusters](#deployments-across-kubernetes-clusters), and use them to
+host "passive" PostgreSQL replica clusters. As with other Kubernetes workloads in
+this scenario, promotion of a Kubernetes cluster as primary must be done
```
jsilvela, line 131: promotion of a CloudNativePG cluster as primary (not Kubernetes cluster) ... right?

docs/src/architecture.md
gbartolini: The point here is to clarify that normally state is replicated via storage. For PostgreSQL we should avoid volume replication, and recommend application level replication. I have noticed that from an architecture's point of view, we take it for granted, but then, decisions about storage have already been taken and cannot be changed. The goal here is to make users aware of that - and it is highlighted later in the doc when we mention shared nothing architectures.

docs/src/architecture.md
gbartolini: It is primarily from an "application point of view". With the active/passive scenario that some large Kubernetes providers suggest when facing a 2 data centre infrastructure, only one Kubernetes cluster can be considered active at any time - even though the other one could be for example replicating data using Postgres replica clusters to "pre-heat" the database cluster having the designated primary ready to be promoted. However, applications are redirected only to the first Kubernetes cluster.

I am sharing some links here:

- https://cloud.redhat.com/blog/a-guide-to-high-availability/disaster-recovery-for-applications-on-openshift
- https://www.ibm.com/cloud/architecture/articles/hadr-containers/hadr-containers-3-two-clusters-active-active-passive/

Thoughts?

docs/src/architecture.md
gbartolini: I think this is actually crucial in terms of risk management - which this page ultimately is about (without mentioning it). The SPoF is definitely the data center. However, if users end up placing all Postgres instances on the same storage, for example, that becomes the SPoF for that Postgres database.

docs/src/architecture.md
gbartolini: Nope, this is exactly the use case that I am trying to cover with this page, considering how frequent we are seeing 2 data centre setups. The only possibility to have Postgres running as primary on the other Kubernetes cluster in a 2 data centre setup with two separate K8s clusters, is to promote the passive one as primary and redirect users to that K8s cluster. The aftermath for our operator is that users will be forced to manually promote the designated primary, while local applications will talk to it. This is primarily a DR operation, which means that the other K8s cluster will need heavy manual intervention, including restoring all the replica clusters.

This is very very bad and time consuming on the operational side and we should warn users from Day 0, when the K8s infrastructure is designed. I hope this clarifies.

jsilvela (approved):

docs/src/architecture.md
jsilvela: makes sense. I've been adding a bit more verbosity to clarify some of the more technical wording.

PR #1524, opened 2023-02-21
feat: add role management to CRD

jsilvela:


api/v1/cluster_types.go
```diff
@@ -1559,6 +1561,63 @@ func (in ExternalCluster) GetServerName() string {
...
+	Replication bool `json:"replication,omitempty"`
+	BypassRLS   bool `json:"bypassrls,omitempty"` // Row-Level Security
+
+	// connection Limit defaults to `-1`
+	// +kubebuilder:default:=-1
+	ConnectionLimit int64 `json:"connectionLimit,omitempty"`
+
+	ValidUntil string   `json:"validUntil,omitempty"`
+	InRoles    []string `json:"in_role,omitempty"`
+	WithRoles  []string `json:"role,omitempty"`
```
mnencia, line 1613: We should change this to be only `Roles` in place of `InRoles`. The `WithRoles` is redundant.

```suggestion
	Roles    []string `json:"role,omitempty"`
```

jsilvela: As per discussion, deleted `WithRoles` and updated the JSON for InRoles `in_role` -> `inRoles`

litaocdl:
rebase against main

docs/src/declarative_role_management.md
```diff
@@ -0,0 +1,61 @@
+# Database Role Management
+
+From its inception, CloudNativePG has managed the creation of specific roles required in PostgreSQL instances:
+
+- The superuser `postgres` and the `streaming_replica` role
```
gbartolini, line 5: What about `cnpg_pooler_pgbouncer` - when requested?

NiccoloFei (approved):

PR #1533, opened 2023-02-22
fix: backup_controller checks if cluster.spec.backup section is present

### LINKED ISSUE #1531: adding a backup resource when cluster.spec.Backup is not defined causes a panic
the panic happens because of this line:
https://github.com/cloudnative-pg/cloudnative-pg/blob/56c5f4cc4122be874bf61422a53bfffd6d9f97ec/controllers/backup_controller.go#L213

We should proceed only if the backup struct exists

armru:
Closes #1531


mnencia (approved):

armru:
for the backup status bug, see: #1535 #1536

sxd (approved):

PR #1536, opened 2023-02-22
fix: update the backup status in case of errors

### LINKED ISSUE #1535: backup_controller should update the backup status in case of errors
Example: https://github.com/cloudnative-pg/cloudnative-pg/blob/8d025bc733320924f3863d7b9aee394e6d0bc7dd/controllers/backup_controller.go#L117

There are multiple cases where we change the backup status without then committing the changes to the API server.

We should proceed to update the resource after modifying it

armru:
This patch ensure that each time when the backup encountered some error, the failed status
and reason are committed to the api server before next reconcile loop.

Closes #1535

### DIFF
--- a/controllers/backup_controller.go
+++ b/controllers/backup_controller.go
@@ -18,6 +18,7 @@ package controllers
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"time"
 
@@ -78,14 +79,9 @@ func NewBackupReconciler(mgr manager.Manager) *BackupReconciler {
 // +kubebuilder:rbac:groups="",resources=pods,verbs=get
 
 // Reconcile is the main reconciliation loop
-func (r *BackupReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+func (r *BackupReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) { // nolint: gocognit
 	contextLogger, ctx := log.SetupLogger(ctx)
-
 	contextLogger.Debug(fmt.Sprintf("reconciling object %#q", req.NamespacedName))
-
-	defer func() {
-		contextLogger.Debug(fmt.Sprintf("object %#q has been reconciled", req.NamespacedName))
-	}()
 
 	var backup apiv1.Backup
 	if err := r.Get(ctx, req.NamespacedName, &backup); err != nil {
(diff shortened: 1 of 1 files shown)


armru:
Please note that this PR requires #1531 to be merged first

controllers/backup_controller.go
```diff
@@ -126,6 +122,7 @@ func (r *BackupReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 			clusterName)
 		contextLogger.Warning(message)
 		r.Recorder.Event(&backup, "Warning", "ClusterHasNoBackupConfig", message)
+		tryFlagBackupAsFailed(ctx, r.Client, &backup, errors.New(message))
```
mnencia, line 125: This conflicts with the RequeueAfter: the failed status is final and will never be reconciled again.

controllers/backup_controller.go
```diff
@@ -200,6 +197,8 @@ func (r *BackupReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 		r.Recorder.Eventf(&backup, "Warning", "Error", "Backup exit with error %v", err)
 	}
 
+	contextLogger.Debug(fmt.Sprintf("object %#q has been reconciled", req.NamespacedName))
+
```
mnencia, line 201: I would add `err` to the log. It could be != nil

controllers/backup_controller.go
```diff
@@ -142,7 +139,7 @@ func (r *BackupReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 			backup.Status.Phase = apiv1.BackupPhasePending
 			return ctrl.Result{RequeueAfter: 30 * time.Second}, r.Status().Update(ctx, &backup)
 		}
-		backup.Status.SetAsFailed(fmt.Errorf("while getting pod: %w", err))
+		tryFlagBackupAsFailed(ctx, r.Client, &backup, fmt.Errorf("while getting pod: %w", err))
 		r.Recorder.Eventf(&backup, "Warning", "FindingPod", "Error getting target pod: %s",
 			cluster.Status.TargetPrimary)
 		return ctrl.Result{}, r.Status().Update(ctx, &backup)
```
mnencia, line 145: I don't think returning an error here makes any sense. Since the backup is marked as failed, it will never rerun a reconciliation cycle. Better to log some meaningful message and `return ctrl.Result{}, nil`

controllers/backup_controller.go
```diff
@@ -142,7 +142,7 @@ func (r *BackupReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 		tryFlagBackupAsFailed(ctx, r.Client, &backup, fmt.Errorf("while getting pod: %w", err))
 		r.Recorder.Eventf(&backup, "Warning", "FindingPod", "Error getting target pod: %s",
 			cluster.Status.TargetPrimary)
-		return ctrl.Result{}, r.Status().Update(ctx, &backup)
+		return ctrl.Result{}, nil
```
mnencia, line 145: I think you miss the `r.Status().Update(ctx, &backup)` call here

armru: tryFlagBackupAsFailed does that

mnencia (approved):

litaocdl (approved):

PR #1605, opened 2023-03-01
chore(renovate): fix GOLANG_VERSION and GOLANGCI_LINT

phisco:


mnencia (approved):

internal/cmd/manager/controller/controller.go
```diff
@@ -41,7 +42,6 @@ import (
 	"github.com/cloudnative-pg/cloudnative-pg/pkg/multicache"
 	"github.com/cloudnative-pg/cloudnative-pg/pkg/utils"
 	"github.com/cloudnative-pg/cloudnative-pg/pkg/versions"
-	// +kubebuilder:scaffold:imports
```
sxd, line 44: This requires the new version of golangci right? So everyone should be running the new version ?

sxd (approved):

internal/cmd/manager/controller/controller.go
mnencia: No, it doesn't. It's also suitable for the old version.

sxd: Nice!! Then let's merge it :)

mnencia:
@phisco has more changes to do (we do not want the full versiond but only Major.Minor)

PR #1650, opened 2023-03-03
feat: execute barman backup with `--name` flag when supported

### LINKED ISSUE #1523: Refactoring of the backup command code
The current implementation of the backup code is quite complex and the result of regular addition of code over time. With backup from a standby, for example, we have introduced a few bugs that were only noticed when BigAnimal went in production.

It is also difficult for new developers to start working on this part of the code. It is a general opinion in the team that a refactoring exercise is needed to reduce the cognitive load required to improve and maintain it in the future.

We should start by removing the duplicated code regarding the conditions and the defer of LastFailedBackup

Issue made by @armru @mnencia @gbartolini

armru:
TODO: proper explanation 

Do not merge without a proper commit message

Partially Closes #1523


mnencia (approved):

jlong49 (approved):

jlong49:
New pull request

https://github.com/cloudnative-pg/cloudnative-pg/pull/1704

pkg/management/barman/capabilities/type.go
```diff
@@ -34,3 +38,12 @@ type Capabilities struct {
 	HasAzureManagedIdentity    bool
 	Version                    *semver.Version
 }
+
+// ShouldExecuteBackupWithName returns true if the new backup logic should be executed
+func (c *Capabilities) ShouldExecuteBackupWithName(cluster *apiv1.Cluster) bool {
+	if !c.hasName || cluster == nil {
+		return c.hasName
+	}
```
jsilvela, line 46: IMO pretty obscure.
we could have `c.hasName` be true, then if `cluster == nil` we return ... true.
Seems wrong, no?

api/v1/cluster_types.go
```diff
@@ -2280,6 +2280,14 @@ func (cluster *Cluster) SetInheritedDataAndOwnership(obj *metav1.ObjectMeta) {
...
 }
 
+// ShouldForceLegacyBackup if present takes a backup without passing the name argument even on barman version 3.3.0+.
+// This is needed to test both backup system in the E2E suite
+func (cluster *Cluster) ShouldForceLegacyBackup() bool {
+	const legacyBackupAnnotationName = "cnpg.io/forceLegacyBackup"
+
+	return cluster.Annotations[legacyBackupAnnotationName] == "true"
+}
+
```
jsilvela, line 2290: never called outside of the E2E... so the E2E will respect the "legacy" Annotation, but the operator ignores it

leonardoce (approved):

PR #1665, opened 2023-03-04
fix: affinity.nodeAffinity not properly propagated to pods

### LINKED ISSUE #1662: nodeAffinity is ignored
`nodeAffinity`  is supported in the `Cluster` CRD but it is not applied to pods when no additional affinity configs are defined.

itay-grudev:
Bug Fix: When a user only specifies `affinity.nodeAffinity` no `affinity` section is set in database pods.

This is because of an early return from the `CreateAffinitySection` section that checks whether none of `AdditionalPodAffinity` or `AdditionalPodAntiAffinity` is set, but it doesn't do the same for `NodeAffinity`.

This patch adds an extra spec for this case and the fix to the early return in `pkg/specs/pods.go#CreateAffinitySection`.

P.S. I don't have much experience with Go and I am not sure if I've matched your code-style 100% as I don't completely understand it. Feel free to directly apply changes.

Closes #1662

### DIFF
--- a/pkg/specs/pods.go
+++ b/pkg/specs/pods.go
@@ -221,7 +221,8 @@ func CreateAffinitySection(clusterName string, config apiv1.AffinityConfiguratio
 	affinity := CreateGeneratedAntiAffinity(clusterName, config)
 
 	if config.AdditionalPodAffinity == nil &&
-		config.AdditionalPodAntiAffinity == nil {
+		config.AdditionalPodAntiAffinity == nil &&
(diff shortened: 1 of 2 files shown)


sxd:
@itay-grudev we're missing the DCO here, can you please fix that ? just follow the instructions in the failed test

sxd:
Closed in favor of https://github.com/cloudnative-pg/cloudnative-pg/pull/1663 due to being the first one on creating the PR

PR #1704, opened 2023-03-14
fix: pass an unique name to barman --name flag

jlong49:
Partially Closes https://github.com/cloudnative-pg/cloudnative-pg/issues/1523


armru:
this should be merged in dev/1523-4

PR #1755, opened 2023-03-19
feat: configurable creation of the `-any` service

### LINKED ISSUE #1788: Make the `-any` service optional
The `-any` service was a leftover of an old implementation that isn't too useful by now, the idea it's to make it optional or just delete it

leonardoce:
The creation of the `-any` service can be opted-out via the `createAnyService` configuration parameter, to be used inside the operator Secret or as an environment variable.

Closes #1788


sxd (approved):

mnencia:
I think we should make optional also the `-ro` and `-r` services

litaocdl (approved):

PR #1767, opened 2023-03-20
fix: fix package name for .tar.gz files

### LINKED ISSUE #1766: GoReleaser  upgrade broke tarball archiving
After upgrading to GoReleaser 1.15.0 we removed the `_version_` in the
package name for the tarball files, this broke the install script and
also the Arkade app that was relying on this name to download the archive.

sxd:
After upgrading to GoReleaser 1.15.0 we removed the `_version_` in the package name for the tarball files, this broke the install script and also the Arkade app that was relying on this name to download the archive.

Closes #1766


litaocdl (approved):

jsilvela:
CI run building the plugin https://github.com/cloudnative-pg/cloudnative-pg/actions/runs/4476716175

jsilvela (approved):

Issue #1779, opened 2023-03-21
Add code verification during auto backport


litaocdl:
To avoid sometimes crd is not up-to-date in release branches.

PR #1780, opened 2023-03-21
ci: add code verification in auto backport

litaocdl:
Closes: #1779


.github/workflows/backport.yml
```diff
@@ -103,6 +103,12 @@ jobs:
           git config user.name "${{ env.AUTHOR_NAME }}"
           git fetch
           git cherry-pick -x --mainline 1 ${{ env.COMMIT }}
+          make fmt vet generate apidoc wordlist-ordered
+          if ! git diff --exit-code --quiet
```
sxd, line 107: The option `--quiet` implies `--exit-code` so I don't think it's needed, but it could be nice to have since  it helps to make it clear what we're expecting from this exit code

sxd (approved):

NiccoloFei (approved):

PR #1791, opened 2023-03-22
docs: labels propagation fix

revolunet:
Looks like label addition are not propagated too (using 0.16.1)


gbartolini:
Mhhh ... if that's the case it is a bug and an issue should be created with instructions on how to reproduce the issue.

I checked the code and it is actually reconciling them. See:

https://github.com/cloudnative-pg/cloudnative-pg/blob/main/controllers/cluster_controller.go#L431

I am closing this pull request as - again - we will not update the documentation given that the desired behaviour is for them to be reconciled.

Issue #1373, opened 2023-01-30
Error while bootstrapping data directory


Sapp00:
Hi there,
when deploying a cluster, I receive this error all the time and cannot sucessfully create one. 

Cloudnative-PG is installed using the current helm chart:
```
---
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization
namespace: postgres

helmCharts:
  - name: cloudnative-pg
    repo: https://cloudnative-pg.github.io/charts
    version: 1.18.1
    releaseName: cnpg
    namespace: postgres
```

Full log (cluster-example-1-initdb-<uid>):
```
Defaulted container "initdb" out of: initdb, bootstrap-controller (init)
{"level":"info","ts":1675094246.6801445,"msg":"Creating new data directory","logging_pod":"cluster-example-1","pgdata":"/var/lib/postgresql/data/pgdata","initDbOptions":["--username","postgres","-D","/var/lib/postgresql/data/pgdata","--encoding=UTF8","--lc-collate=C","--lc-ctype=C"]}
{"level":"info","ts":1675094247.1110559,"logger":"initdb","msg":"The files belonging to this database system will be owned by user \"postgres\".
This user must also own the server process.

The database cluster will be initialized with this locale configuration:
  provider:    libc
  LC_COLLATE:  C
  LC_CTYPE:    C
  LC_MESSAGES: en_US.utf8
  LC_MONETARY: en_US.utf8
  LC_NUMERIC:  en_US.utf8
  LC_TIME:     en_US.utf8
The default text search configuration will be set to \"english\".

Data page checksums are disabled.

creating directory /var/lib/postgresql/data/pgdata ... ok
creating subdirectories ... ok
selecting dynamic shared memory implementation ... posix
selecting default max_connections ... 20
selecting default shared_buffers ... 400kB
selecting default time zone ... Etc/UTC
creating configuration files ... ok
running bootstrap script ... ","pipe":"stdout","logging_pod":"cluster-example-1"}
{"level":"info","ts":1675094247.1112525,"logger":"initdb","msg":"Bus error
child process exited with exit code 135
initdb: removing data directory \"/var/lib/postgresql/data/pgdata\"
","pipe":"stderr","logging_pod":"cluster-example-1"}
[...]

PR #1858, opened 2023-04-03
fix: apply the full configuration during the init job

### LINKED ISSUE #1857: initdb should write the full configuration
Currently, we are only writing a partial configuration during the initdb job. We should proceed to write the full configuration.

armru:
Closes #1857


gbartolini:
Related to #1373 ?

litaocdl (approved):

PR #1916, opened 2023-04-12
Update grafana-dashboard.json

itay-grudev:
## New Cluster Overview

A new cluster overview section at the top that features:
* Base backup and last archived WAL information
* Total CPU and memory usage across all nodes
* An Alerts section
* Last failover/switchover time
* Cluster PostgreSQL version
* Transactions per second
* Volume usage and total database size
* Replication/Write/Flush/Replay lag

image

## Server Health

* Added instance zone
* Displaying the full PostgreSQL version
* Bug Fix: When there are redundant `kube-state-metrics` instances there are duplicate status gauges
* Bug Fix: Max Connections not displaying a progress bar in the gauge due to missing `min` and `max` values.

**Before:**

image

**After:**

image

## Configuration

The configuration parameters are transposed such that every row now contains a parameter, while every column contains the parameter setting across individual database instances. This makes it much easier to scroll through settings.

**Before:**

image

**After:**

image

## New Storage & IO Space and Inode Usage metrics

Added volume space and inode usage gauges.

image

## Accumulated Tuple/IO graph to make the data more comprehensible

**Before:**

image

**After:**

image

## General

* Fixed tooltips and set shared crosshairs of all graphs.
* Bug Fix: Cluster variable query picking up any metric ending with cluster

### DIFF
--- a/docs/src/quickstart.md
+++ b/docs/src/quickstart.md
@@ -321,6 +321,11 @@ kubectl apply -f \
 Which will be picked up by the Grafana page in a few seconds. You should now
 see the `CloudNativePG` dashboard.
 
+!!! Warning
+    Some graphs in the previous dashboard make use of metrics that are in alpha stage by the time
+    this was created, like `kubelet_volume_stats_available_bytes` and `kubelet_volume_stats_capacity_bytes`
+    producing some graphs to show `No data`.
+
 ![local grafana](images/grafana-local.png)
 
 Note that in our example setup, both Prometheus and Grafana will pick up
--- a/docs/src/samples/monitoring/grafana-configmap.yaml
+++ b/docs/src/samples/monitoring/grafana-configmap.yaml
@@ -23,3894 +23,5751 @@ metadata:
 data:
   cnp.json: |-
     {
-        "annotations": {
-          "list": [
-            {
-              "builtIn": 1,
-              "datasource": "-- Grafana --",
-              "enable": true,
-              "hide": true,
-              "iconColor": "rgba(0, 211, 255, 1)",
-              "name": "Annotations & Alerts",
-              "target": {
-                "limit": 100,
-                "matchAny": false,
-                "tags": [],
-                "type": "dashboard"
-              },
+      "annotations": {
+        "list": [
+          {
+            "builtIn": 1,
+            "datasource": {
+              "type": "datasource",
+              "uid": "grafana"
+            },
+            "enable": true,
+            "hide": true,
+            "iconColor": "rgba(0, 211, 255, 1)",
+            "name": "Annotations & Alerts",
+            "target": {
+              "limit": 100,
+              "matchAny": false,
+              "tags": [],
               "type": "dashboard"
-            }
-          ]
+            },
+            "type": "dashboard"
+          }
+        ]
+      },
+      "editable": true,
+      "fiscalYearStartMonth": 0,
+      "graphTooltip": 1,
+      "id": 452,
+      "links": [
+        {
+          "asDropdown": false,
+          "icon": "external link",
+          "includeVars": false,
+          "keepTime": false,
+          "tags": [
+            "cloudnativepg"
+          ],
+          "targetBlank": false,
+          "title": "Related Dashboards",
+          "tooltip": "",
+          "type": "dashboards",
+          "url": ""
+        }
+      ],
+      "liveNow": false,
+      "panels": [
+        {
+          "datasource": "${DS_PROMETHEUS}",
+          "gridPos": {
+            "h": 7,
+            "w": 3,
+            "x": 0,
+            "y": 0
+          },
+          "id": 334,
+          "options": {
+            "alertInstanceLabelFilter": "",
+            "alertName": "Database",
+            "dashboardAlerts": false,
+            "folder": "",
+            "groupBy": [],
+            "groupMode": "default",
+            "maxItems": 20,
+            "sortOrder": 1,
+            "stateFilter": {
+              "error": true,
+              "firing": true,
+              "noData": false,
+              "normal": true,
+              "pending": true
+            },
+            "viewMode": "list"
+          },
+          "title": "Alerts",
+          "type": "alertlist"
         },
-        "editable": true,
-        "fiscalYearStartMonth": 0,
-        "gnetId": null,
-        "graphTooltip": 0,
-        "iteration": 1637064390546,
-        "links": [],
-        "liveNow": false,
-        "panels": [
-          {
-            "collapsed": false,
-            "datasource": "${DataSource}",
-            "gridPos": {
-              "h": 1,
-              "w": 24,
-              "x": 0,
-              "y": 0
-            },
-            "id": 12,
-            "panels": [],
-            "title": "Server Health",
-            "type": "row"
+        {
+          "datasource": "${DS_PROMETHEUS}",
+          "gridPos": {
+            "h": 1,
+            "w": 15,
+            "x": 3,
+            "y": 0
           },
-          {
-            "datasource": "${DataSource}",
-            "gridPos": {
-              "h": 1,
-              "w": 3,
-              "x": 0,
-              "y": 1
+          "id": 336,
+          "options": {
+            "code": {
+              "language": "plaintext",
+              "showLineNumbers": false,
+              "showMiniMap": false
             },
-            "id": 191,
-            "options": {
-              "content": "",
-              "mode": "html"
-            },
-            "pluginVersion": "8.2.1",
-            "repeatDirection": "v",
-            "targets": [
-              {
-                "exemplar": true,
-                "expr": "kube_pod_container_status_ready{container=\"postgres\",namespace=~\"$namespace\",pod=~\"$instances\"}",
-                "instant": true,
-                "interval": "",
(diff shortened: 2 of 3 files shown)


jsilvela:
Sorry for the delay. Been deploying this locally.
There is a problem with the datasource references as you have them in your files.

``` json
"datasource": {
        "type": "prometheus",
        "uid": "${DS_PROMETHEUS}"
}
```

These will not work properly with the Grafana as we deploy from the quickstart, using the kube-prometheus-stack.
We can substitute the `"uid"` with `"prometheus"`, as that uid is generated for the data source in the kube-prometheus-stack
deployment.

jsilvela:
Since the JSON file you generated, in particular, you took from a grafana deployed with a different mechanism, I think there is value in keeping it as-is.
The YAML though, I would like to alter to make sure beginners following the quickstart will get the dashboard working out of the box.
So, I plan to add a bit more context in the documentation, and perhaps fix your files but add your JSON untouched but with a new name.

jsilvela:
With Itay's latest commit, the dashboard works out of the box if following the `quickstart` document.
Doing a detail review of the panels...

jsilvela:
@itay-grudev we're seeing errors on missing signoff in one of your commits, could you please fix it?:
> Commit sha: [29c2aed](https://github.com/cloudnative-pg/cloudnative-pg/pull/1916/commits/29c2aedc554d40d2a6ffcfb1dc06a5a59b046d44), Author: Itay Grudev, Committer: Itay Grudev; The sign-off is missing.

jsilvela:
The JSON/YAML diffs being so large, and their contents being produced, I'll put my comments here, rather than inline.
The `Volume Space Usage` panel is coming up empty in the default installation.

The PromQL is:

```
max(max by(persistentvolumeclaim) (1 - kubelet_volume_stats_available_bytes{namespace="$namespace", persistentvolumeclaim=~"$instances"} / kubelet_volume_stats_capacity_bytes{namespace="$namespace", persistentvolumeclaim=~"$instances"}))
```

But those metrics are not available in the kube-prom-stack install and are in ALPHA.
See https://kubernetes.io/docs/reference/instrumentation/metrics/

That said, they don't do damage, and the panel simply won't show much.
We could add some caveats on the `quickstart`, just so beginners are not surprised.

jsilvela:
As mentioned previously in the volume gauge, the new Storage IO section uses metrics that are in ALPHA, and not available in the kube-prometheus-stack

Screen Shot 2023-05-08 at 12 09 01

jsilvela (changes requested):
Brilliant work.
As the diffs are so big (and automated) I have left you comments in the "Conversation" window.
There are some advanced usages you've added that won't work when following the `quickstart` tutorial. My concern is with CloudNativePG beginners being baffled.

I can see two different approaches to get around this:

1. Keep the advanced usages and put a disclaimer on the `quickstart` that some panels won't be visible
2. Remove the advanced features from the ConfigMap, but add them as a new JSON file, with perhaps an `advanced` bit in the filename.

IMO 2 is the better alternative. A step could be added to the quickstart mentioning the advanced features.

gbartolini:
This is really looking good! Thanks @itay-grudev for the great work! And @jsilvela for the reviews.

benoitschipper:
Hey all,

Relatively new to contributing, I just wanted to share the following. And again, dismiss me if necessary. We just started using the CNPG Operator on a development cluster. We exposed the metrics and were looking for a dashboard. This dashboard is beautiful so I tested the one coming form this pull Request. After I imported the dashboard using the latest Grafana Operator It loaded the dashboard without problems. Only to find out that it did not show most of the fields, no data. I then found out that the "instances" variable is setup as follows:

**Query:**
```
cnpg_collector_up{namespace=~"$namespace",pod=~"$cluster-.*"}
```
**Regex:**
```
/pod="(?<text>[^"]+)/g
```

I then found that our test CNPG "Cluster" kind of the CPNG Cluster was called "cluster-example". Meaning that my CNPG Cluster related pods or in this case the variables within "instances" did not show, causing the "no data" fields within Grafana. This means that anyone that does not follow the convention of starting their "Cluster" kind name with the "cluster" field will get this problem. 

**My Cluster yaml Snippet:**
```
kind: Cluster
metadata:
  name: cluster-example
  namespace: my-namespace
spec:
…
```

Is this by convention? Or is it a best practice to start your "Cluster" kind name with the name of your kubernetes cluster? 

I was able to fix this by removing the `,pod=~"$cluster-.*"` from the "instance" variable query. Making it so any instance name would show up. The downside being that if you have multiple clusters with the same namespace, you lose the ability to filter on cluster.

**Before:**
image

**After:**
image

Should i simply rename my clusters to start with the clustername? Is there perhaps an alternative to make this Dashboard work with whatever name you give the CNPG Cluster?

I hope this makes sense, thanks in advance!
And thank you for the amazing Dashboard!

itay-grudev:
@benoitschipper I am having trouble understanding your problem.

The CNPG pods are named: `clusterName-<number>`.
By first selecting the cluster you are monitoring you are populating the `$cluster` variable, which in turn would filter out the pods based on said query:

```
pod=~"$cluster-.*"
```

I have multiple clusters and it's working fine. Maybe the problem is something else, but I'd like to understand what it is.

benoitschipper:
@itay-grudev I'll check again tomorrow and get back to you. Maybe i misunderstood the scope that was meant with "cluster". For some reason I now think: "Oh they meant CNPG Cluster and not the Kubernetes Cluster. Then again, as you could see, i did get "No Data". I'll check again tomorrow and get back to you :). 

Thanks for getting back to me!

benoitschipper:
@itay-grudev I think I found the issue here, it has to do with the variables within the Dashboard.

Here the instances variable is depending on the Prometheus, the namespace and the cluster variable. Let met show you the results all the other variables besides instances gives, then it should be clear on why I cannot see my CNPG-Cluster deployment.

image

- DS_PROMETHEUS: prometheus, default
- namespace: venek01-postgresql
- cluster: cd01
- instances: "empty", all

The Prometheus and the namespace variables work fine, but then we get to the cluster variable. And there the variable is populated with the name of my Kubernetes cluster, and not the name of the CNPG-Cluster. As my CNPG-Cluster is called "cluster-example" and as the "instances" variable is filled with the below query + Regex, it does not show any "instances" to monitor within the Dashboard.

**Query:**
`cnpg_collector_up{namespace=~"$namespace",pod=~"$cluster-.*"}`
**Regex:**
`/pod="(?<text>[^"]+)/g`

This because the result of the above query would be:

**Query:**
`cnpg_collector_up{namespace=~"venek01-postgresql",pod=~"cd01-.*"}`

Which would result with or without the regex in 0 hits, as my CNPG-Cluster's name is "cluster-example" and not "cd01-cluster-example".

Due to an additional label called "abcpcluster" (see below), the variable "cluster" is propagated with "cd01" and not with "cluster-example", see below :). The reason this is happening is because we seem to have "externalLabels" fields within our Prometheus setup. One was called "abcpcluster" with a value of "cd01". This key:value is listed within the Prometheus Configuration under "externalLabels" as the people who manage the cluster wanted to add this to all the metrics to be able to easially be able to seperate metrics by Kubernetes or Kubernetes cluster.
image

I believe we should be able to prevent this from happening by adjusting the regex that is used within the instances variable to the following `\bcluster\b="(?<text>[^"]+)` . The `\b` denotes a word boundary, ensuring that "cluster" is matched as a whole word and not as part of a larger word. See as demonstrated below:

image

That solved the problem for me 😄  Hope this helps! 😃

itay-grudev:
@benoitschipper Thanks for that. I'll test your proposal and incorporate it into the PR.

sxd:
@itay-grudev fallowing @jsilvela advice, I think we should add the disclaimer into the quickstart with this PR adding that some features may fails because they need a really new/early version of Grafana.

I'm also adding my self as a reviewer to this PR because I really want to try ti! :D

benoitschipper:
@sxd ill happily test/review what I can. Already have the dashboard running :)

itay-grudev:
@sxd I discussed offline with @jsilvela that the alpha metrics are enabled by default on Prometheus. I think it's a limitation of KinD that the volume data is not reported. I've tested it on AWS EKS and Digital Ocean and there it works fine.

I still need to add the change @benoitschipper proposed.

sxd:
@itay-grudev I've added a warning message that will help users to understand why the graphs doesn't have data, can you review it please! I understand that those are limited in KinD, but lots of users use KinD for testing and at least we need to be nice and put the warning message

benoitschipper:
> @sxd I discussed offline with @jsilvela that the alpha metrics are enabled by default on Prometheus. I think it's a limitation of KinD that the volume data is not reported. I've tested it on AWS EKS and Digital Ocean and there it works fine.
> 
> I still need to add the change @benoitschipper proposed.

Ill check/review it by testing it as soon as you have added the regex proposal so I can confirm.

benoitschipper (approved):
@itay-grudev I loaded the Dashboard and went through all the graphs on a live cluster in two seperate clusters. All worked without me making any changes. Works like a charm! 

image

benoitschipper (approved):

jsilvela (approved):

sxd (approved):

PR #1937, opened 2023-04-20
Prevent creating Kubernetes client when calling completion command

### LINKED ISSUE #1932: kubectl-cnpg completion bash fails when there is no APIserver
When `KUBECONFIG` is not set, I get:

```console
$ kubectl cnpg completion bash
Error: Get http://localhost:8080/api?timeout=32s: dial tcp 127.0.0.1:8080: connect: connection refused
$ # While it works for `kubectl`:
$ kubectl completion bash
# Copyright 2016 The Kubernetes Authors.
# [...]
```

This prevents usage of `source <(kubectl cnpg completion bash)` in `~/.bashrc`

sxd:
When the `completion` command was called while it was used inside the
`~/.basrc` file or a `~/.profile` this was useless since we were not able to output
the completion becase we were trying to create a Kubernetes client all the time,
which it's not the case during the load of a profile configuration, since this could
happen even before there's any Kubernetes cluster configured in place.

Closes #1932

### DIFF
--- a/cmd/kubectl-cnpg/main.go
+++ b/cmd/kubectl-cnpg/main.go
@@ -57,6 +57,14 @@ func main() {
 		SilenceUsage: true,
 		PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
 			logFlags.ConfigureLogging()
+
+			// If we're invoking the completion command we shouldn't try to create
+			// a Kubernetes client and we just let the Cobra flow to continue
+			if cmd.Name() == "completion" || cmd.Name() == "version" ||
+				cmd.Parent().Name() == "completion" {
+				return nil
+			}
+
 			return plugin.SetupKubernetesClient(configFlags)
 		},
 	}
--- a/cmd/kubectl-cnpg/main.go
+++ b/cmd/kubectl-cnpg/main.go
@@ -61,7 +61,7 @@ func main() {
 			// If we're invoking the completion command we shouldn't try to create
 			// a Kubernetes client and we just let the Cobra flow to continue
 			if cmd.Name() == "completion" || cmd.Name() == "version" ||
-				cmd.Parent().Name() == "completion" {
+				cmd.HasParent() && cmd.Parent().Name() == "completion" {
 				return nil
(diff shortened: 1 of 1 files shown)


sathieu:
@sxd I want to test, but where is the binary?

sxd:
@sathieu there's no binary created during this process, it's compiled but a binary it's not created, sadly, for testing purposed you need to clone the repo and switch to this branch `dev/1932` and then run `make` to compile and the binary will be `bin/kubectl-cnpg`

sathieu:
@sxd I won't have the time to do this today.

leonardoce:
In my opinion this patch is totally fine.
In `main`, I have:

```
$ KUBECONFIG= ./bin/kubectl-cnpg completion bash > /dev/null; echo $?
Error: Get "https://127.0.0.1:6443/api?timeout=32s": dial tcp 127.0.0.1:6443: connect: connection refused
1
```

With this patch applied:

```
$ KUBECONFIG= ./bin/kubectl-cnpg completion bash > /dev/null; echo $?
0
```

I'm approving it.

sathieu:
I can't test it right now, but I wonder if `cmd.Parent()` can be `nil` and generate a `panic`?

cmd/kubectl-cnpg/main.go
```diff
@@ -57,6 +57,14 @@ func main() {
 		SilenceUsage: true,
 		PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
 			logFlags.ConfigureLogging()
+
+			// If we're invoking the completion command we shouldn't try to create
+			// a Kubernetes client and we just let the Cobra flow to continue
+			if cmd.Name() == "completion" || cmd.Name() == "version" ||
+				cmd.Parent().Name() == "completion" {
```
mnencia, line 64:
```suggestion
			if cmd.Name() == "completion" || cmd.Name() == "version" ||
				cmd.HasParent() && cmd.Parent().Name() == "completion" {
```

mnencia (approved):

Issue #1950, opened 2023-04-24
Important changes to default values in 1.20


litaocdl:
Version 1.20 will change the default behaviour of the operator for a few settings, as described in #1707.

Specifically:

- replication slots for HA will be enabled by default, unless explicitly disabled in `.spec.replicationSlots.highAvailability.enabled`
- following a rolling update, the primary will be restarted, unless `switchover` is explicitly defined for `.spec.primaryUpdateMethod`
- base backups will be taken from a standby, if available, unless `.spec.backup.target` is explicitly set to `primary`
- the `-any` service will not be created anymore, unless the `CREATE_ANY_SERVICE` variable in the operator's configuration is set to `true` (already implemented as part of 3798c23aa10baaf677a2aff2731fd3162e409a81).

Make the above changes and add a section in the docs that explains the upgrade path.

gbartolini:
We decided to postpone changing the default for replication slots to 1.21 and prepare for that in this release.

PR #1958, opened 2023-04-24
chore: important changes to default values

litaocdl:
close: #1950


jsilvela (approved):

jlong49 (approved):

armru (approved):

gbartolini:
I would use this as commit message:

```
feat: change in default behavior for primary update method and backup target

CloudNativePG 1.20 introduces some changes to the default behavior of a
few features for newly created `Cluster` resources, compared to previous
versions of the operator. The goal of these changes is to improve the
resilience of a Postgres cluster out of the box through convention over
configuration. For clusters with one or more replicas:

- Backup from standby is now enabled by default, unless `target` is
   explicitly set to `primary`
- Restart of the primary is now the default method to complete the
   unsupervised rolling update procedure (`primaryUpdateMethod`
   defaults to `restart`, unless explicitly set to `switchover`)

Closes #1950
```

PR #1960, opened 2023-04-25
feat: recovery from VolumeSnapshot, and kubectl-cnpg snapshot subcommand

leonardoce:
This is a POC demonstrating what can be obtained by using CSI VolumeSnapshot resources to backup and restore CloudNativePG clusters.

A snapshot can be created with:

```
$ ./bin/kubectl-cnpg snapshot cluster-example -c longhorn
waiting for cluster-example-3 to be fenced
waiting for VolumeSnapshot cluster-example-3-1682433701 to be ready to use
waiting for VolumeSnapshot cluster-example-3-wal-1682433701 to be ready to use
unfencing pod cluster-example-3
```

And later restored with:

```
kind: Cluster
metadata:
  name: cluster-restore
spec:
  instances: 3

  storage:
    size: 1Gi
  walStorage:
    size: 1Gi

  bootstrap:
    recovery:
      dataSource:
        storage:
          name: cluster-example-3-1682433701
          kind: VolumeSnapshot
          apiGroup: snapshot.storage.k8s.io

        walStorage:
          name: cluster-example-3-wal-1682433701
          kind: VolumeSnapshot
          apiGroup: snapshot.storage.k8s.io
```

Closes: #1968


api/v1/cluster_webhook.go
```diff
@@ -799,6 +800,56 @@ func (r *Cluster) validateBootstrapRecoverySource() field.ErrorList {
...
+	if recoveryStanza.DataSource.WalStorage != nil && r.Spec.WalStorage == nil {
+		result = append(
+			result,
+			field.Invalid(
+				field.NewPath("spec", "bootstrap", "recovery", "dataSource", "walStorage"),
+				r.Spec.Bootstrap.Recovery.DataSource.WalStorage,
+				"A WAL storage configuration is required when recoverying using a DataSource for WALs"))
+	}
+
+	return result
```
fcanovai, line 850: Should we check here also if the sizes of the storages are larger than the ones of the snapshots? The pod will remain pending and only the PVC will display the cause of the issue.

fcanovai (commented):
I've tried to take a snapshot and restore it on EKS using EBS snapshots and it worked, nice!

api/v1/cluster_types.go
```diff
@@ -1091,6 +1091,10 @@ type BootstrapRecovery struct {
 	// so it must be set to the name of the source cluster
 	Source string `json:"source,omitempty"`
 
+	// The following PVC data source will be used to restore
+	// data
+	DataSource *DataSource `json:"dataSource,omitempty"`
```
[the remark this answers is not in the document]
gbartolini: I would suggest `volumeSnapshots`

api/v1/cluster_webhook.go
leonardoce: We could, but to do that we need the webhook server to reach the API server. And since the webhook server is handing a request created by the API server itself... I think we could stress the API server too much.

Beside this, in my understanding the VolumeSnapshot specification don't actually enforce this constraint, which is enforced by storage implementations. There could be storage implementations allowing changing the size on the fly, and perhaps we should not prevent that.

api/v1/cluster_types.go
```diff
@@ -1115,6 +1119,17 @@ type BootstrapRecovery struct {
...
 
+// DataSource contains the configuration required to bootstrap a
+// PostgreSQL cluster from an existing storage
+type DataSource struct {
+	// Configuration of the storage of the instances
+	// +optional
+	Storage corev1.TypedLocalObjectReference `json:"storage,omitempty"`
+
+	// Configuration of the storage for PostgreSQL WAL (Write-Ahead Log)
+	WalStorage *corev1.TypedLocalObjectReference `json:"walStorage,omitempty"`
```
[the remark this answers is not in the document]
leonardoce: Yep, this is a mistake. I'm fixing it.

docs/src/cnpg-plugin.md
```diff
@@ -845,3 +845,43 @@ postgres=# \q
...
+
+The `kubectl cnpg snapshot` creates snapshots of a Cluster by:
+
+1. choosing a replica Pod to work on
+2. fence the replica
+3. taking the snapshot
+4. unfence the replica
+
+!!! Important
+    A cluster having fenced instances cannot be snapshotted.
```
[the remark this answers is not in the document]
leonardoce: I'm not 100% sure this is the right behavior, but I'd prefer to snapshot replicas belonging to healthy clusters.
If you have a fenced instances, you may be performing a manual maintenance activity, and this may not be a good moment to take a snapshot.

Not sure I got your point @fcanovai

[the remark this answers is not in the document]
phisco:
```suggestion
!!! Important
    A cluster already having fenced instances cannot be snapshotted.
```

api/v1/cluster_webhook.go
```diff
@@ -799,6 +800,56 @@ func (r *Cluster) validateBootstrapRecoverySource() field.ErrorList {
 	return result
 }
 
+// validateBootstrapRecoveryDataSource is used to ensure that the data
+// source is correctly defined
+func (r *Cluster) validateBootstrapRecoveryDataSource() field.ErrorList {
```
phisco, line 805: As we already discussed, we should also validate `apiGroup` and `kind` to be of resources we can actually handle, given the current implementation is transparently passing it down to the PVC's [DataSource](https://github.com/kubernetes/api/blob/master/core/v1/types.go#L496), these should be: `PersistentVolumeClaim` and `snapshot.storage.k8s.io/VolumeSnapshot`

api/v1/cluster_webhook.go
```diff
@@ -799,6 +800,56 @@ func (r *Cluster) validateBootstrapRecoverySource() field.ErrorList {
...
+				"Recovery from dataSource is not compatible with other types of recovery"))
+	}
+
+	if recoveryStanza.RecoveryTarget != nil {
+		result = append(
+			result,
+			field.Invalid(
+				field.NewPath("spec", "bootstrap", "recovery", "recoveryTarget"),
+				r.Spec.Bootstrap.Recovery.RecoveryTarget,
+				"A recovery target cannot be set while recoverying from a DataSource"))
```
phisco, line 838:
```suggestion
				"A recovery target cannot be set while recovering from a DataSource"))
```

api/v1/cluster_webhook.go
```diff
@@ -799,6 +800,56 @@ func (r *Cluster) validateBootstrapRecoverySource() field.ErrorList {
...
+	}
+
+	if recoveryStanza.VolumeSnapshots.WalStorage != nil && r.Spec.WalStorage == nil {
+		result = append(
+			result,
+			field.Invalid(
+				field.NewPath("spec", "bootstrap", "recovery", "dataSource", "walStorage"),
+				r.Spec.Bootstrap.Recovery.VolumeSnapshots.WalStorage,
+				"A WAL storage configuration is required when recoverying using a DataSource for WALs"))
```
phisco, line 847:
```suggestion
				"A WAL storage configuration is required when recovering using a DataSource for WALs"))
```

docs/src/bootstrap.md
```diff
@@ -435,6 +438,61 @@ spec:
...
+bootstrap:
+    recovery:
+      volumeSnapshots:
+        storage:
+          name: <snapshot name>
+          kind: VolumeSnapshot
+          apiGroup: snapshot.storage.k8s.io
+```
+
+When recoverying from a Snapshot, it is not currently possible to target
```
phisco, line 464:
```suggestion
When recovering from a Snapshot, it is not currently possible to target
```

armru (approved):

gbartolini:
Reviewed the documentation. This is great. I see no reason why this feature should not work for PITR or replica clusters, provided we have a valid WAL archive in a recovery object store. This is the reason why I have removed for now this kind of limitations from the proposed documentation. If they don't work, it should just be due to validation or guards put by the operator, but from a Postgres PoV they should simply work.

PR #1973, opened 2023-04-27
fix: use correct metric name in example dashboard

jcpunk:
The example dashboard was using an incorrect metric query, this should resolve that issue and provide data where expected.


phisco:
There is definitely some inconsistency, however I think the default manifests deploy `cnpg_pg_postmaster_start_time`, so we should make it consistent with that, not the other way around. See https://github.com/cloudnative-pg/cloudnative-pg/blob/main/releases/cnpg-1.20.0.yaml#L12255

jcpunk:
It looks like https://github.com/prometheus-community/postgres_exporter/blob/master/queries.yaml defaults to `start_time_seconds` if that is at all relevant.

phisco (approved):

gbartolini:
Merged in #2021

Issue #1962, opened 2023-04-26
Support HA for logical replication slots


sjuls:
Currently CloudNativePG supports HA replication slots for physical replication slots, but ignores _logical_ replication slots. To better support connectors such as debezium which require the replication slot to be available upon connection, it would make sense to extend the current behaviour to also synchronize logical replication slot from the primary to the standby's and advance the LSN in the configured interval.

This would also protect against excessive WAL piling up on a standby risking running out of disk, since the consumer will likely consumer from the primary instance.

Similar comments have been made to [issue 13](https://github.com/cloudnative-pg/cloudnative-pg/issues/13), though the issue seems to have an overall different goal, hence the separate issue I am creating here.

nmargaritis:
Hello, what's the status of this?

ntx-ben:
Also looking forward to this for Debezium support

justinwyer:
Is any community contribution needed to move this forward?

ironytr:
any update?

ozen:
Looks like this is not an issue anymore sine the pg_failover_slots extension is the solution:

https://github.com/cloudnative-pg/cloudnative-pg/pull/1997#issuecomment-1722519914

sjuls:
@ozen I agree, closing.

PR #1996, opened 2023-05-03
feat: Add support for logical replication slots with HA (#1962)

sjuls:
Implement support for cloning logical replication slots from primary to standbys and advancing the LSN in the same cadence as physical replication slots. Aimed at solving issue #1962.

Is enabled through the `.spec.replicationSlots.highAvailability.enabled` property in the Cluster CRD.

Also;
* Bump manager limits to mitigate restarts in e2e tests due to failed health checks.
* Change `run-e2e.sh` to not fail if cnpg-system namespace is already created.

### DIFF
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -58,11 +58,11 @@ spec:
           value: $(DEFAULT_MONITORING_CONFIGMAP)
         resources:
           limits:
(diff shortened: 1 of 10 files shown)


sjuls:
Sorry, just noticed my e2e tests have gone - will get them set up again.

PR #1997, opened 2023-05-03
feat: Add support for logical replication slots with HA (#1962)

### LINKED ISSUE #1962: Support HA for logical replication slots

sjuls:
Implement support for cloning logical replication slots from primary to standbys and advancing the LSN in the same cadence as the current physical replication slots.

The cloning is performed by reading the logical replication slot state off the disk of the primary through the SQL layer, then restarting the standby instance to pick up the new logical replication slot. From PostgreSQL 16 it should be possible to [create the logical replication slots directly through SQL](https://bdrouvot.github.io/2023/04/19/postgres-16-highlight-logical-decoding-on-standby/), which is also supported in the PR.

The HA logical replication is an extension of the HA physical replication slots, and is therefore enabled through the `.spec.replicationSlots.highAvailability.enabled` property in the Cluster CRD.

Also in the PR;
* Bump manager limits to mitigate restarts in e2e tests due to failed health checks.
* Change run-e2e.sh to not fail if cnpg-system namespace is already created.

Closes #1962

### DIFF
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -58,8 +58,8 @@ spec:
           value: $(DEFAULT_MONITORING_CONFIGMAP)
         resources:
           limits:
-            cpu: 100m
-            memory: 200Mi
+            cpu: 500m
+            memory: 512Mi
           requests:
             cpu: 100m
             memory: 100Mi
--- a/docs/src/replication.md
+++ b/docs/src/replication.md
@@ -212,16 +212,16 @@ primary and have lost their slot.
 
 CloudNativePG fills this gap by introducing the concept of cluster-managed
 replication slots, starting with high availability clusters. This feature
-automatically manages physical replication slots for each hot standby replica
+automatically manages replication slots for each hot standby replica
 in the High Availability cluster, both in the primary and the standby.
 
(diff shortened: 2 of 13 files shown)


ozen:
Hi there. Is this PR forgotten :(

@sxd @mnencia

sjuls:
> Hi there. Is this PR forgotten :(
> 
> @sxd @mnencia

Don't actually think the PR is relevant anymore since pg_failover_slots has become a managed extension in CloudNativePG.

https://cloudnative-pg.io/documentation/1.20/postgresql_conf/#enabling-pg_failover_slots

I'll close this PR.

PR #2013, opened 2023-05-09
feat: add option to override default kubernetes scheduler

### LINKED ISSUE #1863: Add option to override default kubernetes scheduler
The application does not have the ability to overwrite the default scheduler (kube-scheduler) when running pods and jobs. This option is useful when the cluster uses Portworx storage and a custom Stork sheduler is required to be able to use this type of K8S Storage and also might be useful for other specific configurations or deviations applied to different clusters.

This capability can be achieved by adding another optional ClusterSpec parameter and then propagating it to the configuration of pods and jobs. Thanks to this, the end user will be able to freely overwrite this variable and in the case of a default setting, it will take the default cluster value (in most cases kube-scheduler).

Stork (as example) description: https://portworx.com/blog/stork-storage-orchestration-kubernetes/

Adrixop95:
Added the ability to override the default k8s sheduler with a custom one.

Fixes #1863


api/v1/cluster_types.go
```diff
@@ -135,6 +135,12 @@ type ClusterSpec struct {
 	// +optional
 	ImagePullPolicy corev1.PullPolicy `json:"imagePullPolicy,omitempty"`
 
+	// Name for custom sheduler when the default one cannot be used
+	// If not defined, it defaults to `default-scheduler`.
```
phisco, line 139: I think it would be better to default to "", as that would be the same behavior we currently have and doesn't need any assumption to be made

phisco (changes requested):

api/v1/cluster_types.go
Adrixop95: Okay, that makes sense.
I added additional parameter to kubebuilder:
`// +kubebuilder:default:=none`

From what I understand, it should generate by default the target result of an empty string.

phisco: yeah, either that or just not setting a default at all would have the same effect 👍

api/v1/cluster_types.go
```diff
@@ -135,6 +135,13 @@ type ClusterSpec struct {
...
+	// More info: https://kubernetes.io/docs/concepts/scheduling-eviction/kube-scheduler/
+	// +optional
+	// +kubebuilder:default:=none
```
phisco, line 142:
```suggestion
	// Name for custom scheduler when the default one cannot be used.
	// More info: https://kubernetes.io/docs/concepts/scheduling-eviction/kube-scheduler/
	// +optional
```

phisco: this will need to regenerate docs and CRDs

Adrixop95: I modified the description as suggested and regenerated the docs and CRDs.

phisco (approved):
For some reason it looks like I'm you have a few already merged commits, so this should be rebased before merging, other than that, LGTM

phisco (changes requested):
I have to investigate the failures in the e2es

phisco (approved):

Adrixop95:
> /test

Hey, from what I can see, 2 test scenarios didn't execute correctly, it seems to me that this fail is not entirely related to this change, but some timeout/waiting for the cluster to recreate resources. Could you please confirm this and re-run the tests or guide me what might cause such a problem so I can try to fix it?

armru (approved):

mnencia:
I would instead prefer to have a rollout when we change the scheduler. The rationale is that if you break the pod configuration, it is better to immediately have a not-working replica and then notice it after months when you need to do a rollout for an unrelated change.

armru:
I agree with @mnencia

mnencia (approved):

PR #2015, opened 2023-05-10
test: deprecate CreateNamespace

### LINKED ISSUE #2010: E2E: CreateNamespace should create and return a unique namespace
We should alter the func (env TestingEnvironment) CreateNamespace function to ensure uniqueness.
That function should create a unique Namespace and return its name.

We can think of several alternatives:

- use UUID’s to ensure uniqueness (but the resulting names would be long)

- add a go-routine-safe map data structure in TestingEnvironment where we could keep track of all namespaces created, and check for collisions

- have an atomic counter for namespaces. CreateNamespace() would reserve an entry in the counter, say “13”, then append the counter to the namespace argument provided by the user

This will ensure that we avoid conflicts during E2E runs

The issue was originally reported by @jsilvela

armru:
Closes #2010

### DIFF
--- a/tests/e2e/apparmor_test.go
+++ b/tests/e2e/apparmor_test.go
@@ -32,10 +32,11 @@ var _ = Describe("AppArmor support", Serial, Label(tests.LabelNoOpenshift, tests
 	const (
 		clusterName         = "cluster-apparmor"
 		clusterAppArmorFile = fixturesDir + "/apparmor/cluster-apparmor.yaml"
-		namespace           = "cluster-apparmor-e2e"
+		namespacePrefix     = "cluster-apparmor-e2e"
 		level               = tests.Low
 	)
 	var err error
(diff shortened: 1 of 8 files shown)


jsilvela:
I think the solution as is is incomplete, and needs further discussion.
The main issue is we keep basically the same interface: we provide a name, and create a namespace with that name.
Except, this may throw a runtime error if the namespace existed already…
In a way, this makes the new approach more brittle than the current one.

jsilvela (commented):
I think the solution as is is incomplete, and needs further discussion.
The main issue is we keep basically the same interface: we provide a name, and create a namespace with that name.
Except, this may throw a runtime error if the namespace existed already…
In a way, this makes the new approach more brittle than the current one.

jsilvela (approved):

mnencia (approved):

PR #2034, opened 2023-05-16
feat: add podMonitor support to pooler

### LINKED ISSUE #1966: [Feature] `enablePodMonitor` for `Pooler` CRD
The `Cluster` CRD has a helpful flag to enable creation of a `PodMonitor` for the postgresql servers as a part of their creation.

It would be helpful if a similar feature existed for the `Pooler` objects as well.

https://cloudnative-pg.io/documentation/1.19/connection_pooling/#monitoring

armru:
Closes #1966


api/v1/pooler_types.go
```diff
@@ -72,6 +72,10 @@ type PoolerSpec struct {
 
 	// The deployment strategy to use for pgbouncer to replace existing pods with new ones
 	DeploymentStrategy *appsv1.DeploymentStrategy `json:"deploymentStrategy,omitempty"`
+
+	// Enable or disable the `PodMonitor`
+	// +kubebuilder:default:=false
+	EnablePodMonitor bool `json:"enablePodMonitor,omitempty"`
```
phisco, line 78: What about moving this in a dedicated MonitoringConfiguration? This way we would mirror what we are already doing in the Cluster resource, and if down the line we want to add custom queries we can do so without breaking the APIs.

leonardoce (approved):

Issue #2050, opened 2023-05-19
e2e: avoid to use grace-period=0 in E2E test


litaocdl:
Deleting pods with grace-period=0 is unsafe because it immediately deletes the pod from the API server, leaving the container running for some time. That could lead to two pods insisting on the same data volume. 

Use grace-period=1 instead.

PR #2051, opened 2023-05-19
test: use gracePeriodSeconds=1 in e2e test while deleting pod

litaocdl:
closes: #2050

### DIFF
--- a/tests/e2e/asserts_test.go
+++ b/tests/e2e/asserts_test.go
@@ -2281,10 +2281,16 @@ func OfflineResizePVC(namespace, clusterName string, timeout int) {
 				// Deleting standby and replica pods
 				err = env.DeletePod(namespace, pod.Name, forceDelete)
 				Expect(err).ToNot(HaveOccurred())
-				// Ensuring cluster is healthy with three pods
-				AssertClusterIsReady(namespace, clusterName, timeout, env)
 			}
 		}
+		// Make sure the cluster phase changed first
+		AssertClusterEventuallyDepartsPhase(namespace, clusterName, apiv1.PhaseHealthy, timeout)
+		// Ensuring cluster is healthy with three pods, we only need verify here once to safe time,
+		// but we need to make sure the cluster changed to un-health first, otherwise, it will be too
+		// quick, the status of the cluster did not have time to change from healthy to un-health, but
+		// the code reach to deleting primary pod.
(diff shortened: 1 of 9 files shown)


tests/e2e/asserts_test.go
```diff
@@ -2281,10 +2281,16 @@ func OfflineResizePVC(namespace, clusterName string, timeout int) {
...
-				AssertClusterIsReady(namespace, clusterName, timeout, env)
 			}
 		}
+		// Make sure the cluster phase changed first
+		AssertClusterEventuallyDepartsPhase(namespace, clusterName, apiv1.PhaseHealthy, timeout)
+		// Ensuring cluster is healthy with three pods, we only need verify here once to safe time,
+		// but we need to make sure the cluster changed to un-health first, otherwise, it will be too
+		// quick, the status of the cluster did not have time to change from healthy to un-health, but
+		// the code reach to deleting primary pod.
+		AssertClusterIsReady(namespace, clusterName, timeout, env)
```
jsilvela, line 2292: checking for cluster ready after cluster is no longer healthy (after pods have been deleted) is surprising me

tests/e2e/asserts_test.go
```diff
@@ -2281,10 +2281,16 @@ func OfflineResizePVC(namespace, clusterName string, timeout int) {
 				// Deleting standby and replica pods
 				err = env.DeletePod(namespace, pod.Name, forceDelete)
 				Expect(err).ToNot(HaveOccurred())
-				// Ensuring cluster is healthy with three pods
-				AssertClusterIsReady(namespace, clusterName, timeout, env)
 			}
 		}
+		// Make sure the cluster phase changed first
+		AssertClusterEventuallyDepartsPhase(namespace, clusterName, apiv1.PhaseHealthy, timeout)
```
jsilvela, line 2287: are we sure this is always going to happen? I'm testing locally and the transition isn't there...

jsilvela (approved):

tests/e2e/asserts_test.go
mnencia: I think we should do it differently. AssertClusterIsReady should make sure that none of the Cluster's pods is deleted.

tests/e2e/asserts_test.go
litaocdl: Yeah, if we change grace_period=1, this test case failed everytime.

mnencia (approved):

PR #2124, opened 2023-05-25
feat: add startupProbe

### LINKED ISSUE #1964: [Feature] begin using `startupProbe`
With `startupProbe` introduced in Kubernetes 1.17, and the current version being 1.27, can CNPG begin using this feature?

Folks running Kubernetes older than 1.17 are unlikely to be updating CNPG at this point.

armru:
Closes #1964


armru:
The startupProbe implementation is complex without adding enough benefits with the current reconciliation logic that relies on HTTP status calls.

Closing without merging, I think we can revisit the opportunity to add the startupProbe once we simplify the operator reconciliation loop.

PR #2141, opened 2023-05-29
fix: kind version

jsilvela:


sxd (approved):

mnencia (approved):

phisco:
Looks like this is using the Kubernetes version instead of the kind version, see here for a similar config https://github.com/crossplane/crossplane/blob/master/.github/renovate.json5#L59

PR #2145, opened 2023-05-29
fix: halt the failover if status extraction has failed

### LINKED ISSUE #2144: CNPG should not trigger a failover if all the instances are unreachable
Suppose all the instances are reported up from the K8s API but are unreachable (for example, because of a wrong-configured network policy). In that case, the operator always tries to promote the first instance in alphabetic order. The correct behaviour would be not to try changing the cluster status.

Issue originally opened by @mnencia

armru:
This commit addresses an issue where the failover process would continue even if the operator failed to extract status from the instances, potentially due to network connectivity issues. 

The operator now checks if the status has been successfully extracted from the instances before proceeding with the failover. If status extraction fails, the operator logs an error message, halts the current failover attempt, and reschedules another attempt after a delay.

Closes #2144

### DIFF
--- a/controllers/cluster_controller.go
+++ b/controllers/cluster_controller.go
@@ -230,6 +230,20 @@ func (r *ClusterReconciler) reconcile(ctx context.Context, cluster *apiv1.Cluste
 		return ctrl.Result{}, fmt.Errorf("cannot update the instances status on the cluster: %w", err)
 	}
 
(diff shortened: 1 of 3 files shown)


armru:
when executing the merge check that the title and commit message is correct

mnencia (approved):

sxd (approved):

PR #2186, opened 2023-06-02
Add uri and jdbc-uri to Secret

lieryan:
Hi, this PR adds a [Postgres Connection URI](https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNSTRING) and [JDBC URI](https://jdbc.postgresql.org/documentation/use/).

The Secrets's key names is similar to how the [similar values](https://access.crunchydata.com/documentation/postgres-operator/v5/tutorial/connect-cluster/) are named in Crunchy PGO. It makes it easier if people want to migrate from one to the other operator.


sxd:
Hi @lieryan 

First at all sorry for the late reply on this, but this PR was overlooked by me the previous weeks review.
Can you provides an example on how this will be used? I'm totally inclined to add it but probably we can try to actually test it with something? Can you provide us a way to manually tests this ?

Regards

lieryan:
Hi @sxd, thanks for looking into this PR.

The regular URI can be used by psql like so: `psql postgresql://dbmaster:5433/mydb`. My personal use case is for use with sqlalchemy though. The main idea with having these URIs as Secret would be that they would be mapped to an environment variable using secretKeyRef, and then an application can read that environment variable into SQLAlchemy's `create_engine()`.

jdbc-uri essentially very similar, but it's for database connectors that uses JDBC URI syntax instead of standard postgres URI syntax. JDBC is a common database connection mechanism in Java. I'm not too familiar with it since I don't use Java though.

lieryan:
I didn't really have any way to test this manually. I had only got enough test harnesses setup to run the unit tests and not go through the actual full install of the modified operator, but I'll try to see if I can write some simple k8s manifests to setup a psql/sqlalchemy-based pod to validate the change.

lieryan:
So this is the way I was testing this:

```yaml
---
kind: Cluster
metadata:
  name: test-cnpg
spec:
  storage:
    size: 2Gi

---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: test-psql-client
spec:
  selector:
    matchLabels:
      app: test-psql-client
  template:
    metadata:
      labels:
        app: test-psql-client
    spec:
      containers:
        - name: test-psql-client
          image: postgres:15
          command: 
            - psql
            - $(POSTGRES_URI)
            - -c
            - "SELECT version();"
          env:
            - name: POSTGRES_URI
              valueFrom:
                secretKeyRef:
                  name: test-cnpg-app
                  key: uri


---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: test-sqlalchemy
spec:
  selector:
    matchLabels:
      app: test-sqlalchemy
  template:
    metadata:
      labels:
        app: test-sqlalchemy
    spec:
      containers:
        - name: test-sqlalchemy
          image: python:3.10
          command: 
            - bash
            - -c
            - |
                pip install sqlalchemy psycopg2

                python3 -u -c "
                from sqlalchemy import create_engine, text
                import os, contextlib, time
                while True:
                    with contextlib.suppress(Exception) as e:
                        engine = create_engine(os.environ['POSTGRES_URI'])
                        with engine.connect() as conn:
                            print(conn.scalar(text('SELECT version()')))
                        time.sleep(5)
                "
          env:
            - name: POSTGRES_URI
              valueFrom:
                secretKeyRef:
                  name: test-cnpg-app
                  key: uri


---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: test-jdbc-uri
spec:
  selector:
    matchLabels:
      app: test-jdbc-uri
  template:
    metadata:
      labels:
        app: test-jdbc-uri
    spec:
      containers:
        - name: test-jdbc-uri
          image: debian:latest
          command: 
            - bash
            - -c
            - |
                apt update 
                apt install -yq --no-install-recommends sqlline libpostgresql-jdbc-java

                while true; do
                  echo 'SELECT version()' | sqlline -u $POSTGRES_URI
                  sleep 5
                done
          env:
            - name: POSTGRES_URI
              valueFrom:
                secretKeyRef:
                  name: test-cnpg-app
                  key: jdbc-uri
```

There are three tests there:

1. psql test for Postgres URI
2. sqlalchemy test for Postgres URI
3. [sqlline](https://github.com/julianhyde/sqlline) test for JDBC URI

In all three tests, it should print the output of `SELECT version()` to indicate successfully connecting to the Postgres server.

To check the output, run:

1. `kubectl logs --follow deploy/test-psql-client`
2. `kubectl logs --follow deploy/test-sqlalchemy`
3. `kubectl logs --follow deploy/test-jdbc-uri`

lieryan:
Hi @sxd, is there anything else that I should/could do to push this PR forward?

sxd (approved):

leonardoce (approved):

PR #2243, opened 2023-06-09
chore: detect PodSpec drift in pod rollout, and clarify rollout code

jsilvela:
- Add pod annotation with original PodSpec to detect drift.
- Use a standard structure for the functions checking if
  rollout is needed.
- Fix logic that could allow a rollout with a Postgres image
  from a different major version.

### DIFF
--- a/controllers/cluster_upgrade.go
+++ b/controllers/cluster_upgrade.go
@@ -377,21 +377,20 @@ func checkPodImageIsOutdated(
 	}
 
 	if pgCurrentImageName != targetImageName {
+		canUpgradeImage, err := postgres.CanUpgrade(pgCurrentImageName, targetImageName)
+		if err != nil {
+			return Rollout{}, err
+		}
+
+		if !canUpgradeImage {
+			return Rollout{}, nil
+		}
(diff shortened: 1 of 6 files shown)


armru (approved):

controllers/cluster_upgrade.go
```diff
@@ -372,109 +400,193 @@ func getProjectedVolumeConfigurationFromPod(pod corev1.Pod) *corev1.ProjectedVol
...
 
-	pgCurrentImageName, err := specs.GetPostgresImageName(pod)
+	pgCurrentImageName, err := specs.GetPostgresImageName(status.Pod)
 	if err != nil {
-		return "", "", err
+		return rollout{}, err
 	}
 
 	if pgCurrentImageName != targetImageName {
+		canUpgradeImage, err := postgres.CanUpgrade(pgCurrentImageName, targetImageName)
```
jsilvela, line 415: NOTE: this fixes a bug. Previously the check to verify if the upgrade is possible was checked only when the current image equals the target image

mnencia (approved):

PR #2257, opened 2023-06-12
feat: add `NodesUsed` metric and status field

### LINKED ISSUE #2231: Add number of different worker nodes in the status and as a metric
During some support incidents on CloudNativePG that we have had while working at EDB, we have noticed that it is not immediate to detect if a Postgres cluster is actually spanning across different Kubernetes worker nodes. In these specific cases, users were complaining that the Kubernetes upgrade procedure could not proceed because the pod disruption budget was blocking it as pods and PVCs could not be moved on other nodes.

Having in the status and/or as a metric the possibility to count the number of different worker nodes on which the Postgres pods run, would have dramatically helped diagnosing this issue.

In general, this is an aspect where IMHO there's a lot of room for improvements, by being proactive and possibly sending warnings to the user that HA might be impacted.

armru:
Closes #2231


gbartolini (approved):

armru:
The first /test that I ran should fail the second should pass

NiccoloFei (approved):

litaocdl (approved):

Issue #2283, opened 2023-06-14
Update Kustomization files to be compatible with Kustomize v5


NiccoloFei:
Change the scaffolding Kustomization files to make them compatible with v5.

PR #2290, opened 2023-06-14
chore: update kustomization files to use Kustomize v5

cschp:
Closes https://github.com/cloudnative-pg/cloudnative-pg/issues/2283

Testing: tested generated kustomize from before and after the changes to see the diff and to check if the sha's match which they usually do

testing in comments below

How:
k5r script used on each of the kustomization.yaml files
script listed in comments


cschp:
@sxd hopefully this one is more agreeable i did not use the automated tool and did the changes manually

NiccoloFei (approved):

config/helm/kustomization.yaml
```diff
@@ -1,5 +1,6 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
 resources:
 - ../crd
 commonAnnotations:
-  "helm.sh/resource-policy": keep
```
sxd, line 4: Why this was removed ? we can't remove this please keep it

cschp: this should be fixed back to the way it was now.

sxd:
/ok-to-merge adding abel since thi sis a change in the kustomize files

sxd (approved):

PR #2307, opened 2023-06-15
refactor: reorganize instance metadata patch procedure

### LINKED ISSUE #2207: Patch the Pod metadata only once
Patching multiple times increases the load on the API server without any benefit.

https://github.com/cloudnative-pg/cloudnative-pg/blob/b87182bdfff4eefcb2d619f9296ffc8edd159c09/controllers/cluster_controller.go#L440-L457

armru:
Closes #2207


armru:
this PR poses a risk and should be triple checked

litaocdl (approved):

jsilvela:
Huh, merge conflict because of a recent ticket

> test: replace metadata E2E test with a unit test (#2214)
> commit e146fd7954e4d8c48be9d7981defc3bcc0585071

which added the fake Kubernetes reconciler. This is valuable and we should not lose it.
Will handle the conflict with more time or send back to Armando

jsilvela (approved):

mnencia (approved):

PR #2333, opened 2023-06-20
feat: add PhaseReplicaLagThresholdExceeded

### LINKED ISSUE #2323: Improving Pod Intercommunication Visibility within Instance Status
At present, our instance status is restricted to indicating if a specific instance is able to communicate with the operator. However, it does not provide information regarding inter-instance communication, i.e., if one instance is capable of reaching the other instances.

To enhance visibility and troubleshooting capabilities, we could introduce a new field within the instance status. This field will serve to indicate whether a specific instance is able to establish communication with other instances.

armru:
This commit introduces a new phase called PhaseReplicaLagThresholdExceeded. This phase is triggered when one or more replica databases lag beyond a user-defined limit compared to the primary database.

A new `ReplicaLagThreshold` field has been added to the `Cluster` spec, allowing users to define their acceptable replica lag limit. If left unspecified, the operator defaults to a 30MB lag threshold.

Partially Closes #2323


pkg/postgres/status.go
```diff
@@ -300,6 +300,38 @@ func (list PostgresqlStatusList) AreWalReceiversDown(primaryName string) bool {
...
+		if list.Items[idx].Pod.Name == primaryName {
+			continue
+		}
+		if !list.Items[idx].IsWalReceiverActive {
+			return true
+		}
+	}
+
+	return false
+}
```
jsilvela, line 319: not used anywhere

jsilvela (changes requested):
There's a method added that is not used anywhere - I checked locally that it can be removed and everything is fine.
Wondering if this was meant to use, or is a discarded path, I think this needs to be clarified by the author.

jsilvela (approved):

api/v1/cluster_types.go
```diff
@@ -309,6 +309,14 @@ type ClusterSpec struct {
 	// The SeccompProfile applied to every Pod and Container.
 	// Defaults to: `RuntimeDefault`
 	SeccompProfile *corev1.SeccompProfile `json:"seccompProfile,omitempty"`
+
+	// ReplicaLagThreshold is a user-customizable field that represents the acceptable lag limit (in megabytes,
+	// where 1MB = 1,000,000 Bytes) between the primary database and its replicas. If a replica's lag exceeds this
```
litaocdl, line 314: ReplicaLagThreshold is a user-customizable field that represents the acceptable lag limit (in megabytes... )
========
The default value of the ReplicaLagThreshold is 30,000,000, which is in Bytes, not in megabytes. we should either change the descrption or change the unit used for ReplicaLagThreshold

litaocdl (approved):

mnencia:
I've had a meeting with @leonardoce and @armru about this patch. We decided that having a phase that does not change the operator's behavior is pointless. We decided to close this patch without merging, then @armru will open a new PR that will be more complete.

PR #2336, opened 2023-06-20
fix: restrict read service to instance pods only

### LINKED ISSUE #2335: read service `-r` is backed by join\init containers
Whilst a new instance is being configured an init and join pod is created which prepares the actual pod to join the cluster. The `-r` service is incorrectly being backed by these join and init pods.

All services use the selector `cnpg.io/cluster: cluster-name`
The following services use the role label too:
`-rw` does not have this issue as its selector uses `role: primary` too.
`-ro` does not have this issue as its selector uses `role: replica` too.

Because the `-r` service does not use any additional labels it captures the init and join pods too.

Otterpohl:
Addresses the issue where the -r service would be backed by the init and join pods.

Closes  #2335

### DIFF
--- a/pkg/specs/services.go
+++ b/pkg/specs/services.go
@@ -67,6 +67,7 @@ func CreateClusterReadService(cluster apiv1.Cluster) *corev1.Service {
 			Ports: buildInstanceServicePorts(),
 			Selector: map[string]string{
 				utils.ClusterLabelName: cluster.Name,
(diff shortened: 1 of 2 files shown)


sxd:
hi @Otterpohl 

can you do the same for the `-any` servie? (including the tests) even if we don't promote this service and will be deprecated we can keep consistency

Thanks!!

sxd (approved):

PR #2374, opened 2023-06-28
fix: return early from instance manager if PGDATA does not exist

achanda:
This should handle cases where the PGDATA directory is missing and raise an appropriate error


phisco:
@achanda can you add a description or open an issue describing the problem you are addressing? Even if the patch is pretty clear, it would be still helpful

achanda:
@phisco I was testing a failure scenario where PGDATA is completely deleted in a running cluster using rm rf. I saw that the operator detects that one replica is missing, but does nothing after that. I was hoping this change would raise an error and prompt reconciliation. But I have not been able to test this patch yet.

PR #2382, opened 2023-06-29
chore: do not add label backport-requested to PRs created by renovate

litaocdl:
Previously the goal was to backport everything, and this works perfectly, but with renovate we create the PRs with the label do not backport which it's added because those PR will be replicated one per each release branch.

This is the further fix to avoid adding the backport-requested label when the PR is created by renovate

### DIFF
--- a/.github/workflows/backport.yml
+++ b/.github/workflows/backport.yml
@@ -16,14 +16,16 @@ jobs:
   # we backport everything, except those PR that are created or contain `do not backport` explicitly.
   label-source-pr:
     name: Add labels to PR
-    if: github.event.pull_request.merged == false &&
+    if: | 
+        github.event.pull_request.merged == false &&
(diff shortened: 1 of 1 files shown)


phisco (commented):
We add a the "do not backport" label to all renovate's by default, I'm not sure why this check is needed 🤔

litaocdl:
> We add a the "do not backport" label to all renovate's by default, I'm not sure why this check is needed 🤔

Sometimes, the backport workflow run faster than the renovate add labels action, so `do not backport` added after `backport-requested` label :)

litaocdl:
@sxd @phisco 
The reason why it not works every time is due to the checking for `do not backport` label is added in the job level, between job is schedule and steps run, there is time gap. I add redundant check in the label step level, and also, In the final step, add the step to remove the redundant labels if have.

mnencia (approved):

sxd (approved):

PR #2397, opened 2023-06-30
ci: fix `docker manifest inspect` issue

mnencia:
Ref moby/moby#45726


gbartolini (approved):

sxd (approved):

NiccoloFei (approved):

mnencia:
Unfortunately, this fix doesn't work. Only the last image is kept on older systems, and it could be the one with the wrong architecture. We need a better, less hacky way to copy the images in the local registry to fix this issue properly.

PR #2412, opened 2023-07-05
changed the $cluster-.* regex

### LINKED ISSUE #2411: grafana dashboard $cluster-.* regex
When having similar named cluster, the instance regex is not working.
I have keyclaok, keycloak-int, keycloak-test, keyclaok-stg, and when selecting the keyclaok cluster, all instances from all other keycloak-* clusters are shown
I managed to fix it with a different regex in the instance variable
`cnpg_collector_up{namespace=~"$namespace",pod=~"$cluster-[1-99]+$"}`
instead of
`cnpg_collector_up{namespace=~"$namespace",pod=~"$cluster-.*"}`

Unbenannt

smiyc:
Fixes #2411

Signed-off-by: Daniel Chambre <smiyc@pm.me>

### DIFF
--- a/docs/src/samples/monitoring/grafana-configmap.yaml
+++ b/docs/src/samples/monitoring/grafana-configmap.yaml
@@ -919,7 +919,7 @@ data:
             {
               "datasource": "${DS_PROMETHEUS}",
               "editorMode": "code",
-              "expr": "max(cnpg_pg_stat_replication_flush_lag_seconds{namespace=~\"$namespace\",pod=~\"$cluster-[0-99]+$\"})",
+              "expr": "max(cnpg_pg_stat_replication_flush_lag_seconds{namespace=~\"$namespace\",pod=~\"$cluster-[0-9]+$\"})",
               "legendFormat": "__auto",
               "range": true,
               "refId": "A"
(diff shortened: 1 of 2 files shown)


docs/src/samples/monitoring/grafana-configmap.yaml
```diff
@@ -919,7 +919,7 @@ data:
             {
               "datasource": "${DS_PROMETHEUS}",
               "editorMode": "code",
-              "expr": "max(cnpg_pg_stat_replication_flush_lag_seconds{namespace=~\"$namespace\",pod=~\"$cluster-.*\"})",
+              "expr": "max(cnpg_pg_stat_replication_flush_lag_seconds{namespace=~\"$namespace\",pod=~\"$cluster-[0-99]+$\"})",
```
itay-grudev, line 922: This Regex is incorrect. Could you replace it with: `-[1-9][0-9]?$`.

Could you replace it in both documents?

itay-grudev (changes requested):

docs/src/samples/monitoring/grafana-configmap.yaml
itay-grudev: I would also be fine with `-[0-9]+$` for the sake of simplicity.

itay-grudev (approved):

itay-grudev (approved):

phisco (approved):

leonardoce (approved):

PR #2419, opened 2023-07-06
chore: add `LastCheckpointRedoWAL` check during restore process

### LINKED ISSUE #2417: Perform a check of WAL source before starting a restore from a backup/snapshot
If there is an error in the WAL source settings, there is no way for the user to detect the mistake because PostgreSQL is calling Barman directly and discarding the restore_command output.

The issue was originally reported by @mnencia @leonardoce @armru

armru:
This commit ensures the archive contains the 'LastCheckpointRedoWAL' during the instance manager restore process.

This ensures that the database can be restored correctly and it highlights any errors that occur as a result of the archive lacking the WALs.

Closes #2417


litaocdl (approved):

mnencia:
The code I removed in d00f978db3ff911a0da0ed772ed9369ec2310c46 must be added back to the cold snapshot support patch.

pkg/management/postgres/restore.go
```diff
@@ -145,6 +150,49 @@ func (info InitInfo) Restore(ctx context.Context) error {
...
 }
 
+func (info InitInfo) ensureArchiveContainsLastCheckpointRedoWAL(
+	ctx context.Context,
+	cluster *apiv1.Cluster,
+	env []string,
+	backup *apiv1.Backup,
+) error {
+	// it's the full path of the file that will temporarily contain the LastCheckpointRedoWAL
+	const testWALPath = postgresSpec.RecoveryTemporaryDirectory + "/test.wal"
```
jsilvela, line 160: I'm concerned with potential for conflict here, leaving debris in the recovery directory.
Would it not be safer to use `os.TempDir()`?

jsilvela: Looking at the hour ... going to test this change and see how it behaves...

jsilvela (approved):

mnencia (approved):

PR #2428, opened 2023-07-10
ci: enable cloud provider E2E tests

### LINKED ISSUE #2427: Introduce Cloud Provider E2E tests
Extend the continuos-delivery workflow to allow running E2E tests on Cloud Providers (EKS, AKS, GKE).
The Cloud provider tests are disabled by default, and will only run in case the following GH variables are found inside the target repository:

* vars.AKS_ENABLED is set to 'true'
* vars.EKS_ENABLED is set to 'true'
* vars.GKE_ENABLED is set to 'true'

3 new JSON files (`aks_versions`, `eks_versions` and `gke_versions`) will hold each Cloud Provider version on which E2E tests will be executed. These files will add up to the existing `kind_versions`, and will be used by `e2e-matrix-generator.py` to build a matrix of E2E test jobs.

NiccoloFei:
Closes #2427


litaocdl (approved):

.github/gke_versions.json
```diff
@@ -0,0 +1,6 @@
+[
```
jsilvela, line 1: in the CNP repo I see 1.27 at the top of the gke_versions file.
Noting to double-check....

hack/e2e/eks-cluster.yaml
```diff
@@ -0,0 +1,30 @@
+apiVersion: eksctl.io/v1alpha5
```
jsilvela, line 1: I think this file needs to be called `hack/e2e/eks-cluster.yaml.template`.
Checking...

jsilvela: Confirmed. Fixing.

.github/workflows/kind-k8s-versions-check.yml
```diff
@@ -1,54 +0,0 @@
-# Checks the latest kind image version from `https://registry.hub.docker.com/v2/repositories/kindest/node/tags`,
```
jsilvela, line 1: not sure why we're deleting this workflow. There's a public cloud versions check in CNP that wer're not bringing in instead. Wondering if we forgot to.

jsilvela (approved):

mnencia (approved):

PR #2456, opened 2023-07-15
fix: make `umask` call in initdb compatible with Windows plugin

jsilvela:
The `umask` call used in initdb to prevent permissions tampering is
breaking the build of the kubectl-cnpg plugin for Windows.
This fix puts the call in a compatibility layer so windows builds still work.


jsilvela:
CI/CD run with the kubectl plugin being built successfully in windows
https://github.com/cloudnative-pg/cloudnative-pg/actions/runs/5562887079/jobs/10161473968

sxd (approved):

litaocdl (approved):

Issue #2459, opened 2023-07-17
By default, recovery_target_inclusive should be false per document


litaocdl:
if exclusive not specific or exclusive is true, ecovery_target_inclusive should be false.

gbartolini:
The default behaviour in PostgreSQL is to have the inclusive target enabled by default. See https://www.postgresql.org/docs/current/runtime-config-wal.html#GUC-RECOVERY-TARGET-INCLUSIVE

I think we just need to fix the documentation as the code, IMHO, is coherent with Postgres' behaviour:

```
        switch {
        case target.Exclusive == nil:
                result += "recovery_target_inclusive = true\n"
        case *target.Exclusive:
                result += "recovery_target_inclusive = true\n"
        default:
                result += "recovery_target_inclusive = false\n"
        }
```

jsilvela:
Agree with matching the default in Postgres.
As I read the code, I think the code is wrong though:

``` go
    case *target.Exclusive:
        result += "recovery_target_inclusive = true\n"
    default:
        result += "recovery_target_inclusive = false\n"
```

if `target.Exclusive` is FALSE, we append `recovery_target_inclusive = true`.
Does not look right...

PR #2460, opened 2023-07-17
fix: by default recovery should be exclusive

litaocdl:
closes: #2459


mnencia:
This is a behavior change in the public API. We must be very explicit in our release notes if we implement this.
The alternative is to change the documentation to describe the current behavior.

@gbartolini, what do you think is the best way?

jsilvela (approved):

sxd (approved):

PR #2465, opened 2023-07-18
fix: fix windows compile issue

litaocdl:
Fix the windows os compile issue


litaocdl:
Use `GOOS=windows GOARCH=386 CGO_ENABLED=0 go build` local test

sxd (approved):

PR #2503, opened 2023-07-27
chore: Improve error output on snapshot command

sxd:
We were previously waiting for the snapshot to be not ready to use,
with this patch we check if there's an error inside the VolumeSnapshot
object and error out and let the user knows the error directly.


litaocdl (approved):

mnencia (approved):

mnencia:
/ok-to-merge additional error message, no e2e impact

PR #2528, opened 2023-08-01
feat: declarative cold snapshot backup

### LINKED ISSUE #2527: investigate how to execute a backup through volumesnapshots
We should investigate a way to execute backups with volumesnapshots declaratively.

We should probably enrich the ScheduledBackup and Backup resources with new fields:

- `method: barmanObjectStore|volumeSnapshotTemplate`. Specifies how to execute the backup
- `online: true|false`. Specifies if we should fence or not the instance before taking the backup (in the case of volumesnapshots). Online will always be false in the case of WAL archive not being present.

This is a stub and the issue will be further improved

armru:
Closes #2527


leonardoce (approved):

NiccoloFei (approved):

mnencia (approved):

docs/src/backup.md
```diff
@@ -1,56 +1,103 @@
...
-[replica clusters](replica_cluster.md)
-for more information about designated primary instances), or alternatively
-on a [standby](#backup-from-a-standby).
+**Cold backups**, also known as offline backups, are a physical base backup
+taken when the PostgreSQL instance (standby or primary) is shut down. They are
+consistent per definition and they represent a snapshot of the database at the
+time it was shut down.
 
-## Common object stores
+As a result, they do not require the WAL archive to be restarted, but they can
```
NiccoloFei, line 72: to be restarted -> to be restored ?

docs/src/backup.md
```diff
@@ -1,56 +1,103 @@
...
+take advantage of it if available (with all the benefits for the recovery
+explained in the previous section).
```
gbartolini:
```suggestion
As a result, PostgreSQL instances can be restarted from a cold backup without the need of
a WAL archive, even though they can take advantage of it, if available (with all the benefits
on the recovery side highlighted in the previous section).
```

gbartolini: Better now @NiccoloFei ?

NiccoloFei: yes, it seems good to me now

docs/src/operator_capability_levels.md
```diff
@@ -333,11 +331,31 @@ in the archive. In addition to that `Instance Manager` automatically checks
...
+
+Base backups are defined at the cluster level, declaratively,
+through the `backup` parameter in the cluster definition.
+
 You can define base backups in two ways: on-demand (through the `Backup`
 custom resource definition) or scheduled (through the `ScheduledBackup`
-customer resource definition, using a cron-like syntax). They both rely on
-`barman-cloud-backup` for the job (distributed as part of the application
-container image) to relay backups in the same endpoint, alongside WAL files.
+customer resource definition, using a cron-like syntax).
```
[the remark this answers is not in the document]
NiccoloFei: Has been fixed.

PR #2545, opened 2023-08-08
Fix PoolerTypeRO comment

wzzrd:
An "ro" PoolerType creates a pooler with the "ro" service as it's backend, which means it'll provide a connection pool over all replicas, excluding the primary.


leonardoce:
Sorry @wzzrd this fell off my radar. I'm now taking care of it.

leonardoce (approved):

leonardoce:
Hi @wzzrd. Can you fix the DCO in your commit when you have some free time? We'd like to have this fix in the next release.

PR #2548, opened 2023-08-09
feat: create PVC's to hold tablespaces

jsilvela:


krs2k:
great idea, I hope it goes to the master :)

pkg/reconciler/persistentvolumeclaim/requests.go
```diff
@@ -56,8 +56,9 @@ func reconcilePVCQuantity(
 ) error {
 	contextLogger := log.FromContext(ctx)
 	pvcRole := utils.PVCRole(pvc.Labels[utils.PvcRoleLabelName])
+	tablespace := pvc.Labels[utils.PvcTablespaceLabelName]
```
jsilvela, line 59: NOTE: this is TBD. This is assuming that the pvc was labeled with the tablespace name, and with that we can map back to the tablespace definition in the cluster spec.
Tao points out that since the PVC name is normalized `clustername-serial-tbs-tablespacename` we could parse the tablespace from the PVC name. Worth discussing.

pkg/reconciler/persistentvolumeclaim/resources.go
```diff
@@ -264,6 +301,8 @@ func getStorageSource(
 		source = &volumeSnapshots.Storage
 	case utils.PVCRolePgWal:
 		source = volumeSnapshots.WalStorage
+	case utils.PVCRolePgTablespace:
```
jsilvela, line 304:
```suggestion
	// TODO ad the moment, getting the Source for a tablespace will error out. Do something better
	case utils.PVCRolePgTablespace:
```

litaocdl (approved):

PR #2577, opened 2023-08-14
Fix typo in test

sjmiller609:
I'm troubleshooting a situation on Tembo cloud. I'm manually breaking an instance on purpose to check how error reporting will work.

I ran `ALTER SYSTEM SET shared_preload_libraries = 'oopsies';`, then restarted my instance using the restartedAt annotation. Now, my instance is stuck in readiness 0/1 but liveness probe is passing, however I expected liveness probe to fail because pg_isready is not success. I found liveness probe is passing because the condition "MightBeUnavailable" is true, but I don't know why that is yet, currently I suspect it's related to triggering a restart.

I noticed this typo in the test while reviewing how "MightBeUnavailable" works.

### DIFF
--- a/pkg/management/postgres/instance_test.go
+++ b/pkg/management/postgres/instance_test.go
@@ -214,7 +214,7 @@ var _ = Describe("check atomic bool", func() {
 		Expect(canBeChecked).To(BeTrue())
 	})
 
-	It("should recognize whether the instance might be unavailanle based on the setting", func() {
+	It("should recognize whether the instance might be unavailable based on the setting", func() {
 		instance.SetMightBeUnavailable(false)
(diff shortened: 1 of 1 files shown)


sjmiller609:
After deleting pod manually, liveness probe is failing, so I think this is indicates that MightBeUnavailable is set during restart, and it doesn't know if restart is in progress or if pod is failing to start due to some change applied during restart. I can check docs or code for some setting like "restart timeout"

leonardoce (approved):

phisco:
Hi, yes that's intentional, the operator waits for the instance to come back up after a restart, as that could take an arbitrary amount of time depending on how much data it has to replay.

PR #2583, opened 2023-08-16
fix: reconcile PodMonitor `labels` and `annotations`

### LINKED ISSUE #2582: ensure that PodMonitor resources are reconciled if metadata changes
Currently, the PodMonitors do not get updated if there is a change in the cluster inherited labels/annotation given that we check only for the spec.
 
We should fix this.

Line: https://github.com/cloudnative-pg/cloudnative-pg/blob/049e9426f3f8c1c218a3ab9efbd35646f7dce6d3/controllers/cluster_create.go#L841

armru:
Closes #2582

### DIFF
--- a/controllers/cluster_create.go
+++ b/controllers/cluster_create.go
@@ -836,6 +836,9 @@ func createOrPatchPodMonitor(
 	default:
 		origPodMonitor := podMonitor.DeepCopy()
 		podMonitor.Spec = expectedPodMonitor.Spec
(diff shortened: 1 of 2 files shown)


NiccoloFei (approved):

YanniHu1996 (approved):

mnencia (approved):

mnencia:
/ok-to-merge only unrelated failures in the test results

PR #2593, opened 2023-08-17
fix(docs): Cron schedule expression in ScheduledBackup

sjmiller609:
```
  kind: ScheduledBackup
....
  spec:
...
    # Kubernetes would interpret this as a daily backup at 6:42am
    schedule: 42 6 * * *
...
  status:
    # However, it is actually doing hourly backups at 6th minute, 42nd second
    lastCheckTime: "2023-08-17T18:06:42Z"
    lastScheduleTime: "2023-08-17T18:06:42Z"
    nextScheduleTime: "2023-08-17T19:06:42Z"
....
  ```
  
This seems like when I provide a five-term expression, we have the first term as seconds instead of minutes. In my opinion, it should either be interpreted like a kubernetes cron job, or have input validation error if there are only 5 terms.

### DIFF
--- a/docs/src/backup.md
+++ b/docs/src/backup.md
@@ -164,8 +164,9 @@ CloudNativePG. They are managed by the `ScheduledBackup` resource.
     Please refer to [`ScheduledBackupSpec`](cloudnative-pg.v1.md#postgresql-cnpg-io-v1-ScheduledBackupSpec)
     in the API reference for a full list of options.
 
-The `schedule` field allows you to define a *cron schedule* specification,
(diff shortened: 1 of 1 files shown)


leonardoce:
I agree with you @sjmiller609.
We should upgrade our codebase to use `robfig/cron` V3 to fix this issue
Can you file a ticket for that, with the details you shared in this PR?

In the meantime, I'll merge this PR fixing the documentation.

leonardoce (approved):

armru (approved):

mnencia (approved):

PR #2601, opened 2023-08-17
fix(docs): Default setting for replication slots HA configuration

sjmiller609:
The API reference conflicts with the other section of the documentation

> This feature, introduced in CloudNativePG 1.18, is now enabled by default and can be disabled via configuration. For details, please refer to the ["replicationSlots" section in the API reference](https://cloudnative-pg.io/documentation/1.20/api_reference/#ReplicationSlotsConfiguration). Here follows a brief description of the main options:
.spec.replicationSlots.highAvailability.enabled
if true, the feature is enabled (true is the default - since 1.20)

I am assuming this section is correct because it seems like more recent information.

Also, maybe we need to update the Helm chart here: https://github.com/cloudnative-pg/charts/blob/39c99bc63cb2f97e2d2895484da2bbec0ee6c71f/charts/cloudnative-pg/templates/crds/crds.yaml#L3154

There is also a note about this only being possible to set on initial creation, but I don't understand what will happen if an existing kubernetes cluster is updated to a new CRD where the default is true, or if we only have single-instance Clusters then try to set instances to 2 and enable this setting


sxd:
Closing since it's an outdated PR can't be merged

PR #2605, opened 2023-08-21
fix: evaluate PRESERVE_NAMESPACES as suffix

### LINKED ISSUE #2604: PRESERVE_NAMESPACES not working
We forgot to adjust the PRESERVE_NAMESPACES logic with the introduction of CreateUniqueNamespace.

Currently, the namespaces indicated by the env PRESERVE_NAMESPACES are preserved only if the E2E test case uses the old method CreateNamespace.

We should remedy this by introducing a string prefix check

armru:
Closes #2604

### DIFF
--- a/tests/utils/namespace.go
+++ b/tests/utils/namespace.go
@@ -19,6 +19,7 @@ package utils
 import (
 	"errors"
 	"fmt"
+	"strings"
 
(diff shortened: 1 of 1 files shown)


mnencia (approved):

mnencia:
/ok-to-merge does not affect e2e tests in CI environment

PR #2606, opened 2023-08-21
chore: move api documentation generation from k8s-api-docgen to genref

### LINKED ISSUE #2350: Docs: API reference embedded structs rendering issue
The generated API reference documentation is not properly rendering embedded structs with `json:",inline"` annotations.
See [docs](https://cloudnative-pg.io/documentation/current/api_reference/#SecretKeySelector) (code [here](https://github.com/cloudnative-pg/cloudnative-pg/blob/74b460ec02e152253e51d2090bebfde405ede629/docs/src/api_reference.md?plain=1#L1022)), which is shown to only have a `key` field, while it should also have a `name` inherited from the embedded `LocalObjectReference`, see https://github.com/cloudnative-pg/cloudnative-pg/blob/74b460ec02e152253e51d2090bebfde405ede629/api/v1/base_types.go#L34

fcanovai:
Uses [genref](https://github.com/kubernetes-sigs/reference-docs/tree/master/genref) to create the documentation instead of [k8s-api-docgen](https://github.com/EnterpriseDB/k8s-api-docgen). prometheus-operator abandoned the original version of it for [gen-crd-api-reference-docs](https://github.com/ahmetb/gen-crd-api-reference-docs), and that project now suggests to use alternatives due to lack of time for maintenance.

`k8s-api-docgen` was based on `omitempty`  directives, while `genref` uses `+optional`, so this PR includes a review of the API to align the two. It also specifies as mandatory some field that were actually required but not defined as such.

Closes  #2350


gbartolini (approved):
Thanks Francesco!

mnencia:
I think this PR sould not be squashed

.github/workflows/continuous-integration.yml
```diff
@@ -274,7 +274,7 @@ jobs:
 
       - name: Verify apidoc changes
         run: |
-          apidoc_file_path='docs/src/api_reference.md'
+          apidoc_file_path='docs/src/cloudnative-pg.v1.md'
```
sxd, line 277: why changing the name of the file here? I mean because it's the api reference cloudnative-pg sound like an index to make, this is the only changes that rings a bell on me

mnencia (approved):

leonardoce (approved):

Issue #2501, opened 2023-07-26
Proposal: Rename cluster label to cnpg_cluster


sjentzsch:
I would like to discuss a breaking change on observability. Our problem (and I will argue in a moment why it would not only affect us): Metrics from cnpg expose the `cluster` label (for example `cnpg_collector_up` ), and the [Grafana dashboard](https://github.com/cloudnative-pg/cloudnative-pg/blob/main/docs/src/samples/monitoring/grafana-dashboard.json) builds upon that. However, the `cluster` label is a commonly-used general label often referring to the Kubernetes cluster the data originates from, and used by many as an external label when pushing metrics etc.. Those two concept cannot co-exist, and we decided to relabel the cnpg `cluster` label to `app_cluster`, but then we of course cannot use your upstream dashboard anymore (might also affect alerts at some point in time). My point is, with the current naming, the support for multi-cluster scenarios is quite restricted, and I am afraid we are not the only ones running cnpg in multi-cluster scenarios. My proposal is to rename the `cluster` label to `cnpg_cluster`, knowing that it will break existing alerts & dashboards for users. What do you think?

Feel free to also add to the Slack Thread: https://cloudnativepg.slack.com/archives/C03AX0J5P29/p1690284881972749

gbartolini:
I am concerned about the introduction of this breaking change. Can we do something like what we did for the log field names?

https://github.com/cloudnative-pg/cloudnative-pg/commit/936b930eaca7d0c85d571a7eac3fc6f401105551

And then get the plugin to control that in the generate command as part of the options for the deployment of the operator. See:

https://github.com/cloudnative-pg/cloudnative-pg/commit/91efc27e1a514ea1f4e0d003f3296dd611a304a5

Any comments?

That way we won't be breaking existing installations, but give the possibility for users to change the mapping in the installation and update their dashboard. In a few releases we can change the default.

phisco:
I'd suggest adding both for a while, there should be no issue in doing so and the cluster label would be overridden if needed. We can even make the label to use a parameter in the dashboard, to ease the transition.

gbartolini:
Hi Philippe,

> I'd suggest adding both for a while, there should be no issue in doing so and the cluster label would be overridden if needed. We can even make the label to use a parameter in the dashboard, to ease the transition.

I am probably missing something here. But wouldn't that mean that the exporter everytime sends the information twice?

phisco:
[...]

PR #2608, opened 2023-08-21
fix(prometheus): add the cnpg_cluster label

achanda:
Closes: #2501

This PR renames the cluster label in prometheus to cnpg_cluster since that label is commonly used to refer to the k8s cluster. This is a breaking change.


achanda:
I wonder if we should do this in two separate releases: the first one adding the extra label and the second one removing the deprecated label. Thoughts?

sjentzsch:
Small notes from my side (non-maintainer):

We might also want to apply same changes to `grafana-configmap.yaml` as it's the `grafana-dashboard.json` wrapped inside a configmap (would be best of course to not have this redundancy anymore ...)

Instead of adding both labels at the same time as an intermediate step, we could also make us of (optional) relabeling at `PodMonitor` level (https://cloudnative-pg.io/documentation/1.20/monitoring/#prometheus-operator-example), so e.g. 

```
  podMetricsEndpoints:
  - port: metrics
    relabelings:
    - sourceLabels: [cnpg_cluster]
      targetLabel: cluster
      action: replace
```

to rewrite `cnpg_cluster` back to `cluster`, if people want it. But thinking about it twice, as not everyone might use PodMonitor, probably exposing both metrics at the same time might indeed be a thing. Or, as we might also not want to maintain both dashboard types at the same time, just do the breaking change.

gbartolini:
@achanda can you please try again by following the suggestions that @phisco gave in the original issue #2501? thanks

achanda:
Hi @gbartolini @sxd does this one look good now? Please let me know if this PR needs anything else.

gbartolini (approved):

sxd:
Hi @achanda 

This PR it's actually not working I'm going to throw the observability testing here but you can get the error here:
```
{"level":"info","ts":"2023-09-21T23:30:23Z","msg":"Query with the same name already found. Overwriting the existing one.","logging_pod":"cluster-example-1","queryName":"pg_database"}
{"level":"info","ts":"2023-09-21T23:30:23Z","msg":"Query with the same name already found. Overwriting the existing one.","logging_pod":"cluster-example-1","queryName":"pg_database"}
panic: inconsistent label cardinality: expected 2 label values but got 1 in []string{"cluster-example"}

goroutine 321 [running]:
github.com/prometheus/client_golang/prometheus.(*GaugeVec).WithLabelValues(...)
	/home/zeus/go/pkg/mod/github.com/prometheus/client_golang@v1.16.0/prometheus/gauge.go:238
github.com/cloudnative-pg/cloudnative-pg/pkg/management/postgres/webserver/metricserver.(*Exporter).collectPgMetrics(0xc0002964b0, 0xc00068f6c8?)
	pkg/management/postgres/webserver/metricserver/pg_collector.go:361 +0xff4
github.com/cloudnative-pg/cloudnative-pg/pkg/management/postgres/webserver/metricserver.(*Exporter).Collect(0xc0002964b0, 0xc00068f760?)
	pkg/management/postgres/webserver/metricserver/pg_collector.go:298 +0x2c
github.com/prometheus/client_golang/prometheus.(*Registry).Gather.func1()
	/home/zeus/go/pkg/mod/github.com/prometheus/client_golang@v1.16.0/prometheus/registry.go:455 +0x102
created by github.com/prometheus/client_golang/prometheus.(*Registry).Gather in goroutine 285
	/home/zeus/go/pkg/mod/github.com/prometheus/client_golang@v1.16.0/prometheus/registry.go:547 +0xbab
```

sxd:
There's a new PR

PR #2612, opened 2023-08-22
feat: add phase `PhaseWaitNotActiveInstances`

### LINKED ISSUE #2611: explicitly communicate that we are waiting for the instances to become active
Currently at the line: https://github.com/cloudnative-pg/cloudnative-pg/blob/c9bfd11f09cef524a7379c6659d0e89dcaa06ab0/controllers/cluster_controller.go#L483-L486
We wait for the instances to become ready without setting a proper phase. 

This can lead to the operator remaining in a previous phase (ex: failing over) even when the work was already done because the operator cannot reach the code to set a new phase.

We should add a new phase that communicates that we are waiting for the instances to become active

armru:
Closes #2611


armru:
Very risky PR, has a lot of implications. Adding the label

litaocdl (approved):

mnencia (approved):

mnencia:
/ok-to-merge E2e failure are unrelated (replication slots)

phisco (approved):
except for the debug log which still needs to be updated as we discussed, LGTM! 🎉

PR #2625, opened 2023-08-25
fix: ensure that `ContinuousArchiving` is set as failed in case of errors

### LINKED ISSUE #2624: ensure that `ContinuousArchiving` is set to failing in case of any errors
Currently in our code at: https://github.com/cloudnative-pg/cloudnative-pg/blob/0bed9df55df3657910cb73efadd6be85954fbb41/internal/cmd/manager/walarchive/cmd.go#L97
we flag the ContinuousArchiving as failed only if we encounter certain errors and this is wrong.

We should set the ContinuousArchiving as failed in the case that we encounter any error

armru:
Closes #2624


armru:
@mnencia we should back port to 1.18 too if you agree

litaocdl (approved):

Issue #1313, opened 2023-01-09
No target backup found


danotrampus:
Hi,

I'm trying to replicate what the doc says in [Replica Cluster](https://cloudnative-pg.io/documentation/1.16/replica_cluster/), but when i try to create the replica cluster on a second provider the recovery process gives me the following error:
```
$ kubectl logs -n starten-29-dev odoo-db-1-full-recovery-s85kw
Defaulted container "full-recovery" out of: full-recovery, bootstrap-controller (init)
{"level":"info","ts":1673244415.202184,"msg":"Recovering from external cluster","logging_pod":"odoo-db-1","sourceName":"odoo-db"}
{"level":"error","ts":1673244415.7991278,"msg":"Error while restoring a backup","logging_pod":"odoo-db-1","error":"no target backup found","stacktrace":"github.com/cloudnative-pg/cloudnative-pg/pkg/management/log.(*logger).Error\n\tpkg/management/log/log.go:127\ngithub.com/cloudnative-pg/cloudnative-pg/pkg/management/log.Error\n\tpkg/management/log/log.go:165\ngithub.com/cloudnative-pg/cloudnative-pg/internal/cmd/manager/instance/restore.restoreSubCommand\n\tinternal/cmd/manager/instance/restore/cmd.go:84\ngithub.com/cloudnative-pg/cloudnative-pg/internal/cmd/manager/instance/restore.NewCmd.func2\n\tinternal/cmd/manager/instance/restore/cmd.go:59\ngithub.com/spf13/cobra.(*Command).execute\n\tpkg/mod/github.com/spf13/cobra@v1.6.1/command.go:916\ngithub.com/spf13/cobra.(*Command).ExecuteC\n\tpkg/mod/github.com/spf13/cobra@v1.6.1/command.go:1044\ngithub.com/spf13/cobra.(*Command).Execute\n\tpkg/mod/github.com/spf13/cobra@v1.6.1/command.go:968\nmain.main\n\tcmd/manager/main.go:64\nruntime.main\n\t/opt/hostedtoolcache/go/1.19.4/x64/src/runtime/proc.go:250"}
```
When in fact there are multiple backups already store on my S3 backup store:
![image](https://user-images.githubusercontent.com/11494867/211251902-3a3d62fd-8f0f-419f-9d58-419fa46efd33.png)

**¿What am i doing wrong?**

### Additional Information

Kubernetes version: 1.24.8
CloudNativePG Operation version: 1.18.1

[...]

PR #2630, opened 2023-08-25
fix: use correct endpointURL in Linode example (#2629)

### LINKED ISSUE #2629: Fix linode example in `Other S3-compatible Object Storages providers`
The [Linode example](https://github.com/cloudnative-pg/cloudnative-pg/blob/0bed9df55df3657910cb73efadd6be85954fbb41/docs/src/appendixes/object_stores.md#L106-L117)  in the `Other S3-compatible Object Storages providers` section of the docs specifies to use the following:

```yaml
    destinationPath: "<destination path here>"
    endpointURL: "https://bucket.us-east1.linodeobjects.com"
```

However according to the [Linode docs](https://www.linode.com/docs/products/storage/object-storage/guides/urls/#cluster-url-s3-endpoint) the endpointURL should specify the region and not the bucket.

The example should actually be:

```yaml
    destinationPath: "s3://bucket"
    endpointURL: "https://us-east1.linodeobjects.com"
```

Interestingly, if the example in the docs is followed and the bucket URL is used as the value for `endpointURL`, the resulting behaviour is not outright failure and appears to be undefined. However, as demonstrated in both [#1313](https://github.com/cloudnative-pg/cloudnative-pg/issues/1313#issuecomment-1376594252) and EnterpriseDB/barman#835 this undefined behaviour makes debugging quite time consuming.

mikewallace1979:
Updates the endpointURL for the Linode example in object_stores.md so that it does not include the bucket name and moves the bucket name into the S3 URL provided to destinationPath.

This is based on the recommendations in the current Linode documentation at:

    https://www.linode.com/docs/products/storage/object-storage/guides/urls/#cluster-url-s3-endpoint

Closes #2629

### DIFF
--- a/docs/src/appendixes/object_stores.md
+++ b/docs/src/appendixes/object_stores.md
@@ -110,8 +110,8 @@ kind: Cluster
 spec:
   backup:
     barmanObjectStore:
-      destinationPath: "<destination path here>"
-      endpointURL: "https://bucket.us-east1.linodeobjects.com"
+      destinationPath: "s3://bucket/"
+      endpointURL: "https://us-east1.linodeobjects.com"
       s3Credentials:
         [...]
 ```


sxd:
Hi @mikewallace1979 

I don't see a reason to change this, since when you are asked for the endpoint your are asking for the bucket endpoint, what linode is allowing here, it's a nice feature, but it's common to point to the bucket and then a path inside the bucket, this allow users to have the same endpoint but a different path if they want different directories inside the same bucket.

Regards!

mikewallace1979:
@sxd I wanted to change this because the bucket name should not be included in the endpoint URL. According to [Linode](https://www.linode.com/docs/products/storage/object-storage/guides/urls/#cluster-url-s3-endpoint):

> Cluster URL (S3 Endpoint)
>
> The cluster URL for Object Storage is unique to each data center. This URL can be used for accessing all of your Object Storage resources within a certain region. When used with third party tools, this URL might also be referred to as the S3 endpoint.
>
> Cluster URL: https://[cluster-id].linodeobjects.com/, replacing [cluster-id] with the id of your data center.

The docs currently suggest using the *bucket* URL for the endpoint URL - if a user does this then they will end up with a configuration which works for some requests and does not for others. This is the case when using either barman cloud or the AWS cli to interact with the bucket. Debugging the resulting behaviour is time consuming, as demonstrated in both [#1313](https://github.com/cloudnative-pg/cloudnative-pg/issues/1313#issuecomment-1376594252) and [EnterpriseDB/barman#835](https://github.com/EnterpriseDB/barman/issues/835).

So the line I really care about changing here is:

    endpointURL: "https://bucket.us-east1.linodeobjects.com"

because it suggests to users they should include the bucket name in the endpoint, which is not going to end well.

Keeping an example path in the `destinationPath` line would be fine, so the docs could say something like this:

    destinationPath: "s3://bucket/<destination path here>"
    endpointURL: "https://us-east1.linodeobjects.com"

sxd (approved):

gbartolini (approved):

Issue #2530, opened 2023-08-02
[Feature]: support changing image and PostgreSQL settings simultaneously


GabriFedi97:
Updating the `shared preload libraries` and the `imageName` at the same time, on clusters with one or more replicas and with `primaryUpdateMethod: switchover` can take the cluster in a broken state.
This happens when the new `shared preload libraries` adds values referring to pg extensions that are available on the new image, but they are not in the old one.

Looks like that the operator starts updating the replicas first, which goes fine. When it comes to update the primary, the switchover happens and the old-primary pod starts getting updated. This last pod update is done using the old image along with the new configuration, so if a shared preload libraries's item in the new config is not installed in the old image, the pod start crashing saying it can't find the extension file.

Example:

```
NAME              AGE     INSTANCES   READY   STATUS              PRIMARY
cluster-example   4h17m   3           2       Upgrading cluster   cluster-example-1

NAME                READY   STATUS             RESTARTS       AGE
cluster-example-1   1/1     Running            0              33m
cluster-example-2   0/1     CrashLoopBackOff   11 (44s ago)   39m
cluster-example-3   1/1     Running

k get pod cluster-example-1 -o yaml | yq .spec.containers[0].image
***/postgres:15.3-9-debian --> new image

k get pod cluster-example-2 -o yaml | yq .spec.containers[0].image
***/postgres:15.3-7-debian --> old image

k logs cluster-example-2
...
{"level":"info","ts":"2023-08-02T14:46:44Z","logger":"postgres","msg":"2023-08-02 14:46:44.655 UTC [28] FATAL:  could not access file \"autocluster\": No such file or directory","pipe":"stderr","logging_pod":"cluster-example-2"}
...
```

gbartolini:
This issue was raised again today by an EDB customer during a call when I was present.

Given that we are now working also on extension images in #7188, we should keep this problem in mind (as noted by @itay-grudev as well).

gbartolini:
A first implementation could be to prioritize the update of the container image(s) over the changes in configuration which, in a declarative world, should imply that they are done on the new system (either a new version of Postgres or a new image).

NiccoloFei:
[...]

PR #2650, opened 2023-08-30
fix: check image before reconciling the pg config

GabriFedi97:
Skipping the pg config reconciliation if the pod image differs from the one declared in the cluster spec. During an image upgrade, the cluster pg configuration might be tight to the new image declared in the spec. Hence, if the instance manager applies the pg configuration to the old image, it may cause Postgres to start failing. 

Issue #2530


armru:
Hello @GabriFedi97, the proposed PR is not passing the E2E tests

GabriFedi97:
Hey @armru, I didn't get to check the E2E tests logs before the retention period. Any chances to run them again?

armru:
@GabriFedi97 rebased and started test

armru:
@GabriFedi97 ran an E2E test suite

GabriFedi97:
Thanks @armru.
Looks like e2e are not passing for the `Online Upgrade` and `Declarative Volume Snapshot` features. 
I haven't investigated the second one yet, but the first one is failing because the proposed patch makes the operator changing the environment of the operands, adding the `Image` variable. In that case the pods get upgraded with a restart anyway and not through the binary replace.

PR #2670, opened 2023-09-03
docs: document write amplification issue

weiyentan:
Provide a recommendation on block storage solutions to set replicas to 1 in order to avoid write amplification.

### DIFF
--- a/docs/src/storage.md
+++ b/docs/src/storage.md
@@ -445,7 +445,10 @@ You can use a pre-provisioned volume in CloudNativePG by following these steps:
 ## Block storage considerations (Ceph, Longhorn)
 Most block storage solutions in Kubernetes suggest to have multiple 'replicas' of a volume
 to improve resiliency. This works well for workloads that do not have resiliency built into the 
-application. However Cloud Native Postgres has this resiliency built ino the storage of 
-the Postgres Cluster through the number of instances.
+application. However Cloud Native Postgres has this resiliency built into 
+the Postgres Cluster through the number of instances and the persistent volume that is
+attached to them. 
 
-Therefore 
+In these cases it makes sense to define the  storage class in the cnpg kubernetes cluster
+to be defined aas 1 replica. By having additional replicas defined in the storage solution like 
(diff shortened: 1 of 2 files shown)


docs/src/storage.md
```diff
@@ -435,3 +435,14 @@ You can use a pre-provisioned volume in CloudNativePG by following these steps:
...
+## Block storage considerations (Ceph, Longhorn)
+Most block storage solutions in Kubernetes suggest to have multiple 'replicas' of a volume
+to improve resiliency. This works well for workloads that do not have resiliency built into the 
+application. However Cloud Native Postgres has this resiliency built into 
+the Postgres Cluster through the number of instances and the persistent volume that is
+attached to them. 
+
+In these cases it makes sense to define the  storage class in the cnpg kubernetes cluster
+to be defined as  1 replica. By having additional replicas defined in the storage solution like 
+Longhorn and  Ceph adds on more disk i/o and space being used unnecessarily.
```
gbartolini, line 448:
```suggestion
## Block storage considerations (Ceph, Longhorn)

Most block storage solutions in Kubernetes suggest to have multiple 'replicas' of a volume
to improve resiliency. This works well for workloads that don't have resiliency built into the 
application. However, CloudNativePG has this resiliency built directly into the Postgres `Cluster`
through the number of instances and the persistent volumes that are attached to them. 

In these cases it makes sense to define the  storage class used by the Postgres clusters
to be defined as 1 replica. By having additional replicas defined in the storage solution like 
Longhorn and Ceph you might incur in the issue known as write amplification, unnecessarily
increasing disk I/O and space used.
```

gbartolini (approved):

gbartolini:
@weiyentan do you mind signing your commit so that we can merge it? thanks

see https://github.com/cloudnative-pg/cloudnative-pg/tree/main/contribute#sign-your-work

weiyentan:
Thanks @gbartolini i have resubmitted with this

leonardoce (approved):

PR #2685, opened 2023-09-06
refactor: refactor tablespace to controller

litaocdl:
This patch refactor the tablespace reconciler to controller.


api/v1/cluster_webhook.go
```diff
@@ -1539,6 +1550,21 @@ func (r *Cluster) validateTablespacesNames() field.ErrorList {
...
+				"tablespace name must not be a reserved name"))
+		}
+
+		if !tablespaceNameRegex.MatchString(name) {
+			result = append(result, field.Invalid(
+				field.NewPath("spec", "tablespaces"),
+				name,
+				"tablespace name must consist of lower case alphanumeric characters,'-' or '.', "+
+					"and must start and end with an alphanumeric character"))
+		}
```
jsilvela, line 1566: I don't think we should do this. In Postgres, tablespace names don't have to obey this. We need transform functions to convert back and forth for K8s labels and such.

litaocdl: I agree we should remove this restriction. This is the pvc name restriction, definitely will remove it. as we put the tablespace name into the label of the pvc, we can use the regular expression to valid the converted pvc name.

jsilvela, line 1566: this should have a matching unit test

internal/management/controller/tablespaces/manager.go
```diff
@@ -0,0 +1,60 @@
...
+
+	"k8s.io/apimachinery/pkg/types"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	apiv1 "github.com/cloudnative-pg/cloudnative-pg/api/v1"
+	"github.com/cloudnative-pg/cloudnative-pg/pkg/management/postgres"
+)
+
+// TablespaceReconciler is a Kubernetes manager.Runnable
```
jsilvela, line 14: it's not a Runnable. Missing the Start() method, no?

litaocdl: It is not a runnable, I put the method to create reconciler into manager.go

internal/management/controller/tablespaces/reconciler.go
```diff
@@ -18,52 +18,145 @@ package tablespaces
...
 	"github.com/cloudnative-pg/cloudnative-pg/internal/management/controller/tablespaces/infrastructure"
 	"github.com/cloudnative-pg/cloudnative-pg/pkg/management/log"
-	"github.com/cloudnative-pg/cloudnative-pg/pkg/management/postgres"
 )
 
-// Reconcile triggers reconciliation of declarative tablespaces, gets their status, and
-// updates it into the cluster Status
-func Reconcile(
+// Reconcile is the main reconciliation loop for the instance
+func (r *TablespaceReconciler) Reconcile(
```
litaocdl, line 34: This is the reconciler main method for tablespace reconciler

jsilvela (approved):

internal/management/controller/tablespaces/reconciler.go
```diff
@@ -18,52 +18,145 @@ package tablespaces
...
+		if apierrors.IsNotFound(err) {
+			// The cluster has been deleted.
+			// We just need to wait for this instance manager to be terminated
+			contextLogger.Debug("Could not find Cluster")
+			return reconcile.Result{}, nil
+		}
+		return reconcile.Result{}, fmt.Errorf("could not fetch Cluster: %w", err)
+	}
+
+	if cluster.ShouldCreateTablespaces() {
```
jsilvela, line 67:
```suggestion
	if !cluster.ShouldCreateTablespaces() {
```

mnencia:
We will continue the review in the feature branch. Merging this refactor.

PR #2690, opened 2023-09-06
fix: instances always restarting on operator upgrade

jsilvela:
The PodSpec drift checker was forcing instance rollout
whenever there was an operator upgrade.
This patch restores the logic allowing in-place instance manager
upgrades.


YanniHu1996 (approved):
LGTM

mnencia (approved):

sxd (approved):

PR #2691, opened 2023-09-06
chore: properly quote the values inside LDAP fields for pg_hba.conf

### LINKED ISSUE #2688: ldapBindPassword must be quoted when added to the pg_hba.conf
If the password contains special characters, it can confuse the pg_hba.conf parser and lead to confusing errors.

https://github.com/cloudnative-pg/cloudnative-pg/blob/2a104f9f46004413c4af58d2cf43d5233125eb6a/pkg/management/postgres/configuration.go#L170-L171

We need to quote all the user-provided values we put in pg_hba.conf

sxd:
Only some of the values quoted meaning that that could lead to some issues with the pg_hba.conf parser in PostgreSQL. Also, there was an issue with secrets containing a new line at the end, now we add a `\` before the newline so PostgreSQL will read it only as one line.

Closes #2688

### DIFF
--- a/pkg/management/postgres/configuration.go
+++ b/pkg/management/postgres/configuration.go
@@ -171,7 +171,9 @@ func buildLDAPConfigString(cluster *apiv1.Cluster, ldapBindPassword string) stri
 
 		ldapConfigString += fmt.Sprintf(` ldapbasedn="%s" ldapbinddn="%s" ldapbindpasswd="%s"`,
 			ldapConfig.BindSearchAuth.BaseDN, ldapConfig.BindSearchAuth.BindDN,
-			strings.ReplaceAll(ldapBindPassword, "\n", "\\n"))
+			// if there is a newline, finsh the line with a backslash, and carry on
+			// in the next line. (see https://www.postgresql.org/docs/current/auth-pg-hba-conf.html)
+			strings.ReplaceAll(ldapBindPassword, "\n", "\\\n"))
 		if ldapConfig.BindSearchAuth.SearchFilter != "" {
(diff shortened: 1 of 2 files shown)


jsilvela (approved):

pkg/management/postgres/configuration.go
```diff
@@ -169,15 +169,15 @@ func buildLDAPConfigString(cluster *apiv1.Cluster, ldapBindPassword string) stri
 			"search attribute", ldapConfig.BindSearchAuth.SearchAttribute,
 			"search filter", ldapConfig.BindSearchAuth.SearchFilter)
 
-		ldapConfigString += fmt.Sprintf(" ldapbasedn=\"%s\" ldapbinddn=\"%s\" ldapbindpasswd=\"%s\"",
+		ldapConfigString += fmt.Sprintf(` ldapbasedn="%s" ldapbinddn="%s" ldapbindpasswd="%s"`,
 			ldapConfig.BindSearchAuth.BaseDN, ldapConfig.BindSearchAuth.BindDN,
-			strings.ReplaceAll(ldapBindPassword, "\n", "\\\n"))
+			strings.ReplaceAll(ldapBindPassword, "\n", `\n`)) // escape newlines
```
mnencia, line 174: I find this line confusing. I would have preferred to use the same quote type here, as it was before.

mnencia (approved):

pkg/management/postgres/configuration.go
jsilvela: you're probably right. I wondered too. Reverting.

mnencia (approved):

PR #2699, opened 2023-09-08
fix: old primary pod need restart when rolling update tablespace chan…

litaocdl:
this patch merge to feature branch, make sure when primaryUpdateMethod 
is switchover, add a tablespace, the old primary will restart to get the new volume
mounted.

### DIFF
--- a/pkg/reconciler/persistentvolumeclaim/resources.go
+++ b/pkg/reconciler/persistentvolumeclaim/resources.go
@@ -159,10 +159,8 @@ func getExpectedPVCsFromCluster(cluster *apiv1.Cluster, instanceName string) []e
 		roles = append(roles, utils.PVCRolePgWal)
 	}
 
-	if cluster.ShouldCreateTablespaces() {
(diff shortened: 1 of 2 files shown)


pkg/reconciler/persistentvolumeclaim/resources.go
```diff
@@ -159,6 +159,10 @@ func getExpectedPVCsFromCluster(cluster *apiv1.Cluster, instanceName string) []e
 		roles = append(roles, utils.PVCRolePgWal)
 	}
 
+	if cluster.ShouldCreateTablespaces() {
+		roles = append(roles, utils.PVCRolePgTablespace)
+	}
+
```
jsilvela, line 165: I think this is not necessary. The function buildExpectedPVCs below, does the work...

jsilvela: Confirmed. The unit test below works without this. Deleting.
As far as I know, Marco has feedback about this, and I do think we will need to restructure.

PR #2728, opened 2023-09-11
refactor: use `override.conf` instead of `postgresql.auto.conf`

### LINKED ISSUE #2727: Stop using `postgresql.auto.conf` for server configuration
At the moment, CloudNativePG writes its own settings in the `postgresql.auto.conf` file directly. For example:

```
recovery_target_timeline = 'latest'
restore_command = '/controller/manager wal-restore --log-destination /controller/log/postgres.json %f %p'
primary_slot_name = '_cnpg_freddie_1_initdb'
primary_conninfo = 'host=freddie-rw user=streaming_replica port=5432 sslkey=/controller/certificates/streaming_replica.key sslcert=/controller/certificates/streaming_replica.crt sslrootcert=/controller/certificates/server-ca.crt application_name=freddie-1-initdb sslmode=verify-ca'
```

Use a separate file called `override.conf` for this purpose (feel free to change the name during the analysis if you find a better one). Related to #2725

gbartolini:
Up to now, CloudNativePG has been using the `postgresql.auto.conf` to manage PostgreSQL GUCs that coordinate activities in a PostgreSQL cluster in terms of HA and DR, primarily. Given that `postgresql.auto.conf` is used directly by the `ALTER SYSTEM` command in PostgreSQL, we should use a different file, called `override.conf` for the above purposes.

Closes #2727


gbartolini:
Incomplete ... a stub for now.

gbartolini:
The patch works fine on a new instance. What is missing is the case of the upgrade of the operator on an existing cluster. What is happening now is that:

- the override.conf file is created and managed
- the override.conf file is ignored as it is not included from the postgresql.conf file
- the postgresql.auto.conf is left as it was (which is not good)

mnencia:
Superseded by https://github.com/cloudnative-pg/cloudnative-pg/pull/2812

PR #2796, opened 2023-09-15
fix(doc): repair bench codeblock

thomasboussekey:
... and update 1 job name into bench configuration

Hope It will fix this incorrect rendering 
image


leonardoce (approved):

thomasboussekey:
Fixed by #2621

PR #2799, opened 2023-09-15
fix: avoid snapshot executor timeout

armru:
This patch makes the wait.Backoff of the snapshot executor more frequent avoiding potential timeouts


leonardoce:
This is not needed with the latest changes we merged.

PR #2812, opened 2023-09-19
feat: stop using postgresql.auto.conf to control Postgres

### LINKED ISSUE #2727: Stop using `postgresql.auto.conf` for server configuration

YanniHu1996:
CloudNativePG was writing its own settings in the `postgresql.auto.conf` file directly.
For example:

```
recovery_target_timeline = 'latest'
restore_command = 'XXXX'
primary_slot_name = 'XXXX'
primary_conninfo = 'XXXX'
```

We are now using a separate file called `override.conf`.

Closes #2727


pkg/management/postgres/configuration.go
```diff
@@ -304,24 +305,64 @@ func createStandbySignal(pgData string) error {
...
+		"migratedOptions", migratedOptions)
+	changed, err = configfile.UpdatePostgresConfigurationFile(overrideConfigFile, migratedOptions)
+	if err != nil {
+		return false, err
+	}
+
+	// add include `override.conf` at the end of the `postgresql.conf` file
+	err = fileutils.AppendStringToFile(
+		path.Join(pgData, "postgresql.conf"),
+		fmt.Sprintf("# load CloudNativePG override configuration\ninclude '%v'\n", constants.PostgresqlOverrideConfigurationFile),
```
litaocdl, line 357: we should log instead of fmt

litaocdl (approved):

pkg/management/postgres/instance_replica.go
```diff
@@ -34,30 +34,29 @@ func (instance *Instance) RefreshReplicaConfiguration(
...
-	// The "archive_mode" setting was used to be overridden in the "postgresql.auto.conf"
-	// if the server was a designated primary. We need make sure to remove it
-	// and fall back on the value defined in "custom.conf".
-	// TODO: Removed this code together the RemoveArchiveModeFromPostgresAutoConf function
-	// TODO: when enough time passed since 1.12 release
-	changed, err = removeArchiveModeFromPostgresAutoConf(instance.PgData)
+	// TODO: Remove this code when enough time has passed since 1.21 release
+	//       This is due to the operator switching from postgresql.auto.conf
+	//       to override.conf for coordinating replication configuration
+	changed, err = migratePostgresAutoConfFile(ctx, instance, true)
```
armru, line 40: @litaocdl @YanniHu1996 
Can't we assume that if the new file exists, the migration is not needed? This would avoid running the code when not needed.

litaocdl: Good suggestion! In the case of inplace upgrade, i think yes, we can leverage the existance of override.conf to skip the migration. for the. restore case, probably difficult as the file is created as empty in temp directory. but the two case can easy split, I will try.

litaocdl: @armru I made a new commit which skip the migration by checking if the override.conf exist or not. this is for upgrade test case. for migration case, as override.conf is created in advance and only migrate one, I will let it force migrate, but migration will also be ignored if there is no content in postgresql.auto.conf (which means backup is already based on override.conf)

mnencia:
E2E test on EDB fork https://github.com/EnterpriseDB/cloudnative-pg/actions/runs/6326944385

mnencia:
E2E on EDB fork https://github.com/EnterpriseDB/cloudnative-pg/actions/runs/6336886076

gbartolini:
Given that the current implementation, differently from the initial one, should be entirely transparent for the end user, I think we could make this slip on main after 1.21.0 is released, so that we have more bandwidth for testing and reduce the risk of breaking existing environments.

gbartolini (approved):

leonardoce:
I don't like the way we parse and manipulate the PostgreSQL config files. Unfortunately, it looks like this is a needed step.
I hope in the future we'll be able to prevent `ALTER SYSTEM` and make 80% of this code not needed anymore.

leonardoce (approved):

gbartolini:
We agree that this solution is not unfortunately ideal, but we agreed in a call that for the time being we keep this behavior and remove it in a future version when we have stabilized the alter/disable system stuff.

leonardoce:
E2e tests are fine

PR #2816, opened 2023-09-19
fix: Skip replication slot reconciliation if fenced

### LINKED ISSUE #2784: Skip replication slot synchronization on a fenced instance
If an instance is fenced, the log fills with the following lines:

```
{"level":"info","ts":"2023-09-13T14:51:23Z","logger":"Replicator","msg":"synchronizing replication slots","logging_pod":"kubecon-na-23-eu-central-1-2","err":"getting replication slot status from local: failed to connect to `host=/controller/run user=postgres database=postgres`: dial error (dial unix /controller/run/.s.PGSQL.5432: connect: no such file or directory)"}
```

We should be smarter, and prevent that from happening - similarly to what we do with metrics collection, which correctly reports:

``` 
{"level":"info","ts":"2023-09-13T14:58:22Z","msg":"metrics collection skipped due to fencing","logging_pod":"kubecon-na-23-eu-central-1-2"}
```

YanniHu1996:
closes #2784

### DIFF
--- a/internal/management/controller/slots/runner/runner.go
+++ b/internal/management/controller/slots/runner/runner.go
@@ -78,12 +78,6 @@ func (sr *Replicator) Start(ctx context.Context) error {
 				updateInterval = newUpdateInterval
 			}
 
-			// If the instance is fenced, skip reconciliation
-			if sr.instance.IsFenced() {
-				contextLog.Info("skip replication slots reconciliation due to fencing")
-				continue
-			}
-
 			err := sr.reconcile(ctx, config)
 			if err != nil {
 				contextLog.Warning("synchronizing replication slots", "err", err)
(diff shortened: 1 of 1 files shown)


armru (approved):

internal/management/controller/slots/runner/runner.go
```diff
@@ -98,6 +98,13 @@ func (sr *Replicator) reconcile(ctx context.Context, config *apiv1.ReplicationSl
 		}
 	}()
 
+	contextLog := log.FromContext(ctx)
+
+	if sr.instance.IsFenced() {
+		contextLog.Debug("Replication slots reconciliation skipped: instance is fenced.")
```
mnencia, line 104: I would set it to `Trace`

litaocdl: yeah, trace is fine, let me change

litaocdl (approved):

PR #2830, opened 2023-09-20
feat: add ephemeral-storage limit for emptyDir's

### LINKED ISSUE #1548: Pods should set Ephemeral storage limit
As discussed with EDB team during an interlock call between EDB and IBM,  hosted by Prasa U,  we agreed to open this ticket to track this request.  

Even though EDB pods are not using ephemeral storage today and no plan to use it down the row, to be safe and avoid any potential trouble down the road,  EDB team will look at adding Ephemeral limit setting  with very minimal value (such as 10MB or less). 

Some certification requirements in Cloud pak for data depend on such settings to pass (instead of documenting with some exceptions)

jlong49:
Closes #1548


api/v1/cluster_types.go
```diff
@@ -315,6 +315,10 @@ type ClusterSpec struct {
 	// +optional
 	Resources corev1.ResourceRequirements `json:"resources,omitempty"`
 
+	// EmptyDirLimit set the limit for the ephemeral volumes
+	// +optional
+	EmptyDirLimit *string `json:"emptyDirLimit,omitempty"`
```
mnencia, line 320: I do not like this field name. What about

```suggestion
	EmptyDirLimit *string `json:"ephemeralVolumeSizeLimit,omitempty"`
```

leonardoce: I like `ephemeralVolumeSizeLimit` too

leonardoce (approved):

leonardoce (approved):

mnencia (approved):

.wordlist-en-custom.txt
```diff
@@ -114,6 +114,7 @@ EKS
 EOF
 EOL
 EmbeddedObjectMetadata
+EmptyDirLimit
```
mnencia, line 117: This is an old name. We need to recheck the wordlist.

leonardoce: Good catch. I'm regenerating it.

leonardoce:
We still miss the documentation, I cannot merge it unfortunately

leonardoce:
@jlong49 the patch has been fixed to address the difference between the `emptyDir` volume, which is storage-based, and the `shm` volume, which is allocated in memory.
The implementation has been changed deeply, and I'd appreciate it if you could take another look and write the documentation so that we can merge it.

leonardoce (changes requested):
Missing documentation.

leonardoce:
Thank you @jlong49!

gbartolini:
Mhh ... why did we end up creating a new page for the volumes? For example, we already define projected volumes here: https://cloudnative-pg.io/documentation/current/cluster_conf/

I'd personally avoid creating a new page for just those few entries. What do others think?

leonardoce (approved):

PR #2838, opened 2023-09-21
feat: bring backup status of snapshots to parity with barman

### LINKED ISSUE #2813: Enrich the `Backup` status with volume snapshots
At the moment, the status of a `Backup` resource is not coherent between the object store case and the volume snapshot one. For example, this is a backup taken with object stores:

```
Status:
  Backup Id:         20230918T213041
  Backup Name:       backup-1695072640
  Begin LSN:         1C/B4000028
  Begin Wal:         000000010000001C0000002D
  Destination Path:  s3://angus/
  End LSN:           1C/C0000000
  End Wal:           000000010000001C00000030
  Instance ID:
    Container ID:  containerd://49914699b198326195252adbc0d1b46c8b8b156c7021f5cd2ad573c02c4cb8e1
    Pod Name:      angus-2
  Method:          barmanObjectStore
  Phase:           completed
  s3Credentials:
    Inherit From IAM Role:  true
  Server Name:              angus
  Started At:               2023-09-18T21:30:41Z
  Stopped At:               2023-09-18T22:36:29Z
```

On the other hand, this is the status of a volume snapshot backup:

```
Status:
  Backup Id:  angus-backup-02
  Instance ID:
    Container ID:  containerd://49914699b198326195252adbc0d1b46c8b8b156c7021f5cd2ad573c02c4cb8e1
    Pod Name:      angus-2
  Method:          volumeSnapshot
  Phase:           completed
  Snapshot Backup Status:
    Snapshots:
      angus-2-1695047917
      angus-2-wal-1695047917 
```

As you can see, relevant information is missing, in particular:

- Started At
- Stopped At
- Begin LSN
- Begin WAL
- End LSN
- End WAL

Check also the discrepancy when treating the Id and Name of the backup in the two methods.

The information here MUST be coherent with the one included as labels and annotations in the volume snapshots (#2794).

jsilvela:
Closes #2813


pkg/utils/parser.go
```diff
@@ -0,0 +1,33 @@
...
+	lines := strings.Split(data, "\n")
+	for _, line := range lines {
+		frags := strings.Split(line, ":")
+		if len(frags) != 2 {
+			continue
+		}
+		pairs[strings.TrimSpace(frags[0])] = strings.TrimSpace(frags[1])
+	}
+	return pairs
+}
```
mnencia, line 33: This function looks very similar to part of the `GetEnforcedParametersThroughPgControldata` function inside the `pkg/management/postgres/restore.go` file. I think that the parsing part should be in a common function.

armru (approved):

mnencia:
/ok-to-merge E2E tests green on EDB fork https://github.com/EnterpriseDB/cloudnative-pg/actions/runs/6313760209/job/17142807170

PR #2840, opened 2023-09-21
chore: add webhook validation for `shared_buffers`

### LINKED ISSUE #2664: CNPG should validate memory resources settings against the `shared_memory` Postgres parameter
In #2661, we added a validator to catch some possible mistakes in resource definition.

We can expand it to catch more obvious mistakes in the memory settings, comparing them to the `shared_memory` Postgres configuration parameter. 

If defined, the memory resources must be greater than the Postgres shared memory.

We could also talk about it if we want to let the user know (with a warning) if the values are too narrow, but that would be a heuristic and may require another ticket to introduce it.

armru:
Partially Closes #2664

### DIFF
--- a/api/v1/cluster_webhook.go
+++ b/api/v1/cluster_webhook.go
@@ -58,6 +58,10 @@ const (
 	DefaultApplicationUserName = DefaultApplicationDatabaseName
 )
 
+const (
+	sharedBuffersParameter = "shared_buffers"
+)
+
 // clusterLog is for logging in this package.
 var clusterLog = log.WithName("cluster-resource").WithValues("version", "v1")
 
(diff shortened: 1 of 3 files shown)


api/v1/cluster_webhook.go
```diff
@@ -1009,6 +1028,17 @@ func (r *Cluster) validateConfiguration() field.ErrorList {
 		}
 	}
 
+	if value := r.Spec.PostgresConfiguration.Parameters[sharedBuffersParameter]; value != "" {
+		if _, err := parsePostgresQuantityValue(value); err != nil {
```
YanniHu1996, line 1032: the error is hidden, it is better to let the user know what went wrong,  or what is the correct format

YanniHu1996 (approved):

api/v1/cluster_webhook.go
armru: Updated the message, is it any better?

YanniHu1996 (approved):

litaocdl (approved):

litaocdl:
run from fork https://github.com/EnterpriseDB/cloudnative-pg/actions/runs/6297393445

api/v1/cluster_webhook.go
```diff
@@ -944,25 +948,40 @@ func (r *Cluster) validateImagePullPolicy() field.ErrorList {
...
-		memoryRequestGtThanLimit := r.Spec.Resources.Requests.Memory().Cmp(*r.Spec.Resources.Limits.Memory()) > 0
+	memoryRequest := r.Spec.Resources.Requests.Memory()
+	rawSharedBuffer := r.Spec.PostgresConfiguration.Parameters[sharedBuffersParameter]
+	if !memoryRequest.IsZero() && rawSharedBuffer != "" {
+		if sharedBuffers, err := parsePostgresQuantityValue(rawSharedBuffer); err == nil {
+			if memoryRequest.Cmp(sharedBuffers) < 0 {
+				result = append(result, field.Invalid(
+					field.NewPath("spec", "resources", "requests", "memory"),
+					memoryRequest.String(),
+					"Memory request is lower than the shared_buffer value",
```
gbartolini, line 972:
```suggestion
					"Memory request is lower than PostgreSQL `shared_buffers` value",
```

api/v1/cluster_webhook_test.go
```diff
@@ -1052,6 +1052,20 @@ var _ = Describe("configuration change validation", func() {
 		}
 		Expect(clusterNew.validateConfiguration()).To(HaveLen(1))
 	})
+
+	It("should detect an invalid shared_buffer value", func() {
```
gbartolini, line 1056:
```suggestion
	It("should detect an invalid `shared_buffers` value", func() {
```

gbartolini (approved):

PR #2854, opened 2023-09-23
feat: avoid fencing instance unnecessarily in volume snapshots

### LINKED ISSUE #2853: Volume snapshot reconciler should fence instances only if there are no snapshots
The volumesnapshot reconciler starts by fencing the target pod right away.
This can lead to trouble if the top reconciliation loop has an error downstream that prevents a backup from being marked as completed.
In such case, the backup reconciler will get called once again, and will start by fencing the instance, then doing a listing where it might well detect that there are voluesnapshots already created, and would then proceed to unfence the instance.
We should only be fencing the pod if we have positively identified that there are no volumesnapshots created.

---

The logic of the snapshot backup reconciler is:

1. Step 1: fencing
2. Step 2: create snapshots
3. Step 3: wait for snapshots to be ready

as implemented in the volumesnapshot reconciler code (`pkg/reconciler/backup/volumesnapshot/resources.go`):

``` go
// Execute the volume snapshot of the given cluster instance
func (se *Reconciler) Execute(
```

jsilvela:
Closes #2853

### DIFF
--- a/pkg/reconciler/backup/volumesnapshot/reconciler.go
+++ b/pkg/reconciler/backup/volumesnapshot/reconciler.go
@@ -144,8 +144,8 @@ func (se *Reconciler) Execute(
 	}
 
 	// Step 2: create snapshot
-	// we execute the snapshots only if we don't find any
 	if len(volumeSnapshots) == 0 {
+		// we execute the snapshots only if we don't find any
 		if err := se.createSnapshotPVCGroupStep(ctx, cluster, pvcs, backup, targetPod); err != nil {
 			return nil, err
 		}
--- a/pkg/reconciler/backup/volumesnapshot/reconciler.go
+++ b/pkg/reconciler/backup/volumesnapshot/reconciler.go
@@ -123,6 +123,10 @@ func (se *Reconciler) Execute(
 	targetPod *corev1.Pod,
 	pvcs []corev1.PersistentVolumeClaim,
 ) (*ctrl.Result, error) {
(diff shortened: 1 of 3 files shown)


pkg/reconciler/backup/volumesnapshot/reconciler.go
```diff
@@ -123,10 +123,20 @@ func (se *Reconciler) Execute(
 	targetPod *corev1.Pod,
 	pvcs []corev1.PersistentVolumeClaim,
 ) (*ctrl.Result, error) {
+	if cluster.Spec.Backup == nil || cluster.Spec.Backup.VolumeSnapshot == nil {
```
YanniHu1996, line 126: we don't need to check the `Backup` here, the previous code handles this already, 
please refer to https://github.com/cloudnative-pg/cloudnative-pg/blob/main/controllers/backup_controller.go#L194

jsilvela: The problem is that the unit tests won't go through that previous checking.
If you take it out and you test the code with a cluster with no backup, you get a panic.

YanniHu1996: thanks, @jsilvela , it makes sence

YanniHu1996 (approved):

armru (approved):

mnencia (approved):

mnencia:
/ok-to-merge E2E green on EDB fork https://github.com/EnterpriseDB/cloudnative-pg/actions/runs/6314358265/job/17144644959

Issue #2800, opened 2023-09-16
primary failure and timeline change can lead to replica failure which is not respected correctly


Lennart01:
# Configuration and Cluster overview
Our cluster consists of 3 replicas (cluster-2, cluster-5 and cluster-7).
This is our configuration:
```yaml
apiVersion: postgresql.cnpg.io/v1
kind: Cluster
metadata:
  name: postgres-cluster
spec:
  imageName: ghcr.io/cloudnative-pg/postgresql:15.1
  instances: 3
  minSyncReplicas: 2
  maxSyncReplicas: 2
  storage:
    storageClass: csi-cinder-sc-retain
    size: 10Gi
  affinity:
    enablePodAntiAffinity: true
    topologyKey: failure-domain.beta.kubernetes.io/zone
  replicationSlots:
    highAvailability:
      enabled: true
  backup:
    retentionPolicy: "60d"
    barmanObjectStore:
      destinationPath: s3://storage
      endpointURL: https://s3-compatible-object-store
      s3Credentials:
        accessKeyId:
          name:  s3-postgres-backup
          key: ACCESS_KEY_ID
        secretAccessKey:
          name: s3-postgres-backup
          key: SECRET_ACCESS_KEY
  postgresql:
    parameters:
      shared_buffers: 256MB
      max_connections: '200'
  bootstrap:
    initdb:
      database: db
      owner: db
      secret:
         name: postgresql-creds
```
# Observed behavior
The Primary instance (cluster-2) failed and cluster-5 was selected as the new primary. The failover went as usual. 
cluster-5 now decided to switch the timeline (from 38 to 39)
```
{"level":"info","ts":"2023-09-15T20:06:20Z","logger":"postgres","msg":"record","logging_pod":"postgres-cluster-5","record":{"log_time":"2023-09-15 20:06:20.253 UTC","process_id":"31","session_id":"650001d4.1f","session_line_num":"15","session_start_time":"2023-09-12 06:14:44 UTC","virtual_transaction_id":"1/0","transaction_id":"0","error_severity":"LOG","sql_state_code":"00000","message":"selected new timeline ID: 39","backend_type":"startup","query_id":"0"}}
```
cluster-7 acknowledged this change
```
[...]

PR #2869, opened 2023-09-25
feat: added default alert case for failing replication

Lennart01:
Hey,
as long as #2800 is in the wild I thought it would be a good idea to add a default alert case to warn the user if this edge case occurs.
The alert was tested against a broken and a working cluster and performed as expected (also tested with single replica cluster)


docs/src/samples/monitoring/alerts.yaml
```diff
@@ -55,3 +55,12 @@ groups:
     for: 1m
     labels:
       severity: warning
+  - alert: ReplicaFailingReplication
+    annotations:
+      description: Replica {{ $labels.pod }} is failing to replicate
+      summary: Checks if the replica is failing to replicate
+    expr: |-
+      (1 - cnpg_pg_replication_in_recovery + cnpg_pg_replication_is_wal_receiver_up) < 1
```
gbartolini, line 63: Not familiar with Prometheus alerts, but why not:

```
cnpg_pg_replication_in_recovery > cnpg_pg_replication_is_wal_receiver_up
```

Lennart01: Yup, works will change accordingly 👍 .

phisco (approved):

phisco:
@Lennart01, looks like the DCO step is failing, can you fix that? see https://github.com/cloudnative-pg/cloudnative-pg/pull/2869/checks?check_run_id=17206660410 for more details 🙏

Lennart01:
@phisco Sorry about that. Should be fixed 👍

leonardoce (approved):

PR #2879, opened 2023-09-26
fix: ensure that restore snapshot cmd omits excluded files

### LINKED ISSUE #2865: Cleanup content of PGDATA after recovery from snapshots
After the recovery process from a volume snapshot has completed, we need to clean the content of PGDATA.

Be inspired by the list of files that Barman uses (taken from pg_basebackup source code partially). See https://github.com/EnterpriseDB/barman/blob/367d5e5aec634075a3f891a00b79d7ef1e1b11d6/barman/postgres_plumbing.py#L28

armru:
After the recovery process from a volume snapshot completes,
we need to cleanup the content of PGDATA, similarly to what
backup tools like Barman do. This is the list of files that we
remove from PGDATA:

- core.*
- pgsql_tmp*
- postgresql.auto.conf.tmp
- current_logfiles.tmp
- pg_internal.init
- postmaster.pid
- postmaster.opts
- recovery.conf
- recovery.signal
- standby.signal
- log/*
- pg_dynshmem/*
- pg_log/*
- pg_notify/*
- pg_replslot/*
- pg_serial/*
- pg_snapshots/*
- pg_stat_tmp/*
- pg_subtrans/*

Closes #2865

### DIFF
--- a/docs/src/samples/cluster-restore-snapshot-full.yaml
+++ b/docs/src/samples/cluster-restore-snapshot-full.yaml
@@ -1,7 +1,7 @@
 apiVersion: postgresql.cnpg.io/v1
 kind: Cluster
 metadata:
-  name: cluster-restore-pitr
+  name: cluster-restore-full
 spec:
   instances: 3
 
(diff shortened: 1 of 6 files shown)


gbartolini (approved):

leonardoce (approved):

leonardoce:
The E2e were canceled, I ran them on the EnterpriseDB repo

Issue #2794, opened 2023-09-15
Add standard backup labels/annotations to volume snapshots


gbartolini:
Once the volume snapshot through cold backup is taken, we should add some labels to the object to facilitate search via selectors, as well as some annotations containing the relevant information. This is just a basic initial idea.

Labels:
- cnpg.io/backupTimeline
- cnpg.io/backupYear: format YYYY
- cnpg.io/backupMonth: format YYYMM
- cnpg.io/onlineBackup: true/false - false in case of cold backup (meaning the backup is consistent)

Annotations:
- cnpg.io/backupStartWAL
- cnpg.io/backupEndWAL
- cnpg.io/backupStartTime
- cnpg.io/backupEndTime
- cnpg.io/snapshotStartTime - taken before invoking the creation of the snapshot
- cnpg.io/snapshotEndTime - taken when the volume snapshot becomes ready (approximation).

My suggestion is to do further analysis on pg_controldata and what we will expect from future support of backup labels returned by pg_start_backup() and pg_stop_backup(). More ideas can be taken from the backup info file in Barman.

PR #2880, opened 2023-09-26
fix: uniquely attribute snapshots to a backup

### LINKED ISSUE #2876: Volume snapshots should be labeled with a historically unique backup ID
In the function to get the volume snapshots taken for a given backup, we are looking for those matching the backup name.

``` go
func GetBackupVolumeSnapshots(
<SNIPPED>

	if err := cli.List(
		ctx,
		&list,
		client.InNamespace(namespace),
		client.MatchingLabels{utils.BackupNameLabelName: backupLabelName},
```

The problem is that the backup name may not be historically unique, for example if the backup manifest was created manually instead of via the plugin.
In such case, we could delete the backup object, create one with the same name, and right away, we would have matching volume snapshots.

armru:
Closes #2876


mnencia:
This patch has been superseded by implementing #2881, #2882, #2883, and #2794.

PR #2903, opened 2023-09-27
feat: enable replications slots for HA by default

### LINKED ISSUE #2673: Enable replication slots for HA by default
Cluster managed replication slots for high availability were first introduced by commit:10f630e4f62cd9700182842e6d8dd777c785e8c4 in CloudNativePG 1.18.

Given that all currently supported versions of Postgres now have the [`pg_replication_slot_advance()` required function](https://www.postgresql.org/docs/current/functions-admin.html), and that replication slots are fundamental to achieve self-healing when no WAL archive is configured, it makes sense to enable this feature by default starting with version 1.21.

Please remember to mention this change in the release notes, and provide instructions on how to disable them explicitly in the documentation. This might be related to #2671 too.

gbartolini:
IMPORTANT: the `.spec.replicationSlots.highAvailability.enabled` parameter is now by default set to `true`. This change was already anticipated in the upgrade instructions of CloudNativePG 1.20.

Closes #2673

### DIFF
--- a/api/v1/cluster_webhook.go
+++ b/api/v1/cluster_webhook.go
@@ -145,7 +145,7 @@ func (r *Cluster) setDefaults(preserveUserSettings bool) {
 		r.defaultMonitoringQueries(configuration.Current)
 	}
 
-	// If the API or validation webhooks are disable we set the replicationSlots to true if nil
+	// If the ReplicationSlots or HighAvailability stanzas are nil, we create them and enable slots
 	if r.Spec.ReplicationSlots == nil {
 		r.Spec.ReplicationSlots = &ReplicationSlotsConfiguration{}
 	}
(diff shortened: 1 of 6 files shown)


sxd (approved):

YanniHu1996 (approved):

api/v1/cluster_webhook.go
```diff
@@ -143,6 +144,17 @@ func (r *Cluster) setDefaults(preserveUserSettings bool) {
...
+	// If the API or validation webhooks are disable we set the replicationSlots to true if nil
+	if r.Spec.ReplicationSlots == nil {
+		r.Spec.ReplicationSlots = &ReplicationSlotsConfiguration{}
+	}
+	if r.Spec.ReplicationSlots.HighAvailability == nil {
+		r.Spec.ReplicationSlots.HighAvailability = &ReplicationSlotsHAConfiguration{
+			Enabled:    ptr.To(true),
+			SlotPrefix: "_cnpg_",
+		}
+	}
```
jsilvela, line 157: The logic here is a bit obscure. Line 149... we create an empty ReplicationSlots config. Line 152 WILL get executed right after, and will add the replication slots by default.
Why not merge the two? Or am I missing something?

jsilvela (approved):

PR #2905, opened 2023-09-27
feat: disable `postgres` super user access by default

### LINKED ISSUE #2899: Change default value of `enableSuperuserAccess` to `false`
In order to enhance the secure by default approach of a cluster created by CloudNativePG, we should provide superuser access only if requested by the user.

gbartolini:
IMPORTANT: Change the default value of `.spec.enableSuperuserAccess` to `false`, in order improve the security by default posture of CloudNativePG by disabling superuser access via the network.

The rationale behind this is to foster a microservice approach where changes to PostgreSQL are performed in a declarative way through the `spec` of the `Cluster` resource, while providing developers with full powers inside the database through the database owner user. At the same time, give the possibility to users to enable it by explicitly setting `.spec.enableSuperuserAccess: true`.

Closes #2899

### DIFF
--- a/api/v1/cluster_types.go
+++ b/api/v1/cluster_types.go
@@ -240,8 +240,8 @@ type ClusterSpec struct {
 	// not present, the operator will automatically create one). When this
 	// option is disabled, the operator will ignore the `SuperuserSecret` content, delete
 	// it when automatically created, and then blank the password of the `postgres`
-	// user by setting it to `NULL`. Enabled by default.
-	// +kubebuilder:default:=true
+	// user by setting it to `NULL`. Disabled by default.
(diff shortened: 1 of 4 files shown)


sxd (approved):

NiccoloFei (approved):

gbartolini:
IMPORTANT: First we need to merge #3015 as tests need to be backported.

ComaVN:
I highly agree with this approach in general, but this *is* a BC break with v1.20.x, because Cluster resources without a `.spec.enableSuperuserAccess` key that rely on the superuser will stop working. Am I mistaken in assuming that cnpg follows semantic versioning?

PR #2915, opened 2023-09-28
feat: add `cnpg.io/instanceRole` label

### LINKED ISSUE #2825: Rename the `role` label into `cnpg.io/instanceRole`
Currently we use a generic `role` label. In order to continue our standardization process, we should consider renaming it into a more canonical `cnpg.io/instanceRole` name.

We should consider - for back compatibility - to keep the `role` label for a few releases, similarly to what we did for the `postgresql` label.

armru:
This patch adds `cnpg.io/instanceRole` and deprecates the existing `role` label.

This new addition results from an ongoing effort to use only namespaced labels/annotations for uniformity and to avoid potential name clashes with other domains.

The deprecated `role` label will be removed in a future release.

Closes #2825

### DIFF
--- a/pkg/utils/labels_annotations.go
+++ b/pkg/utils/labels_annotations.go
@@ -53,7 +53,11 @@ const (
 	PgbouncerNameLabel = MetadataNamespace + "/poolerName"
 
 	// ClusterRoleLabelName is the name of label applied to instances to mark primary/replica
+	// Deprecated: Use ClusterInstanceRoleLabelName.
 	ClusterRoleLabelName = "role"
+
+	// ClusterInstanceRoleLabelName is the name of label applied to instances to mark primary/replica
+	ClusterInstanceRoleLabelName = MetadataNamespace + "/instanceRole"
 
 	// ImmediateBackupLabelName is the name of the label applied to backups to tell if the first scheduled backup is
 	// taken immediately or not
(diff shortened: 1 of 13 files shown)


pkg/utils/labels_annotations.go
```diff
@@ -361,3 +365,24 @@ func MergeMap(receiver, giver map[string]string) {
 		receiver[key] = value
 	}
 }
+
+// GetInstanceRole tries to fetch the ClusterRoleLabelName andClusterInstanceRoleLabelName value from a given labels map
+func GetInstanceRole(labels map[string]string) (string, bool) {
+	if value := labels[ClusterRoleLabelName]; value != "" {
```
gbartolini, line 371: shouldn't we swap the two checks here? also please add a TODO that we will remove this from 1.22.1 or later.

armru: It makes no difference, given that the contain virtually the same value.

A TODO is not needed here IMO, because once we remove the old ClusterRoleLabelName, the project won't compile because the referenced constant is used here

gbartolini:
@armru Can you please cover https://github.com/cloudnative-pg/cloudnative-pg/issues/2930 in this patch?

jlong49 (approved):

sxd (approved):

PR #2946, opened 2023-09-30
feat: include replicas in volume snapshot recovery

### LINKED ISSUE #2947: Include replicas in volume snapshot recovery
**Premise:** we knew we would get to this point / great we are able to fix this just by leveraging Kubernetes storage resources.

While testing volume snapshot recovery with medium (500GB) and large size databases (4TB), as expected the recovery time to spin up the primary reduces dramatically (I am working on a blog post to cover this). However, as expected, replicas are then created using the default available method at the moment: pg_basebackup.

We should change the replica creation method to use the same volume snapshots set to spin them up.

As you can imagine, this possibility opens up a tons of interesting scenarios that gravitate around using volume snapshots and PVC cloning in general as a way to generate replicas, not only when boostrapping from a volume snapshot backup (i.e. a replica dies we could use the latest snapshot provided WAL files exist, or the whole cluster dies, we could spin it up from the last available snapshot and use the WAL archive to self-heal). All these scenarios are premature now.

However, if possible, we should investigate and add the possibility to at least build the replicas in the case of bootstrap recovery from a volume snapshot. If we see that this patch can land, we should add it to 1.21.

leonardoce:
When bootstrapping a cluster from a volume snapshot backup,
reuse the volume snapshot to build the replicas.

Closes #2947


gbartolini:
I ❤️ this!

armru (approved):

leonardoce:
Thank you @armru for your review. I'd like to change a few more things in this PR so I'm assigning it to me.

leonardoce:
@NiccoloFei, unfortunately, I had to remove the improvement you did in the E2e test since now we (correctly) require WAL archiving to be active when we bootstrap replicas using snapshots.
Perhaps we could create an E2e test just for this scenario?

pkg/reconciler/persistentvolumeclaim/storagesource_test.go
```diff
@@ -0,0 +1,281 @@
...
+					Phase: apiv1.BackupPhaseCompleted,
+				},
+			},
+		},
+	}
+
+	When("bootstrapping from a VolumeSnapshot", func() {
+		When("we don't have backups", func() {
+			When("there's no source WAL archive", func() {
+				It("should fail when looking for a wrong role", func(ctx context.Context) {
```
jsilvela, line 122: oooh, nice! Very readable. I had never thought of nesting When's

tests/utils/job.go
```diff
@@ -0,0 +1,41 @@
...
+import (
+	"errors"
+	"fmt"
+
+	batchv1 "k8s.io/api/batch/v1"
+)
+
+// GetJob gets a Job by namespace and name
+func (env TestingEnvironment) GetJob(namespace, jobName string) (*batchv1.Job, error) {
+	wrapErr := func(err error) error {
```
jsilvela, line 28: This function seems unused. Still, good idea to have it, and the linter is not complaining...

jsilvela (approved):

tests/utils/job.go
leonardoce: I think this will be needed when we add E2e tests for replica cloning via Volume Snapshots.

gbartolini:
We should IMHO have documented this behaviour.

PR #2995, opened 2023-10-05
Embed the version number in the header of the docs

djw-m:
The current docs do not indicate what version the docs were created for (outside release notes). This small change must be manually updated. It embeds that information in the page header via the site_name configuration variable, which is only (according to the docs) used as the main title.

### DIFF
--- a/docs/mkdocs.yml
+++ b/docs/mkdocs.yml
@@ -1,4 +1,4 @@
-site_name: CloudNativePG
+site_name: CloudNativePG<br/><sub>v1.20.2</sub>
 site_author: The CloudNativePG Contributors
 docs_dir: src
 


jsilvela:
Thanks @djw-m ! Adding the release version to the docs is very valuable.
Let's have a look at possibly scripting this.
Nice, no? @gbartolini

djw-m:
Ok, noted on the automation so reworked to simplify things

1) No more embedded HTML - Version number is displayed through the readthedocs theme's extra.version support
2) The setting of extra.version is done via an environment variable CNPG_VERSION (adjust to taste/standards in mkdocs.yml). If not set, no version is rendered. This should allow any release process to set the env var and run generating the appropriate docs set for that version.
3) The version number styling is tuned in override.css - I've made it a little more prominent yet smaller over the default theme (which is very low contrast).

Hopefully this makes things easier to integrate.

jsilvela:
Going to try this out sometime this morning. Thanks @djw-m

jsilvela:
I think we need to iterate on this.
On one hand, the automation aspect - env vars vs something else.

On the other hand, the version number is ensconced in a corner.
Definitely better than not having it, but while we're mucking here, we could try
to make it more prominent. Perhaps with a link back to the version list page https://cloudnative-pg.io/docs/
Screenshot 2023-10-06 at 15 24 27

djw-m:
Ok well, I've been trying to avoid diving into tweaking the theme itself or
adding some dependencies....

Theoretically it should be easy to override a block in the readthedocs
template but.... the thing we'd want to override is not in a block. I may
submit a fix upstream to mkdocs to resolve that as its very irritating :)

So anyway, I'm pushing a defaulted CSS setting and offering this...

Run mkdocs with

VERSION=1.20.2 CNPG_VERSION="<a
href='https://cloudnative-pg.io/docs/'>$VERSION</a>"
mkdocs serve -a localhost:6900

And it will build with a link on the version number. It's a bit kludgy, but
it does work and if its being automated, almost invisible.

Dj

sxd:
Hi! CNPG_VERSION it's not an environment variable that exists when building the documentation so this patch will not work on a standard build, am I right @djw-m ? if I'm not wrong, when the documentation is built there's a variable with the version with the name `VERSION`, probably we can try and test with that one ? What do you think @jsilvela and @gbartolini ?

jsilvela:
This is  a valuable change to make for the documentation. I don't think
we're ready to merge this for today (releasing 1.21).
However, it should go in soon and we can amend the documentation in
https://cloudnative-pg.io without cutting a new release.

Also, @djw-m , the automation is a bit more involved than I had told you,
as there is a separate repo used to house the documents for the website.
We need to have an approach that works with the commands used to
export the documents to that repo.

This ticket merits a good look from several of the CNPG core team members.

djw-m:
> Hi! CNPG_VERSION it's not an environment variable that exists when building the documentation so this patch will not work on a standard build, am I right @djw-m ? 

Yes, I was assuming it would get mated with an appropriate environment variable in review as this is more a proof of concept.

sxd:
Closing this PR since it's not valid anymore

PR #3002, opened 2023-10-06
docs: upgrade instructions to 1.21.0, 1.20.3 and 1.19.5

### LINKED ISSUE #2988: Upgrade instructions for 1.21
1.21 changes the default value of several settings. We need to inform users of that and how they can ensure the previous behaviour is not changed.

gbartolini:
List all the significant changes applied to the default values in 1.21.0, 1.20.3, and 1.19.5, while providing instructions on how to keep the existing behavior on new clusters too.

Closes #2988

### DIFF
--- a/docs/src/installation_upgrade.md
+++ b/docs/src/installation_upgrade.md
@@ -300,6 +300,7 @@ If you want to retain the old behavior, you need to set explicitly:
 spec:
    ...
    stopDelay: 30
+   smartShutdownTimeout: 15
 ```
 
 #### Delay for PostgreSQL startup
--- a/docs/src/installation_upgrade.md
+++ b/docs/src/installation_upgrade.md
@@ -236,15 +236,22 @@ When versions are not directly upgradable, the old version needs to be
 removed before installing the new one. This won't affect user data but
 only the operator itself.
 
-### Upgrading to 1.21 from a previous minor version
+### Upgrading to 1.21.0, 1.20.3 or 1.19.5
 
 !!! Important
-    If you are upgrading from 1.19 or older, please read the
-    [instructions to upgrade to version 1.20](#upgrading-to-120-from-a-previous-minor-version)
-    first.
+    We encourage all existing users of CloudNativePG to upgrade to version
+    1.21.0 or at least to the latest stable version of the minor release you are
+    currently using (namely 1.20.3 or 1.19.5).
+
+!!! Warning
+    Everytime you are upgrading to a higher minor release, make sure you
+    go through the release notes and upgrade instructions of all the
+    intermediate minor releases. For example, if you want to move
+    from 1.19.x to 1.21, make sure you go through the release notes
+    and upgrade instructions for 1.20 and 1.21.
 
 With the goal to keep improving out-of-the-box the *convention over
-configuration* behavior of the operator, CloudNativePG 1.21 changes the default
+configuration* behavior of the operator, CloudNativePG changes the default
 value of several knobs in the following areas:
 
 - startup and shutdown control of the PostgreSQL instance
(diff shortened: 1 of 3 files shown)


docs/src/installation_upgrade.md
```diff
@@ -236,6 +236,150 @@ When versions are not directly upgradable, the old version needs to be
...
+with both the archive and the streaming replicas to ensure the cluster
+doesn't lose any data.
+
+If you want to retain the old behavior, you need to set explicitly:
+
+```yaml
+spec:
+   ...
+   stopDelay: 30
+```
```
mnencia, line 303: AFAIU, we cannot achieve the previous behavior with the current code. Setting stopDelay to 30 will leave the smartStopDelay set to 30 too, which means we are not giving any time to Postgres to run the shutdown fast before killing the pod (besides the 2-second grace period gifted by the kubelet). I do not know if that's intentional, but I think we should at least write it in the documentation.

gbartolini: What we can do then is change the minimum slot to 15 seconds instead of 30 in the formula that calculates the minimum time.

mnencia (approved):

docs/src/installation_upgrade.md
```diff
@@ -236,6 +236,179 @@ When versions are not directly upgradable, the old version needs to be
...
+```yaml
+spec:
+   ...
+   startDelay: 30
+```
+
+!!! Important
+    This change has been backported to all supported minor releases. As a
+    result, it will also be available in versions 1.20.3 and 1.19.5.
+
```
jsilvela, line 361: hierarchically, this is under the "Upgrading to 1.21" section.
This warning means that this is not accurate.
I think we probably want to make that one "Upgrading to 1.21, 1.20.3 or 1.19.5",
And then have a sub-section, or a box, for 1.21 only.

docs/src/installation_upgrade.md
```diff
@@ -236,6 +236,179 @@ When versions are not directly upgradable, the old version needs to be
...
+```yaml
+spec:
+   ...
+   switchoverDelay: 40000000
+```
+
+!!! Important
+    This change has been backported to all supported minor releases. As a
+    result, it will also be available in versions 1.20.3 and 1.19.5.
```
jsilvela, line 381: same comment as before. Under "upgrading to 1.21"

docs/src/installation_upgrade.md
```diff
@@ -236,6 +236,179 @@ When versions are not directly upgradable, the old version needs to be
 removed before installing the new one. This won't affect user data but
 only the operator itself.
 
+### Upgrading to 1.21 from a previous minor version
```
jsilvela, line 239: See other comments. This top section should cover the new 1.19 and 1.20 patches,
or should be a subsection under a wider header

jsilvela (changes requested):
I've made some changes, but I think we need to revise the section hierarchy.
See comments.

docs/src/installation_upgrade.md
jsilvela: And. What happens if I upgrade from a previous release to 1.19.5? The warnings in this section apply, definitely. And upgrading from 1.19.5 to 1.21 should then be smooth sailing.

docs/src/installation_upgrade.md
gbartolini: Ohh ... so we should reverse everything. Upgrade from 1.19.4 or before, 1.20.2 or before, or any previous version?

docs/src/installation_upgrade.md
```diff
@@ -236,6 +236,179 @@ When versions are not directly upgradable, the old version needs to be
...
+```
+
+Once the upgrade is completed, also add:
+
+```yaml
+spec:
+   ...
+   smartShutdownTimeout: 30
+```
+
```
leonardoce, line 287:
```suggestion
The `smartShutdownTimeout` field has been introduced in the version
1.21, hence the need to add it after upgrading the operator.
```

docs/src/installation_upgrade.md
```diff
@@ -236,6 +236,179 @@ When versions are not directly upgradable, the old version needs to be
...
+!!! Important
+    This change has been backported to all supported minor releases. As a
+    result, it will also be available in versions 1.20.3 and 1.19.5.
+
+#### Delay for PostgreSQL startup
+
+Until now, [the `startDelay` parameter](instance_manager.md#startup-liveness-and-readiness-probes)
+was set to 30 seconds, and CloudNativePG was using this parameter as the initial
+delay for the Kubernetes liveness probe. Given that all the supported Kubernetes
```
leonardoce, line 330:
```suggestion
was set to 30 seconds, and CloudNativePG used this parameter as the 
`initialDelaySeconds` for the Kubernetes liveness probe. Given that all the supported Kubernetes
```

docs/src/installation_upgrade.md
```diff
@@ -236,6 +236,179 @@ When versions are not directly upgradable, the old version needs to be
...
+we have examined during support incidents or community issues show that this
+value is left unchanged. Given that this parameter influences the startup of
+a PostgreSQL instance, a low value of `startDelay` would cause Postgres
+never to reach a consistent recovery state and be restarted indefinitely.
+
+For this reason, `startDelay` has been [raised by default to 3600 seconds](https://github.com/cloudnative-pg/cloudnative-pg/commit/4f4cd96bc6f8e284a200705c11a2b41652d58146),
+the equivalent of 1 hour.
+
+You can achieve similar behavior using the startup probe instead of
+liveness probes with an initial delay by explicitly setting:
```
leonardoce, line 350:
```suggestion
if you want to retain the old behavior using the new implementation, you can
do that by explicitly setting:
```

While reading this, I didn't realize we were talking about retaining the old behavior, and my mind started to wonder how to emulate the startup probe with the liveness probe :) But I don't know if it's just me.

docs/src/installation_upgrade.md
mnencia: It will also be available in versions 1.20.3 and 1.19.5, so we should conform this to the others places where we mention this.

docs/src/installation_upgrade.md
jsilvela: Not sure what the best approach is. Was thinking something like below.
What do you think?

```
### Upgrading to 1.21, 1.20.3, 1.19.5

New defaults are being set for various things in these three releases.
Justification about users leaving the old defaults on, despite our recommendations.
Explanation of the one-time rolling upgrade.

#### New functionality in 1.21 only

#### Upgrading to 1.21 from 1.19.5 or 1.20.3

#### stopDelay parameter

#### startDelay

#### 
```

docs/src/installation_upgrade.md
gbartolini: This makes sense. Please check my new version. Thanks for pointing this out.

docs/src/installation_upgrade.md
gbartolini: Agree. Check the new proposal please.

docs/src/installation_upgrade.md
```diff
@@ -236,6 +236,222 @@ When versions are not directly upgradable, the old version needs to be
...
+value, almost all the cases we have examined during support incidents or
+community issues show that this value is left unchanged.
+
+The [new default value is 1800 seconds](https://github.com/cloudnative-pg/cloudnative-pg/commit/9f7f18c5b9d9103423a53d180c0e2f2189e71c3c),
+the equivalent of 30 minutes.
+
+The new `smartShutdownTimeout` parameter has been introduced to define
+the maximum time window within the `stopDelay` value reserved to complete
+the `smart` shutdown procedure in PostgreSQL. During this time, the
+Postgres servers denies any new connection while waiting for all regular
```
jsilvela, line 329:
```suggestion
Postgres servers rejects any new connection while waiting for all regular
```

jsilvela (approved):

PR #3030, opened 2023-10-09
chore: rename `smartStopDelay` to `smartShutdownTimeout`

### LINKED ISSUE #3027: Fix formula to calculate remaining time for fast shutdown
In #2844 we introduced a formula that wanted to ensure that the minimum amount of time for fast shutdown was 30 seconds. However, it is actually returning the wrong value.

With the occasion, in order to restore the previous default behaviour of having a fast shutdown of 15 seconds (50% of the default stop delay that was set to 30 seconds), let's ensure that the minimum amount of time dedicated to fast shutdown is 15 seconds.

mnencia:
Rename the newly introduced `smartStopDelay` parameter (never released)
into a clearer one called `smartShutdownTimeout`, which exactly describes
the intention of that option: control for how long we wait for PostgreSQL
to execute a graceful shutdown when requested, before trying a fast one
(during shutdown or fencing, for example).

Closes #3027

### DIFF
--- a/.wordlist-en-custom.txt
+++ b/.wordlist-en-custom.txt
@@ -1011,7 +1011,7 @@ sig
 sigs
 singlenamespace
 slotPrefix
-smartStopDelay
+smartShutdownTimeout
 snapshotBackupStatus
 snapshotOwnerReference
 snapshotted
--- a/api/v1/cluster_types.go
+++ b/api/v1/cluster_types.go
@@ -281,10 +281,10 @@ type ClusterSpec struct {
 
 	// The time in seconds that controls the window of time reserved for the smart shutdown of Postgres to complete.
 	// Make sure you reserve enough time for the operator to request a fast shutdown of Postgres
-	// (that is: `stopDelay` - `smartStopDelay`).
+	// (that is: `stopDelay` - `smartShutdownTimeout`).
 	// +kubebuilder:default:=180
 	// +optional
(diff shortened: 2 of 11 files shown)


gbartolini (approved):

internal/cmd/manager/instance/run/lifecycle/shutdown.go
```diff
@@ -51,24 +51,31 @@ func tryShuttingDownFastImmediate(timeout int32, instance *postgres.Instance) er
...
+	var err error
+	if timeout > 0 {
+		log.Info("Requesting smart shutdown of the PostgreSQL instance")
+		err = instance.Shutdown(postgres.ShutdownOptions{
+			Mode:    postgres.ShutdownModeSmart,
+			Wait:    true,
+			Timeout: &timeout,
+		})
+		if err != nil {
+			log.Warning("Error while handling the smart shutdown request̋", "err", err)
```
mnencia, line 63:
```suggestion
			log.Warning("Error while handling the smart shutdown request", "err", err)
```

gbartolini:
We have decided to first refactor the code and complete work related to #3032

litaocdl:
after solve conflict, only doc changes, will going to merge

gbartolini (approved):

leonardoce (approved):

PR #3037, opened 2023-10-10
feat: allow user to reduce the storage size that is equal to the PVC

### LINKED ISSUE #2829: Cannot rollback from resizing a volume when not allowed
While doing some tests on volume snapshot backups, I ran out of disk on the WAL volume and tried to increase the size of it using `kubectl edit cluster`, passing the webhook check.

However, the underlying storage class did not support expansion - the underlying PVC confirmed that the size was still the original one. I tried to rollback the change by decreasing the size to the one of the current PVC, but the webhook prevented that with: `can't shrink existing storage`.

I was able to exit this catch22 situation by deleting the webhook, changing the cluster spec, and restoring the webhook. However ideally the operator should allow restoring the size to the previous value, provided that none of the related PVCs in the cluster have been resized in the meantime.

YanniHu1996:
Closes #2829 

As the operator lacks sufficient information to determine whether a PVC resize is valid, the following changes have been made:

Removed validation of changes to storage.

Introduced a new type of condition called `PVCResizeSucceeded` to record the resize result

### DIFF
--- a/api/v1/cluster_types.go
+++ b/api/v1/cluster_types.go
@@ -862,6 +862,8 @@ const (
 	ConditionBackup ClusterConditionType = "LastBackupSucceeded"
 	// ConditionClusterReady represents whether a cluster is Ready
 	ConditionClusterReady ClusterConditionType = "Ready"
+	// ConditionPVCResize represents whether a PVC resize is successful
+	ConditionPVCResize ClusterConditionType = "PVCResizeSucceeded"
 )
 
 // A Condition that can be used to communicate the Backup progress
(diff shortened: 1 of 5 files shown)


mnencia:
This implementation is too complex and violates the principle of keeping the webhook code self-contained and not accessing the API server. We need to rewrite it using a field in the status to work as a proxy for the status of PVCs.

YanniHu1996:
Got it @mnencia , I will rewrite this pull request based on status way

litaocdl (approved):

litaocdl:
resolve conflict and force push

PR #3073, opened 2023-10-11
fix: add the a timeout of 60 minutes to GoReleaser in the CI

### LINKED ISSUE #3072: Increase the timeout of the GoReleaser in all the pipelines

sxd:
Closes #3072


mnencia:
It is not needed to build the manager. Only the plugin requires long time because it has a lot of architectures

PR #2954, opened 2023-09-30
fix(prometheus): add the cnpg_cluster label

achanda:
This PR adds a new label to PgVersion and PostgreSQLUp called cnpg_cluster The cluster label is commonly used to refer to the k8s cluster. That label can be removed in a later version.

Closes: #2501

PR #3075, opened 2023-10-11
feat: add PodMonitor relabeling options

npdgm:
Hi,
Prometheus relabeling rules are often used to deal with deployment specific cases where exporter metrics have to be altered or filtered before recording.
These changes add two new options `monitoring.podMonitorMetricRelabelings` and `monitoring.podMonitorRelabelings` to pass slices of `RelabelConfig` to the PodMonitor.

This contrib is probably still missing some docs or other required changes I'm not aware of. Looking forward for your feedbacks.
Thanks!

Related to: https://github.com/cloudnative-pg/charts/pull/148


npdgm:
I'll implement it for `PoolerMonitoringConfiguration` too. Just figured out it's missing...

npdgm:
This should be a generic fix for #2501 (#2954) with no API breakage

npdgm:
@armru should I be worried about the failed e2e jobs?

armru:
Hello @npdgm, nothing to worry about for the moment.
Thanks for the contribution, we will review it ASAP

mnencia:
E2E tests running in EDB fork https://github.com/EnterpriseDB/cloudnative-pg/actions/runs/6638192579

sjentzsch:
> This should be a generic fix for #2501 (#2954) with no API breakage.
> 
> Users wanting to free the `cluster` label because it conflicts with federation could use:
> 
> ```yaml
> monitoring:
>   podMonitorEnabled: true
>   podMonitorMetricRelabelings:
>   - sourceLabels: [cluster]
>     targetLabel: cnpg_cluster
>   - sourceLabels: [cluster]
>     action: labeldrop
> ```

As the creator of #2501, I can say this approach of PodMonitor relabeling does not really fix my issue.

As a user of centralized observability (= one view on multiple sources where cnpg-based psql DBs are created) we then still cannot re-use alerts and dashboards, as those still use the cluster label for referring to the psql cluster.

We cannot really be backwards-compatible here; at least with respect to monitoring artifacts (alerts and dashboards).

My main question is: Do you even want to offer first-level support for those use-cases?
So far, we forked alerts and dashboards and maintain them ourselves moving forward. That's basically the price to pay for not supporting it :)

phisco:
@sjentzsch, let me recap:
1. We currently emit metrics setting a label "cluster" for 2 metrics returned by instances (not the operator), `cnpg_collector_up` and `cnpg_collector_postgres_version`.
2. Other multi-cluster tools expect the "cluster" label to be used for the Kubernetes cluster the metric is coming from.
3. Our dashboard expect the "cluster" label on `cnpg_collector_up` to get the available clusters and show them in the drop down.
4. The only alerts we [provide](https://raw.githubusercontent.com/cloudnative-pg/cloudnative-pg/main/docs/src/samples/monitoring/prometheusrule.yaml), don't use the cluster label at all atm.
5. Other people's custom dashboards and alerts could potentially expect the "cluster" label as we intend today to be there.
6. Just adding a second label would not solve your issue because the other tools would still not overwrite the old one.
7. It would be awesome if we could avoid introducing a breaking change both on people using our own dashboard or a customised version of it and their own custom alerts.

The Prometheus Operator allows you to rewrite labels via both PodMonitor and ServiceMonitor, if you are not using those you can still configure Prometheus to do so via [metric_relabel_configs](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#metric_relabel_configs), which if I got it correctly, @sjentzsch, you are already doing and you would be fine continuing doing, right? So your issue, @sjentzsch, is actually that you have to fork the dashboard, correct? Because if that's the issue we could just find another way to infer all the available clusters in the dashboard and leave the "cluster" label as is, providing people suggestions about how to rewrite it if needed. WDYT?

npdgm:
> can you integrate the PR with some documentation?
Sure, I'll look into that.

As for the discussion on whether this `cluster` label should be renamed, I also have a hard time forging an opinion on what's best for the project and how to deal with a potentially breaking change. Here we tend to tweak most upstream dashboard anyway. So I won't claim relabelings are the solution for all, but they surely have other uses and still be nice to have.

api/v1/cluster_types.go
```diff
@@ -1919,6 +1920,14 @@ type MonitoringConfiguration struct {
...
 	// +optional
 	EnablePodMonitor bool `json:"enablePodMonitor,omitempty"`
+
+	// The list of metric relabelings for the `PodMonitor`. Applied to samples before ingestion.
+	// +optional
+	PodMonitorMetricRelabelConfigs []*monitoringv1.RelabelConfig `json:"podMonitorMetricRelabelings,omitempty"`
+
+	// The list of relabelings for the `PodMonitor`. Applied to samples before scraping.
+	// +optional
+	PodMonitorRelabelConfigs []*monitoringv1.RelabelConfig `json:"podMonitorRelabelings,omitempty"`
```
phisco, line 1930: we could even think to add the whole podmonitorspec here probably, overriding only what we care about (the port and the label selector mainly)

phisco (approved):

npdgm:
> I added support for them in the pooler, too. Are you okay with that?
Of course, thanks a lot!

leonardoce (approved):

api/v1/cluster_types.go
npdgm: Sounds good!
Asking as a Go newbie, is there a facility to ease with merging, taking into account that PodMetricsEndpoint are defined in a list and users may want to target one element to reach RelabelConfigs? Thanks

phisco: No, there is not unfortunately, but given that we are opinionated on the endpoint to hit, we could just enforce our own for that. But this implementation is fine, we can move to the full spec later on if we want to

sjentzsch:
> @sjentzsch, let me recap:
> 
>     1. We currently emit metrics setting a label "cluster" for 2 metrics returned by instances (not the operator), `cnpg_collector_up` and `cnpg_collector_postgres_version`.
> 
>     2. Other multi-cluster tools expect the "cluster" label to be used for the Kubernetes cluster the metric is coming from.
> 
>     3. Our dashboard expect the "cluster" label on `cnpg_collector_up` to get the available clusters and show them in the drop down.
> 
>     4. The only alerts we [provide](https://raw.githubusercontent.com/cloudnative-pg/cloudnative-pg/main/docs/src/samples/monitoring/prometheusrule.yaml), don't use the cluster label at all atm.
> 
>     5. Other people's custom dashboards and alerts could potentially expect the "cluster" label as we intend today to be there.
> 
>     6. Just adding a second label would not solve your issue because the other tools would still not overwrite the old one.
> 
>     7. It would be awesome if we could avoid introducing a breaking change both on people using our own dashboard or a customised version of it and their own custom alerts.
> 
> 
> The Prometheus Operator allows you to rewrite labels via both PodMonitor and ServiceMonitor, if you are not using those you can still configure Prometheus to do so via [metric_relabel_configs](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#metric_relabel_configs), which if I got it correctly, @sjentzsch, you are already doing and you would be fine continuing doing, right? So your issue, @sjentzsch, is actually that you have to fork the dashboard, correct? Because if that's the issue we could just find another way to infer all the available clusters in the dashboard and leave the "cluster" label as is, providing people suggestions about how to rewrite it if needed. WDYT? CC @achanda

Good summary, yes, I do agree that my issue is currently with the dashboard that we require to fork.
However, in general, if we decide to stick with the `cluster` label, new metrics & alerts might arise in the future, making it even harder to align alerts & dashboards for both use cases (introducing non-breaking changes while supporting multi-cluster setups). I can totally understand both perspectives.
I think one of the things to put on the roadmap maybe is: configurability of alerts & dashboards, using jsonnet or other tooling, in order to allow for different environments.

However, this discussion might be out-of-scope for this PR 👍

onedr0p:
@npdgm I have added the following to my cluster but I am not seeing the `cnpg_cluster` label being populated.

https://github.com/onedr0p/home-ops/blob/556982317381ddb0800601a0da7816b898b7e3e8/kubernetes/main/apps/database/cloudnative-pg/cluster/cluster16.yaml#L32-L39

I didn't know if it was worth commenting here, or in the linked issue, or creating a new issue.

PR #3087, opened 2023-10-13
docs: minor typo in applications.md

bittermandel:


mnencia:
This is not a typo. It's an ammonition of type `Seealso`. See https://python-markdown.github.io/extensions/admonition/

Issue #3011, opened 2023-10-06
Properly manage `pg_basebackup` in `cpng status`


gbartolini:
Currently, if you request the status of a cluster that is creating a replica, the streaming replication section reports misleading information, as in this example:

```
Streaming Replication status
Replication Slots Enabled
Name                               Sent LSN     Write LSN    Flush LSN    Replay LSN  Write Lag       Flush Lag       Replay Lag       State      Sync State  Sync Priority  Replication Slot
----                               --------     ---------    ---------    ----------  ---------       ---------       ----------       -----      ----------  -------------  ----------------
kubecon-na-23-eu-central-1-2-join                                                     00:00:00        00:00:00        00:00:00         backup     async       0              -
kubecon-na-23-eu-central-1-2-join                                                     00:00:00        00:00:00        00:00:00         startup    async       0              -
kubecon-na-23-eu-central-1-2-join  6A/24000000  6A/24000000  6A/24000000              00:00:01.22384  00:00:01.22384  00:28:02.053341  streaming  async       0              -
```

We should filter out the `backup` and `startup` rows and add a section before that, similar to the following:

```
Progress of physical backups
Name                                  Phase                      Started at                   Total    Transferred     Progress  Tablespaces
kubecon-na-23-eu-central-1-2-join     streaming database files   2023-10-06 13:12:01.76563Z   470 GB   120 GB          25.5%     0/1
```

The following query could be used (under normal circumstances, this should return an empty set):

```sql
SELECT
   usename, application_name, backend_start, phase
   backup_total, backup_streamed,
   tablespaces_total, tablespaces_streamed
FROM pg_stat_progress_basebackup b
   JOIN pg_stat_activity a USING (pid)
WHERE application_name ~ '-join$' ORDER BY 1, 2;
```

This should be run on the primary or designated primary only. And only on PostgreSQL 13 or above.

PR #3101, opened 2023-10-17
feat: enhance the `plugin status` command 

YanniHu1996:
closes: #3011

While working on this issue, I found that we have an opportunity to reduce the response time  of `status` API by the way used in this PR https://github.com/cloudnative-pg/cloudnative-pg/blob/af347f76cd234a1f715393c62fb6915904f7d159/pkg/management/postgres/probes.go#L66-L73 

Opened a new issue #3121 for tracing it


mnencia:
@YanniHu1996 to fix the linter issues you should run `golangci-lint run --fix`

pkg/management/postgres/probes.go
```diff
@@ -415,8 +510,8 @@ func (instance *Instance) fillWalStatusFromConnection(result *postgres.Postgresq
 			coalesce(sync_state, ''),
 			coalesce(sync_priority, 0)
 		FROM pg_catalog.pg_stat_replication
-		WHERE application_name LIKE $1 AND usename = $2`,
-		fmt.Sprintf("%s-%%", instance.ClusterName),
+		WHERE application_name ~ $1 AND usename = $2`,
+		fmt.Sprintf("%s-[0-9]+$", instance.ClusterName),
```
armru, line 514: Personally I don't know if this changes should be in this patch

armru (approved):

pkg/management/postgres/probes.go
```diff
@@ -270,9 +271,71 @@ func (instance *Instance) fillStatus(result *postgres.PostgresqlStatus) error {
...
+		FROM pg_stat_progress_basebackup b
+		   JOIN pg_stat_activity a USING (pid) 
+		WHERE application_name ~ '-join$'
+		ORDER BY 1, 2`)
+	if err != nil {
+		return err
+	}
+	defer func() {
+		if closeErr := rows.Close(); closeErr != nil && err == nil {
+			log.Error(err, "while closing rows")
```
litaocdl, line 311: should we log the closeErr here

pkg/management/postgres/probes.go
YanniHu1996: the change to this SQL is to filter out rows that are not represented as replication to standby instances

leonardoce (approved):

leonardoce:
We've problems with the PRs opened from repositories other than `cloudnative-pg/cloudnative-pg`. The build containers step is failing. See #3154

PR #3139, opened 2023-10-21
docs: make superuser access explicit in the examples

### LINKED ISSUE #3113: Samples aren't working as expected because of missing enableSuperuserAccess values
Setting up any cluster that needs to have a superuser account is documented here https://github.com/cloudnative-pg/cloudnative-pg/blob/main/docs/src/security.md#postgresql but the samples are lacking enableSuperuserAccess value altogether. 

Maybe add a commented enableSuperuserAccess value to show that since 1.21 it's required if a superuser is created?

Here's is the supposed cluster-example-full.yaml for which the superuser won't work for 1.21. 
https://github.com/cloudnative-pg/cloudnative-pg/blob/main/docs/src/samples/cluster-example-full.yaml
Also, here the superuser won't be taken into account at all for 1.21:
https://github.com/cloudnative-pg/cloudnative-pg/blob/main/docs/src/samples/cluster-example-secret.yaml

Also documentation here could be misleading if not noted that enableSuperuserAccess must be enabled for it to be completely true: https://github.com/cloudnative-pg/cloudnative-pg/blob/main/docs/src/applications.md#secrets

gbartolini:
Closes #3113


docs/src/applications.md
```diff
@@ -59,11 +59,12 @@ you can use the following environment variables in your applications:
...
 ### Secrets
 
-The PostgreSQL operator will generate two `basic-auth` type secrets for every
-PostgreSQL cluster it deploys:
+The PostgreSQL operator will generate up to two `basic-auth` type secrets for
+every PostgreSQL cluster it deploys:
 
-* `[cluster name]-superuser`
-* `[cluster name]-app`
+* `[cluster name]-app` (unless you have provide an existing secret through `.spec.bootstra.initdb.secret.name`)
```
vstariradev, line 65: “provded” maybe?

vstariradev (approved):
Looks good thanks!

litaocdl (approved):

leonardoce (approved):

PR #3152, opened 2023-10-24
feat: use extended query protocol for PostgreSQL

leonardoce:
This patch makes the operator use two different connection profiles for PostgreSQL and for PgBouncer. The first one is using the extended query protocol, the second one the simple one.

The reasoning behind is that for PostgreSQL we'll need to use the more secure query cancelation method. PgBouncer needs the simple query protocol against the admin DB, and we use that for metrics.


mnencia (approved):

mnencia:
E2E tests running in EDB fork https://github.com/EnterpriseDB/cloudnative-pg/actions/runs/6625119688

armru (approved):

PR #3214, opened 2023-10-26
feat: support recovering from a volume snapshot backup object

### LINKED ISSUE #2964: Properly manage volume snapshot method when recovering from a backup object
Currently, when recovering from a backup object (which will be deprecated at some point) we assume that the backup is in an object store. Introduce a check to ensure that recovery can also cater for volume snapshots. Blocked by #2963

mnencia:
Closes #2964

### DIFF
--- a/pkg/reconciler/persistentvolumeclaim/storagesource.go
+++ b/pkg/reconciler/persistentvolumeclaim/storagesource.go
@@ -143,7 +143,7 @@ func getCandidateSourceFromBackup(backup *apiv1.Backup) *StorageSource {
 	result := &StorageSource{
 		DataSource: corev1.TypedLocalObjectReference{
 			APIGroup: ptr.To(volumesnapshot.GroupName),
-			Kind:     "VolumeSnapshot",
+			Kind:     apiv1.VolumeSnapshotKind,
 			Name:     GetName(backup.Name, utils.PVCRolePgData),
 		},
 	}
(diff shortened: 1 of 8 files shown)


mnencia:
E2E tests running on EDB fork https://github.com/EnterpriseDB/cloudnative-pg/actions/runs/6656938711

armru (approved):

pkg/reconciler/persistentvolumeclaim/storagesource.go
```diff
@@ -131,33 +133,35 @@ func getCandidateSourceFromBackupList(ctx context.Context, backupList apiv1.Back
...
-				Kind:     "VolumeSnapshot",
-				Name:     GetName(backup.Name, utils.PVCRolePgData),
-			},
+		return getCandidateSourceFromBackup(backup)
+	}
+
+	return nil
+}
+
+func getCandidateSourceFromBackup(backup *apiv1.Backup) *StorageSource {
```
armru, line 142: in future, we should also return an error if there is no match for the element.type

mnencia: Yes. Especially when we add the tablespaces, which have one more degree of freedom (the tablespace name)

leonardoce (approved):

gbartolini:
I am not a big fan of this approach, but I see that others are OK with it, so I approve it.

PR #3251, opened 2023-10-31
fix: defer WAL archiving during a recovery job

### LINKED ISSUE #3259: [Bug]: recovery job discarding WAL files

### Contact Details

_No response_

### Version

1.21.0

1.28

YAML manifest

### What happened?

The recovery job uses `cd .` as `archive_command`. While this is OK in most cases that involve recovery of a standby, this could prevent the operator to create the replicas after recovering a primary from a standby. This issue has been highlighted by taking a backup on object store, taking a volume snapshot backup, restoring the cluster from the object store (the recovery job, once promoted the primary, did not archive any WAL file).

The operator would start creating the replica from the existing volume snapshot and then, when trying to bring it up, this would complain that the new timeline was not in the archive.

### Cluster resource

_No response_

### Relevant log output

_No response_

### Code of Conduct

leonardoce:
Up to now, the recovery job was discarding WAL files by setting `archive_command` to `cd .` (that is success). We now set it to `false`, so that we defer the decision about archiving a WAL file to when the instance is actually started as a regular pod.

Closes #3259

### DIFF
--- a/pkg/management/postgres/restore.go
+++ b/pkg/management/postgres/restore.go
@@ -585,10 +585,12 @@ func (info InitInfo) writeRecoveryConfiguration(recoveryFileContents string) err
 	}
 
 	log.Info("Generated recovery configuration", "configuration", recoveryFileContents)
-	// Disable archiving
+	// Temporarily suspend WAL archiving. We set it to `false` (which means failure
+	// of the archiver) in order to defer the decision about archiving to PostgreSQL
+	// itself once the recovery job is completed and the instance is regularly started.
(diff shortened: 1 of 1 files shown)


leonardoce:
https://github.com/EnterpriseDB/cloudnative-pg/actions/runs/6705408615

gbartolini (approved):

mnencia (approved):

armru (approved):

gbartolini (approved):

PR #3271, opened 2023-11-02
fix: use proto unmarshal podSpec if json failed

litaocdl:
wip


leonardoce:
This is nice! We can merge it in main and backport in every release

Issue #3273, opened 2023-11-02
[Bug]: e2e rolling out upgrade case is failing 


litaocdl:
### Is there an existing issue already for this bug?

- [X] I have searched for an existing issue, and could not find anything. I believe this is a new bug.

### I have read the troubleshooting guide

- [X] I have read the troubleshooting guide and I think this is a new bug.

### I am running a supported version of CloudNativePG

- [X] I have read the troubleshooting guide and I think this is a new bug.

### Version

1.21.0

### What version of Kubernetes are you using?

1.28

### What is your Kubernetes environment?

Self-managed: kind (evaluation)

### How did you install the operator?

YAML manifest

### What happened?

rolling out update test case is failing on 1.20 and 1.19. main and 1.21 is fine

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #3274, opened 2023-11-02
fix: skip podSpec rollout checker if pod annotation is invalid

litaocdl:
Fixes a bug where the PodSpec annotation was stored as Protobuf, but after
upgrade, was expected as JSON. This fix skips the PodSpec rollout check in
this case, defaulting to the pre-PodSpec rollout checkers.

Closes: #3273

### DIFF
--- a/controllers/cluster_upgrade.go
+++ b/controllers/cluster_upgrade.go
@@ -277,8 +277,8 @@ func isPodNeedingRollout(
 		return podRollout
 	}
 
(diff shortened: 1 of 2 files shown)


litaocdl:
If this is approveled, need backport to all releases branch and verify the upgrade test case in release-1.20 and 1.19 branch.

mnencia (approved):

armru (approved):

jsilvela (approved):

PR #3315, opened 2023-11-07
edits to declarative hibernation and declarative role management

ebgitelman:


.vscode/spellright.dict
```diff
@@ -0,0 +1,3 @@
+postgres
```
josh-heyer, line 1: @ebgitelman try to avoid committing files that aren't part of the work you're doing. You can add the names of files like this to your local .gitignore if that makes it easier

PR #3321, opened 2023-11-09
fix: correctly evaluate the request and response for the backup mode

armru:
In the old code for the status endpoint, we evaluated the body errors as an error response code. This needed to be corrected, given that this would lead the old code to an infinite stuck-on error response when the held backup connection contained an error.

With the new implementation, the caller determines if the returned body error is relevant to the backup that is trying to run.

The patch also ensures that requests for outdated backups aren't processed.

### DIFF
--- a/pkg/management/postgres/webserver/backup_client.go
+++ b/pkg/management/postgres/webserver/backup_client.go
@@ -54,14 +54,14 @@ func NewBackupClient() *BackupClient {
 }
 
 // StatusWithBodyErrors the current status of the backup..
-func (c *BackupClient) StatusWithBodyErrors(ctx context.Context, podIP string) (*BackupResultData, error) {
+func (c *BackupClient) StatusWithBodyErrors(ctx context.Context, podIP string) (*Response[BackupResultData], error) {
 	httpURL := url.Build(podIP, url.PathPgModeBackup, url.StatusPort)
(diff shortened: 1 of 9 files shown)


armru:
mirror repo only snapshot run:

pkg/management/postgres/webserver/backup_client.go
```diff
@@ -61,7 +72,11 @@ func (c *BackupClient) Status(ctx context.Context, podIP string) (*BackupResultD
 		return nil, err
 	}
 
-	return executeRequest[BackupResultData](ctx, c.cli, req)
+	res, err := executeRequestWithError[BackupResultData](ctx, c.cli, req, false)
```
armru, line 75: We evaluated the body errors in the old code as an error response code. This seems wrong.
The caller should determine if that error is relevant to the backup that is trying to run.

armru:
E2E SNAPSHOT: https://github.com/EnterpriseDB/cloudnative-pg/actions/runs/6851431481

armru:
E2E SNAPSHOT: https://github.com/EnterpriseDB/cloudnative-pg/actions/runs/6852608811/job/18631669217

armru:
running the third test to be sure that the bug is fixed

mnencia (approved):

jsilvela (approved):

Issue #3331, opened 2023-11-11
[Bug]: postgres doesn't care what UID it runs in, but initdb does. user does not exist


wibed:
### Is there an existing issue already for this bug?

- [X] I have searched for an existing issue, and could not find anything. I believe this is a new bug.

### I have read the troubleshooting guide

- [X] I have read the troubleshooting guide and I think this is a new bug.

### I am running a supported version of CloudNativePG

- [X] I have read the troubleshooting guide and I think this is a new bug.

### Version

1.21.0

### What version of Kubernetes are you using?

1.28

### What is your Kubernetes environment?

Self-managed: k3s

### How did you install the operator?

Helm

### What happened?

i found a configuration item.

https://github.com/cloudnative-pg/cloudnative-pg/blob/06871b2735823111a3f78ccc57c384447ee1cde5/config/crd/bases/postgresql.cnpg.io_clusters.yaml#L2517

upon using postgresGID and postgresUID, the uid is not changed the user is:
```shell
# tl;dr: "logger":"initdb","msg":"initdb: could not look up effective user ID 501: user does not exist\n",
# full log below:
[...]

PR #3332, opened 2023-11-12
added init container for createBootstrapContainer function,

wibed:
check if user exists and create it if not, for initdb to initialize the database along the UID GID passed in the spec successfully

related issue: https://github.com/cloudnative-pg/cloudnative-pg/issues/3331

### DIFF
--- a/pkg/specs/containers.go
+++ b/pkg/specs/containers.go
@@ -27,6 +27,32 @@ import (
 	"github.com/cloudnative-pg/cloudnative-pg/pkg/utils"
 )
 
+
+// createBootstrapInitContainer creates the init container bootstrapping the
+// initdb user inside the generated Pods
+func createInitDBUserContainer(cluster apiv1.Cluster) corev1.Container {
+	container := corev1.Container{
+		Name: 						BootstrapControllerContainerName,
+		Image: 						configuration.Current.OperatorImageName,
+		ImagePullPolicy:  cluster.Spec.ImagePullPolicy,
+		
+		Command: []string{
+			"/bin/sh",
+			"-c",
+			fmt.Sprintf("getent passwd %s &> /dev/null", cluster.Spec.postgresUID),
+			"||",
+			fmt.Sprintf("adduser --uid %s --gid %s --no-create-home --system --disabled-login --quiet %s", 
+				cluster.Spec.postgresUID, 
+				cluster.Spec.postgresGID,
+				"postgres"
+			 )
+		},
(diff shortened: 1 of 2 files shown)


armru:
Hello @wibed, thanks for the PR. We will take a look at it ASAP.

onedr0p:
Wouldn't these changes mean the container has to run as root?

wibed:
> Hello @wibed, It seems that the PR depends on the presence of `getent` and `adduser`. Those binaries could not be available in the container.
> 
> also CC @leonardoce , @mnencia and @phisco if you want to add your opinions

as it is based on debian, those packages are indeed present.
the command i ran to test is:
```shell
docker run -it -u 0 ghcr.io/cloudnative-pg/postgresql:15.2 bash
```

i did need root privileges to add a user or group to the system though.

wibed:
> Hello, I don't think we should tie CNP with root permissions, but this is my opinion. I would love to hear more opinions from the community and the maintainers.

id love to have multiple images to choose from. it comes down to who will maintain those over the timeline though

wibed:
@armru 

any update on this?

armru:
@wibed , 
Hello, probably the fastest way to proceed is to bring this up at the community meeting to gather feedback from the community/ other maintainers. Would you be available to join one of the upcoming ones?

@mnencia, @leonardoce any input on this patch?

mnencia:
We are trying to solve the wrong problem here. Having the `postgres` user inside the Postgres container is the responsibility of whoever builds the container image. If the environment changes the user to something different (like open shift does with restricted SCC), it is responsible for ensuring that that user exists inside the /etc/password file, and our operator supports it.

CNPG allows you to pass the desired user as part of the cluster configuration, which is all we need to do, IMHO.

wibed:
i add my two cents here as well as i did in the issue last week.

i see where you coming from, that any system related changes should not bleed into the responsibility of a postgres related deployment.

yet does initdb, a fundamental part of postgres, do expect a user which can be set within the settings already available from within the repo. 

i ask myself if it wouldnt be smarter to exclude that option and just force uid/gid nobody on everyone as it should.
what do you think?

sxd:
Please refer to https://github.com/cloudnative-pg/cloudnative-pg/issues/3331#issuecomment-1961476853 for my thoughts on this

mnencia:
I'm closing this pull request because it is outdated, and the maintainers agree that this should not be addressed as an init container.

PR #3343, opened 2023-11-15
ci: expose workflow's input properties

litaocdl:
This patch abstracts the input properties of several GHA workflows, improving their reusability and simplifying the way of overriding default values.


NiccoloFei (approved):

.github/workflows/continuous-delivery.yml
```diff
@@ -545,8 +549,10 @@ jobs:
       E2E_CSI_STORAGE_CLASS: csi-hostpath-sc
       E2E_DEFAULT_VOLUMESNAPSHOT_CLASS: csi-hostpath-snapclass
       LOG_DIR: ${{ github.workspace }}/kind-logs/
-      DOCKER_REGISTRY_MIRROR: https://mirror.gcr.io
+      REGISTRY_MIRROR: https://mirror.gcr.io
```
NiccoloFei, line 552: This needs to be changed also in the underlying scripts like `setup-cluster.sh` and so on

Issue #3339, opened 2023-11-14
[Bug]: cnpg.io/cluster label on pooler makes pods hard to distinguish and breaks podAntiAffinity


baurmatt:
### Is there an existing issue already for this bug?

- [X] I have searched for an existing issue, and could not find anything. I believe this is a new bug.

### I have read the troubleshooting guide

- [X] I have read the troubleshooting guide and I think this is a new bug.

### I am running a supported version of CloudNativePG

- [X] I have read the troubleshooting guide and I think this is a new bug.

### Version

1.21.0

### What version of Kubernetes are you using?

1.27

### What is your Kubernetes environment?

Cloud: Other

### How did you install the operator?

Helm

### What happened?

The introduction of the `cnpg.io/cluster` label in #3153 for pooler pods lead to 2 unintended consequences:

1. It's harder to label select the expected resource (PostgreSQL or PGBouncer):

```
$ kubectl get pods cnpg-db-7e2a3083-74ae-44df-8393-f01438a949ae-1 -o json | jq '.metadata.labels'
{
  "cnpg.io/cluster": "cnpg-db-7e2a3083-74ae-44df-8393-f01438a949ae",
  "cnpg.io/instanceName": "cnpg-db-7e2a3083-74ae-44df-8393-f01438a949ae-1",
  "cnpg.io/podRole": "instance"
}
$ kubectl get pods cnpg-pooler-7e2a3083-74ae-44df-8393-f01438a949ae-6dc5f89fbnlmmm -o json | jq '.metadata.labels'
{
  "cnpg.io/cluster": "cnpg-db-7e2a3083-74ae-44df-8393-f01438a949ae",
  "cnpg.io/poolerName": "cnpg-pooler-7e2a3083-74ae-44df-8393-f01438a949ae",
  "pod-template-hash": "6dc5f89fb4"
}
```

Before 1.21.1 we used label selectors `cnpg.io/cluster: $id` to select PostgreSQL pods. This isn't possible anymore and actually broken our CiliumNetworkPolicy :)

2. We run DBs/PGBouncer deployments on dedicated node deployments, thus having affinity rules set:

```
  affinity:
    nodeSelector:
      dedicated: {{ .Values.cnpgCluster.uuid }}
    tolerations:
    - key: "dedicated"
      operator: "Equal"
      value: "{{ .Values.cnpgCluster.uuid }}"
      effect: "NoSchedule"
```

[...]

PR #3346, opened 2023-11-15
fix: use app labels for affinity policy (#3339)

ceriath:
This resolves the issue described in #3339 by adding a label `cnpg.io/app` as suggested and use it in the anti affinity policy, to distinguish between postgresql itself and the pgbouncer pods.

### DIFF
--- a/docs/src/scheduling.md
+++ b/docs/src/scheduling.md
@@ -32,7 +32,8 @@ running in those nodes.
 This is technically known as **inter-pod affinity/anti-affinity**.
 
 CloudNativePG by default will configure the cluster's instances
-preferably on different nodes, resulting in the following `affinity` definition:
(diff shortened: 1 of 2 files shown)


gbartolini:
We should use the `podRole` label for this purpose.

gbartolini:
Can you please rebase your commit and try with podRole? Make sure you also update the docs. Thanks.

ceriath:
I rebased it ad used the `podRole` label. I am not exactly sure which docs i need to update, i hope this is the right place (as it seemed to be outdated aswell)

StopMotionCuber:
We're seeing the same issue and I was just in the process of creating an MR for that. I wanted to ask whether there are blockers to get this one merged?

mnencia:
Superseded by #5156

PR #3349, opened 2023-11-15
fix: reconcile `Pooler` service and add labels.

armru:
This patch fixes two issues that we currently have with the Pooler service.

a) we never reconcile the service once is created
b) we don't add the proper labels to the created service


gabriele-wolfox:
LGTM

gabriele-wolfox (approved):

PR #3350, opened 2023-11-15
fix: add proper labels to the `Pooler` deployment

armru:
This patch ensures that we add the proper labels to the deployment. In a previous patch we ensured that the labels were present on the pods but not on the actual deployment.


baurmatt:
This sounds like it fixes the problem described in #3339. There is also #3346 which introduces dedicated app labels for PostgreSQL and PGBouncer.

gabriele-wolfox (approved):

sxd (approved):

Issue #3353, opened 2023-11-15
[Bug]: graceful shutdown not being set on manager


jlong49:
### Is there an existing issue already for this bug?

- [X] I have searched for an existing issue, and could not find anything. I believe this is a new bug.

### I have read the troubleshooting guide

- [X] I have read the troubleshooting guide and I think this is a new bug.

### I am running a supported version of CloudNativePG

- [X] I have read the troubleshooting guide and I think this is a new bug.

### Contact Details

john.long@enterprisedb.com

### Version

1.21.0

### What version of Kubernetes are you using?

1.28

### What is your Kubernetes environment?

Self-managed: kind (evaluation)

### How did you install the operator?

YAML manifest

### What happened?

I noticed that the smart shutdown is not being applied to the controller manager. This could cause hard stops to happen as the controller manager will still default to 30s if this is not implemented. 

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

mnencia:
The fix for this issue has been reverted because it introduces a regression.

We need to discuss more about what we need to do to fix this issue.

PR #3354, opened 2023-11-15
bug: adding shutdown delay to controller manager

jlong49:
#3353


internal/cmd/manager/instance/run/cmd.go
```diff
@@ -115,6 +116,8 @@ func runSubCommand(ctx context.Context, instance *postgres.Instance) error {
 		"version", versions.Version,
 		"build", versions.Info)
 
+	gracefulShutdownTimeout := time.Duration(int64(instance.SmartStopDelay) * int64(time.Second))
```
YanniHu1996, line 119: I recommend using `instance.MaxStopDelay` instead of `instance.SmartStopDelay`, as the latter is solely for smart shutdowns. When a smart shutdown exceeds its timeout limit, it automatically switches to a fast shutdown. However, if the graceful shutdown of the instance also times out, it results in the same problem

YanniHu1996 (approved):

PR #3384, opened 2023-11-19
feat: allow setting a service type and annotations for a pooler resource

### LINKED ISSUE #3141: [Feature]: Allow exposing a pooler resource externally using a loadbalancer

In our use-case, CNPG runs in a specific K8S cluster while apps might be on a different K8S cluster. Hence, it is necessary to expose the pooler service externally using a service type `LoadBalancer`. Other people might have a similar use-case and will benefit from having this built into CNPG.

I am proposing that the `Pooler` resource be modified to specify a flag that denotes if the pooler should be exposed. Also, add a new field `loadBalancerSourceRanges` that is then used to create the pooler service.

```
kind: Pooler
metadata:
  name: pooler-example-rw
spec:
  cluster:
    name: cluster-example
  instances: 1
  type: rw
  external: true
  loadBalancerSourceRanges:
  - 192.168.122.0/24
  - 0.0.0.0/32
  pgbouncer:
    poolMode: transaction
```

So far we created a new service of type loadbalancer to expose the pooler externally. While this works fine, this introduces complexity in managing the deployment with two services, one managed by CNPG and one not managed by CNPG. It will be nice to have everything in the same umbrella.

### Additional context

_No response_

### Backport?

No

### Code of Conduct

achanda:
Closes #3141


achanda:
Todo:

- [ ] Make annotations optional
- [ ] More tests
- [ ] Docs update

api/v1/pooler_types.go
```diff
@@ -80,6 +80,14 @@ type PoolerSpec struct {
 	// The configuration of the monitoring infrastructure of this pooler.
 	// +optional
 	Monitoring *PoolerMonitoringConfiguration `json:"monitoring,omitempty"`
+
+	// What type of service to create for the pooler. Default: ClusterIP.
+	// +optional
+	ServiceType corev1.ServiceType `json:"servicetype,omitempty"`
```
phisco, line 86: Ideally this could fall under a serviceTemplate, providing a full service spec and metadata for labels and annotations, we could still apply the labels/annotation inheritance, but allow setting service specific ones here.

armru: I agree

pkg/servicespec/builder.go
```diff
@@ -0,0 +1,78 @@
...
+// Package servicespec contains various utilities to deal with Service Specs
+package servicespec
+
+import (
+	corev1 "k8s.io/api/core/v1"
+
+	apiv1 "github.com/cloudnative-pg/cloudnative-pg/api/v1"
+)
+
+// Builder enables to user to create a PodTemplate starting from a baseline
```
phisco, line 26:
```suggestion
// Builder enables users to create a serviceTemplate starting from a baseline
```

pkg/servicespec/builder.go
```diff
@@ -0,0 +1,78 @@
...
+// and adding patches
+type Builder struct {
+	status apiv1.ServiceTemplateSpec
+}
+
+// New creates a new empty podTemplate builder
```
phisco, line 32:
```suggestion
// New creates a new empty serviceTemplate builder
```

pkg/servicespec/builder.go
```diff
@@ -0,0 +1,78 @@
...
+	return builder
+}
+
+// WithServiceType adds a label to the current status
+func (builder *Builder) WithServiceType(serviceType corev1.ServiceType) *Builder {
+	builder.status.Spec.Type = serviceType
+	return builder
+}
+
+// Build gets the final Pod template
```
phisco, line 75:
```suggestion
// Build gets the final ServiceTemplate
```

pkg/specs/pgbouncer/services.go
```diff
@@ -23,23 +23,28 @@ import (
...
-			Labels: map[string]string{
-				utils.PgbouncerNameLabel: pooler.Name,
-				utils.ClusterLabelName:   cluster.Name,
-			},
+			Name:        pooler.Name,
+			Namespace:   pooler.Namespace,
+			Labels:      serviceTemplate.ObjectMeta.Labels,
+			Annotations: serviceTemplate.ObjectMeta.Annotations,
 		},
 		Spec: corev1.ServiceSpec{
```
phisco, line 46: Why not moving all these to the builder too and only setting that here?

phisco (commented):
Awesome work @achanda! I would get a little bit further with the templating before merging!🙏

pkg/specs/pgbouncer/services.go
achanda: Thanks, done that now. I also included the copy paste fixes you suggested. Once we agree on the implementation, I will work on tests.

pkg/servicespec/builder.go
```diff
@@ -0,0 +1,114 @@
...
+
+// WithServiceType adds a label to the current status
+func (builder *Builder) WithServiceType(serviceType corev1.ServiceType) *Builder {
+	builder.status.Spec.Type = serviceType
+	return builder
+}
+
+// WithPorts adds a label to the current status
+func (builder *Builder) WithPorts(port int) *Builder {
+	builder.status.Spec.Ports = []corev1.ServicePort{
```
phisco, line 79: Here it should get a whole ServicePort, iterate over the provided ports, if any, and if a port with the given name is found, update it, otherwise append it at the end of the list.

pkg/servicespec/builder.go
```diff
@@ -0,0 +1,114 @@
...
+			TargetPort: intstr.FromInt(port),
+			Port:       int32(port),
+		},
+	}
+	return builder
+}
+
+// WithSelector adds a label to the current status
+func (builder *Builder) WithSelector(name string) *Builder {
+	builder.status.Spec.Selector = map[string]string{
```
phisco, line 92: Here it should get a bool and overwrite only if true, only adding it otherwise.

achanda:
@phisco what usecase do you have in mind to enable adding to selectors and ports? I was thinking we don't enable adding to these and keep these internal. Later, if there is a usecase for modifying this then we revisit.

phisco:
@achanda just for the sake of completeness, as long as we handle that properly, I don't see a reason not to allow that right away. Not a strong requirement though.

achanda:
@phisco done and pushed, PTAL 🙏

phisco:
Awesome! Thanks, @achanda! The code looks good, I'll give it a spin tomorrow 🙏

pkg/specs/pgbouncer/services.go
```diff
@@ -18,39 +18,34 @@ package pgbouncer
...
 
 // Service create the specification for the service of
 // pgbouncer
 func Service(pooler *apiv1.Pooler, cluster *apiv1.Cluster) *corev1.Service {
+	serviceTemplate := servicespec.NewFrom(pooler.Spec.ServiceTemplate).
+		WithName(pooler.Name).
+		WithNamespace(pooler.Namespace).
+		WithLabel(utils.PgbouncerNameLabel, pooler.Name).
+		WithLabel(utils.ClusterLabelName, cluster.Name).
+		WithServiceType(pooler.Spec.ServiceTemplate.Spec.Type).
```
phisco, line 37: Here you should be setting the "default" service type instead, as the one from `pooler.Spec.ServiceTemplate.Spec.Type` is already set by `NewFrom` implicitly.
Given that you don't want to overwrite the default value, you should also add a boolean `overwrite` parameter, preferably as last parameter to adhere to the standard we already have for the podspec builder.
```suggestion
		WithServiceType(corev1.ClusterIP, true).
```

pkg/specs/pgbouncer/services.go
```diff
@@ -18,39 +18,34 @@ package pgbouncer
...
+		WithServicePort(&corev1.ServicePort{
+			Name:       "pgbouncer",
+			Port:       pgBouncerConfig.PgBouncerPort,
+			TargetPort: intstr.FromString("pgbouncer"),
+			Protocol:   corev1.ProtocolTCP,
+		}).
+		WithSelector(true, pooler.Name).
```
phisco, line 44: Please invert the two parameters to have the `overwrite` one last.
```suggestion
		WithSelector(pooler.Name, true).
```

config/crd/bases/postgresql.cnpg.io_poolers.yaml
phisco, line 1: unfortunately it looks like this CRD became too long and can not be applied anymore as the `kubectl.kubernetes.io/last-applied-configuration` annotation `kubectl apply` adds automatically, would not fit anymore in the maximum allowed size for annotations, at least on Kubernetes `v1.28.0`:
```
The CustomResourceDefinition "poolers.postgresql.cnpg.io" is invalid: metadata.annotations: Too long: must have at most 262144 bytes
```
I'll have to think about a workaround.

phisco (changes requested):
can you run a `make reviewable` to be sure the linter is happy and that the unit tests are all green too?

config/crd/bases/postgresql.cnpg.io_poolers.yaml
phisco: Unfortunately that's a pretty difficult issue to solve properly for an embedded resource :( https://book.kubebuilder.io/faq#the-error-too-long-must-have-at-most-262144-bytes-is-faced-when-i-run-make-install-to-apply-the-crd-manifests-how-to-solve-it-why-this-error-is-faced.

The best option would be to drop all descriptions for the fields in the service (and pod) template, but unfortunately I didn't find any available tool to do that. Usually you don't care about required fields either, so such tool would have to be able to "vendor" a struct from an upstream project, stripping annotations. We have a similar need in Crossplane, so I could take the chance to develop something able to do that. But that will require time. In the meantime we could either hack something together to cleanup the crds after having generated them using something like `yq` for example 🤔

achanda: I need to figure this one out, I will get to this next week.

achanda:
@phisco I checked that I can apply these CRDs on a kind cluster. This should be ready for review now.

```
╭─Abhishek@AbhishekMacbookPro ~/src/cloudnative-pg ‹external-pooler›
╰─$ kubectl apply -f config/crd/bases/
customresourcedefinition.apiextensions.k8s.io/backups.postgresql.cnpg.io created
customresourcedefinition.apiextensions.k8s.io/clusters.postgresql.cnpg.io unchanged
customresourcedefinition.apiextensions.k8s.io/poolers.postgresql.cnpg.io created
customresourcedefinition.apiextensions.k8s.io/scheduledbackups.postgresql.cnpg.io created
```

phisco:
Depending on the decision taken by https://github.com/cloudnative-pg/cloudnative-pg/issues/3697 we could be fine with the full CRD for the pooler too here.

phisco:
@achanda, as you can see we agreed to use server-side apply for CRDs: https://github.com/cloudnative-pg/cloudnative-pg/pull/3729

This means here we can have the full schema produced, could you revert the last commit? Sorry for the back and forth, but it was a big decision to take.

achanda:
No worries @phisco I have updated this branch now. Thanks for all the help 🙏

api/v1/pooler_types.go
```diff
@@ -126,6 +130,20 @@ type PodTemplateSpec struct {
 	Spec corev1.PodSpec `json:"spec,omitempty"`
 }
 
+// ServiceTemplateSpec is a structure allowing the user to set
+// a template for Service generation.
+type ServiceTemplateSpec struct {
+	// Standard object's metadata.
+	// More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#metadata
+	// +optional
+	ObjectMeta metav1.ObjectMeta `json:"metadata,omitempty"`
```
phisco, line 139:
```suggestion
	ObjectMeta Metadata `json:"metadata,omitempty"`
```
Same as the pod template above, will need to be converted

pkg/specs/pgbouncer/services.go
```diff
@@ -23,34 +23,45 @@ import (
...
+	"github.com/cloudnative-pg/cloudnative-pg/pkg/utils/hash"
 )
 
 // Service create the specification for the service of
 // pgbouncer
-func Service(pooler *apiv1.Pooler, cluster *apiv1.Cluster) *corev1.Service {
+func Service(pooler *apiv1.Pooler, cluster *apiv1.Cluster) (*corev1.Service, error) {
+	serviceTemplate := servicespec.NewFrom(pooler.Spec.ServiceTemplate).
+		WithLabel(utils.PgbouncerNameLabel, pooler.Name).
+		WithLabel(utils.ClusterLabelName, cluster.Name).
```
phisco, line 36:
```suggestion
        poolerHash, err := hash.ComputeVersionedHash(pooler.Spec, 3)
	if err != nil {
		return nil, err
	}

	serviceTemplate := servicespec.NewFrom(pooler.Spec.ServiceTemplate).
		WithLabel(utils.PgbouncerNameLabel, pooler.Name).
		WithLabel(utils.ClusterLabelName, cluster.Name).
		WithLabel(utils.PodRoleLabelName, string(utils.PodRolePooler)).
		WithAnnotation(utils.PoolerSpecHashAnnotationName, poolerHash).
```

pkg/specs/pgbouncer/services.go
```diff
@@ -23,34 +23,45 @@ import (
...
-					Port:       pgBouncerConfig.PgBouncerPort,
-				},
-			},
-			Selector: map[string]string{
 				utils.PgbouncerNameLabel: pooler.Name,
+				utils.PodRoleLabelName:   string(utils.PodRolePooler),
+			},
+			Annotations: map[string]string{
+				utils.PoolerSpecHashAnnotationName: poolerHash,
 			},
```
phisco, line 63:
```suggestion
			Annotations: serviceTemplate.ObjectMeta.Annotations,
```

phisco: Same should be done for the Labels above, github didn't let me:
```
			Labels:      serviceTemplate.ObjectMeta.Labels,
```

pkg/specs/pgbouncer/services.go
```diff
@@ -23,34 +23,45 @@ import (
...
+			Protocol:   corev1.ProtocolTCP,
+		}).
+		WithSelector(pooler.Name, true).
+		Build()
+
+	poolerHash, err := hash.ComputeVersionedHash(pooler.Spec, 3)
+	if err != nil {
+		return nil, err
+	}
+
```
phisco, line 51: moved above
```suggestion
```

phisco (changes requested):

achanda:
@phisco gentle ping on this, do you need anything else from me?

pkg/servicespec/builder.go
```diff
@@ -0,0 +1,104 @@
...
+	}
+
+	builder.status.ObjectMeta.Labels[name] = value
+
+	return builder
+}
+
+// WithServiceType adds a service type to the current status
+func (builder *Builder) WithServiceType(serviceType corev1.ServiceType, overwrite bool) *Builder {
+	if overwrite {
```
phisco, line 72: This should set the type also if it's not set:
```suggestion
	if overwrite || builder.status.Spec.Type == "" {
```

pkg/servicespec/builder.go
```diff
@@ -0,0 +1,104 @@
...
+// WithSelector adds a selector to the current status
+func (builder *Builder) WithSelector(name string, overwrite bool) *Builder {
+	if overwrite {
+		builder.status.Spec.Selector = map[string]string{
+			utils.PgbouncerNameLabel: name,
+		}
+	}
+
+	return builder
+}
```
phisco, line 99: Same as above, overwrite means if present overwrite:
```suggestion
func (builder *Builder) WithSelector(name string, overwrite bool) *Builder {
	s := builder.status.Spec.Selector

	if len(s) == 0 {
		s = make(map[string]string, 1)
	}

	if _, exists := s[utils.PgbouncerNameLabel]; overwrite || !exists {
		s[utils.PgbouncerNameLabel] = name
	}

	builder.status.Spec.Selector = s

	return builder
}
```

phisco (approved):
Thanks @achanda! Sorry for taking so long, I had a pretty busy period at work 😅

achanda:
@phisco no worries at all, I know last week was kubecon :) The failing tests look unrelated to me, mind running the tests again please? I don't have permission

phisco:
Yes, I think they are too, I'll let the second reviewer decide if they want to run them again or just mark it as good to go straight away 👍

sxd:
Hi @achanda would you like to add more documentation? I just added a small example but maybe you want to create a file in docs/src/samples/ directory that can be used ?

achanda:
@sxd done now, added a new file named pooler-external.yaml and referenced that in the doc

docs/src/connection_pooling.md
```diff
@@ -256,6 +256,36 @@ spec:
...
+    spec:
+      type: LoadBalancer
+  pgbouncer:
+    poolMode: session
+    parameters:
+      max_client_conn: "1000"
+      default_pool_size: "10"
+```
+
+An example of this is in [`pooler-external.yaml`](samples/pooler-external.yaml)
```
sxd, line 287: there's a full section of samples in samples.md where this should go

achanda: Made that change, thanks!

sxd (approved):

PR #3458, opened 2023-12-05
review: Change MergeMap

YanniHu1996:


YanniHu1996:
Close this pull request as we should not change the function to sometimes mutate an object  sometime not

PR #3469, opened 2023-12-05
Check for correct pooler secret name

### LINKED ISSUE #2479: "Pooler not automatically configured, manual configuration required" issue popping up on all poolers
I am seeing this issue pop up for all of my poolers, despite having what I believe to be correctly configured poolers.  According to the [docs](https://cloudnative-pg.io/documentation/1.15/connection_pooling/) all I need is for the pooler to set `cluster.name` to be that of the cluster, and to not let its own name conflict.

The image I'm using is: `ghcr.io/cloudnative-pg/cloudnative-pg:1.20.0`

Here is a sample message from the operator logs:
```
{"level":"info","ts":"2023-07-21T10:22:04Z","msg":"pooler not automatically configured, manual configuration required","cluster":"cnpg-cluster","pooler":"cnpg-pooler"}
```

The pooler configuration:

```
---
# Source: cnpg-pooler/templates/pooler.yaml
kind: Pooler
metadata:
  name: cnpg-pooler
  namespace: "database"
  labels:
    app.kubernetes.io/name: cnpg-pooler
    helm.sh/chart: cnpg-pooler-0.1.0
    app.kubernetes.io/instance: pooler
    app.kubernetes.io/managed-by: Helm
spec:
  cluster:
    name: cnpg-cluster
  instances: 1
  type: rw
  pgbouncer:
    poolMode: session
    parameters:
      null
```

The cluster config:

```
---
# Source: cnpg-cluster/templates/cluster.yaml
kind: Cluster
metadata:
  name: cnpg-cluster
  namespace: "database"
  labels:
    app.kubernetes.io/name: cnpg-cluster
    helm.sh/chart: cnpg-cluster-0.1.2
    app.kubernetes.io/instance: cluster
    app.kubernetes.io/managed-by: Helm
spec:
  instances: 2

  superuserSecret:
    name: superuser-creds

  managed:
    roles:
      - name: readonly
        ensure: present
        login: true
        superuser: false
        passwordSecret:
          name: readonly-creds
        inRoles:
          - pg_read_all_data

  storage:
    size: 1Gi
```

Am I missing something?

baurmatt:
Otherwise CloudnativePG will always log:

> pooler not automatically configured, manual configuration required

Fixes #2479

### DIFF
--- a/controllers/cluster_status.go
+++ b/controllers/cluster_status.go
@@ -395,7 +395,7 @@ func (r *ClusterReconciler) getPoolerIntegrationsNeeded(ctx context.Context,
 	pgbouncerPoolerIntegrations, err := r.getPgbouncerIntegrationStatus(ctx, cluster, poolers)
 
 	for _, pooler := range poolers.Items {
-		if !slices.Contains(pgbouncerPoolerIntegrations.Secrets, pooler.Name) {
+		if !slices.Contains(pgbouncerPoolerIntegrations.Secrets, pooler.GetAuthQuerySecretName()) {
 			log.Info("pooler not automatically configured, manual configuration required",
 				"cluster", pooler.Spec.Cluster.Name, "pooler", pooler.Name)
 		}


controllers/cluster_status.go
```diff
@@ -395,7 +395,7 @@ func (r *ClusterReconciler) getPoolerIntegrationsNeeded(ctx context.Context,
 	pgbouncerPoolerIntegrations, err := r.getPgbouncerIntegrationStatus(ctx, cluster, poolers)
 
 	for _, pooler := range poolers.Items {
-		if !slices.Contains(pgbouncerPoolerIntegrations.Secrets, pooler.Name) {
+		if !slices.Contains(pgbouncerPoolerIntegrations.Secrets, pooler.GetAuthQuerySecretName()) {
```
phisco, line 398: You are right about the missing -pooler, however this would defeat the purpose of the check, as it would always return true by definition. What we want to check here is that the default secret name is in the list, which is probably a too weak condition to start with.

baurmatt: Oh oO Understood, this effectively checks all possible combinations 🤔 Would it make sense to check for `pooler.Name + DefaultPgBouncerPoolerSecretSuffix`? 🤔

phisco: That would still mean a log line is printed for every reconciliation loop if the pooler auth is not automatically handled. Moving the logging to a place where we are able to tell if we actually created the secret and do that only on creation would be the optimal solution.

phisco:
FYI: https://github.com/cloudnative-pg/cloudnative-pg/pull/3585

baurmatt:
Looks good, thank you! Closing this in favor of #3585.

PR #3500, opened 2023-12-07
Edits to new tablespaces and samples content

ebgitelman:


docs/src/tablespaces.md
```diff
@@ -291,21 +291,21 @@ spec:
...
 currently the main `PGDATA` volume.
 
 When you specify more than one temporary tablespace, PostgreSQL randomly picks
-one the first time a temporary object needs to be created in a transaction,
-then sequentially iterates through the list.
+one the first time a temporary object needs to be created in a transaction.
+Then it sequentially iterates through the list.
 
-Temporary tablespaces work like regular tablespaces, also regarding backups.
+Temporary tablespaces also work like regular tablespaces with regard to backups.
```
jsilvela, line 303: I think part of the meaning is lost here. The main message is that temporary tablespaces work like regular tablespaces.
The backups part is just for reassurance, not the main take-away.
Not sure how to say it better though...

docs/src/tablespaces.md
```diff
@@ -125,12 +125,12 @@ status:
 
 ## Storage classes and tablespaces
 
-As for PGDATA and WAL volumes, you can use different storage classes for your
-tablespaces too. This is a very convenient way of optimizing your resources,
+As you can for PGDATA and WAL volumes, you can use different storage classes for your
+tablespaces. This is a convenient way of optimizing your resources,
```
jsilvela, line 129: Maybe this would be simpler to parse?
```suggestion
You can use different storage classes for your tablespaces, just as you can for PGDATA and
WAL volumes. This is a convenient way of optimizing your resources,
```

docs/src/tablespaces.md
```diff
@@ -153,18 +153,18 @@ spec:
...
 
-The `yardbirds` cluster example above requests 4 persistent volume claims using
+The `yardbirds` cluster example requests 4 persistent volume claims using
 3 different storage classes:
 
-- default storage class: used by the `PGDATA` and WALs
-- `fastest`: used by the `current` tablespace to store the most active and
-  demanding set of data in the database
-- `balanced`: used by the `this_year` tablespace to store older partitions of
+- Default storage class – Used by the `PGDATA` and WALs.
```
jsilvela, line 159:
```suggestion
- Default storage class – Used by the `PGDATA` and WAL volumes.
```

docs/src/tablespaces.md
```diff
@@ -239,31 +239,31 @@ spec:
...
 
 Replica clusters must have the same tablespace definition as their origin.
 The reason is that tablespace management commands like `CREATE TABLESPACE`
-are WAL logged and will be replayed by any physical replication client (streaming and/or via WAL shipping).
+are WAL logged and are replayed by any physical replication client (streaming or by way of WAL shipping).
 
-It is your responsibility to ensure that replica cluster have the same list of
-tablespaces, with the same name (storage class and size might vary).
+It's your responsibility to ensure that replica clusters have the same list of
+tablespaces, with the same name. (Storage class and size might vary.)
```
gbartolini, line 266:
```suggestion
tablespaces, with the same name. Storage class and size might vary.
```

docs/src/tablespaces.md
gbartolini: I agree

docs/src/tablespaces.md
```diff
@@ -291,21 +291,21 @@ spec:
 
 PostgreSQL allows you to define one or more temporary tablespaces to create
 temporary objects (temporary tables and indexes on temporary tables) when a
-`CREATE` command does not explicitly specify a tablespace, as well as temporary
+`CREATE` command doesn't explicitly specify a tablespace. It also allows you to define temporary
```
gbartolini, line 294:
```suggestion
`CREATE` command doesn't explicitly specify a tablespace, and to create temporary
```

docs/src/tablespaces.md
```diff
@@ -318,22 +318,22 @@ spec:
 ```
 
 They can be created at initialization time or added later, requiring a
-rolling update. The `temporary: true/false` option simply adds/removes the
-tablespace name to/from the list of tablespaces in the `temp_tablespaces`
-option (which doesn't require a restart of PostgreSQL to be changed).
+rolling update. The `temporary: true/false` option adds or removes the
+tablespace name to or from the list of tablespaces in the `temp_tablespaces`
+option. (This change doesn't require a restart of PostgreSQL.)
```
gbartolini, line 323:
```suggestion
option. This change doesn't require a restart of PostgreSQL.
```

gbartolini (approved):

ebgitelman:
> Thanks @ebgitelman , good changes. I have left a couple of suggestions. I see there are some very consistent fixes. For example, keeping the informal contractions, or using the present tense rather than the future. Taking note.
> 
> One change you've also consistently made is adding periods at the end of list items. We've been fairly consistent in the document files about not adding periods. IMO if we're moving to add periods, which I'm not against, we might think of doing a sweep of all the documents in a PR dedicated to doing just that.

I'm actually making this change in every doc I'm editing, which is the entire content. 

The ting with not using periods in what I call a definition list is that people go through contortions to avoid adding the period. This results in a complex structure involving parentheses and semi-colons. So I prefer to just use the period, so you have the flexibility to have multiple sentences. There are some exceptions, but basically for definitions, I like the periods.

PR #3532, opened 2023-12-12
 feat: Allow shrinking of storage

### LINKED ISSUE #2405: Allow shrinking of storage
It is currently not possible to shrink a PVC. The webhook is restricting this with the following message:
```
cannot patch "<name>" with kind Cluster: admission webhook "vcluster.kb.io" denied the request: Cluster.cluster.cnpg.io "<name>" is invalid: spec.storage: Invalid value: resource.Quantity{i:resource.int64Amount{value:0, scale:0}, d:resource.infDecAmount{Dec:(*inf.Dec)(0xc001615470)}, s:"10Gi", Format:"BinarySI"}: can't shrink existing storage from 20Gi to 10Gi
```

I understand that the PVC is not supporting shrinking, since this is a limitation of the K8s.

Would a downsizing be possible with re-creating the underlying storage?
So to have a similar experience as when expanding with the flag `resizeInUseVolumes`?

YanniHu1996:
Unfasten the restriction of shrinking of storage while the `resizeInUseVolumes` is set to false 
Introduced a new type of condition `PVCResizeSucceeded` to record the result of offline resizing pvcs

closes: #2405

### DIFF
--- a/api/v1/cluster_types.go
+++ b/api/v1/cluster_types.go
@@ -862,6 +862,8 @@ const (
 	ConditionBackup ClusterConditionType = "LastBackupSucceeded"
 	// ConditionClusterReady represents whether a cluster is Ready
 	ConditionClusterReady ClusterConditionType = "Ready"
+	// ConditionPVCResize represents whether a PVC resize is successful
+	ConditionPVCResize ClusterConditionType = "PVCResizeSucceeded"
 )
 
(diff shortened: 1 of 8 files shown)


YanniHu1996:
the e2e result from the folk
https://github.com/EnterpriseDB/cloudnative-pg/actions/runs/7197032167

PR #3533, opened 2023-12-12
feat: reconciler for `externalClusters` entries

### LINKED ISSUE #2386: Enhance externalClusters management of password to use `.pgpass` files
At the moment, whenever we use the `password` field contained in the declared `externalCluster.password` secret to build the connection information to the PostgreSQL server we insert the password in the `DSN`.

The proposal here is to actually leverage the [PostgreSQL `.pgpass` file](https://www.postgresql.org/docs/current/libpq-pgpass.html) for this purpose by omitting the `password` field in the DSN, therefore triggering the password file capability.

The idea is to generate for each `externalCluster` that contains a `password` field an entry in the `~/.pgpass` file, with the following content:

```
hostname:port:*:username:password
```

Note that the `dbname` is fixed to `*` as we do not want to discriminate based on the target database, just the pair host:port.

As part of this feature, please make sure that the code is refactored to use this technique everytime we use the `password` field in a way that is potentially and unnecessarily exposed.

leonardoce:
Whenever a password is supplied within an `externalClusters` entry, the instance manager autonomously manages a PostgreSQL password file for it, residing at `/controller/external/NAME/pgpass` in each instance (where "NAME" is the name of the external cluster entry).

This approach empowers CloudNativePG to securely establish connections with an external server without exposing any passwords in the connection string. Instead, the connection safely references the aforementioned file through the
`passfile` connection parameter.

Closes #2386


leonardoce:
E2e: https://github.com/EnterpriseDB/cloudnative-pg/actions/runs/7181924883

litaocdl (approved):

PR #3534, opened 2023-12-12
feat: configuration of PostgreSQL ident map file

### LINKED ISSUE #3362: [Feature]: allowing additional lines to be added to pg_ident file

It would be nice to have a way to add user maps, so we can use certificate based authentication where the common name of the certificate is different from the postgres role name.

For example:

In pg_hba.conf:
`hostssl  all  +users  all  cert  map=user-map`

In pg_ident.conf
`user-map  /^(.*)\.user\.example\.com  \1`

There's a related issue where a certificate issuer I'm using doesn't allow the `_` in `streaming_replica`, which could be solved by mapping `streaming-replica` to `streaming_replica`.

My preferred solution would be to add a `pg_ident` list to the Custer CRD, where each line adds a line to the pg_ident file. This mirrors the current way to customize the pg_hba file.

An alternative way could be to the manual addition of a `pg_ident.conf` file as volumeMount and allow to override the `ident_file` postgres setting, which is currently blocked from being modified.

### Additional context

No other context.

### Backport?

No

### Code of Conduct

nilsbrenkman:
As described in the issue, we would like a way to add lines to the pg_ident.conf file.

I followed the same approach as pg_hba, it's basically a 1 on 1 mapping.

The pull request contains unit test and e2e test. Documentation needs to be added still.

Closes #3362


armru:
Hello @nilsbrenkman We will look into this ASAP

pkg/management/postgres/configuration.go
```diff
@@ -197,6 +198,42 @@ func quoteHbaLiteral(literal string) string {
...
+		pgIdentContent,
+		constants.PostgresqlIdentFile)
+	if err != nil {
+		return postgresIdentChanged, fmt.Errorf(
+			"installing postgresql Ident rules: %w",
+			err)
+	}
+
+	return postgresIdentChanged, err
+}
```
mnencia, line 235: I think that this code overlaps with the [WritePostgresUserMaps](https://github.com/cloudnative-pg/cloudnative-pg/blob/main/pkg/management/postgres/ident.go#L31) function. We should really have only one place doing this work.

armru (approved):

armru:
Hello @nilsbrenkman , can you add the documentation?

mnencia (approved):

nilsbrenkman:
> Hello @nilsbrenkman , can you add the documentation?

Will do!

Issue #2725, opened 2023-09-11
Allow users to disable `ALTER SYSTEM`


gbartolini:
**Background information:** I have submitted a patch for PostgreSQL to disable `ALTER SYSTEM` at startup time. For more information and also the reason behind it, please refer to the thread in the pgsql-hackers mailing list below:

https://www.postgresql.org/message-id/CA%2BVUV5rEKt2%2BCdC_KUaPoihMu%2Bi5ChT4WVNTr4CD5-xXZUfuQw%40mail.gmail.com

If ever, this patch will not land in PostgreSQL before version 17. There are chances though that it will never land.

In the meantime, it would be good to introduce a workaround for all currently available Postgres versions, including the upcoming 16. Add a new option in the cluster's spec called `enableAlterSystem` with a current default value of `true` for back-compatibility (in 1.22 we should make the default change to `false`).

The idea is that the `postgresql.auto.conf` file is made read-only. Add proper documentation too.

leonardoce:
I'd disable `ALTER SYSTEM` in CNPG. I can't imagine a case where this should be used when the cluster is working.

I understand this would be useful to fix a broken cluster, but in this case, the user could:

1. Fence the instances
2. Update the `postgres.conf` file as needed

By doing this the user will have total control over the GUCs, and over where/which configuration file we include.

PR #3535, opened 2023-12-12
feat: allow users to disable ALTER SYSTEM

leonardoce:
Introduce the `.spec.postgresql.enableAlterSystem` to enable/disable the usage of the `ALTER SYSTEM` command for a PostgreSQL instance. Internally, the operator disables writes on the `$PGDATA/postgresql.auto.conf` file when `.spec.postgresql.enableAlterSystem` is set to `false`.

For back compatibility, the default is `true`.

Closes: #2725

### DIFF
--- a/.wordlist-en-custom.txt
+++ b/.wordlist-en-custom.txt
@@ -634,6 +634,7 @@ dx
 ecdsa
 edb
 eks
+enableAlterSystem
 enablePodAntiAffinity
 enablePodMonitor
 enableSuperuserAccess
--- a/.wordlist-en-custom.txt
+++ b/.wordlist-en-custom.txt
@@ -859,6 +859,7 @@ olm
 ongoingBackups
 onlineConfiguration
 onlineUpdateEnabled
+onwards
 openldap
 openshift
 operability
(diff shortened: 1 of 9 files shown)


leonardoce:
E2e: https://github.com/EnterpriseDB/cloudnative-pg/actions/runs/7184102912

gbartolini:
I have tested it locally. By default this is the behaviour:

```
$ kubectl exec -ti pg1-1 -c postgres -- ls -la /var/lib/postgresql/data/pgdata/postgresql.auto.conf
-r-------- 1 postgres tape 88 Dec 13 01:55 /var/lib/postgresql/data/pgdata/postgresql.auto.conf
```

Then, once I add:

```
...
  postgresql:
    enableAlterSystem: true
...
```

I get this:

```
$ kubectl exec -ti pg1-1 -c postgres -- ls -la /var/lib/postgresql/data/pgdata/postgresql.auto.conf
-rw------- 1 postgres tape 88 Dec 13 01:55 /var/lib/postgresql/data/pgdata/postgresql.auto.conf
```

I manually changed the file permissions and noticed that with the next reconciliation loop they were restored.

leonardoce:
I changed the patch to set `enableAlterSystem` to `true` by default. This PR is safe to backport now.
After system we can file a new PR to change the default to `false` without backporting it.

mnencia:
E2E tests running in EDB fork:
- Default to false https://github.com/EnterpriseDB/cloudnative-pg/actions/runs/7195570970
- Default to true https://github.com/EnterpriseDB/cloudnative-pg/actions/runs/7195840409

mnencia (approved):

mnencia:
/ok-to-merge all E2E green, both with default off and on

armru (approved):

Issue #3563, opened 2023-12-14
[Bug]: When creating a post-import 3 instance cluster pg_rewind grabs a Volumesnapshot with an incorrect db ID


chrisalavoinehmm:
### Is there an existing issue already for this bug?

- [X] I have searched for an existing issue, and could not find anything. I believe this is a new bug.

### I have read the troubleshooting guide

- [X] I have read the troubleshooting guide and I think this is a new bug.

### I am running a supported version of CloudNativePG

- [X] I have read the troubleshooting guide and I think this is a new bug.

### Contact Details

chris.alavoine@humanmademachine.com

### Version

1.21.1

### What version of Kubernetes are you using?

1.28

### What is your Kubernetes environment?

Cloud: Google GKE

### How did you install the operator?

YAML manifest

### What happened?

 I see this error when I run my post-import YAML (we are importing from PG 11 to 16). The first instance of my 3 instance cluster spins up ok and I can connect etc but the second instance gets stuck restarting and I'm seeing this in the logs:
`{"level":"info","ts":"2023-12-13T18:21:53Z","logger":"pg_rewind","msg":"pg_rewind: error: source and target clusters are from different systems","pipe":"stderr","logging_pod":"prod16-2"} `

I discovered that it was taking a volumesnapshot backup with the same cluster name (from a previous incarnation of the cluster) and trying to create the standby instance from there. This caused problems with pg_rewind as above.

@leonardoce has requested I raise this bug. He also said "We could fix it by avoiding reusing VolumeSnapshots that have been created before the Cluster, unless they are specified in the Cluster Spec itself (in the recovery from VolumeSnapshot section)"

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #3570, opened 2023-12-15
fix: avoid considering old snapshots when creating a replica

leonardoce:
This patch prevents CNPG from using backups that are older than the created Cluster when bootstrapping a new replica.

Closes: #3563


mnencia (approved):

gbartolini (approved):

leonardoce:
E2e: https://github.com/EnterpriseDB/cloudnative-pg/actions/runs/7260068202

leonardoce:
Azurite-based tests are failing, we already know that and are working on it. I'm merging this PR.

PR #3573, opened 2023-12-15
fix: use the correct version of Go when building OLM catalog

leonardoce:


.github/workflows/continuous-integration.yml
```diff
@@ -639,6 +639,14 @@ jobs:
         with:
           platforms: ${{ needs.buildx.outputs.platforms }}
 
+      - name: Install Go
+        uses: actions/setup-go@v5
+        with:
+          # Disable setup-go caching. Cache is better handled by the golangci-lint action
+          cache: false
```
mnencia, line 646: Shouldn't we drop these two lines? They are meant only for the linter step. The olm builder would benefit from the cache.

PR #3583, opened 2023-12-15
fix(e2e): avoid uploading multiple artifacts with same name

leonardoce:
This is needed to support upload-artifact and download-artifact v4.


leonardoce:
E2e: https://github.com/EnterpriseDB/cloudnative-pg/actions/runs/7222728440

sxd (approved):

PR #3589, opened 2023-12-18
fix(doc): indentation of example in `recovery.md`

BaudouinH:
Some examples in the documentation page `recovery.md` have the wrong indentation:
The `bootstrap` block is at the same indentation level as the spec block, when it should belong to the `spec` block according to the API documentation: https://cloudnative-pg.io/documentation/1.21/cloudnative-pg.v1/#postgresql-cnpg-io-v1-ClusterSpec

### DIFF
--- a/docs/src/recovery.md
+++ b/docs/src/recovery.md
@@ -129,8 +129,8 @@ metadata:
   name: cluster-restore
 spec:
   [...]
-
-bootstrap:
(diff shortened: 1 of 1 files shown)


leonardoce (approved):

leonardoce:
Thanks @BaudouinH !

PR #3624, opened 2023-12-21
test: support upgrade test from the current version

### LINKED ISSUE #3592: [Feature]: Add E2E test for upgrade from the version under test

Our current upgrade E2E test only checks for errors when upgrading from the latest stable version to the version under test. However, the version under test could break the upgrade mechanics to any other future version. We need to catch this kind of bug before merging the changes.

We should add a further step to our upgrade E2E test, where we try updating from the version under test to itself (same code, different image/hash). In this way, we can test if we break the upgrade logic.

N/A

### Additional context

We merged PR #3354, which broke the online upgrade. We had to revert it with PR #3590. This test would have caught the issue. 

### Backport?

Yes

Yes

### Code of Conduct

YanniHu1996:
Adds two E2E test cases where the operator is updated from the
current source build to a future version (which is simulated with a
modified build from the current source tree).
This will allow us to catch code that would break future upgrades.

Closes #3592


hack/e2e/run-e2e.sh
```diff
@@ -79,10 +79,6 @@ echo "E2E tests are running with the following filters: ${LABEL_FILTERS}"
 RC=0
 RC_GINKGO1=0
 if [[ "${TEST_UPGRADE_TO_V1}" != "false" ]]; then
-  # Getting the operator images need a pull secret
-  kubectl delete namespace cnpg-system || :
-  kubectl create namespace cnpg-system
-  ensure_image_pull_secret
```
mnencia, line 85: Why has this code been deleted? AFAIR it is important.

tests/e2e/upgrade_test.go
```diff
@@ -95,6 +96,34 @@ var _ = Describe("Upgrade", Label(tests.LabelUpgrade, tests.LabelNoOpenshift), O
...
 			Skip("Test depth is lower than the amount requested for this test")
 		}
+
+		err := env.DeleteNamespace(operatorNamespace)
+		if !apierrs.IsNotFound(err) {
+			Expect(err).NotTo(HaveOccurred())
+		}
+
+		err = env.CreateNamespace(operatorNamespace)
+		Expect(err).NotTo(HaveOccurred())
```
jsilvela, line 106: This is going to happen on top of the `cleanupNamespace` and `assertCreateNamespace` we already have.
Locally I'm seeing problems with the operator namespace being there or not in unexpected ways.
Why is this necessary?

tests/e2e/upgrade_test.go
```diff
@@ -95,6 +96,34 @@ var _ = Describe("Upgrade", Label(tests.LabelUpgrade, tests.LabelNoOpenshift), O
...
+			--docker-server="%v"
+			--docker-username="%v"
+			--docker-password="%v"`,
+				operatorNamespace,
+				dockerServer,
+				dockerUsername,
+				dockerPassword,
+			))
+			Expect(err).NotTo(HaveOccurred())
+		}
```
jsilvela, line 126: isn't this covered in `ensure_image_pull_secret` in the run-e2e.sh script?
Why is it here?

jsilvela:
OK, I think I have a simplification of this that will work. @YanniHu1996 let me address the comments so far in my next PR.

hack/e2e/run-e2e.sh
YanniHu1996: please look at https://github.com/cloudnative-pg/cloudnative-pg/pull/3624#issuecomment-1875083582

tests/e2e/upgrade_test.go
YanniHu1996: please look at https://github.com/cloudnative-pg/cloudnative-pg/pull/3624#issuecomment-1875083582

tests/e2e/upgrade_test.go
YanniHu1996: As the comment https://github.com/cloudnative-pg/cloudnative-pg/pull/3624#issuecomment-1875083582 said, creating a namespace before creating a pull image secret

jsilvela (approved):

hack/e2e/run-e2e.sh
```diff
@@ -79,27 +79,30 @@ echo "E2E tests are running with the following filters: ${LABEL_FILTERS}"
...
-      cd "${CONFIG_TMP_DIR}/default"
-      "${KUSTOMIZE}" edit add patch --path manager_image_pull_secret.yaml
-      cd "${CONFIG_TMP_DIR}/manager"
-      "${KUSTOMIZE}" edit set image "controller=${CONTROLLER_IMG}"
-      "${KUSTOMIZE}" edit add patch --path env_override.yaml
-      "${KUSTOMIZE}" edit add configmap controller-manager-env \
-        --from-literal="POSTGRES_IMAGE_NAME=${POSTGRES_IMG}"
-  )
-  "${KUSTOMIZE}" build "${CONFIG_TMP_DIR}/default" > "${ROOT_DIR}/tests/e2e/fixtures/upgrade/current-manifest.yaml"
+if [[ "${TEST_UPGRADE_TO_V1}" != "false" ]]; && [[ "${OPENSHIFT}" != "true" ]]; then
```
jsilvela, line 82:
```suggestion
if [[ "${TEST_UPGRADE_TO_V1}" != "false" ]] && [[ "${OPENSHIFT}" != "true" ]]; then
```

NiccoloFei (approved):

PR #3639, opened 2023-12-22
chore(crd): reduce CRD size

leonardoce:
This patch declare the `pooler.spec.template.spec` field as SchemaLess, result in a much lightweight CRD.

This allows our users to upgrade to new versions using `kubectl apply -f` as we won't hit anymore the maximum annotation length.

The downside of it is that there's no validation done by the API server.


leonardoce:
Please bear with me; I don't love this as much as you do!
But this will help us fixing #3604

phisco (commented):
Makes sense to me, it's unfortunate, but it solves the issue, we'll do the same for the service template. We can always add back validation in the webhook later on.

leonardoce:
Yeah, and we'll fix #3552 too. About adding validation in the webhook: that would require reaching the API server while we're being reached to validate the `Cluster` resource or directly reusing the [Kubernetes source code](https://github.com/kubernetes/kubernetes/blob/3f7a50f38688eb332e2a1b013678c6435d539ae6/pkg/apis/apps/validation/validation.go#L647) do to that.

I don't like the second choice because we'll bring in many dependencies, and I don't like the first one because we'll put probably too much stress on the API server.

Do you have any other insights @phisco?

phisco:
🤔 we could embed the deployment template openAPI schema and use that to do the validation in the webhook. But I don't like that much either.

leonardoce:
Reusing Kubernetes validation is a bad idea, and it's [not supported too](https://github.com/kubernetes/kubernetes/#to-start-using-k8s).
See https://github.com/kubernetes/kubernetes/issues/79384#issuecomment-505627280.

I'll see what happens when embedding the Deployment OpenAPI schema.

leonardoce:
https://github.com/kubernetes/kubernetes/issues/82292 is interesting too

leonardoce:
Unfortunately, the OpenAPI schema is huge, and we don't receive stray fields since we're deserializing them as PodTemplateSpec. It seems like the cure is worse than the disease.

sxd:
@leonardoce @phisco I think we should merge this otherwise we will be stopping the support of Kubernetes 1.29, BTW the reduction only with the schemaless goes from 562315 -> 21329 that's a really really big change

sxd (approved):

leonardoce:
Yeah, this was meant to have a discussion.
It's not a solution I really like, but I can't find another one.

mnencia:
We abandon this approach in favor of the `--server-side` apply. See #3697

PR #3656, opened 2024-01-02
feat(plugin): add options for database user an password in `pgbench` command

baurmatt:
We want pgbench to run against a different DB/user. DB is already configurable. This MR allows to configure the user and password pgbench will use.


armru:
@mnencia we should put this in our work queue

gbartolini:
@mnencia is there a way for us to use PGPASSFILE instead of PGPASSWORD when specifying a password?

Issue #317, opened 2022-06-27
Backup temp dir in scratch data fills the host disk


francoispqt:
When we do base backups barman writes its data into `/controller/backup` but this directory is on an EmptyDir volume and in our case the .part buffer file created by barman fills up the host disk and leads to Kubernetes terminating the database pod because it uses ephemeral storage it didn't request. The problem is our host disk isn't big enough to accommodate the .part file and it grows quite big because there is high latency between our object storage location and the database location (and we can't do anything about it).

As a solution we could specify a different path for backups and have a volume for that or maybe there is some barman configuration that could help?

sxd:
I think this issue was fixed in the newer version, please reopen if that's not the case

PR #3678, opened 2024-01-09
feat: add ephemeralVolumeSource option to cluster spec.

### LINKED ISSUE #3677: [Feature]: Allow configuration of EphemeralStorage.

As mentioned in #317 Barman `.part` files fill the `scratch-data` ephemeral storage. A comment there says that it has been fixed, however in the certain environments, for example GKE Autopilot this still happens. We are unable to set the `ephemeralVolumesSizeLimit` high enough to allow a base backup to happen.

I would like the cluster spec to optionally allow the setting of `ephemeralStorage` which would allow users to configure how the `scratch-data` volume is configured. This would allow the use of [generic ephemeral volumes](https://kubernetes.io/docs/concepts/storage/ephemeral-volumes/#generic-ephemeral-volumes). For example

```yaml
kind: Cluster
metadata:
  name: ephemeral-example
spec:
  instances: 3

  storage:
    size: 1Gi

  ephemeralStorage:
    volumeClaimTemplate:
      spec:
        accessModes: ["ReadWriteOnce"]
        storageClassName: "standard"
        resources:
          requests:
            storage: 1Gi
```

I have considered the option of having a separate mount for Barman, or configuring barman to output into the walStorage volume. I don't think these are good options, allowing the use of generic ephemeral volumes feels like the best approach.

### Additional context

_No response_

### Backport?

No

Yes

### Code of Conduct

jfyne:
Allow the user to optionally set the ephemeralStorage option in the cluster spec.

Closes #3677

Refs:
 - https://github.com/cloudnative-pg/cloudnative-pg/issues/3677
 - https://github.com/cloudnative-pg/cloudnative-pg/issues/317

### DIFF
--- a/.wordlist-en-custom.txt
+++ b/.wordlist-en-custom.txt
@@ -126,6 +126,7 @@ EnterpriseDB's
 EnvFrom
 EnvFromSource
 EnvVar
+EphemeralVolumeSource
 EphemeralVolumesSizeLimit
 EphemeralVolumesSizeLimitConfiguration
 ExternalCluster
@@ -649,6 +650,7 @@ endpointURL
 enterprisedb
 env
 envFrom
+ephemeralVolumeSource
 ephemeralVolumesSizeLimit
 executables
 expirations
--- a/api/v1/cluster_types.go
+++ b/api/v1/cluster_types.go
@@ -338,6 +338,10 @@ type ClusterSpec struct {
 	// +optional
 	WalStorage *StorageConfiguration `json:"walStorage,omitempty"`
 
+	// EphemeralVolumeSource allows the user to configure the source of ephemeral volumes.
+	// +optional
+	EphemeralVolumeSource *corev1.EphemeralVolumeSource `json:"ephemeralVolumeSource,omitempty"`
+
 	// The time in seconds that is allowed for a PostgreSQL instance to
 	// successfully start up (default 3600).
 	// The startup probe failure threshold is derived from this value using the formula:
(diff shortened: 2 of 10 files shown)


armru:
Hello, I've made several minor adjustments, the most notable one is to avoid increasing the CRD size.

whats your opinion on the patch @mnencia @leonardoce @gbartolini ?

gbartolini:
I like this patch. However, this conflicts with #2830.

My suggestion is to:

- clearly document this behaviour and make it the preferred way to configure ephemeral storage
- highlight the relationship with `EphemeralVolumesSizeLimit` (which relies on emptydir)
- if needed, add validation that excludes one or the other

If the ephemeral PVC is defined, this has higher precedence than the existing behavior. The `EphemeralVolumesSizeLimit` option is evaluated only if the PVC is not specified or is removed during the standard operations of the cluster.

jfyne:
@armru would you like me to make the requested changes?

armru:
Yes, thanks very much

jfyne:
Looks like there was no code change required, just a documentation update. I have done that now

armru:
I think we should also add a validation webhook to prevent both EphemeralVolumesSizeLimit and EphemeralVolumeSource from being defined simultaneously.

jfyne:
@armru I've had a go, I haven't written an operator before. Is what I have done enough, and also the correct error?

armru:
Hello @jfyne, could you add unit tests for the webhook validation?

jfyne:
Added the validation unit tests. Changes look good to me

armru (approved):

mnencia (approved):

mnencia:
Unit tests are failing

Issue #1691, opened 2023-03-12
PG Certificates - configurable values for threshold \ certificateDuration


edgarkz:
Hello,
Currently those are hardcoded values for certificate renewal threshold and also certificate duration.
Can you please make those value adjustable via helm chart?

expiringCheckThreshold:
https://github.com/cloudnative-pg/cloudnative-pg/blob/main/pkg/certs/certs.go#L49
[certs.go](https://github.com/cloudnative-pg/cloudnative-pg/blob/main/pkg/certs/certs.go)
    expiringCheckThreshold = 7 * 24 * time.Hour

certificateDuration: 
https://github.com/cloudnative-pg/cloudnative-pg/blob/main/pkg/certs/certs.go#L40
[certs.go](https://github.com/cloudnative-pg/cloudnative-pg/blob/main/pkg/certs/certs.go)
    certificateDuration = 90 * 24 * time.Hour

`        Issuer: OU=postgresql, CN=archive
        Validity
            Not Before: Jan 25 12:40:34 2023 GMT
            Not After : Apr 25 12:40:34 2023 GMT
        Subject: OU=postgresql, CN=archive
`

gbartolini:
Any volunteers in the community to try and tackle this and submit a pull request?

Otterpohl:
I will take a look at this over the next couple of weeks!

PR #3686, opened 2024-01-11
feat: allow configuration of certificate duration and expiring check time

zekiyeaydemir:
Allowed the configuration of certificationDuration and expiringCheckThreshold parameters.

Closes: #1691


leonardoce:
Hi @zekiyeaydemir. We want to merge this in the next release.
Can you please test what happens when you set an incorrect value for `CERTIFICATE_DURATION,` for example, a string instead of a number? We need to be sure we'll have an understandable error message.
Thanks!

sxd (approved):

litaocdl:
> Hi @zekiyeaydemir. We want to merge this in the next release. Can you please test what happens when you set an incorrect value for `CERTIFICATE_DURATION,` for example, a string instead of a number? We need to be sure we'll have an understandable error message. Thanks!

if the invalid value is set for cetification duration or threhold, default will be used  and info msg will reports in operator log

"Skipping invalid integer value parsing configuration","field":"CertificateDuration","value":"100s"

litaocdl (approved):

PR #3693, opened 2024-01-12
fix: tests tablespaces file from the RunAsUser spec from container

sxd:
Since the user may be changed by a SecurityContext, we detect if there's a user set and we compare against it, otherwise, we use the default UID for the `postgres` user to compare with.


jsilvela (approved):

litaocdl (approved):

litaocdl:
https://github.com/EnterpriseDB/cloudnative-pg/actions/runs/7621829224/job/20758921421

mnencia (approved):

PR #3702, opened 2024-01-16
[docs] fix azure connection string

srikiz:
Minor doc issue when creating a secret. The secret key should be `AZURE_CONNECTION_STRING` instead of `AZURE_STORAGE_CONNECTION_STRING`


jsilvela:
Thanks for your fix @srikiz !

Not sure it's quite right. There is actual operator code using `AZURE_STORAGE_CONNECTION_STRING`.
See extract below from `./pkg/management/barman/credentials/env.go`.
On the other hand, `AZURE_CONNECTION_STRING` only appears in test code, not operator code.
We need to figure this out, something is definitely wrong, we should consolidate on one...

``` go
	if configuration.BarmanCredentials.Azure.ConnectionString != nil {
		connString, err := extractValueFromSecret(
			ctx,
			c,
			configuration.BarmanCredentials.Azure.ConnectionString,
			namespace,
		)
		if err != nil {
			return nil, err
		}
		env = append(env, fmt.Sprintf("AZURE_STORAGE_CONNECTION_STRING=%s", connString))
	}
```

PR #3710, opened 2024-01-18
feat: synchronize user-defined physical replication slots

### LINKED ISSUE #2385: Manage failover of additional physical replication slots
Since version 1.18, CloudNativePG supports automated management of physical replication slots that the operator directly builds to improve both self-healing and high availability of a PostgreSQL cluster, which also includes surviving after a failover, a capability that Postgres alone cannot offer.

A stanza called `.spec.replicationSlots.highAvailability` in the `Cluster` resource spec controls this feature. For details, please refer to the “Additional Background” section below.

The current implementation, however, does not synchronize on the replicas any additional replication slots that have been independently created and are managed by users, preventing them from persisting after a failover.

The proposal is to add a new stanza called `synchronizeReplicas` in the `replicationSlots` stanza:

- `enabled`: when set to `true`, every replication slot that is on the primary is synchronized on each standby (by default set to `true`); the value is changed from `true` to `false`, the operator must remove from each standby any replication slot that was previously created by itself
- `excludePatterns`: list of regular expression patterns to match the names of replication slots to be excluded (by default empty)

These two parameters will not control the replication slots for high availability mentioned above. However, this feature will reuse the same logic for synchronizing slots on the replicas.

Users that want to take advantage of this feature must monitor such replication slots to ensure they don’t interfere with the operations on the primary.

Here follows an excerpt of an example to configure this feature in the configuration:

```
replicationSlots:
 synchronizeReplicas:
  enabled: true
  excludePatterns:
  - "^foo"
```

### Additional Background

Replication slots are a native PostgreSQL feature introduced in 9.4 that provides an automated way to ensure that the primary does not remove WAL segments until all the attached streaming replication clients have received them and that the primary does not remove rows which could cause a recovery conflict even when the standby is (temporarily) disconnected. A replication slot exists solely on the instance that created it, and PostgreSQL does not replicate it on the standby servers. As a result, after a failover or a switchover, the new primary does not contain the replication slot from the old primary. Streaming replication clients previously connected to the old primary cannot connect to the new one due to missing replication slots.

CloudNativePG 1.18 introduced the concept of cluster-managed replication slots for High Availability purposes only by automatically managing physical replication slots for each hot standby replica in the High Availability cluster, both in the primary and the standby instances. Specifically, the operator manages slots in the primary (for each of the HA standbys) and in each standby (for any other HA standby in the cluster, by looking at the content of the `pg_replication_slots` view in the primary and subsequently advancing the local slots at regular intervals using `pg_replication_slot_advance()`)

Users can create and drop replication slots through the `pg_create_physical_replication_slot` and `pg_drop_replication_slot` functions in PostgreSQL.

armru:
Extend the existing synchronization of physical replication slots (currently limited to HA replication slots) to include those defined by the user on the primary, enhancing self-healing, high availability, and third-party application integration.

The feature is enabled by default but can be disabled or customized through exclusion patterns by configuring the newly introduced stanza: `replicationSlots.synchronizeReplicas`.

Closes #2385


armru:
I don't think we intend to backport this feature, but I will let you decide @mnencia @gbartolini @mnencia

gbartolini:
I have a change proposal on the documentation side. Will work on that:

- Rename the "Replication slots for High Availability" section into a broader "Replication slots" section
- Leave the introductory part, then:
    - place here the "Replication slots for High Availability" section
    - add another section "User managed replication slots" (containing more or less the new content)
- Move at the end of this new section the "Capping the WAL size retained for replication slots" part
- Review the existing links

What do you think? @armru

armru:
I think we could address a more broad documentation refactor with a new PR

gbartolini:
I am happy to give it a try

gbartolini (approved):

internal/management/controller/slots/reconciler/replicationslot.go
```diff
@@ -40,13 +40,15 @@ func ReconcileReplicationSlots(
 		return reconcile.Result{}, nil
 	}
 
+	isPrimary := cluster.Status.CurrentPrimary == instanceName || cluster.Status.TargetPrimary == instanceName
+
 	// if the replication slots feature was deactivated, ensure any existing
 	// replication slots get cleaned up
 	if !cluster.Spec.ReplicationSlots.HighAvailability.GetEnabled() {
```
armru, line 47: this should be !cluster.Spec.ReplicationSlots.GetEnabled()

litaocdl: if both HA and user-defined slot are disabled:
  - user defined slot will be delete from standby in reconcling loop
  - HA slot will be deleted from both primary/standby in reconciling loop too.
 
if only HA slot is disabled
  - HA slot will be delete from primary/standby by reconcile loop
  - user defined slot will not be deleted from standby
  
if only user defined slot is been disabled
  - user defined slot will be delete from standby by runnable

litaocdl (approved):

litaocdl:
enhance the slot deleting and now have the following assumption

- If both HA and user-defined slot are disabled:

  - User-defined slots will be deleted by reconcling loop
  - HA slots will be deleted from both primary/standby in reconciling loop too.
  
- if only HA slots is disabled

  - HA slot will be deleted from primary/standby by reconcile loop
  - User-defined slots will NOT be deleted
  
- if only user defined slot is been disabled
  - user defined slot will be deleted by runnable

mnencia (approved):

mnencia:
E2E tests in EDB fork https://github.com/EnterpriseDB/cloudnative-pg/actions/runs/7799644093

mnencia:
/ok-to-merge all failures in E2E are unrelated

PR #3713, opened 2024-01-21
fix: escape reserved characters in pgpass fields (resolves #3712)

### LINKED ISSUE #3712: [Bug]: Import of external cluster fails if password contains reserved characters

### Contact Details

_No response_

### Version

1.22.0

1.28

OLM

### What happened?

When importing a database from an external cluster (bootstrap/initdb/import) and the source database-user has a password with a reserved character (e.g. colon; see https://www.postgresql.org/docs/current/libpq-pgpass.html) then the authentication to the source-cluster fails.

The reason for the bug seems to be a missing escaping of `:` and `\`, e.g. if the password contains a colon (`pass:word`) the pgpass file would look like this:
```
pgtest.com:5432:*:postgres:pass:word
```

### Cluster resource

_No response_

### Relevant log output

_No response_

### Code of Conduct

dlob:
Escape `:` and `\` characters according to pgpass file escaping rules.

Ref: https://www.postgresql.org/docs/current/libpq-pgpass.html

Fixes #3712

### DIFF
--- a/pkg/management/external/internal/pgpass/conninfo.go
+++ b/pkg/management/external/internal/pgpass/conninfo.go
@@ -16,8 +16,10 @@ limitations under the License.
 
 package pgpass
 
-import "fmt"
-import "strings"
+import (
+	"fmt"
+	"strings"
+)
 
 // ConnectionInfo contains the information identifying
 // a PostgreSQL server whom credentials need to be included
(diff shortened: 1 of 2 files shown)


leonardoce:
I started a run of E2e tests: https://github.com/EnterpriseDB/cloudnative-pg/actions/runs/7626649056

leonardoce (approved):

PR #3740, opened 2024-01-29
fix: skip deleting ha replicationslot when it is active

litaocdl:
This patch enfore when a HA replication slot is not the expected one, cnpg delete the 
replicationslot only when it changes to inacive status. 

Closes: #3739


mnencia (approved):

mnencia:
E2E running on EDB fork https://github.com/EnterpriseDB/cloudnative-pg/actions/runs/7699207998

PR #3743, opened 2024-01-29
fix: disable WAL archiving during import bootstrap

### LINKED ISSUE #3741: [Feature]: disable WAL archiving during import bootstrap

When using the `import` facility of the `initdb` bootstrap method in a job, the PostgreSQL instance has the following values:

- `archive_mode` is set to `on` (fixed)
- `archive_command` is left empty
- `wal_level` is set to `logical` (fixed)

The empty `archive_command` is fine with both standard `initdb` and `recovery` bootstraps, as it will have the effect to accumulate WAL files for the short duration of the instance run time.

However, in the `import` phase, the Postgres instance may run for quite some time and accumulate a lot of WAL files, increasing the risk of running out of disk in the WAL partition with subsequent failure of the cluster deployment.

In the specific `import` phase, we should set, just for the job duration:

- `archive_mode` is set to `off`
- `wal_level` is set to `minimal`

This will also improve the performance of the loading phase.

We should also consider setting `full_page_writes` to `off` in this phase.

From a HA & disaster point of view, once the job is completed and the database is imported, the pod starts with WAL archiving, replicas are created from there, and - if requested - the first backup is taken. We should inform users that, depending on the size of the database, the time taken for the first backup and for replica cloning might take longer.

Look at #3731

### Additional context

N/A

### Backport?

Yes

Yes

### Code of Conduct

gbartolini:
When using the import facility of the initdb bootstrap method in a job, optimize the PostgreSQL instance for the sole duration of the job, by setting the following GUCs:

- `archive_mode` to `off`
- `fsync` to `off`
- `full_page_writes` to `off`
- `max_wal_senders` to `0`
- `wal_level` to `minimal`

Then, restore the configuration and invoke `initdb --sync-only` before the end of the import job.

This avoids unnecessary and systematic piling of WAL files during any import job.

Closes #3741

### DIFF
--- a/docs/src/database_import.md
+++ b/docs/src/database_import.md
@@ -240,3 +240,30 @@ There are a few things you need to be aware of when using the `monolith` type:
 - After the clone procedure is done, `ANALYZE VERBOSE` is executed for every
   database.
 - `postImportApplicationSQL` field is not supported
+
+## Import optimizations
+
+During the logical import of a database, CloudNativePG optimizes the
+configuration of PostgreSQL in order to prioritize speed versus data
+durability, by forcing:
+
+- `archive_mode` to `off`
+- `fsync` to `off`
+- `full_page_writes` to `off`
+- `wal_level` to `minimal`
+
+Before completing the import job, CloudNativePG restores the expected
+configuration, then runs `initdb --sync-only` to ensure that data is
+permanently written on disk.
+
+!!! Important
+    WAL archiving, if requested, and WAL level will be honored after the
+    database import process has completed. Similarly, replicas will be cloned
+    after the bootstrap phase, when the actual cluster resource starts.
+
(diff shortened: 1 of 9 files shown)


mnencia:
E2E test on EDB fork https://github.com/EnterpriseDB/cloudnative-pg/actions/runs/7697496524

armru (approved):

gbartolini:
Rerun E2E tests after fixing wal senders GUC. See https://github.com/EnterpriseDB/cloudnative-pg/actions/runs/7699971017

gbartolini:
Run E2E tests on EDB account: https://github.com/EnterpriseDB/cloudnative-pg/actions/runs/7709440784

mnencia (approved):

Issue #3751, opened 2024-01-30
[Feature]: Privide a condition for the resuming cluster from hibernation 


litaocdl:
### Is there an existing issue already for this bug?

- [X] I have searched for an existing issue, and could not find anything. I believe this is a new bug.

### I have read the troubleshooting guide

- [X] I have read the troubleshooting guide and I think this is a new bug.

### I am running a supported version of CloudNativePG

- [X] I have read the troubleshooting guide and I think this is a new bug.

### Version

1.22.0

### What version of Kubernetes are you using?

1.28

### What is your Kubernetes environment?

Self-managed: kind (evaluation)

### How did you install the operator?

YAML manifest

### What happened?

Cluster in resuming is in healthy phase, and there is no condition to show it is resuming either. the hibernate condition `cnpg.io/hibernation` is removed at the first place when hibernate annotation is removed. this task defer the removing of hibernating condtion, introduce the `HibernationConditionReasonResumingPods` reason for hibernate condition. once all pods are resumed, then remove the hibernate condtion.  

### Cluster resource

```shell
N/A
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

mnencia:
After some discussions with @gabriele-wolfox and @litaocdl, we agreed that the operator should not try to be more specific about the "resuming" state of the Cluster because, from the operator's point of view, it is indistinguishable from its normal operations.

PR #3752, opened 2024-01-30
feat: add condition for a resuming cluster

litaocdl:
when a cluster is resuming from hibernated status, we remove the 
`cnpg.io/hibernation` condition at the first place, as the cluster at this 
time is in `healthy` phase, there is no way to distinguish a cluster resuming
from hibernated status and  a cluster doing the switchover.

this patch defer the removing of `cnpg.io/hibernation` condition. when 
cluster is been resumed, we change the hibernate condtion status to 
`false` and change the condtion reason to `ResumingPods`, when all 
pods are get resumed and ready, we then remove the hibernated condition. 

Closes: #3751

### DIFF
--- a/pkg/reconciler/hibernation/reconciler.go
+++ b/pkg/reconciler/hibernation/reconciler.go
@@ -46,9 +46,10 @@ func Reconcile(
 	switch hibernationCondition.Reason {
 	case HibernationConditionReasonDeletingPods:
 		return reconcileDeletePods(ctx, c, instances)
-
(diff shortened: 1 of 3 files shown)


gabriele-wolfox (approved):

YanniHu1996 (approved):

mnencia:
After some discussions with @gabriele-wolfox and @litaocdl, we agreed that the operator should not try to be more specific about the "resuming" state of the Cluster because, from the operator's point of view, it is indistinguishable from its normal operations.

Issue #3762, opened 2024-02-01
[Feature]: Split out cnpg-api to be used from CNPG-I plugins


mnencia:
### Is there an existing issue already for this feature request/idea?

- [X] I have searched for an existing issue, and could not find anything. I believe this is a new feature request to be evaluated.

### What problem is this feature going to solve? Why should it be added?

With the plugin interface, every plugin must depend on the cloudnative-pg API, but depending to the main cloudnative-pg repository would bring all the dependencies of the project.

We also want to include some helper code in cnpg-i repository, but that's impossible now because it would depend on the main cloudnative-pg API, and with the present state it would create a circular dependency.

### Describe the solution you'd like

We could create a `cloudnative-pg/cnpg-api` repository and create automation to mirror in that repository the `cloudnative-pg/cloudnative-pg/api` directory.

### Describe alternatives you've considered

We have also considered moving the API to a separate package, removing it from the `cloudnative-pg/cloudnative-pg` repository. However, this solution would significantly increase the complexity of introducing changes in CloudnativePG.

### Backport?

N/A

### Are you willing to actively contribute to this feature?

Yes

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #3814, opened 2024-02-09
feat: make CNPG API exportable

mnencia:
Ref #3762


mnencia:
Abandoned. We used a different method.

sxd:
Do not delete the original branch until @mnencia is agree

PR #3837, opened 2024-02-19
feat: make PDB configurable

### LINKED ISSUE #2570: [Feature Request] Allow configurable PDBs
**Use Case:** We want to allow eviction of the primary DB instance in order to perform upgrades on the underlying infrastructure automatically, for example updating a kubernetes node. While that will cause degraded performance during the failover, that is aligned with our SLOs for our uses of a majority of the clusters.

**Problem:** The current operator sets up a PDB that never allows the primary to be evicted. No configuration parameters are exposed to override this default behavior.

Edit:

I have expanded on additional use cases and reasons why this is critically important in a follow-up [comment](https://github.com/cloudnative-pg/cloudnative-pg/issues/2570#issuecomment-1858650998).

achanda:
Adds a new field to the CRD allowing the deactivation of the default
Pod Disruption Budgets, thereby allowing primary instances to be
evicted during maintenance operations. While this could entail
temporary service degradation, there are use cases where this is
an acceptable tradeoff.

Closes #2570


achanda:
Putting this up for an initial review to make sure this is the right approach. cc @gbartolini 

Things to figure out:
1. Will this affect existing clusters that have this flag changed to false?
2. Tests

gbartolini:
I think in the longer term we need to understand how this feature will replace for good the `nodeMaintenanceWindow` option. Can you please rename the option `enablePDB`?

I will leave the other maintainers to check whether yours is the right approach on the code. Thanks.

leonardoce (approved):

leonardoce:
I agree with the implementation, but I think we should clarify better in the documentation the intersection between the maintenance window stuff and `enablePDB`.

armru (approved):

config/crd/bases/postgresql.cnpg.io_clusters.yaml
```diff
@@ -1877,6 +1877,13 @@ spec:
               description:
                 description: Description of this PostgreSQL cluster
                 type: string
+              enablePDB:
```
pgodowski, line 1880: Please make sure that the new CRD field is also added into the older release (incl 1.18), as per the other discussion about 1.22 and 1.18 CRD compatibility. Thanks!
@mnencia

pgodowski (changes requested):

docs/src/kubernetes_upgrade.md
```diff
@@ -1,54 +1,105 @@
...
+    If your database files reside on shared storage accessible over the
+    network, the default self-healing behavior of the operator can efficiently
+    handle scenarios where volumes are reused by pods on different nodes after a
+    drain operation. In such cases, you can skip the remaining sections of this
+    document.
+
+## Pod Disruption Budgets
+
+By default, CloudNativePG ensures that the primary node in a PostgreSQL cluster
+prohibits its own draining, safeguarding uninterrupted operations.
```
jsilvela, line 53: IMO this is correct but possibly confusing. The user does have the ability to select a node for draining.
Suggestion
```suggestion
By default, CloudNativePG ensures that if a node to be drained contains the
cluster's primary instance, a switchover happens ahead of the drain. Once the
instance in the node is downgraded to replica, the draining can proceed.
For single-instance clusters, CloudNativePG will prevent draining the node where
the instance is housed, safeguarding operations.
```

docs/src/kubernetes_upgrade.md
```diff
@@ -1,54 +1,105 @@
...
+Additionally, in the presence of replicas, it guarantees that only one replica
+at a time is gracefully shut down during a drain operation.
```
jsilvela, line 55: IMO this is correct but possibly confusing. The user does have the ability to select a node for draining.
Suggestion
```suggestion
By default, CloudNativePG safeguards Postgres cluster operations. If a node is
to be drained and contains a cluster's primary instance, a switchover happens
ahead of the drain. Once the instance in the node is downgraded to replica, the
draining can resume.
For single-instance clusters, a switchover is not possible, so CloudNativePG
will prevent draining the node where the instance is housed.
Additionally, in multi-instance clusters, CloudNativePG guarantees that only
one replica at a time is gracefully shut down during a drain operation.
```

gbartolini:
I have removed the backport of this functionality to 1.22, but my suggestion is to backport the feature to 1.22 once 1.23.0 is released.

pgodowski:
> I have removed the backport of this functionality to 1.22, but my suggestion is to backport the feature to 1.22 once 1.23.0 is released.

I am not sure what is the 1.23 timeline, but we're definitely interested in this feature in 1.22.

config/crd/bases/postgresql.cnpg.io_clusters.yaml
```diff
@@ -1877,6 +1877,13 @@ spec:
...
+                default: true
```
NiharDudam, line 1881: Please move the default values to webhook 

cc: @mnencia

jsilvela:
I'm not a maintainer...

jsilvela:
To be honest I don't know how to do it @achanda ... I may ask a maintainer to commit for me...

jsilvela:
@achanda all figured. I checked out your branch with `gh` as opposed to `git` and had no issue. 🤷 
Going to check a bit and hopefully we merge your patch today!

docs/src/kubernetes_upgrade.md
```diff
@@ -1,54 +1,109 @@
...
-degraded cluster and wait for the upgraded node to be up again.
+## Temporary PostgreSQL Cluster Degradation
 
-In particular, if your PostgreSQL cluster relies on **node-local storage**
-\- that is *storage which is local to the Kubernetes worker node where
-the PostgreSQL database is running*.
-Node-local storage (or simply *local storage*) is used to enhance performance.
+While the standard approach ensures service reliability and leverages
+Kubernetes' self-healing capabilities, there are scenarios where operating with
+a temporarily degraded cluster may be advisable. This is particularly relevant
```
jsilvela, line 38: "... may be advisable" ... doesn't sound right.
Perhaps I'm misunderstanding. @gbartolini perhaps you meant "acceptable" not "advisable"?

jsilvela (approved):

pgodowski:
Can I please ask to backport this to 1.22 as well as per earlier comment https://github.com/cloudnative-pg/cloudnative-pg/pull/3837#issuecomment-1986048216 ?

@jsilvela

PR #3840, opened 2024-02-19
feat: multi-arch in-place update of the operator

NiccoloFei:
From the catalog of all the available architectures contained in the running operator image, retrieve the matching architecture's binary during in-place updates, enabling the operator to inject the correct binary in any Pod having a supported architecture.

Closes #3868 
Closes #3717

PR #3973, opened 2024-03-01
test(e2e): add coverage for multi-arch in-place update

### LINKED ISSUE #3972: [Feature]: Enhance architecture E2E tests to consolidate how available architectures are retrieved

This patch aims at solving the following TODO of the architecture's E2E tests:
```
// TODO: instead of fetching the current architectures using the
// PLATFORMS env variable, we should have a manager command which
// returns all the architectures available in the current image.
```

The problem with relying on the PLATFORMS env variable is that it's only realistically being set by our CI/CD, so it's useless when running E2Es locally or in a different environment.

Discover the available architectures by executing a `manager` command that reports back each manager's architecture of the current image.

I've placed the new manager's command under `debug show-architectures` (creating a new debug command).
An alternative would be to use the already existing `show` command, which currently contains only `show walarchivequeue` (e.g. `show architectures`)

### Additional context

_No response_

### Backport?

Yes

Yes

### Code of Conduct

NiccoloFei:
This patch adds end-to-end (E2E) coverage for the multi-architecture in-place update feature. A debug command was introduced to fetch available architectures in the operator container to facilitate this E2E testing in any environment.

Related To #3840 
Closes #3972

### DIFF
--- a/internal/cmd/manager/debug/architectures/cmd.go
+++ b/internal/cmd/manager/debug/architectures/cmd.go
@@ -48,8 +48,9 @@ func run() error {
 	if err := utils.DetectAvailableArchitectures(); err != nil {
 		return err
 	}
-	architectures := make([]string, 0, len(utils.GetAvailableArchitectures()))
-	for _, arch := range utils.GetAvailableArchitectures() {
+	availableArchitectures := utils.GetAvailableArchitectures()
+	architectures := make([]string, 0, len(availableArchitectures))
+	for _, arch := range availableArchitectures {
 		architectures = append(architectures, arch.GoArch)
 	}
 	val, err := json.MarshalIndent(architectures, "", "    ")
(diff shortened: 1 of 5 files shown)


internal/cmd/manager/debug/architectures/cmd.go
```diff
@@ -0,0 +1,64 @@
...
+func run() error {
+	binaries, err := filepath.Glob("bin/manager_*")
+	if err != nil {
+		return err
+	}
+	architectures := make([]string, 0, len(binaries))
+	for _, b := range binaries {
+		goArch := strings.Split(filepath.Base(b), "manager_")[1]
+		architectures = append(architectures, goArch)
+	}
```
YanniHu1996, line 57:
```suggestion
	if err := utils.DetectAvailableArchitectures(); err != nil {
		return err
	}
	architectures := make([]string, 0, len(utils.GetAvailableArchitectures()))
	for _, arch := range utils.GetAvailableArchitectures() {
		architectures = append(architectures, arch.GoArch)
	}
```

To centralize the code logic, using DetectAvailableArchitectures.

YanniHu1996 (approved):

armru (approved):

litaocdl (approved):

PR #4020, opened 2024-03-06
chore: calculate `archive_mode` and `wal_level` values

### LINKED ISSUE #3746: [Feature]: Enable configuration of `wal_level`

Following #3731 we have agreed to enable users to configure `wal_level`, provided some validation.

As explained in #3731 we should implement a function that, based on the configuration of the cluster, determines the minimum required value for `wal_level` and also ensures that, if `wal_level` is set to `minimal`, `max_wal_senders` should be set to 0.

See #3731 

### Additional context

N/A

### Backport?

Yes

Yes

### Code of Conduct

armru:
Closes #3746


armru:
This patch will most likely cause a restart of the instances

api/v1/cluster_webhook.go
```diff
@@ -1079,6 +1079,24 @@ func (r *Cluster) validateConfiguration() field.ErrorList {
 		}
 	}
 
+	const walLevelParameter = "wal_level"
+	walLevel := sanitizedParameters[walLevelParameter]
+	if r.Spec.Instances > 1 && walLevel != "logical" {
```
gbartolini, line 1084: Actually, `replica` is allowed too.

pkg/postgres/configuration.go
```diff
@@ -428,6 +428,7 @@ var (
 			{130000, MajorVersionRangeUnlimited}: {
 				"wal_keep_size":      "512MB",
 				"shared_memory_type": "mmap",
+				"wal_level":          "logical",
```
leonardoce, line 431: I think we need to set `wal_level` to `logical` by default in every PostgreSQL version. Logical replication is always supported for PostgreSQL versions that we use in CNPG.

api/v1/cluster_webhook.go
```diff
@@ -1079,6 +1079,29 @@ func (r *Cluster) validateConfiguration() field.ErrorList {
 		}
 	}
 
+	walLevel := postgres.WalLevelValue(sanitizedParameters[postgres.WalLevelParameter])
+	hasWalLevelRequirement := r.Spec.Instances > 1 || r.Spec.Backup.IsBarmanBackupConfigured() || r.IsReplica()
```
gbartolini, line 1083: In my opinion, instead of checking `r.Spec.Backup.IsBarmanBackupConfigured()` we need to simply check if `archive_mode` is `on`.

gbartolini (approved):

Issue #3774, opened 2024-02-02
[Feature]: Enable configuration of `archive_mode`


gbartolini:
### Is there an existing issue already for this feature request/idea?

- [X] I have searched for an existing issue, and could not find anything. I believe this is a new feature request to be evaluated.

### What problem is this feature going to solve? Why should it be added?

Following #3731, now that Postgres adoption increases, it is time to enable users to configure `archive_mode`, provided some validation.

### Describe the solution you'd like

We should enable users to set `archive_mode` to either `off` or `on` (default and recommended value).

We forbid setting `always` in a Cluster definition - at least for now. `always` indeed has a special meaning for CloudNativePG: we use it as a synonym of `on` in a replica cluster, by setting it `archive_mode = always` in the designated primary of a replica cluster (see [pkg/postgres/configuration.go](https://github.com/cloudnative-pg/cloudnative-pg/blob/main/pkg/postgres/configuration.go#L582)). The instance manager has a logic already that prevents a cascading standby from archiving a WAL file when `archive_mode` is set to `always`.

We should document the dangerous effects of setting `archive_mode` to `off`.

### Describe alternatives you've considered

See #3731

### Additional context

See #3731

### Backport?

Yes

### Are you willing to actively contribute to this feature?

Yes

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

gbartolini:
With Leonardo we came to the conclusion that we cannot block users from manually setting `archive_mode` to `always`. If we do so, we must then introduce a breaking change in the API of a replica cluster, requiring users to manually set `archive_mode` to `on` after promotion.

As a result we decided to abandon PR #4053 and proposed a different approach: use the `cnpg.io/skipWalArchiving` annotation. This is not intrusive at all. Please check PR #4055.

PR #4053, opened 2024-03-09
feat: enable configuration of `archive_mode`

leonardoce:
Allow users to disable `archive_mode`.

Closes: #3774

### DIFF
--- a/api/v1/cluster_webhook.go
+++ b/api/v1/cluster_webhook.go
@@ -2145,7 +2145,7 @@ func (r *Cluster) validateWALLevelChange(old *Cluster) field.ErrorList {
 		errs = append(errs, field.Invalid(
 			field.NewPath("spec", "postgresql", "parameters", "wal_level"),
 			"minimal",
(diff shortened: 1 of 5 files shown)


gbartolini (approved):

sxd (approved):

gbartolini:
It looks like we need to fix the `always` check as E2E tests fail with:

```
#1: admission webhook "vcluster.cnpg.io" denied the request: Cluster.postgresql.cnpg.io "cluster-replica-tls" is invalid: spec.postgresql.parameters.archive_mode: Invalid value: "always": archive_mode=always is directly managed by the operator
```

leonardoce:
E2e tests are now green

gbartolini (approved):

gbartolini:
Trying this patch, it is clear that it requires a change of API in the case of promotion of a replica cluster.

gbartolini:
We should either go with a new option or - probably better in this phase - an annotation (similarly to what we already have for the skip check of the WAL archive)

gbartolini:
Superseded by #4055

Issue #3459, opened 2023-12-05
[Bug]: `maxParallel` doesn't work during recovery


anthonator:
### Is there an existing issue already for this bug?

- [X] I have searched for an existing issue, and could not find anything. I believe this is a new bug.

### I have read the troubleshooting guide

- [X] I have read the troubleshooting guide and I think this is a new bug.

### I am running a supported version of CloudNativePG

- [X] I have read the troubleshooting guide and I think this is a new bug.

### Version

1.21.1

### What version of Kubernetes are you using?

1.26

### What is your Kubernetes environment?

Cloud: Other

### How did you install the operator?

Helm

### What happened?

With the provided config we'd expect barman to load WAL files in parallel during a full recovery.

```yaml
bootstrap:
    recovery:
        source: core-db-15

externalClusters:
    - barmanObjectStore:
            destinationPath: {{ .Values.dbBackupPath }}

            s3Credentials:
                accessKeyId:
                    key: ***

                    name: ***
                secretAccessKey:
                    key: ***

                    name: ***
            wal:
                maxParallel: 15

        name: core-db-15
```

Despite this configuration only 1 WAL is being restored at a time.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

Knorreman:
I am seeing the same behavior and our recovery times are very slow

Grubla:
Yeah, this is affecting us as well!

pzeinlinger:
@mnencia Sorry to comment on a closed issue, but I'm observing behavior that suggests the parameter is ignored during a restore, which seems related to the discussion here. Should I open a new issue?

mnencia:
@pzeinlinger Thank you for reporting this issue. I'd like to give you some clarifications regarding the functional and our current direction: the parallel WAL retrieval during recovery functions correctly when using CNPG-I backup plugins, such as `plugin-barman-cloud`, and we are deprecating the in-tree backup implementation in favor of CNPG-I backup plugins. Documentation updates related to this transition are being addressed in #6876 and #7065.

PR #4088, opened 2024-03-12
feat: make use of the maxParallel parameter during wal-restore

Knorreman:
Addresses https://github.com/cloudnative-pg/cloudnative-pg/issues/3459


Weihrauch:
Hello, using EDB enterprise with 40 TB database we are very waiting for parallel recovery really implemented. Is the reviewing of this merge have been planned ?

PR #4132, opened 2024-03-18
fix: skip updateTargetPrimary if no instance status is available

### LINKED ISSUE #4049: [Bug]: Recreate the primary pod if no pod is ready
The PR #3930 solves the PVC label issue. However, there is still a way the cluster could fail when you delete all the instances in sequence. If the timing is just right, you can end up in a situation where only one not-ready pod remains (a former primary that needs to run pg_rewind). 

The cluster remains stuck in a phase like `Instance Status Extraction Error: HTTP communication issue`, and the operator logs the following message over and over:

```
{"level":"info","ts":"2024-03-08T18:00:54Z","msg":"Cannot update target primary: operation cannot be fulfilled. An immediate retry will be scheduled","controller":"cluster","controllerGroup":"postgresql.cnpg.io","controllerKind":"Cluster","Cluster":{"name":"cluster-example","namespace":"default"},"namespace":"default","name":"cluster-example","reconcileID":"4e0c2cca-d287-45ff-87c5-edd28b494b53","uuid":"cde5207a-dd75-11ee-9130-62304a1b1e3f","error":"unable to evaluate failover logic, unable to fetch the instances status"}
```

However, the situation is easy to recover: if you delete the not-ready pod, the operator recreates the primary first, and the cluster becomes healthy again.

_Originally posted by @mnencia in https://github.com/cloudnative-pg/cloudnative-pg/issues/3930#issuecomment-1986186678_

armru:
The current code assumes that `updateTargetPrimaryFromPodsPrimaryCluster` is always executed when an HTTP status is present, but this is not the case. This patch ensures that the function concludes with a `noop` instead of an error allowing the reconciliation loop to continue.
Closes #4049

### DIFF
--- a/controllers/cluster_controller.go
+++ b/controllers/cluster_controller.go
@@ -416,7 +416,7 @@ func (r *ClusterReconciler) handleSwitchover(
 
 	// Update the target primary name from the Pods status.
 	// This means issuing a failover or switchover when needed.
-	selectedPrimary, err := r.updateTargetPrimaryFromPods(ctx, cluster, instancesStatus, resources)
+	selectedPrimary, err := r.reconcileTargetPrimaryFromPods(ctx, cluster, instancesStatus, resources)
 	if err != nil {
 		if errors.Is(err, ErrWaitingOnFailOverDelay) {
 			contextLogger.Info("Waiting for the failover delay to expire")
--- a/controllers/cluster_controller_test.go
+++ b/controllers/cluster_controller_test.go
@@ -119,7 +119,7 @@ var _ = Describe("Updating target primary", func() {
 		})
 
 		By("updating target primary pods for the cluster", func() {
-			selectedPrimary, err := env.clusterReconciler.updateTargetPrimaryFromPods(
+			selectedPrimary, err := env.clusterReconciler.reconcileTargetPrimaryFromPods(
 				ctx,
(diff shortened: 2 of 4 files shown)


controllers/cluster_controller.go
```diff
@@ -286,7 +286,7 @@ func (r *ClusterReconciler) reconcile(ctx context.Context, cluster *apiv1.Cluste
 		return ctrl.Result{}, err
 	}
 
-	if instancesStatus.AllReadyInstancesStatusUnreachable() {
+	if instancesStatus.AllReadyAndActiveInstancesAreUnreachable() {
```
armru, line 289: IMO, this is one of the root causes of code being extremely hard to understand and could lead to dangerous bugs.

In this check, we are only evaluating Ready and Active instances, but after that we operate on the original list containing the inactive or not ready instances.

We should start thinking of refactoring this in a filter that returns a new object and adapt the following code to it

armru: @mnencia @gbartolini @leonardoce

armru:
this PR should be reviewed and tested accurately

mnencia (approved):

mnencia (approved):

leonardoce (approved):

PR #4141, opened 2024-03-19
fix: ensure pods with no ownership are deleted during cluster restore

armru:
At present, during the cluster restore from orphan PVCs , it is not possible for the operator to bring back a cluster to a functional state if any orphan pods are present.

This is because the existing pods will linger indefenetly and won't be be recreated with the correct data. To address this issue, this patch ensures that no orphan pods are present during the cluster restore from PVCs by deleting them.

### DIFF
--- a/controllers/cluster_restore.go
+++ b/controllers/cluster_restore.go
@@ -22,6 +22,7 @@ import (
 
 	corev1 "k8s.io/api/core/v1"
 	apierrs "k8s.io/apimachinery/pkg/api/errors"
+	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	apiv1 "github.com/cloudnative-pg/cloudnative-pg/api/v1"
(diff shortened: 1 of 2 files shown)


controllers/cluster_restore.go
```diff
@@ -148,6 +168,38 @@ func getOrphanPVCs(
 	return orphanPVCs, nil
 }
 
+func ensureOrphanPodsAreDeleted(ctx context.Context, c client.Client, cluster *apiv1.Cluster) error {
+	var podList corev1.PodList
+	if err := c.List(
+		ctx,
+		&podList,
+		client.InNamespace(cluster.Namespace),
+		client.MatchingLabels{utils.ClusterLabelName: cluster.Name},
```
YanniHu1996, line 177: should match label `cnpg.io/podRole` to `instance` to filter out  pooler pod

armru: Hey Yanni, thanks, this poses the question: how would you handle the orphan pooler pod if it exists?

leonardoce:
E2e: https://github.com/EnterpriseDB/cloudnative-pg/actions/runs/8423797788

leonardoce:
E2e tests are green

leonardoce (approved):

PR #4152, opened 2024-03-22
refactor: improve metadata fencing code organization

armru:
Currently, the metadata fencing code is challenging for inexperienced developers to use, given that it is fragmented across the codebase without a clear pattern.
    
This patch aggregates the logic and adds a fluent builder to execute the fencing metadata logic.
    
It also improves handling certain operations like fencing or unfecing an already present or deleted instance. In those cases, the executor does not return an error anymore.


pkg/utils/fencing.go
```diff
@@ -80,58 +79,142 @@ func SetFencedInstances(object *metav1.ObjectMeta, data *stringset.Data) error {
...
 
-	return SetFencedInstances(object, fencedInstances)
+	return true, setFencedInstances(object, fencedInstances)
 }
 
-// RemoveFencedInstance removes the given server name from the FencedInstanceAnnotation annotation
+// removeFencedInstance removes the given server name from the FencedInstanceAnnotation annotation
 // returns an error if the instance was already unfenced
-func RemoveFencedInstance(serverName string, object *metav1.ObjectMeta) error {
-	if serverName == FenceAllServers {
```
armru, line 118: This behavior is debatable, but I'm preserving it in this refactor.

leonardoce (approved):

mnencia (approved):

PR #4197, opened 2024-04-02
test: remove `envtest` from the controller package

armru:
This patch removes `envtest` from the controller packages. This is done for several reasons:
- is only used in a few of the controller package tests
- acts as a singleton, reducing the predictability of the unit tests
- makes it harder to use the debugger
- does not currently add any real benefit that isn't already covered by webhook unit tests and E2E tests

Note for the reviewers:
In my opinion, `envtest` could be reintroduced  as an integration testing layer with a different PR

### DIFF
--- a/controllers/backup_controller_test.go
+++ b/controllers/backup_controller_test.go
@@ -35,6 +35,11 @@ import (
 )
 
 var _ = Describe("backup_controller barmanObjectStore unit tests", func() {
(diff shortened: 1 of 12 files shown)


leonardoce:
This PR is fine but we need to merge #4212 before merging it.

leonardoce (approved):

sxd (approved):

PR #4212, opened 2024-04-05
fix: align enableSuperuserAccess default

leonardoce:
Within the `GetEnableSuperuserAccess` function, the codebase defaulted to `true` while the `cluster.spec.enableSuperuserAccess` in the CRD was defaulted to `false`.

This discrepancy does not lead to any runtime issues because when the operator retrieves Custom Resources from the API server, the structure has already been populated with the correct values.

However, the problem can still surface during unit tests when utilizing a mock Kubernetes client.

Closes: #4211

### DIFF
--- a/api/v1/cluster_types.go
+++ b/api/v1/cluster_types.go
@@ -3292,7 +3292,7 @@ func (cluster *Cluster) GetEnableSuperuserAccess() bool {
 		return *cluster.Spec.EnableSuperuserAccess
 	}
 
-	return true
+	return false
 }
 
 // LogTimestampsWithMessage prints useful information about timestamps in stdout
--- a/api/v1/cluster_types_test.go
+++ b/api/v1/cluster_types_test.go
@@ -41,7 +41,7 @@ var _ = Describe("PostgreSQL cluster type", func() {
 
 	It("correctly get if the superuser is enabled", func() {
 		postgresql.Spec.EnableSuperuserAccess = nil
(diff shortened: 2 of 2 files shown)


leonardoce:
This shouldn't be backported to CNPG releases where `cluster.spec.enableSuperuserAccess` was defaulting to `false`.

sxd (approved):

mnencia (approved):

mnencia:
@sxd, why did you remove the backport request? Both the 1.21 and 1.22 have the same issue.

mnencia:
> This shouldn't be backported to CNPG releases where `cluster.spec.enableSuperuserAccess` was defaulting to `false`.

do you mean `true`? The trunk and all the supported versions default to `false`! The last version defaulting to `true` was 1.18.

leonardoce:
> do you mean `true`? The trunk and all the supported versions default to `false`! The last version defaulting to `true` was 1.18.

This should be back ported.

leonardoce:
> do you mean `true`? The trunk and all the supported versions default to `false`! The last version defaulting to `true` was 1.18.

Do you mean the Go code or the CRD? The first one was defaulting to true in trunk too

mnencia:
@leonardoce, @sxd, I mean that all the supported versions (and the trunk) are intended to default to false. The CRD is correct, but the function returns the wrong value as a default for all of them. That's why I'm adding back the backport request.

Issue #4041, opened 2024-03-08
[Feature]: Enable configuration of `wal_log_hints`


gbartolini:
### Is there an existing issue already for this feature request/idea?

- [X] I have searched for an existing issue, and could not find anything. I believe this is a new feature request to be evaluated.

### What problem is this feature going to solve? Why should it be added?

Currently, the [`wal_log_hints`](https://www.postgresql.org/docs/current/runtime-config-wal.html#GUC-WAL-LOG-HINTS) GUC is fixed to `on`.

### Describe the solution you'd like

Given that currently `wal_log_hints` is required by `pg_rewind`, we could allow users to disable `wal_log_hints` if the number of `instance` is 1.

### Describe alternatives you've considered

N/A

### Additional context

See #4026

### Backport?

Yes

### Are you willing to actively contribute to this feature?

Yes

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #4218, opened 2024-04-06
feat: enable configuration of `wal_log_hints`

leonardoce:
Enable setting `wal_log_hints` to `off` for clusters with a single instance.

Previously, `wal_log_hints` was fixed to `on` for every cluster due to `pg_rewind,` which is not required with single-instance clusters.

Closes: #4041

### DIFF
--- a/api/v1/cluster_webhook.go
+++ b/api/v1/cluster_webhook.go
@@ -1201,7 +1201,7 @@ func (r *Cluster) validateConfiguration() field.ErrorList {
 			field.Invalid(
 				field.NewPath("spec", "postgresql", "parameters", postgres.ParameterWalLogHints),
 				r.Spec.PostgresConfiguration.Parameters[postgres.ParameterWalLogHints],
-				"`wal_log_hints` should be set to `on` when having replica instances"))
+				"`wal_log_hints` must be set to `on` when `instances` > 1"))
 	}
 
 	// verify the postgres setting min_wal_size < max_wal_size < volume size
--- a/api/v1/cluster_webhook.go
+++ b/api/v1/cluster_webhook.go
@@ -1195,6 +1195,15 @@ func (r *Cluster) validateConfiguration() field.ErrorList {
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


leonardoce:
This is still a draft. When I finish testing it manually, I'll open it for review.

leonardoce:
I tested it manually. It looks fine. Running an E2e test suite too.

leonardoce:
E2e tests are fine

gbartolini (approved):

sxd (approved):

api/v1/cluster_webhook.go
```diff
@@ -1195,6 +1195,15 @@ func (r *Cluster) validateConfiguration() field.ErrorList {
 		}
 	}
 
+	if r.Spec.Instances > 1 && r.Spec.PostgresConfiguration.Parameters[postgres.ParameterWalLogHints] == "off" {
+		result = append(
```
mnencia, line 1199: I think this needs to be revised.

From the Postgres docs about the config file:
> Boolean: Values can be written as on, off, true, false, yes, no, 1, 0 (all case-insensitive) or any unambiguous prefix of one of these.

That means that a user could set it to "0", "no", or even "OFF" and go ahead with a broken configuration.

If we don't want to parse the value, we can use `!= "on"` in the condition, forcing the user to use `on` as the only valid value for true.

Also, is there any possibility that the parameter is not present and we compare an empty value?

leonardoce: Same thing with archive_mode, where we choose this behavior.

If we like something different, let's address this but not in this PR.

mnencia: So, we should also change the `archive_mode` check to be more precise. Doing that in another PR is fine for me. Let's merge this feature and create another issue to improve the checks involving boolean parameters.

leonardoce: Will you create an issue for that @mnencia?

leonardoce: When you do it, assign it to me.

mnencia (approved):

PR #4268, opened 2024-04-11
feat: Improve PodMonitor spec

CiraciNicolo:
# Summary

This PR introduces new PodMonitor specs to customise even more the monitor capabilities, I propose a new spec that is intended to allow more and more integration between Prometheus CRD and CNPG monitoring support. 

This in done **deprecating** the old Monitoring field - relabing configs - in favour of a new `PodMonitoringConfiguration` dictionary that implement the same fields plus more. This allow us to reduce the complexity of reconciling the PodMonitor with a unified interface+struct and to allow more and more customisation in the upcoming future.

# Bug Fixes

This pr fix a [bug](https://github.com/cloudnative-pg/cloudnative-pg/blob/6e423eb50b71ced1fe78aaecdb1cf0411aefdac5/internal/cnpi/plugin/client/lifecycle.go#L61) introduced by the Plugin API which caused the reconcile failure for object not recognized by the operator, such as PodMonitor. The lifecycle hook now stop if the resource is not known but returns the original object so the reconcile can continue.

# Description

This pr introduce new specs for PodMonitor functionality with the plan to implement 1:1 functionality as PodMonitor spec by also guard-railing the creation. This has been done by completely rewrite the PodMonitor logic implementing a new kind of Controller that will implement all the current and future logic. We choose to **not** use the prometheus CRD spec to prevent accidental issue with some fields calculated by the Cluster spec - such as PodMonitor endpoint port name. 

This pr also propose the deprecation of the original PodMonitor field in favour of a more standardized sub-spec which can be improved more easily in the future; finally e2e test for PodMonitor are implemented to intercept eventual regression more easily.

Some other quirks and fixes have been implemented:
- Add commodities for testing
- Implement Prometheus CRD installation for e2e testing

# Testing Done
- Implemented monitoring_test.go
- Implemented e2e tests


phisco:
I think the approach so far has been to avoid providing too many knobs for podmonitors, you can always define your own separately if you need more control, you just need to match the labels on the pods, which you can control from the Cluster.

CiraciNicolo:
It would be nice that managing the PodMonitor would be done by the operator; if I recall correctly a discussion about 1:1 functionality is present somewhere in this repo. Can you describe why you would prefer letting the user manage the PodMonitor?

phisco:
Because we have to draw the line somewhere 😅 and this operator's job is not to handle Prometheus operator's resources, we do that for simple scenarios because it might be convenient for less experienced users, but we should not try to offer full customization. Otherwise we risk having to keep up to date with all the features they might introduce there, just to let users with more complex requirements deploy a single resource instead of two.

LucaCominoli21:
> Because we have to draw the line somewhere 😅 and this operator's job is not to handle Prometheus operator's resources, we do that for simple scenarios because it might be convenient for less experienced users, but we should not try to offer full customization. Otherwise we risk having to keep up to date with all the features they might introduce there, just to let users with more complex requirements deploy a single resource instead of two.

I agree that a line should be drawn and the pod monitor should not be parameterized too much, but I think that in a simple scenario and for a novice user, it is useful to be able to change the frequency of data scraping without having to re-implement it from scratch. I also do not see any big changes that require more effort to keep them up to date.
In addition, tests on the pod monitor have also been added with this PR, which identified a bug on the operator caused by the API Plugin implementation.

phisco:
Either way, could you please split the fix to its own PR?

CiraciNicolo:
Hi. Will split the PR ASAP, I will also refer this PR in the other two so we can continue the discussion. 

Thanks

gbartolini:
Hello @CiraciNicolo, I regret to inform you that we, the project maintainers, have decided to deprecate the native integration with PodMonitor resources starting from version 1.28. We will delegate all functionality to the official upstream PodMonitor implementation. For more details, please refer to issue #8075. Thank you for your contribution.

Issue #4266, opened 2024-04-11
[Bug]: kubectl destroy plugin should report error when cluster and instance not found


litaocdl:
### Is there an existing issue already for this bug?

- [X] I have searched for an existing issue, and could not find anything. I believe this is a new bug.

### I have read the troubleshooting guide

- [X] I have read the troubleshooting guide and I think this is a new bug.

### I am running a supported version of CloudNativePG

- [X] I have read the troubleshooting guide and I think this is a new bug.

### Version

1.22.2

### What version of Kubernetes are you using?

1.29

### What is your Kubernetes environment?

Self-managed: kind (evaluation)

### How did you install the operator?

YAML manifest

### What happened?

when issue 
kubectl cnpg destroy foo bar
it report nothing, neither run failed none any message report if instance not found
and we accept instance number as the second params,it is not easy to figure out the params gave is wrong. 

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #4280, opened 2024-04-15
fix: destroy command show meaningful message if pod not exists

litaocdl:
this patch enhance the destroy, fencing and promote plugin command to show
context help if argument number is not met. Also in this patch we make the destroy
command accept both node id and node name as the second parameter. 

Closes: #4266

### DIFF
--- a/internal/cmd/plugin/destroy/cmd.go
+++ b/internal/cmd/plugin/destroy/cmd.go
@@ -26,9 +26,9 @@ import (
 	"github.com/cloudnative-pg/cloudnative-pg/internal/cmd/plugin"
 )
 
-// NewCmd create the new "promote" subcommand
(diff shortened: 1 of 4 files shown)


internal/cmd/plugin/destroy/cmd.go
```diff
@@ -18,22 +18,30 @@ package destroy
...
+	"strconv"
 
 	"github.com/spf13/cobra"
+
+	"github.com/cloudnative-pg/cloudnative-pg/internal/cmd/plugin"
 )
 
 // NewCmd create the new "promote" subcommand
 func NewCmd() *cobra.Command {
 	promoteCmd := &cobra.Command{
```
YanniHu1996, line 31: Not from this PR, but we can improve

YanniHu1996 (approved):

PR #4283, opened 2024-04-15
fix: empty LSN handling in instance sorting

leonardoce:
`receivedLSN` is empty for instances that have been never attached to streaming replication.

The instance sorting code considers this an error, and the `Less` implementation for LSNs is not antisymmetric, resulting in a non-stable sorting.

This could result in the operator preferring a replica that is never attached to streaming replication over a replica that is more advanced but attached to streaming replication.

Closes: #4284


leonardoce:
E2e tests: https://github.com/EnterpriseDB/cloudnative-pg/actions/runs/8689767885

sxd (approved):

mnencia (approved):

Issue #4285, opened 2024-04-15
[Bug]: streaming connection is not required when electing a replica to switch over to


leonardoce:
### Is there an existing issue already for this bug?

- [X] I have searched for an existing issue, and could not find anything. I believe this is a new bug.

### I have read the troubleshooting guide

- [X] I have read the troubleshooting guide and I think this is a new bug.

### I am running a supported version of CloudNativePG

- [X] I have read the troubleshooting guide and I think this is a new bug.

### Contact Details

leonardo.cecchi@gmail.com

### Version

1.22.2

### What version of Kubernetes are you using?

1.29

### What is your Kubernetes environment?

Self-managed: kind (evaluation)

### How did you install the operator?

YAML manifest

### What happened?

Similar to #4284 but should happen less frequently.

When selecting a replica to switch over to, we don't ensure that the replica is attached to the streaming connection.
Being not attached to it, the code that waits for the WAL receivers to be down before electing it will not ensure that the replica it actually up-to-date.

This could lead to an early promotion of a replica.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #4288, opened 2024-04-15
fix: wait for elected replica to be in streaming mode before a switchover

leonardoce:
Before switching over to a replica to apply a configuration change of the primary, we now wait for that replica to be connected in streaming replication.

We do that to ensure that the replica has all the data that is written on the primary.

Closes: #4285


mnencia (approved):

armru (approved):

mnencia:
The backport labels were missing, so I backported manually.

PR #4309, opened 2024-04-16
chore: increase waiting for a mounted secret timeout

### LINKED ISSUE #4075: [Bug]: Do not stop reconciliation loop when webhook secret aren't mounted

### Contact Details

_No response_

### Version

1.22.1

1.28

YAML manifest

### What happened?

When the certificates are renewed, and the secret it's not refreshed in time the reconciliation loop is stopped and so it's the entire certification refresh process.
We should avoid stopping the reconciliation loop just because we were waiting too little for that secret to be mounted, this could be easily reach in a really busy cluster 

### Cluster resource

_No response_

### Relevant log output

_No response_

### Code of Conduct

mnencia:
Increase the timeout the operator waits for the Kubelet to
refresh a mounted secret from ~10 to ~60 sec to better cope with
busy Kubernetes clusters.

This timeout is used during certificate maintenance and when it expires
the operator will retry after the configured time (default one day).

Closes:  #4075

### DIFF
--- a/pkg/certs/k8s.go
+++ b/pkg/certs/k8s.go
@@ -46,7 +46,7 @@ var (
 	errSecretsMountNotRefreshed = errors.New("secrets mount still not refreshed")
 
 	mountedSecretCheckBackoff = wait.Backoff{
-		Duration: 10 * time.Millisecond,
+		Duration: 60 * time.Millisecond,
 		Jitter:   0.1,
 		Factor:   2,
 		Steps:    10,


litaocdl:
e2e is running here https://github.com/EnterpriseDB/cloudnative-pg/actions/runs/8733107460

litaocdl:
the failed case is a known issue which is address by separate ticket, merge this one

Issue #4317, opened 2024-04-17
[Bug]: replace version is not generated in make file in mac


litaocdl:
### Is there an existing issue already for this bug?

- [X] I have searched for an existing issue, and could not find anything. I believe this is a new bug.

### I have read the troubleshooting guide

- [X] I have read the troubleshooting guide and I think this is a new bug.

### I am running a supported version of CloudNativePG

- [X] I have read the troubleshooting guide and I think this is a new bug.

### Version

1.22.2

### What version of Kubernetes are you using?

1.29

### What is your Kubernetes environment?

Self-managed: kind (evaluation)

### How did you install the operator?

YAML manifest

### What happened?

when run make file on mac saw 
```
Building operator from current worktree
fatal: Not a valid object name ^
```
and replace_version is not correctly generated
fine in linux

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #4318, opened 2024-04-17
fix: replace version is not generated in make file

litaocdl:
In patch #4196 , we introduce the REPLACE_VERSION in make file, 
which is not work well when run make in mac/OS,  this patch fix the 
problem.

Closes: #4317


NiccoloFei:
Closing as duplicate of https://github.com/cloudnative-pg/cloudnative-pg/commit/ce3f07236d3633bcdae37ff60039bef148390a79

PR #4340, opened 2024-04-19
fix: regression on the creation of the bundle and catalog for OLM

sxd:
We must always add to the CONTROLLER_IMG variable the words `catalog` and
`bundle` otherwise, we can't have the TAG if the CONTROLLER_IMG was already
provided by the user.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,10 +23,10 @@ ifeq (,$(CONTROLLER_IMG))
 IMAGE_TAG = $(shell (git symbolic-ref -q --short HEAD || git describe --tags --exact-match) | tr / -)
 ifneq (,${IMAGE_TAG})
 CONTROLLER_IMG = ${IMAGE_NAME}:${IMAGE_TAG}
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -23,11 +23,10 @@ ifeq (,$(CONTROLLER_IMG))
 IMAGE_TAG = $(shell (git symbolic-ref -q --short HEAD || git describe --tags --exact-match) | tr / -)
 ifneq (,${IMAGE_TAG})
 CONTROLLER_IMG = ${IMAGE_NAME}:${IMAGE_TAG}
-BUNDLE_IMG = ${IMAGE_NAME}:bundle-${IMAGE_TAG}
-CATALOG_IMG = ${IMAGE_NAME}:catalog-${IMAGE_TAG}
 endif
 endif
 
+
```
mnencia, line 29:
```suggestion
```

NiccoloFei (approved):

PR #4343, opened 2024-04-19
docs: community support covering the latest minor release

### LINKED ISSUE #4342: Proposal to support only the latest minor release
### Discussed in https://github.com/cloudnative-pg/cloudnative-pg/discussions/4258

<div type='discussions-op-text'>

<sup>Originally posted by **gbartolini** April  9, 2024</sup>
In the ["Supported release" page](https://cloudnative-pg.io/documentation/current/supported_releases/) we currently state:

> We support the latest two minor releases, and we aim to create a new minor release every two months.

We supported the latest two minor releases, given that we could release a minor version every two months. Unfortunately, we have not been able to respect this timeframe. 1.21 took 6 months. 1.23 is already 4 months on the go.

I propose to support only the latest version as a community and keep ~~the additional month~~ two additional months for the previous version.
</div>

gbartolini:
We have adjusted our support policy to focus on one supported minor release at a time, instead of two. Concurrently, we have extended the supplementary support period for the previous minor release to 3 months.

Closes #4342

### DIFF
--- a/docs/src/supported_releases.md
+++ b/docs/src/supported_releases.md
@@ -6,11 +6,12 @@
 *This page lists the status, timeline and policy for currently supported
 releases of CloudNativePG*.
 
-We support the latest two minor releases, and we aim to create a new minor
-release every two months. Each release is supported until the declared
-"End of Life" date in the [table below](#support-status-of-cloudnativepg-releases),
-which considers an additional month of assistance to allow for upgrade
-planning.
+We are committed to providing support for the latest minor release, with a
+dedication to launching a new minor release every two months. Each release
+remains fully supported until reaching its designated "End of Life" date, as
+outlined in the [support status table for CloudNativePG releases](#support-status-of-cloudnativepg-releases).
+This includes an additional 3-month assistance window to facilitate seamless
+upgrade planning.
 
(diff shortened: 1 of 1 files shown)


docs/src/supported_releases.md
```diff
@@ -67,7 +68,7 @@ Git tags for versions are prepended with `v`.
 |-----------------|----------------------|-------------------|---------------------|-------------------------------|---------------------------|-----------------------------|
 | 1.22.x          | Yes                  | December 21, 2023 | ~ July/August, 2024 | 1.26, 1.27, 1.28              | 1.23, 1.24, 1.25          | 12 - 16                     |
 | 1.21.x          | Yes                  | October 12, 2023  | ~ May 23, 2024      | 1.25, 1.26, 1.27, 1.28        | 1.23, 1.24                | 12 - 16                     |
-| main            | No, development only |                   |                     |                               |                           | 11 - 16                     |
+| main            | No, development only |                   |                     |                               |                           | 12 - 16                     |
```
mnencia, line 71: We could also populate the "Tested, but not supported" column with all the versions we test and set the other columns to N/A or just `-`.

leonardoce (approved):

phisco (approved):

PR #4388, opened 2024-04-26
feat: add `.spec.replicaCluster.primary` field

### LINKED ISSUE #3942: [Feature]: Introduce `replica.primary` and `replica.enabled = auto`

Currently, we need to manually transition from replication to primary mode, by setting `replica.enabled = true` or viceversa.

We should enable users to define a distributed topology, for example in 2 or 3 regions, and coordinate the scheduled promotion of a given cluster at regular intervals. We should also account for circular configurations based on cascading replication.

Consider the example in #3796, with the following changes.

In `pg1`:

```yaml
  replica:
    primary: pg1
    source: pg2
    enabled: auto
    # self: pg1 # this should be set by default to `metadata.name` (but overridable by users)
```

In `pg2`:

```yaml
  replica:
    primary: pg1
    source: pg1
    enabled: auto
    # self: pg2 # same as above
```

Similarly to `source`, the `primary` option matches a key in the `externalClusters` section. This option defines which `Cluster` is defined to be the `primary` in the distributed PostgreSQL cluster, based on the topology specified in `externalClusters`.

When `primary` is set, `enabled` must be set to `auto`.

When `primary` and `self` match, the cluster is accepting write operations (is a primary), so `replica` is the equivalent of `false`. When `primary` and `self` don't match, the cluster is in recovery mode from its `source`, so `replica` is the equivalent of `true`.

**IMPORTANT:** The solution should work with our without streaming replication (so just by relying on WAL shipping if required).

We had thought about using the database to share the topology. We believe that instead we should leverage configuration and rely on the `externalClusters` section.

### Additional context

We must cater for additional clusters. For example, there could be a `pg3` cluster in a third region, with the following circular architecture based on cascading replication:

```
pg1  ->  pg2  -> pg3 -
 ^                   |
  ------------------- 
```

### Backport?

No

No

### Code of Conduct

armru:
This patch introduces changes to the handling of the `.spec.replica.enabled` field, allowing it to be set to `nil`. Prior to this patch, only `true` and `false` were permitted, with `false` being the default. When the `.spec.replica.enabled` field is set to `nil`, the cluster's classification as a primary or replica cluster will be determined by two new fields, `.spec.replica.primary` and `.spec.replica.self`.

The `.spec.replica.self` field specifies the name of the current cluster, which will be compared with the `.spec.externalClusters[].name` field. The `.spec.replica.primary` field specifies the name of the current primary cluster, which will be searched in the same list.
Under the new behavior, when the `.spec.replica.enabled` field is set to `nil` and the `.spec.replica.primary` and `.spec.replica.self` fields contain the same value, the cluster is classified as a primary cluster.

Conversely, when the `.spec.replica.enabled` field is `nil` and the values of `.spec.replica.primary` and `.spec.replica.self` differ, the cluster will be classified as a replica cluster.

The purpose of this patch is to streamline the configuration process for a distributed PostgreSQL topology, where the `.spec.externalClusters` section is synchronized across different data centers.

Closes #3942


leonardoce (approved):

litaocdl (approved):

leonardoce:
E2e: https://github.com/EnterpriseDB/cloudnative-pg/actions/runs/9580425444

fcanovai (approved):

leonardoce (approved):

PR #4392, opened 2024-04-26
feat: add `cnpg.io/userType` label to generated secrets

### LINKED ISSUE #2631: When creating secrets for a cluster's users, add a label indicating what kind of user it's for
I am hoping to be able to rely on label searches alone to identify the superuser vs owner secret for a given cluster. So I'm imagining a label similar to `cnpg.io/user-role` where the value would be one of `superuser` or `owner`.

That way, I could find the secret using the pair of `cnpg.io/cluster` and this new label without relying on the assumption that the secret's name will always be the cluster name appended with "-app".

What do you think?

pchovelon:
This patch adds a `cnpg.io/userType` label to the secrets containing users' credentials whose value tells whether the user is a superuser or an application role.

Support for this label is limited to users created by the operator.

Closes #2631

### DIFF
--- a/docs/src/labels_annotations.md
+++ b/docs/src/labels_annotations.md
@@ -77,9 +77,11 @@ These predefined labels are managed by CloudNativePG.
 : Available on `ConfigMap` and `Secret` resources. When set to `true`,
   a change in the resource is automatically reloaded by the operator.
 
-`cnpg.io/user-role`
-: Available on `Secret` resources. Indicates the kind of user the Secret refers
-to, such as `superuser` or `default`.
+`cnpg.io/userType`
+: Specifies the type of PostgreSQL user associated with the
+  `Secret`, either `superuser` (Postgres superuser access) or `app`
+  (application-level user in CloudNativePG terminology), and is limited to the
+  default users created by CloudNativePG (typically `postgres` and `app`).
 
 `role` - **deprecated**
 :  Whether the instance running in a pod is a `primary` or a `replica`.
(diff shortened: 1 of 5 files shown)


pchovelon:
Do you think this PR could be in the next release ?

pchovelon:
Any idea when this PR can be integrated in a future release ?

docs/src/labels_annotations.md
```diff
@@ -77,6 +77,10 @@ These predefined labels are managed by CloudNativePG.
 : Available on `ConfigMap` and `Secret` resources. When set to `true`,
   a change in the resource is automatically reloaded by the operator.
 
+`cnpg.io/user-role`
```
jsilvela, line 80: `user` and `role`  are quasi-synonyms in Postgres.
How about `cnpg.io/user-class` or similar?

pkg/utils/labels_annotations.go
```diff
@@ -75,6 +75,9 @@ const (
 	// or not, use for Secrets or ConfigMaps
 	WatchedLabelName = MetadataNamespace + "/reload"
 
+	// UserRoleLabeName the name of the label which tell if a Secret refers to a superuser database role or a default one
+	UserRoleLabelName = MetadataNamespace + "/user-role"
```
jsilvela, line 79: Same comment as above, maybe "/user-class" or kind

gbartolini (approved):

pkg/utils/labels_annotations.go
gbartolini: I thought about `userClass` too, but then I opted for `userType`. Had I read your input earlier, I would have named it `userClass`. :)

pchovelon: UserType is nice and clear to understand !

jsilvela (approved):

leonardoce (approved):

gbartolini:
Let's wait for E2E tests

mnencia (approved):

mnencia:
/ok-to-merge tests failed due to a bug introduced in #5898

PR #4399, opened 2024-04-29
fix: replica cluster should restart after promotion

### LINKED ISSUE #4172: [Bug]: Replica cluster require a restart after promote

### Contact Details

_No response_

### Version

1.22.2

1.29

YAML manifest

### What happened?

when replica cluster promoted, `archive_mode` changed from always to on but primary is not restart, once scale up the cluster, primary will restart. 

### Cluster resource

_No response_

### Relevant log output

_No response_

### Code of Conduct

litaocdl:
When a replica cluster is promoted, the `archive_mode` is changed from
`always` to `on`. This change requires a restart because Postgres 
does not reload the configuration during the promotion.

Closes: #4172

### DIFF
--- a/internal/management/controller/instance_controller.go
+++ b/internal/management/controller/instance_controller.go
@@ -171,8 +171,8 @@ func (r *InstanceReconciler) Reconcile(
 		return reconcile.Result{RequeueAfter: time.Second}, nil
 	}
 
-	// If the instance is promoted, it will not automatically load the changed configuration files
-	// so promoted should not stop the instance reload requires
+	// Instance promotion will not automatically load the changed configuration files.
+	// Therefore it should not be counted as "a restart" to prevent instance restart.
 	if _, err := r.reconcilePrimary(ctx, cluster); err != nil {
 		return reconcile.Result{}, err
 	}
--- a/internal/management/controller/instance_controller.go
+++ b/internal/management/controller/instance_controller.go
@@ -173,7 +173,7 @@ func (r *InstanceReconciler) Reconcile(
 
 	// Instance promotion will not automatically load the changed configuration files.
 	// Therefore it should not be counted as "a restart" to prevent instance restart.
-	if _, err := r.reconcilePrimary(ctx, cluster); err != nil {
+	if err := r.reconcilePrimary(ctx, cluster); err != nil {
 		return reconcile.Result{}, err
 	}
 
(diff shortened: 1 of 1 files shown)


internal/management/controller/instance_controller.go
```diff
@@ -1114,7 +1113,7 @@ func (r *InstanceReconciler) reconcilePrimary(ctx context.Context, cluster *apiv
 		if err := r.handlePromotion(ctx, cluster); err != nil {
 			return false, err
 		}
-		restarted = true
+		promoted = true
```
jsilvela, line 1116: This is not your PR but the existing code.
This is asymmetric. We decide that the instance is a primary with IsPrimary, which checks for the existence of "recovery.conf"

Let's say we:
1. successfully promote the instace
2. when trying to update the Status or drop the connections, we fail

Then the next time around we'll see the instance is a primary, and we won't retry.

internal/management/controller/instance_controller.go
```diff
@@ -171,18 +171,17 @@ func (r *InstanceReconciler) Reconcile(
 		return reconcile.Result{RequeueAfter: time.Second}, nil
 	}
 
-	restarted, err := r.reconcilePrimary(ctx, cluster)
-	if err != nil {
+	// If the instance is promoted, it will not automatically load the changed configuration files
+	// so promoted should not stop the instance reload requires
+	if _, err := r.reconcilePrimary(ctx, cluster); err != nil {
```
jsilvela, line 176: This seems to make sense, and the E2E are looking very green.

jsilvela (approved):

internal/management/controller/instance_controller.go
```diff
@@ -171,18 +171,17 @@ func (r *InstanceReconciler) Reconcile(
 		return reconcile.Result{RequeueAfter: time.Second}, nil
 	}
 
-	restarted, err := r.reconcilePrimary(ctx, cluster)
-	if err != nil {
+	// If the instance is promoted, it will not automatically load the changed configuration files
+	// so promoted should not stop the instance reload requires
```
jsilvela, line 175:
```suggestion
	// Instance promotion will not automatically load the changed configuration files.
	// Therefore it should not be counted as "a restart" to prevent instance restart.
```

internal/management/controller/instance_controller.go
```diff
@@ -171,18 +171,17 @@ func (r *InstanceReconciler) Reconcile(
 		return reconcile.Result{RequeueAfter: time.Second}, nil
 	}
 
-	restarted, err := r.reconcilePrimary(ctx, cluster)
-	if err != nil {
+	// Instance promotion will not automatically load the changed configuration files.
+	// Therefore it should not be counted as "a restart" to prevent instance restart.
```
jsilvela, line 175:
```suggestion
	// Instance promotion will not automatically load the changed configuration files.
	// Therefore it should not be counted as "a restart".
```

litaocdl:
update the pr by removing the unused variable

fcanovai (approved):

mnencia (approved):

internal/management/controller/instance_controller.go
mnencia: If we don't update the status, the Status patch step is retried.
The drop connection is not vital. It is there only to prevent a long-running session through the read-only service from silently becoming a read-write one.

Issue #4286, opened 2024-04-15
[Bug]: Bug while creating a PodMonitor


CiraciNicolo:
### Is there an existing issue already for this bug?

- [X] I have searched for an existing issue, and could not find anything. I believe this is a new bug.

### I have read the troubleshooting guide

- [X] I have read the troubleshooting guide and I think this is a new bug.

### I am running a supported version of CloudNativePG

- [X] I have read the troubleshooting guide and I think this is a new bug.

### Contact Details

nicolo.ciraci@elmec.it

### Version

trunk (main)

### What version of Kubernetes are you using?

1.29

### What is your Kubernetes environment?

Self-managed: kind (evaluation)

### How did you install the operator?

Other

### What happened?

While reviewing the PodMonitor functionality as described in #4268, we found a regression caused by the new Plugin API implementation that crashed the controller while creating the PodMonitor. The issue can be found [here](https://github.com/cloudnative-pg/cloudnative-pg/blob/6e423eb50b71ced1fe78aaecdb1cf0411aefdac5/internal/cnpi/plugin/client/lifecycle.go#L61).

The new lifecycle hook does not recognise the new PodMonitor CR and return nil to the Kubernetes API client; this triggers a reconcile error in which the cluster is not created. 

A fix is implemented in the PR #4282.

### Cluster resource

```shell
---
apiVersion: postgresql.cnpg.io/v1
kind: Cluster
metadata:
  name: cluster-default-monitoring
spec:
  instances: 1

  monitoring:
    enablePodMonitor: true

  storage:
    size: 1Gi
```

### Relevant log output

```shell
[...]

PR #4282, opened 2024-04-15
fix: PodMonitor regression

CiraciNicolo:
# Summary

This pr fix a regression introduced by the new Plugin API which caused the PodMonitor - and probably other CRD - to not be created. Some E2E test have been introduced to detect issues with PodMonitors.

Furthermore some commodities have been added and general code tweaks.

# Additional Info

Kindly note that this pr is a cherry picked version of https://github.com/cloudnative-pg/cloudnative-pg/pull/4268, which contains a broader review of PodMonitor specs.

Fixes: #4286

PR #4402, opened 2024-04-29
feat: allow injecting schema into lifecycle hook

phisco:
An attempt at fixing the root cause of #4286, mostly a POC, I feel it could be improved.

### DIFF
--- a/internal/cnpi/plugin/client/lifecycle.go
+++ b/internal/cnpi/plugin/client/lifecycle.go
@@ -25,6 +25,7 @@ import (
 	"github.com/cloudnative-pg/cnpg-i/pkg/lifecycle"
 	jsonpatch "github.com/evanphx/json-patch/v5"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/serializer"
 	"k8s.io/apimachinery/pkg/util/json"
 	"k8s.io/client-go/kubernetes/scheme"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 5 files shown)


internal/cnpi/plugin/client/lifecycle.go
```diff
@@ -53,15 +49,20 @@ func (data *data) LifecycleHook(
...
+			return object, nil
+		}
+		gvk, err = apiutil.GVKForObject(object, schema)
 		if err != nil {
-			contextLogger.Trace("skipping unknown object", "object", object)
+			contextLogger.Info("missing gvk for object, got error trying to infer it from schema",
+				"object", object, "error", err)
 			// Skip unknown object
-			return nil, nil
+			return object, nil
```
phisco, line 62: this is already covered by https://github.com/cloudnative-pg/cloudnative-pg/pull/4282, I do think it's the right thing to do.

internal/cnpi/plugin/client/lifecycle.go
```diff
@@ -144,7 +145,7 @@ func (data *data) LifecycleHook(
 		return object, nil
 	}
 
-	decoder := scheme.Codecs.UniversalDeserializer()
+	decoder := serializer.NewCodecFactory(scheme.Scheme).UniversalDeserializer()
```
phisco, line 148: would be better to move this to a builder and only create it once.

internal/cnpi/plugin/client/lifecycle.go
```diff
@@ -34,15 +35,10 @@ import (
...
-
-func init() {
-	_ = scheme.AddToScheme(runtimeScheme)
-}
-
-func (data *data) LifecycleHook(
+func (data *data) LifecycleHook( //nolint:gocognit
 	ctx context.Context,
 	operationType plugin.OperationVerb,
+	schema *runtime.Scheme,
```
phisco, line 41: I feel it could be passed through a builder or something instead of adding one more parameter.

PR #4404, opened 2024-04-29
feat: prevent failovers when disk space is exhausted

### LINKED ISSUE #4521: [Feature]: Graceful handling of WAL disk space exhaustion

PostgreSQL will cleanly shut down when there's no space left for WAL files.
The operator will perceive that condition as a failure of the primary instance, leading to a failover.

Failing over will not help because every other replica will have the same error condition.

Having the operator automatically fence the primary would help because it would prevent any automatic failover from happening and would give the user time to increase the WAL disk space, fixing the root cause of the issue.

Automatically scaling up the disk space will work, too, but I feel this as a different and broader topic.

### Additional context

_No response_

### Backport?

No

Yes

### Code of Conduct

leonardoce:
PostgreSQL will shut down cleanly when there is not enough disk space to store WAL files.

The operator did not recognize this condition and, since the primary failed, was performing a failover to the most advanced replica. This action will not fix the underlying issue.

Only a manual disk resize, initiated by the user, can ultimately lead to a fully working PostgreSQL cluster.

This patch makes the instance manager recognize this condition and report it to the operator. Upon detecting it, the operator will not trigger a switchover and set a phase describing the situation.

After the PVCs are resized, the cluster will restart working correctly.

Closes: #4521


leonardoce:
I tested this using Longhorn in a Fedora VM, but any storage enforcing the PV capacity will do the trick.

To test the patch, you need to finish your WAL storage. To keep things easy, I used:

```yaml
kind: Cluster
metadata:
  name: cluster-example
spec:
  instances: 1

  storage:
    size: 256Mi
```

And then:

```sql
CREATE TABLE storage_area (t text);

-- repeat the following query 20-30 times (you need to be fast!)
INSERT INTO storage_area (t) (select repeat('Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do', 5*1024*1024));
```

With the predefined WAL settings, you'll finish your WAL disk space before you finish the space for PGDATA.

pkg/management/postgres/instance.go
```diff
@@ -187,6 +188,9 @@ type Instance struct {
 	// mightBeUnavailable specifies whether we expect the instance to be down
 	mightBeUnavailable atomic.Bool
 
+	// noWALDiskSpaceLeft specified whether there's no more free disk space
+	noWALDiskSpaceLeft atomic.Bool
```
gbartolini, line 192: I'd suggest either `walDiskNoSpaceLeft` or `noSpaceLeftOnWALDisk`

controllers/cluster_controller.go
```diff
@@ -305,6 +305,35 @@ func (r *ClusterReconciler) reconcile(ctx context.Context, cluster *apiv1.Cluste
...
+			"podName", podName,
+		)
+
+		fencingErr := utils.NewFencingMetadataExecutor(r.Client).
+			ForInstance(podName).
+			AddFencing().
+			Execute(ctx, client.ObjectKeyFromObject(cluster), cluster)
+		if fencingErr != nil {
+			contextLogger.Error(
+				err,
```
armru, line 320: at quick glance this should be fencingErr

armru (approved):

pkg/fileutils/directory.go
```diff
@@ -0,0 +1,139 @@
...
+		path:           path,
+		createFileFunc: createFileWithSize,
+	}
+}
+
+// createFileWithSize creates a file with a certain name and
+// a certain size. It will fail if the file already exists.
+//
+// To allocate the file, the specified number zero bytes will
+// be written.
```
jsilvela, line 60:
```suggestion
// To allocate the file, the specified number of bytes will
// be written, set to zero.
```

pkg/fileutils/directory.go
```diff
@@ -0,0 +1,139 @@
...
+		}
+		writtenBytes += b
+	}
+
+	return nil
+}
+
+// HasSpaceInDirectory checks if there's enough disk space to store a
+// file with a specified size inside the passed directory. It does that
+// by using createFileWithSize to create such a file in the directory
```
jsilvela, line 100:
```suggestion
// file with a specified size inside the directory. It does that
// by using createFileFunc to create such a file in the directory
```

pkg/fileutils/directory.go
```diff
@@ -0,0 +1,139 @@
...
+		return false, nil
+	} else if err != nil {
+		return false, err
+	}
+	return true, nil
+}
+
+// IsNoSpaceLeftOnDevice returns true when there's no more
+// space left
+func IsNoSpaceLeftOnDevice(err error) bool {
```
jsilvela, line 132: I think this could be private

pkg/management/postgres/instance.go
```diff
@@ -247,11 +251,41 @@ func (instance *Instance) SetCanCheckReadiness(enabled bool) {
 	instance.canCheckReadiness.Store(enabled)
 }
 
+// HasWALDiskSpace checks if we have enough disk space to store two WAL files
+func (instance *Instance) HasWALDiskSpace(ctx context.Context) (bool, error) {
```
jsilvela, line 255: How do we know that the issue is the "WAL disk"? If the cluster does not have a separate volume for WAL storage, then this would be the PG_DATA disk, no?

pkg/management/postgres/instance.go
```diff
@@ -247,11 +251,41 @@ func (instance *Instance) SetCanCheckReadiness(enabled bool) {
...
 func (instance *Instance) SetMightBeUnavailable(enabled bool) {
 	instance.mightBeUnavailable.Store(enabled)
 }
 
+// SetNoDiskSpaceLeft marks whether PostgreSQL exited because there's no more
+// disk space left for WALs
+func (instance *Instance) SetNoDiskSpaceLeft(enabled bool) {
+	instance.noWALDiskSpaceLeft.Store(enabled)
+}
+
```
jsilvela, line 288: Firstly, the method is hard to parse, since it begins with "set no disk ..." which gives a wrong impression.
Then, `enabled` doesn't look like the correct name for this boolean.

```suggestion
// SetDiskIsFull marks whether PostgreSQL exited because there's no more
// disk space left for WALs
func (instance *Instance) SetDiskIsFull(diskIsFull bool) {
	instance.noWALDiskSpaceLeft.Store(diskIsFull)
}

```

pkg/utils/fencing.go
```diff
@@ -165,37 +166,46 @@ func (fb *FencingMetadataExecutor) RemoveFencing() *FencingMetadataExecutor {
...
 }
 
 // Execute executes the instructions given with the fluent builder, returns any error encountered
 func (fb *FencingMetadataExecutor) Execute(ctx context.Context, key types.NamespacedName, obj client.Object) error {
-	if fb.instanceName == "" {
+	if len(fb.instanceNames) == 0 {
 		return errors.New("chose an operation to execute")
 	}
+	if len(fb.instanceNames) > 1 && slices.Contains(fb.instanceNames, FenceAllInstances) {
+		return errors.New("the fence all instance token cannot be used along other instances")
```
jsilvela, line 185:
```suggestion
		return errors.New("the fence-all-instances token cannot be used along other instances")
```

jsilvela (commented):
About to start adding documentation and going over the E2E, but left a few comments on the implementation bits.
IMO the "WALDisk" nomenclature could get confusing as it seems to imply there is a separate WAL volume, which may or may not be the case.

controllers/cluster_controller.go
```diff
@@ -394,6 +398,47 @@ func (r *ClusterReconciler) reconcile(ctx context.Context, cluster *apiv1.Cluste
 	return hookResult.Result, hookResult.Err
 }
 
+func (r *ClusterReconciler) ensureSufficientDiskSpace(
```
jsilvela, line 401: IMO this name would imply that if disk space is running out, we resize the volume.
At present we don't "ensure".
```suggestion
func (r *ClusterReconciler) reconcileDiskFull(
```

pkg/management/postgres/instance.go
leonardoce: Yes! But that was meant to check if there's free space for WALs, besides in which volume they're hosted.

controllers/cluster_controller.go
leonardoce: @armru, what about you? Which name would you prefer?

internal/cmd/manager/instance/run/lifecycle/run.go
```diff
@@ -115,7 +121,25 @@ func (i *PostgresLifecycle) runPostgresAndWait(ctx context.Context) <-chan error
...
+			switch {
+			case err != nil:
+				contextLogger.Error(err, "unable to check for free disk space")
+
+			case !freeDiskSpace:
+				i.instance.SetNoDiskSpaceLeft(true)
+				errChan <- ErrNoFreeWALSpace
+
+			default:
+				errChan <- err
```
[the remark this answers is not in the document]
leonardoce: No, that's a mistake. It should not be `err` but `pgExitStatus`. Let me fix it.

pkg/management/postgres/instance.go
leonardoce: I renamed it `SetWALDiskIsFull`, is that fine with you?

jsilvela: Yes, perfect

pkg/management/postgres/instance.go
jsilvela: I know that what we mean here is "Space on the disk to store WALs", but the workding is "Has space on the WAL disk". It's close but not the same, and my reflex impression before reading the code is that this requires the dedicated WAL volume.
How about `HasDiskSpaceForWAL`?
Or, if we keep the method name, we can add a comment to clarify

jsilvela (approved):
I still think it's worth renaming the `ensureSufficientDiskSpace` method, but otherwise give this an enthusiastic 👍

docs/src/instance_manager.md
```diff
@@ -94,3 +94,44 @@ the WAL files. By default it is set to `3600` (1 hour).
 
 In case of primary pod failure, the cluster will go into failover mode.
 Please refer to the ["Failover" section](failover.md) for details.
+
+## Behavior when disk becomes full
```
jsilvela, line 98:
```suggestion
## Behavior when disk storage becomes full
```

docs/src/troubleshooting.md
```diff
@@ -638,6 +638,19 @@ suggest this could be due to a full disk, you probably have to increase the
 size of the instance's `PersistentVolumeClaim`. Please look at the
 ["Volume expansion" section](storage.md#volume-expansion) in the documentation.
 
+#### Automatically fenced cluster on WAL storage becoming full
+
+If the space for WAL segments is full, the instance manager will
```
jsilvela, line 643:
```suggestion
If the space for WAL segments is exhausted, the instance manager will
```

pkg/fileutils/directory.go
```diff
@@ -0,0 +1,139 @@
...
+
+type fileCreatorFunc = func(ctx context.Context, name string, size int) error
+
+// Directory represents a filesystem directory and provides methods to interact
+// with it, such as checking for available disk space by attempting to create
+// a file of a specified size.
+type Directory struct {
+	path           string
+	createFileFunc fileCreatorFunc
+}
```
jsilvela, line 45: sorry, yet another comment.
I think this name, Directory, for this object, is far too wide.
Especially since this would be `fileutils.Directory` and there is a `fileutils.EnsureParentDirectoryExist` function.
This Directory is not a general directory implementation but a very specific use case.

jsilvela:
```suggestion
type DiskProbe struct {
	path           string
	createFileFunc fileCreatorFunc
}
```

pkg/fileutils/directory.go
```diff
@@ -0,0 +1,139 @@
...
+	}
+
+	return nil
+}
+
+// HasSpaceInDirectory checks if there's enough disk space to store a
+// file with a specified size inside the directory. It does that
+// by using createFileFunc to create such a file in the directory
+// and then removing it.
+func (d Directory) HasSpaceInDirectory(ctx context.Context, size int) (bool, error) {
```
jsilvela, line 102: following the class rename proposal:
```suggestion
func (d DiskProbe) HasStorageAvailable(ctx context.Context, size int) (bool, error) {
```

controllers/cluster_controller.go
leonardoce: I still think `ensure` is the best "prefix" because we aren't really reconciling something.
Perhaps `ensureSufficientDiskSpace` is not the correct name, too; we aren't scaling up the volume size.

What about `ensureWALSpaceFencing`? @jsilvela

leonardoce:
The documentation looks good to me

controllers/cluster_controller.go
jsilvela: Hmm. If we must use the ensure prefix, then we might want to be more explicit.
How about `ensureNoFailoverOnFullDisk` or `ensureFencingOnFullDisk` @leonardoce ?
Or (maybe AND) worth having a method comment.

leonardoce: `ensureNoFailoverOnFullDisk` is nice! I'll use that.

controllers/cluster_controller.go
```diff
@@ -394,6 +398,47 @@ func (r *ClusterReconciler) reconcile(ctx context.Context, cluster *apiv1.Cluste
...
+		Execute(ctx, client.ObjectKeyFromObject(cluster), cluster)
+	if fencingErr != nil {
+		contextLogger.Error(
+			fencingErr,
+			"Error while fencing instance because disk space is finished",
+		)
+		return ctrl.Result{}, fencingErr
+	}
+
+	reason := "Found one or more Pods without enough disk space for PostgreSQL to continue running." +
```
gbartolini:
```suggestion
	reason := "Insufficient disk space detected in one or more pods is preventing PostgreSQL from running." +
```

gbartolini (approved):

leonardoce:
I'd like to test it manually with LongHorn after all the changes.
After that, I'll merge it.
Thank you!

leonardoce:
I tested it manually, and it works just file.
I run an E2e test suite for the cloud environments in the EDB repo: https://github.com/EnterpriseDB/cloudnative-pg/actions/runs/9190831899

leonardoce:
E2e tests: https://github.com/EnterpriseDB/cloudnative-pg/actions/runs/9191020863

leonardoce:
The E2e tests spotted this bug: #4636.
We need to fix it before merging.

leonardoce:
E2e: https://github.com/EnterpriseDB/cloudnative-pg/actions/runs/9206156623

leonardoce:
https://github.com/EnterpriseDB/cloudnative-pg/actions/runs/9284417523

leonardoce:
E2e tests are green!!!!

leonardoce:
I started an entire E2e run, both local and on cloud: https://github.com/EnterpriseDB/cloudnative-pg/actions/runs/9286574473

leonardoce:
We removed the `WALSpaceAvailable` because we're worried that it could be misunderstood and used as a metric instead of proper monitoring infrastructure.

The operator sets this value from the exit code of the instance Pods. The instance Pods check whether they have enough WAL disk space before starting PG and after PostgreSQL exits with an error condition.
This is not a way to monitor disk space.

leonardoce:
Full E2e run: https://github.com/EnterpriseDB/cloudnative-pg/actions/runs/9363810554

leonardoce:
E2e tests are fine!

PR #4424, opened 2024-05-02
feat: Add plugin method to ScheduledBackup

### LINKED ISSUE #4422: [Feature]: ScheduledBackup resource should be able to backup using plugin method

It will allow the use of backup plugins in the ScheduledBackup resource

The pluginConfiguration field is already in the CRD, so it only needs to add  "plugin" to the enum for method and the logic to handle the plugin type similar to the Backup controller.

The only alternative I can think of is having a separate cron mechanism to call backups with the required configuration

### Additional context

_No response_

### Backport?

Yes

Yes

### Code of Conduct

dougkirkley:
Closes #4422

### DIFF
--- a/api/v1/scheduledbackup_types.go
+++ b/api/v1/scheduledbackup_types.go
@@ -60,10 +60,10 @@ type ScheduledBackupSpec struct {
 	// +optional
 	Target BackupTarget `json:"target,omitempty"`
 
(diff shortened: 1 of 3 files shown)


mnencia (approved):

armru (approved):

dougkirkley:
@mnencia I just saw that I missed the API doc update, so I pushed that change.

PR #4425, opened 2024-05-02
chore: avoid hitting the API server when updating plugins status

leonardoce:
When not needed, this patch avoids hitting the API server with a PATCH request.

### DIFF
--- a/controllers/cluster_plugins.go
+++ b/controllers/cluster_plugins.go
@@ -19,6 +19,7 @@ package controllers
 
 import (
 	"context"
+	"reflect"
 
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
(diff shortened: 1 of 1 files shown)


sxd (approved):

controllers/cluster_plugins.go
```diff
@@ -54,5 +55,10 @@ func (r *ClusterReconciler) updatePluginsStatus(ctx context.Context, cluster *ap
 		cluster.Status.PluginStatus[i].BackupCapabilities = entry.BackupCapabilities
 	}
 
+	// If nothing changes, there's no need to hit the API server
+	if reflect.DeepEqual(oldCluster.Status.PluginStatus, cluster.Status.PluginStatus) {
+		return nil
+	}
+
 	return r.Client.Status().Patch(ctx, cluster, client.MergeFrom(oldCluster))
```
phisco, line 63: Shouldn't this already avoid hitting the api server if the patch is empty (which should always be the case if deepequal)?

leonardoce: Nice question. Turns out it doesn't :(

leonardoce: @sxd have more details

sxd: @phisco I did explain you this or not? because I don't remember :S

mnencia (approved):

controllers/cluster_plugins.go
phisco: Yes you did, the issue wasn't hitting the api server, which it already did for no-op changes, but all the json marshaling/unmarshaling to build patches and understand it was a no-op. So the pr description could be updated, but the fix is right.

phisco (approved):

sxd:
/ok-to-merge some test failed but it's not related at all

PR #4434, opened 2024-05-03
fix: instance manager stuck if PostgreSQL doesn't start

### LINKED ISSUE #4433: [Bug]: The instance manager is not able to unfence PG after a previous unfence operation failed

### Contact Details

### Version

1.23.0

1.30 (unsuppprted)

YAML manifest

### What happened?

If an unfence request fails because PostgreSQL cannot start up, subsequent unfence requests will fail, too, because the instance manager won't start PostgreSQL up again.

These steps will reproduce this weird situation programmatically and every time:

1. Applying the attached cluster definition (this is the minimal example, where instances have been replaced by 1 [even if the number of instances should not be relevant])
2. Fence the primary instance (`./bin/kubectl-cnpg fencing on cluster-example 1`) -- This will result in PG to shut down
3. Remove every WAL file from PGDATA (`kubectl exec -ti cluster-example-1 -- bash -c 'rm $PGDATA/pg_wal/*'`)
4. Lift fencing (`./bin/kubectl-cnpg fencing off cluster-example 1`) -- This will result in PG to start up again, and fail to start up
5. Fence the instance again (`./bin/kubectl-cnpg fencing on cluster-example 1`)
6. Unfence the instance (`./bin/kubectl-cnpg fencing off cluster-example 1`) -- The instance manager will not even try to start PostgreSQL

I'm working on a patch to fix this issue.

### Cluster resource

```shell
kind: Cluster
metadata:
  name: cluster-example
spec:
  instances: 1

  storage:
    size: 1Gi
```

### Relevant log output

_No response_

### Code of Conduct

leonardoce:
If a primary PostgreSQL instance fails to start after being stopped, the instance manager is indefinitely waiting for it to start running the configuration queries.

This patch creates a separate context for each postmaster process and runs the configuration queries in a parallel goroutine, canceling it as soon as the postmaster finishes.

Fixes: #4433

### DIFF
--- a/internal/cmd/manager/instance/run/lifecycle/run.go
+++ b/internal/cmd/manager/instance/run/lifecycle/run.go
@@ -43,14 +43,10 @@ func (i *PostgresLifecycle) runPostgresAndWait(ctx context.Context) <-chan error
 		defer close(errChan)
 
 		var wg sync.WaitGroup
-		defer func() {
-			wg.Wait()
-		}()
+		defer wg.Wait()
 
 		postgresContext, postgresContextCancel := context.WithCancel(ctx)
-		defer func() {
-			postgresContextCancel()
-		}()
+		defer postgresContextCancel()
 
 		err := i.instance.VerifyPgDataCoherence(postgresContext)
 		if err != nil {
--- a/internal/cmd/manager/instance/run/lifecycle/run.go
+++ b/internal/cmd/manager/instance/run/lifecycle/run.go
@@ -39,26 +39,50 @@ func (i *PostgresLifecycle) runPostgresAndWait(ctx context.Context) <-chan error
 	contextLogger := log.FromContext(ctx)
 	errChan := make(chan error, 1)
 
+	// The following goroutine runs the postmaster process, and stops
+	// when the process exits.
 	go func() {
 		defer close(errChan)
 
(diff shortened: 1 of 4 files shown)


internal/cmd/manager/instance/run/lifecycle/run.go
```diff
@@ -67,12 +73,13 @@ func (i *PostgresLifecycle) runPostgresAndWait(ctx context.Context) <-chan error
 		}
 
 		// once the database will be up we'll connect and setup everything required
-		err = configureInstancePermissions(ctx, i.instance)
-		if err != nil {
-			contextLogger.Error(err, "Unable to update PostgreSQL roles and permissions")
-			errChan <- err
-			return
-		}
+		go func() {
```
jsilvela, line 76: I worry about this being a goroutine without a waitgroup or similar.
Seems if there is an error here but it doesn't happen super quick, we could ignore it.

leonardoce:
E2e tests are all green!

armru (approved):

mnencia:
I think the concurrency in this code is difficult to understand correctly. I would like to have some comments to help understanding it.

mnencia (approved):

fcanovai (approved):

PR #4442, opened 2024-05-06
feat: TLS communication between the operator and the instance manager

### LINKED ISSUE #4441: [Feature]: The communication between the operator and the instance manager should be encrypted

Even if no sensitive data is sent in the status connection, using SSL/TLS in all HTTP connections is desirable.

Both the end of the status connection are controlled by the CNPG code, so it should be possible to update an existing cluster without losing the connectivity. The status connection, however, is also used for the probes, so it will require recreating the pods to upgrade the protocol from HTTP to HTTPS.

Online upgrades of the instance manager will require particular attention to be supported. The new instance manager should continue to serve HTTP until the pod is recreated to avoid breaking the probes.

We could encrypt the connection using an external tool like Istio, but it would not be by default and would complicate the deployment.

### Additional context

After implementing this feature, we should also consider adding TLS capabilities to the metrics exporter.

### Backport?

No

Yes

### Code of Conduct

mnencia:
Closes #4441

### DIFF
--- a/docs/src/security.md
+++ b/docs/src/security.md
@@ -302,7 +302,7 @@ System           | Port number  | Exposing            |  Name               |  C
 operator         | 9443         | webhook server      | `webhook-server`    |  TLS                           | Yes
 operator         | 8080         | metrics             | `metrics`           |  no TLS                        | No
 instance manager | 9187         | metrics             | `metrics`           |  no TLS                        | No
-instance manager | 8000         | status              | `status`            |  TLS (no TLS in old releases)  | Yes
+instance manager | 8000         | status              | `status`            |  TLS (no TLS in old releases)  | No
 operand          | 5432         | PostgreSQL instance | `postgresql`        |  optional TLS                  | Yes
 
 ### PostgreSQL
--- a/docs/src/security.md
+++ b/docs/src/security.md
@@ -297,13 +297,13 @@ section of the Kubernetes documentation for further information.
 CloudNativePG exposes ports at operator, instance manager and operand
 levels, as listed in the table below:
 
-System           | Port number  | Exposing            |  Name               |  Certificates  |  Authentication
-:--------------- | :----------- | :------------------ | :------------------ | :------------  | :--------------
-operator         | 9443         | webhook server      | `webhook-server`    |  TLS           | Yes
-operator         | 8080         | metrics             | `metrics`           |  no TLS        | No
-instance manager | 9187         | metrics             | `metrics`           |  no TLS        | No
-instance manager | 8000         | status              | `status`            |  no TLS        | No
-operand          | 5432         | PostgreSQL instance | `postgresql`        |  optional TLS  | Yes
+System           | Port number  | Exposing            |  Name               |  Certificates                  |  Authentication
+:--------------- | :----------- | :------------------ | :------------------ | :----------------------------  | :--------------
+operator         | 9443         | webhook server      | `webhook-server`    |  TLS                           | Yes
+operator         | 8080         | metrics             | `metrics`           |  no TLS                        | No
+instance manager | 9187         | metrics             | `metrics`           |  no TLS                        | No
+instance manager | 8000         | status              | `status`            |  TLS (no TLS in old releases)  | Yes
+operand          | 5432         | PostgreSQL instance | `postgresql`        |  optional TLS                  | Yes
 
 ### PostgreSQL
 
--- a/internal/cmd/manager/instance/run/cmd.go
+++ b/internal/cmd/manager/instance/run/cmd.go
@@ -75,7 +75,7 @@ func NewCmd() *cobra.Command {
 	var podName string
 	var clusterName string
 	var namespace string
-	var tlsStatus bool
+	var statusPortTLS bool
 
 	cmd := &cobra.Command{
 		Use: "run [flags]",
@@ -93,7 +93,7 @@ func NewCmd() *cobra.Command {
 			instance.Namespace = namespace
 			instance.PodName = podName
 			instance.ClusterName = clusterName
-			instance.StatusTLS = tlsStatus
+			instance.StatusPortTLS = statusPortTLS
 
 			err := retry.OnError(retry.DefaultRetry, isRunSubCommandRetryable, func() error {
 				return runSubCommand(ctx, instance)
@@ -121,7 +121,7 @@ func NewCmd() *cobra.Command {
 		"current cluster in k8s, used to coordinate switchover and failover")
 	cmd.Flags().StringVar(&namespace, "namespace", os.Getenv("NAMESPACE"), "The namespace of "+
 		"the cluster and of the Pod in k8s")
-	cmd.Flags().BoolVar(&tlsStatus, "tls-status", false,
+	cmd.Flags().BoolVar(&statusPortTLS, "status-port-tls", false,
 		"Enable TLS for communicating with the operator")
 	return cmd
 }
--- a/internal/cmd/manager/instance/run/cmd.go
+++ b/internal/cmd/manager/instance/run/cmd.go
@@ -122,7 +122,7 @@ func NewCmd() *cobra.Command {
 	cmd.Flags().StringVar(&namespace, "namespace", os.Getenv("NAMESPACE"), "The namespace of "+
 		"the cluster and of the Pod in k8s")
 	cmd.Flags().BoolVar(&tlsStatus, "tls-status", false,
-		"Enable TLS communicating with the operator")
+		"Enable TLS for communicating with the operator")
 	return cmd
 }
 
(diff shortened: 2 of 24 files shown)


controllers/cluster_upgrade.go
```diff
@@ -770,6 +771,7 @@ func (r *ClusterReconciler) upgradeInstanceManager(
 // upgradeInstanceManagerOnPod upgrades an instance manager of a Pod via an HTTP PUT request.
 func upgradeInstanceManagerOnPod(
 	ctx context.Context,
+	statusClient *instance.StatusClient,
```
mnencia, line 774: I think this needs to be a different client. The timeout we use for the status client could hurt here.

pkg/resources/instance/client.go
```diff
@@ -210,20 +227,81 @@ func (r *StatusClient) GetPgControlDataFromInstance(
...
+	availableArchitecture *utils.AvailableArchitecture,
+) error {
+	contextLogger := log.FromContext(ctx)
+
+	binaryFileStream, err := availableArchitecture.FileStream()
+	if err != nil {
+		return err
+	}
+	defer func() {
+		if binaryErr := binaryFileStream.Close(); binaryErr != nil {
```
armru, line 243: we now assign a new variable to avoid overwriting the returned error

pkg/resources/instance/client.go
```diff
@@ -210,20 +227,81 @@ func (r *StatusClient) GetPgControlDataFromInstance(
...
+	if isEOF(err) {
+		return nil
+	}
+	if err != nil {
+		return err
+	}
+
+	if resp.StatusCode == http.StatusOK {
+		// Currently the instance manager should never return StatusOK
+		return errors.New("instance manager has returned an unexpected status code")
```
armru, line 268: instead of returning nil I made the function return an error, given that status is in fact an unexpected return value

pkg/resources/instance/client.go
```diff
@@ -210,20 +227,81 @@ func (r *StatusClient) GetPgControlDataFromInstance(
...
+	req, err := http.NewRequestWithContext(ctx, http.MethodPut, updateURL, nil)
+	if err != nil {
+		return err
+	}
+	req.Body = binaryFileStream
+
+	resp, err := r.Client.Do(req)
+	// This is the desired response. The instance manager will
+	// synchronously update and this call won't return.
+	if isEOF(err) {
```
armru, line 259: I still think that is weird that the expected function outcome is not at the end, probably I will refactor this further

pkg/management/postgres/instance.go
```diff
@@ -199,6 +200,11 @@ type Instance struct {
 
 	// tablespaceSynchronizerChan is used to send tablespace configuration to the tablespace synchronizer
 	tablespaceSynchronizerChan chan map[string]apiv1.TablespaceConfiguration
+
+	StatusTLS bool
```
mnencia, line 204:
```suggestion
  // StatusTLS enables TLS on the status port used to communicate with the operator
	StatusTLS bool
```

mnencia:
E2E tests running on EDB fork https://github.com/EnterpriseDB/cloudnative-pg/actions/runs/9192103205

controllers/cluster_upgrade.go
mnencia: This is solved by configuring the timeout before executing the request.

pkg/management/postgres/instance.go
```diff
@@ -199,6 +200,12 @@ type Instance struct {
 
 	// tablespaceSynchronizerChan is used to send tablespace configuration to the tablespace synchronizer
 	tablespaceSynchronizerChan chan map[string]apiv1.TablespaceConfiguration
+
+	// StatusTLS enables TLS on the status port used to communicate with the operator
+	StatusTLS bool
```
jsilvela, line 205: IMO since "status" is so overloaded already, this would be better clarified.
How about `StatusPortTLS`?

internal/cmd/manager/instance/run/cmd.go
```diff
@@ -105,7 +107,8 @@ func NewCmd() *cobra.Command {
 		"current cluster in k8s, used to coordinate switchover and failover")
 	cmd.Flags().StringVar(&namespace, "namespace", os.Getenv("NAMESPACE"), "The namespace of "+
 		"the cluster and of the Pod in k8s")
-
+	cmd.Flags().BoolVar(&tlsStatus, "tls-status", false,
```
jsilvela, line 110: IMO `tls-status` makes it seem like "status of the TLS".
Like my other comment, I'd suggest `status-port-tls` or `status-via-tls` or `enable-status-via-tls`

pkg/specs/pods.go
```diff
@@ -381,7 +389,7 @@ func PodWithExistingStorage(cluster apiv1.Cluster, nodeSerial int) *corev1.Pod {
 
 	envConfig := CreatePodEnvConfig(cluster, podName)
 
-	podSpec := CreateClusterPodSpec(podName, cluster, envConfig, gracePeriod)
+	podSpec := CreateClusterPodSpec(podName, cluster, envConfig, gracePeriod, true)
```
jsilvela, line 392: This is the key entry point. I think a boolean here obscures things.
Might be worth creating a public type here.

jsilvela: E.g.
```
type TLScommunication bool

const (
	TLSDisabled TLScommunication = false
	TLSEnabled  TLScommunication = true
)

...
func PodWithExistingStorage(cluster apiv1.Cluster, nodeSerial int) *corev1.Pod {

	podSpec := CreateClusterPodSpec(podName, cluster, envConfig, gracePeriod, TLSEnabled)
```

jsilvela (commented):
It seems we're always going to try to use HTTPS.
See
``` go
podSpec := CreateClusterPodSpec(podName, cluster, envConfig, gracePeriod, true)
```
I thought this was supposed to be an option, and possibly and opt-in type of thing?

internal/cmd/manager/instance/run/cmd.go
mnencia: I would use `--status-port-tls`

pkg/specs/pods.go
mnencia: This function is called only in two places. Here, it must always be true, as you cannot opt out of TLS for new pods. It is false only if we compare the generated spec with the spec of an old pod where an old version of the operator is not TLS capable.

Would it be clearer if we create two function

```
func CreateClusterPodSpec(podName, cluster, envConfig, gracePeriod) ...
func CreateClusterPodSpecTLS(podName, cluster, envConfig, gracePeriod, tlsEnabled) ...
```

mnencia:
> It seems we're always going to try to use HTTPS.

That's correct. The conditional part is only for supporting Online Upgrades.

pkg/specs/pods.go
jsilvela: Now that I understand better, I don't think two functions are necessary.
I do think boolean args, sent as "true" always leave you confused, esp with multiple args. What is true?
For the invocation in `PodWithExistingStorage` we could add a var
```
    podSpec := CreateClusterPodSpec(podName, cluster, envConfig, gracePeriod, tlsEnabled)
```
Or, add a comment.

docs/src/security.md
```diff
@@ -297,13 +297,13 @@ section of the Kubernetes documentation for further information.
...
-operator         | 8080         | metrics             | `metrics`           |  no TLS        | No
-instance manager | 9187         | metrics             | `metrics`           |  no TLS        | No
-instance manager | 8000         | status              | `status`            |  no TLS        | No
-operand          | 5432         | PostgreSQL instance | `postgresql`        |  optional TLS  | Yes
+System           | Port number  | Exposing            |  Name               |  Certificates                  |  Authentication
+:--------------- | :----------- | :------------------ | :------------------ | :----------------------------  | :--------------
+operator         | 9443         | webhook server      | `webhook-server`    |  TLS                           | Yes
+operator         | 8080         | metrics             | `metrics`           |  no TLS                        | No
+instance manager | 9187         | metrics             | `metrics`           |  no TLS                        | No
+instance manager | 8000         | status              | `status`            |  TLS (no TLS in old releases)  | Yes
```
mnencia, line 305: Authentication is still no. We verify the server part, but the client side is not authenticated.

jsilvela: 🤦

jsilvela (approved):

litaocdl:
run e2e here: https://github.com/EnterpriseDB/cloudnative-pg/actions/runs/9281563495

mnencia:
I don't know if we want to backport this pull request. I would like to have an additional review from @fcanovai and know his opinion.

litaocdl (approved):

litaocdl (approved):

mnencia:
In today's team meeting, we agreed not to backport this feature. It will be part of 1.24

mnencia:
Running another round of tests after the rebase with conflicts

armru (approved):

PR #4494, opened 2024-05-09
chore: admit general PG booleans for configuration

### LINKED ISSUE #4252: [Feature]: Enahance parsing of postgres configuration booleans in webhooks
From the Postgres docs about the config file:
> Boolean: Values can be written as on, off, true, false, yes, no, 1, 0 (all case-insensitive) or any unambiguous prefix of one of these.

In webhooks, we compare the `archive_mode` and `wal_log_hints` parameter values to "off".

That means that a user could set it to "0", "no", or even "OFF" and go ahead with a broken configuration.

_Originally posted by @mnencia in https://github.com/cloudnative-pg/cloudnative-pg/pull/4218#discussion_r1554892367_

jsilvela:
Allows any valid Postgres boolean for configuration in the webhooks.
Currently, only "off" and "on" are properly understood by the webhooks.

Closes #4252


mnencia (changes requested):
I don't think it's correct to have two different functions (IsTrue and IsFalse) to parse a Postgres boolean.

We should have a single function `func ParsePostgresBoolean(in string) (bool, error)` and then manage the parse errors in the webhook.

Also, the PostgreSQL documentation says "Values can be written as on, off, true, false, yes, no, 1, 0 (all case-insensitive) or any unambiguous prefix of one of these.", and this implementation doesn't recognise "unambiguous prefixes".

jsilvela:
What is an "unambiguous prefix"? I've been looking for explanations or examples and found none.

jsilvela:
Because strictly speaking, an unambiguous prefix could be "fa", or "of". Is that something we really want?

mnencia:
> Because strictly speaking, an unambiguous prefix could be "fa", or "of". Is that something we really want?

Postgres does it https://doxygen.postgresql.org/bool_8c_source.html#l00036

jsilvela:
Realized, `archive_mode` is not a boolean but an enum. Leaving that one as was.

api/v1/cluster_webhook.go
```diff
@@ -2398,14 +2411,27 @@ func (r *Cluster) validatePgFailoverSlots() field.ErrorList {
...
+	if hasHotStandbyFeedback {
+		var err error
+		hotStandbyFeedbackActivated, err = postgres.ParsePostgresBoolean(hotStandbyFeedback)
+		if err != nil {
+			result = append(
+				result,
+				field.Invalid(
+					field.NewPath("spec", "postgresql", "parameters", hotStandbyFeedbackKey),
+					hotStandbyFeedback,
+					fmt.Sprintf("invalid %s value. Must be a postgres boolean", hotStandbyFeedbackKey)))
```
gabriele-wolfox, line 2424: I would suggest to change with 
`fmt.Sprintf("invalid `%s` value. Must be a postgres boolean", hotStandbyFeedbackKey)))`

So that the log format looks similar to the one at line 1208

gabriele-wolfox:
Beside my suggestion, PR looks good to me.

gabriele-wolfox (approved):

mnencia:
> Realized, `archive_mode` is not a boolean but an enum. Leaving that one as was.

You are right. Archive mode can only be `off`, `on` and `always`, so we can safely compare it to `off`.

mnencia (approved):

pkg/postgres/booleans.go
```diff
@@ -0,0 +1,44 @@
...
+	sanitized = strings.ToLower(sanitized)
+
+	if slices.Contains(falseValues, sanitized) {
+		return false, nil
+	}
+	if slices.Contains(trueValues, sanitized) {
+		return true, nil
+	}
+	return false, fmt.Errorf("configuration value is not a postgres boolean: %s", in)
+}
```
mnencia, line 44: What about this implementation?

```suggestion
func ParsePostgresBoolean(in string) (bool, error) {
	booleans := map[string]bool{
		"1":     true,
		"on":    true,
		"y":     true,
		"ye":    true,
		"yes":   true,
		"t":     true,
		"tr":    true,
		"tru":   true,
		"true":  true,
		"0":     false,
		"n":     false,
		"no":    false,
		"of":    false,
		"off":   false,
		"f":     false,
		"fa":    false,
		"fal":   false,
		"fals":  false,
		"false": false,
	}

	sanitized := strings.TrimSpace(in)
	sanitized = strings.ToLower(sanitized)

	if val, ok := booleans[sanitized]; ok {
		return val, nil
	}

	return false, fmt.Errorf("configuration value is not a postgres boolean: %s", in)
}
```

I find it easier to read.

api/v1/cluster_webhook.go
mnencia: Applied

pkg/postgres/booleans.go
gabriele-wolfox: I prefer it, as you say it improve the readability.
+1

armru: In my opinion, we should only accept two types and not try to do more, even if Postgres accepts them. This is because, in the end, they express only two values.
I propose that we accept only "true" "false" or "on" and "off" and we enable a strict validation on the webhook

mnencia: Accepting only `on` and off` could surprise the user. Suppose you have set `hot_standby_feedback` to `yes` in your configuration, and it works because you don't have `pg_failover_slots` enabled.
You enable the pg_failover_slots extension, and the validation kicks in, rejecting a parameter that was valid before.

jsilvela: I don't think this is an improvement. In the original, you can easily see the listing of "truthy" and "falsy" values.
With the map, you need to read the column on the right.
But, a `switch` statement would work well:
```suggestion
func ParsePostgresBoolean(in string) (bool, error) {
	sanitized := strings.TrimSpace(in)
	sanitized = strings.ToLower(sanitized)

	switch sanitized {
	case "1", "on", "yes", "true", "y", "t", "ye", "tr", "tru":
		return true, nil
	case "0", "no", "off", "false", "n", "f", "of", "fa", "fal", "fals":
		return false, nil
	default:
		return false, fmt.Errorf("configuration value is not a postgres boolean: %s", in)
	}
}
```

pkg/postgres/booleans_test.go
```diff
@@ -0,0 +1,49 @@
...
+*/
+
+package postgres
+
+import (
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+var _ = DescribeTable("Moar PostgreSQL booleans parsing",
```
jsilvela, line 24:
```suggestion
var _ = DescribeTable("Test PostgreSQL booleans parsing",
```

pkg/postgres/booleans_test.go
```diff
@@ -0,0 +1,49 @@
...
+*/
+
+package postgres
+
+import (
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+var _ = DescribeTable("Test PostgreSQL booleans parsing",
```
jsilvela, line 24:
```suggestion
var _ = DescribeTable("Test parsing of PostgreSQL booleans",
```

leonardoce (approved):

Issue #4499, opened 2024-05-09
[Feature]: For each metric to be collected define a mechanism to check if the query needs to be executed or not 


gabrosys:
### Is there an existing issue already for this feature request/idea?

- [X] I have searched for an existing issue, and could not find anything. I believe this is a new feature request to be evaluated.

### What problem is this feature going to solve? Why should it be added?

Postgres is a database that is prone to be extended via plugins and extensions. Would be great to define metrics that will be collected only if certain conditions in the database are verified. 
Some scenario where this could be applied:
- Extension dependent queries.
- Extension version-dependent queries.
- Pg flavour dependent queries (new col in the system view, new func added, new system view added etc).
- Enabling queries dynamically when a customer installs an optional extension.
- Supporting changing the PGD database name since we can detect the extension instead of relying on a pattern matching the DB name.
- Metrics that are conditional on runtime state.scenarios

### Describe the solution you'd like

Implement a new property along the `query` that is named `predicate_query`. This property is used to execute a query in the same context as the one to collect the metrics. 
The logic applied should be something like:
- `predicate_query` empty -> `true` executes the query to collect the metrics.
- `predicate_query` not empty -> execute the query and evaluate the result. Behave accordingly to execute or not the query for the metrics.

### Describe alternatives you've considered

Update the `query` property definition to deliver an array of queries where an error on the first blocks the executions of the next.

### Additional context

The need for this feature is also related to an issue with PG that does not allow execution in a single statement of more than one query. Also, using the `DO` we are not able to get in return a table with a single statement and so we need to end up with something that executes another query to achieve what we need.

### Backport?

N/A

### Are you willing to actively contribute to this feature?

Yes

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #4503, opened 2024-05-09
feat: implement predicate_query to validate which are the metrics to be collected

gabrosys:
### What

Implement a validation logic to verify if a query to collect metrics should be executed or not.

- ref: #4499 

### How

- Define a property `predicate_query`.
- Define a method `isCollectable` that is invoked by `collect` passing the same `tx` (to reduce the performance consumption) that returns a `bool`.
- Add a new e2e test, with related files, to check the scenarios of `predicate_query`.

### Test

- Execute the `e2e-test-local` with `FEATURE_TYPE= observability` and `TEST_DEPTH=3`.

### Open points - Questions

- Do you think there is a need to add any specific log when a query is not executed due to the `false` response of the `predicate_query`? I want to avoid bloating PG with this kind of log so this is why currently it is not present.
- `isCollectable` is intended to be invoked only within the `collect` method, maybe I could define it as anonymous method inside the `collect`?
- We can simplify a lot the logic in the `isCollectable` method if we assume strong and tight rules regarding the output of the `predicate_query`, what do you think about that?
- Do I have to update the doc within this PR or, when approved, I will create another PR to document what is and it is working the `predicate_query`?

Thanks at disposal for any question 🙇 🚀


pkg/management/postgres/metrics/collector.go
```diff
@@ -378,6 +387,66 @@ func (c QueryCollector) collect(conn *sql.DB, ch chan<- prometheus.Metric) error
 	return nil
 }
 
+// isCollectable checks if a query to collect metrics can be executed or not.
```
armru, line 390: Is there any reason why we are accepting multiple types?
In my opinion, we should only evaluate booleans, and it should be the responsibility of the query to return the intended type

leonardoce: I agree with Armando. It's better to rely on booleans to do that.
True means it's collectible, false means it's not, and NULL means we don't know.
And if we don't know, perhaps it's better not to collect it.

pkg/management/postgres/metrics/collector.go
```diff
@@ -22,11 +22,11 @@ import (
...
 	"database/sql"
 	"fmt"
-	"path"
-	"regexp"
-
 	"github.com/blang/semver"
 	"github.com/prometheus/client_golang/prometheus"
+	"path"
+	"regexp"
+	"strings"
```
leonardoce, line 29: I'd leave the import order as it was. This will make our linter fail.

pkg/management/postgres/metrics/collector.go
```diff
@@ -22,11 +22,11 @@ import (
...
 
 	"github.com/cloudnative-pg/cloudnative-pg/pkg/management/log"
 	"github.com/cloudnative-pg/cloudnative-pg/pkg/management/postgres"
```
mnencia, line 32: Please run the linter with `make checks`. This hunk causes the linter to fail due to the import order.

pkg/management/postgres/metrics/collector.go
gabrosys: So to summarize we want to support ONLY bools so that even numbers are not tolerated. Also, do we want to force the developers to deliver a query that returns ONLY one column? If so we can simplify the whole logic with a single `QueryRow` and a `Scan` that accepts a bool variable.

armru: Yes, I think evaluating a bool should be enough

pkg/management/postgres/metrics/parser.go
```diff
@@ -29,6 +29,7 @@ type UserQueries map[string]UserQuery
 // UserQuery represent a query created by the user
 type UserQuery struct {
 	Query           string    `yaml:"query"`
+	PredicateQuery  string    `yaml:"predicate_query"`
```
armru, line 32: Perhaps a more explicit name would be better?
writing this to gather ideas

pkg/management/postgres/metrics/collector.go
gabrosys: Update the comment and the method accordingly. Now assume that the type is a `bool` if not we return `false` and log a warning.

pkg/management/postgres/metrics/parser.go
ringerc: To me "predicate query" is the right term - it's a query that is used as a predicate to determine whether the main query should run.

Suggestions are welcome. `ConditonalExecutionQuery` ? IDK.

ringerc:
> Do you think there is a need to add any specific log when a query is not executed due to the false response of the `predicate_query`? I want to avoid bloating PG with this kind of log so this is why currently it is not present.

No, it definitely should not log, or not at above `debug` level. That's part of the point, so unwanted log spam can be avoided when the target server doesn't support a specific view, column, function etc.

In general there is IMO a need to improve how CNP's metrics scraper diagnostics work and make it easier to understand its behaviour + detect errors. That's why I wrote a little tool to run a one-shot scrape against a pre-configured postgres recently. I'll see if I can tidy that up for a community contribution. But logging isn't IMO the answer to that. I already have enough problems with it quietly logging a complaint and happily continuing with some confusing behaviour.

It might make sense to emit metrics as scrape output to report on skips, but they should probably be optional, and I don't personally think they're necessary as part of the basic feature. They could easily bloat scrape results and add unwanted scrape cardinality, requiring more filtering by scrape result consumers to discard the expected ones.

> We can simplify a lot the logic in the isCollectable method if we assume strong and tight rules regarding the output of the predicate_query, what do you think about that?

The problem with that is that the user has to reliably assume the query has the expected results under all circumstances. Part of the point of this is to make the CNP scraper robust in the face of differing server versions, extension installs, vendor flavours, etc etc.

OTOH if the query does something unexpected, a clear error log might be better than just silently returning false and not running the query. IDK. Right now it's hard to properly test CNP scraper configs across a wide matrix of possible server configs, so I prefer to be error-tolerant, but I can see the argument for strictness too.

I'd personally prefer it to accept a 0-row result as false, but it's not hard to wrap that in a `SELECT EXISTS (...)` if you want to require a strict boolean result.

pkg/management/postgres/metrics/collector.go
gabrosys: The updated code handles:
- `bool` checking the first row first column.
- no rows, returning `false`.
- null value, returning `false`.

pkg/management/postgres/metrics/parser.go
gabrosys: `predicate_query` is good to me however feel free to propose the name that you think fits better the cnp config 😄

ringerc: @armru Would you be happier with `ConditionQuery` or something? Suggestions welcome if you don't like the current naming.

armru: We can leave `PredicateQuery` then :) I was opening a discussion to gather ideas

ringerc:
@mnencia What can we do to progress this PR?

ringerc:
@gabrosys I see a test failure

>    [FAIL] Metrics [It] can gather metrics according with predicate query [observability]
>  /home/runner/work/cloudnative-pg/cloudnative-pg/tests/e2e/asserts_test.go:2926

https://github.com/cloudnative-pg/cloudnative-pg/actions/runs/9131921366/job/25112487661#step:9:5716

It failed in some of the other runs too, but passed in others.

Details https://github.com/cloudnative-pg/cloudnative-pg/actions/runs/9131921366/job/25112490237#step:9:2344

```
   Timeline >>
  Putting Tail on the operator log
  STEP: verifying the custom metrics ConfigMaps and Secrets exist @ 05/17/24 18:30:22.086
  STEP: setting up curl client pod @ 05/17/24 18:30:22.245
  STEP: having a predicate-query-metrics-e2e-9171 namespace @ 05/17/24 18:30:27.285
  STEP: creating a Cluster in the predicate-query-metrics-e2e-9171 namespace @ 05/17/24 18:30:27.291
  STEP: having a Cluster postgresql-metrics with each instance in status ready @ 05/17/24 18:30:27.335
  Cluster ready, took 1m12.298617472s
  STEP: ensuring metrics with positive predicate are collected @ 05/17/24 18:31:39.665
  STEP: checking metrics for pod: postgresql-metrics-1 @ 05/17/24 18:31:39.672
  [FAILED] in [It] - /home/runner/work/cloudnative-pg/cloudnative-pg/tests/e2e/asserts_test.go:2926 @ 05/17/24 18:31:40.062
  DUMPING tailed Operator Logs with error/warning (at most 10 lines ). Failed Spec: Metrics can gather metrics according with predicate query

  ================================================================================
  -- no error / warning logs --
  ================================================================================
  << Timeline

  [FAILED] Found no match for metric cnpg_pg_predicate_query_return_true_and_multiple_rows_fixed

  Priting rawMetricsOutput:
...
  cnpg_errors_total{errorUserQueries="pg_predicate_query_return_true_and_multiple_rows on db app: ERROR: subquery in FROM must have an alias (SQLSTATE 42601)"} 1
...
  # HELP cnpg_pg_predicate_query_return_true_and_multiple_columns_fixed Always 42, used to test predicate_query
  # TYPE cnpg_pg_predicate_query_return_true_and_multiple_columns_fixed gauge
  cnpg_pg_predicate_query_return_true_and_multiple_columns_fixed 42
  # HELP cnpg_pg_predicate_query_return_true_and_multiple_columns_multiple_rows_fixed Always 42, used to test predicate_query
  # TYPE cnpg_pg_predicate_query_return_true_and_multiple_columns_multiple_rows_fixed gauge
  cnpg_pg_predicate_query_return_true_and_multiple_columns_multiple_rows_fixed 42
  # HELP cnpg_pg_predicate_query_return_true_fixed Always 42, used to test predicate_query
  # TYPE cnpg_pg_predicate_query_return_true_fixed gauge
  cnpg_pg_predicate_query_return_true_fixed 42
```

so it looks like an issue with the spec for `pg_predicate_query_return_true_and_multiple_rows`

tests/e2e/fixtures/metrics/custom-queries-with-predicate-query.yaml
```diff
@@ -0,0 +1,89 @@
...
+      query: "SELECT 42 as fixed"
+      predicate_query: "SELECT false as result, 0 as column_one, 'something' as column_two"
+      primary: false
+      metrics:
+        - fixed:
+            usage: "GAUGE"
+            description: "Always 42, used to test predicate_query"
+    pg_predicate_query_return_true_and_multiple_rows:
+      query: "SELECT 42 as fixed"
+      predicate_query: "SELECT result FROM (SELECT true as result UNION SELECT false as result) order by result desc"
```
ringerc, line 60: This fails with ` ERROR: subquery in FROM must have an alias (SQLSTATE 42601)`

It could be formulated as

```
SELECT result FROM (VALUES (true), (false)) AS r(result) ORDER BY 1  DESC;
```

which produces

```
14/main craig@craig > SELECT result FROM (VALUES (true), (false)) AS r(result) ORDER BY 1 DESC;
 result 
--------
 t
 f
(2 rows)
```

ringerc (commented):
Identified query bug causing test failure

tests/e2e/fixtures/metrics/custom-queries-with-predicate-query.yaml
gabrosys: Thanks a lot for your suggestion, I will update the PR according!

gabrosys: Solved!

pkg/management/postgres/metrics/collector.go
```diff
@@ -378,6 +387,65 @@ func (c QueryCollector) collect(conn *sql.DB, ch chan<- prometheus.Metric) error
...
+func (c QueryCollector) isCollectable(tx *sql.Tx) (bool, error) {
+	if c.userQuery.PredicateQuery == "" {
+		return true, nil
+	}
+
+	rows, err := tx.Query(c.userQuery.PredicateQuery)
+	if err != nil {
+		return false, err
+	}
+
```
jsilvela, line 403: was wondering why you're using `Query` rather than `QueryRow`, and I think it's because you want to query for `Columns()` below... but.
Is it worth it?

pkg/management/postgres/metrics/collector.go
```diff
@@ -378,6 +387,65 @@ func (c QueryCollector) collect(conn *sql.DB, ch chan<- prometheus.Metric) error
...
+		return false, nil
+	}
+
+	result, ok := columnData[0].(bool)
+	if !ok {
+		log.Warning("Unsupported column type",
+			"predicate_query", c.userQuery.PredicateQuery,
+			"type", fmt.Sprintf("%T", result))
+		return false, nil
+	}
```
jsilvela, line 444: why not use `QueryRow` and scan into a boolean, and then the driver will let us know the first argument on the firs row is not a predicate?

jsilvela (commented):
couldn't `isCollectable` be substantially simplified?

pkg/management/postgres/metrics/collector.go
jsilvela: I don't think it's too much to as, given a field called `predicate_query`, that the query should be a predicate.

pkg/management/postgres/metrics/collector.go
gabrosys: Yep, you're right! I think there is something related to what we want to achieve. As said above we can assume that the predicate is one single row and deliver the `QueryRow` or be open on what the query provides and just have a look at the first row in the first column.

gabrosys:
Update the PR as requested. Now we accept only `predicate_query` with at most one row and with a single `bool` column.
In detail, we handle:
- `true` and `false`
- no row as `false`
- null as `false`

@jsilvela ready to be reviewed, thanks 🙇

jsilvela (approved):
nice work, Gabriele

armru (approved):

armru:
Hello @gabrosys could you please integrate the `User defined metrics` doc section with the new field and explain how to use it?

docs/src/monitoring.md
```diff
@@ -546,6 +546,8 @@ Here is a short description of all the available fields:
     - `target_databases`: a list of databases to run the `query` against,
       or a [shell-like pattern](#example-of-a-user-defined-metric-running-on-multiple-databases)
       to enable auto discovery. Overwrites the default database if provided.
+    - `predicate_query`: a SQL query that returns at most one row and one `boolean` column to run on the target database.
+       The system evaluates the predicate and if `true` executes the `query`.
```
gabrosys, line 550: Good enough? Or I will also add a paragraph like the one for multiple databases with an example? @armru Thanks a lot in advance 🙇

gabrosys:
Update the doc with a new paragraph with a simple example and add the new prop to the prop list. Thank you all for your support 🙇

PR #4504, opened 2024-05-09
fix: timeout when restarting PostgreSQL and while lifting fencing

### LINKED ISSUE #4501: [Bug]: Instance manager reconciler loop stuck if PostgreSQL restart fails

### Contact Details

### Version

1.23.0

1.30 (unsuppprted)

YAML manifest

### What happened?

The instance manager reconciler loop can be blocked by a PostgreSQL that fails to start up after being shut down.

### Cluster resource

_No response_

### Relevant log output

_No response_

### Code of Conduct

leonardoce:
The instance manager starts PostgreSQL:

1. when it starts up
2. when configuration changes are being applied (after stopping it)
3. when fencing is lifted.

In the second and third examples, the operator is requested by the embedded cluster reconciler loop, and performed without any timeout.

If PostgreSQL won't start up again because of a wrong configuration or missing disk space, the reconciler loop will be stuck waiting for a dead postmaster to be up.

This patch handles this condition by using a combination of the timeout parameters that are already set in the cluster.

Fixes: #4501

### DIFF
--- a/api/v1/cluster_types.go
+++ b/api/v1/cluster_types.go
@@ -2939,6 +2939,12 @@ func (cluster *Cluster) GetSmartShutdownTimeout() int32 {
 	return 180
 }
 
+// GetRestartTimeout is used to have a timeout for operations that involve
+// a restart of a PostgreSQL instance
+func (cluster *Cluster) GetRestartTimeout() int32 {
+	return cluster.GetMaxStopDelay() + cluster.GetMaxStartDelay()
+}
+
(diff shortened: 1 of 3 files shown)


leonardoce:
Hi! This one seems harmless and needed, but I encourage you to take a deep look.

After much thinking, I still didn't make up my mind about whether it was better to fix it or leave it as it is and employ the energies on refactoring, to make the code flow clearer.

armru (approved):

mnencia:
/ok-to-merge E2E green. Only expected unrelated failures.

mnencia (approved):

Issue #2337, opened 2023-06-21
recover through max_connection parameter change


smiyc:
Operator version: 1.20.1
K8s version: 1.23.12 (PKS 1.14.3-build.16)

Description:
I started a PITR recover which failed because I changed the parameter max_connections inbetween the base backup and the time i wanted to recover to.
I was not able to reproduce the error on an 1.20.0 Operator Cluster.

Slack Link:
https://cloudnativepg.slack.com/archives/C03AX0J5P29/p1687160914284189

Error:
> {
  "level": "info",
  "ts": "2023-06-19T07:40:39Z",
  "logger": "postgres",
  "msg": "record",
  "logging_pod": "keycloak-test-1-full-recovery",
  "record": {
    "log_time": "2023-06-19 07:40:39.991 UTC",
    "process_id": "40",
    "session_id": "64900673.28",
    "session_line_num": "11",
    "session_start_time": "2023-06-19 07:40:35 UTC",
    "virtual_transaction_id": "1/0",
    "transaction_id": "0",
    "error_severity": "WARNING",
    "sql_state_code": "22023",
    "message": "hot standby is not possible because of insufficient parameter settings",
    "detail": "max_connections = 200 is a lower setting than on the primary server, where its value was 300.",
    "context": "WAL redo at 8/8D0000A0 for XLOG/PARAMETER_CHANGE: max_connections=300 max_worker_processes=32 max_wal_senders=10 max_prepared_xacts=0 max_locks_per_xact=64 wal_level=logical wal_log_hints=on track_commit_timestamp=off",
    "backend_type": "startup",
    "query_id": "0"
  }
}
{
  "level": "info",
  "ts": "2023-06-19T07:40:39Z",
  "logger": "postgres",
  "msg": "record",
  "logging_pod": "keycloak-test-1-full-recovery",
  "record": {
    "log_time": "2023-06-19 07:40:39.991 UTC",
    "process_id": "40",
    "session_id": "64900673.28",
    "session_line_num": "12",
    "session_start_time": "2023-06-19 07:40:35 UTC",
    "virtual_transaction_id": "1/0",
    "transaction_id": "0",
    "error_severity": "LOG",
    "sql_state_code": "00000",
    "message": "recovery has paused",
    "detail": "If recovery is unpaused, the server will shut down.",
    "hint": "You can then restart the server after making the necessary configuration changes.",
[...]

PR #4564, opened 2024-05-17
fix: allow raising replication parameters when restoring a backup

### LINKED ISSUE #2478: Recovery failure with barman due to max_connections
CNPG version: 1.20.1

I'm trying to recover from a backup that was produced with barmanObjectStore but recovering the primary instance fails.
In postgresql.parameters.max_connections I use a value of 40, and because of that the following error is produced:
"message":"hot standby is not possible because of insufficient parameter settings","detail":"max_connections = 10 is a lower setting than on the primary server, where its value was 40."
Meaning that recovery from a backup must use at least as large max_connections as what was used by the backup.

Tracking down the code I figured that the max_connections=10 setting comes from pg_controldata([src](https://github.com/cloudnative-pg/cloudnative-pg/blob/c4a369fff121f4060db9c53df99bae73292b2caf/pkg/management/postgres/restore.go#L71)), which strictly overwrites the user setting.

I guess the user should be able to override max_connections when running a recovery to match that was backed up?

I've checked and /var/lib/postgresql/data/pgdata/custom.conf indeed shows max_connections=10.
What's confusing though is that running pg_controldata displays "max_connections setting: 40" so I'm confused where that 10 comes from.
The logs show this: {..., "msg":"enforcing parameters found in pg_controldata","parameters":{"max_connections":"10","max_locks_per_transaction":"64","max_prepared_transactions":"0","max_wal_senders":"10","max_worker_processes":"32"}}

litaocdl:
Ensure the PostgreSQL replication parameters are set to the higher value between
the ones specified in the cluster specification and the ones stored in the backup.

This will ensure that the backup will be restored correctly while allowing the users
to raise their value to accommodate changes in the configuration that have
happened after the backup was taken.

Partially closes #2478 #2337

### DIFF
--- a/internal/management/controller/instance_controller.go
+++ b/internal/management/controller/instance_controller.go
@@ -394,30 +394,25 @@ func (r *InstanceReconciler) verifyParametersForFollower(cluster *apiv1.Cluster)
 		return err
 	}
 	log.Info("Found previous run flag", "filename", filename)
-	enforcedParams, err := postgresManagement.GetEnforcedParametersThroughPgControldata(r.instance.PgData)
+	controldataParams, err := postgresManagement.LoadEnforcedParametersFromPgControldata(r.instance.PgData)
+	if err != nil {
+		return err
+	}
+	clusterParams, err := postgresManagement.LoadEnforcedParametersFromCluster(cluster)
 	if err != nil {
 		return err
 	}
 
-	clusterParams := cluster.Spec.PostgresConfiguration.Parameters
 	options := make(map[string]string)
-	for key, enforcedparam := range enforcedParams {
+	for key, enforcedparam := range controldataParams {
 		clusterparam, found := clusterParams[key]
 		if !found {
 			continue
 		}
-		enforcedparamInt, err := strconv.Atoi(enforcedparam)
-		if err != nil {
-			return err
-		}
-		clusterparamInt, err := strconv.Atoi(clusterparam)
-		if err != nil {
-			return err
-		}
 		// if the values from `pg_controldata` are higher than the cluster spec,
 		// they are the safer choice, so set them in config
-		if enforcedparamInt > clusterparamInt {
-			options[key] = enforcedparam
+		if enforcedparam > clusterparam {
+			options[key] = strconv.Itoa(enforcedparam)
 		}
 	}
 	if len(options) == 0 {
--- a/pkg/management/postgres/restore.go
+++ b/pkg/management/postgres/restore.go
@@ -634,7 +634,12 @@ func (info InitInfo) writeRecoveryConfiguration(cluster *apiv1.Cluster, recovery
 		enforcedParams,
 	)
 	if changed {
-		log.Info("enforcing parameters found in pg_controldata and cluster spec", "parameters", enforcedParams)
+		log.Info(
+			"Aligned PostgreSQL configuration to satisfy both pg_controldata and cluster spec",
+			"enforcedParams", enforcedParams,
+			"controldataParams", controldataParams,
+			"clusterParams", clusterParams,
+		)
 	}
 	if err != nil {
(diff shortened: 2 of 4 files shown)


litaocdl:
e2e is running here https://github.com/EnterpriseDB/cloudnative-pg/actions/runs/9136158988

mnencia:
E2E test running on the fork https://github.com/EnterpriseDB/cloudnative-pg/actions/runs/9251317637

mnencia (approved):

YanniHu1996 (approved):

pkg/management/postgres/restore.go
```diff
@@ -601,22 +602,31 @@ func (info InitInfo) writeRecoveryConfiguration(recoveryFileContents string) err
 		return fmt.Errorf("cannot write recovery config: %w", err)
 	}
 
-	enforcedParams, err := GetEnforcedParametersThroughPgControldata(info.PgData)
+	controldataParams, err := LoadEnforcedParametersFromPgControldata(info.PgData)
```
leonardoce, line 605:
```suggestion
	// Now we need to choose which parameters to use to complete the recovery
	// of this PostgreSQL instance.
	// We know the values that these parameters had when the backup was started
	// from the `pg_controldata` output.
	// We don't know how these values were set in the newer WALs.
	//
	// The only way to proceed is to rely on the user-defined configuration,
	// with the caveat of ensuring that the values are high enough to be
	// able to start recovering the backup.
	//
	// To be on the safe side, we'll use the largest setting we find
	// from `pg_controldata` and the Cluster definition.
	//
	// https://www.postgresql.org/docs/16/hot-standby.html#HOT-STANDBY-ADMIN
	controldataParams, err := LoadEnforcedParametersFromPgControldata(info.PgData)
```

internal/management/controller/instance_controller.go
```diff
@@ -394,30 +394,25 @@ func (r *InstanceReconciler) verifyParametersForFollower(cluster *apiv1.Cluster)
...
+	}
+	clusterParams, err := postgresManagement.LoadEnforcedParametersFromCluster(cluster)
 	if err != nil {
 		return err
 	}
 
-	clusterParams := cluster.Spec.PostgresConfiguration.Parameters
 	options := make(map[string]string)
-	for key, enforcedparam := range enforcedParams {
+	for key, enforcedparam := range controldataParams {
```
leonardoce, line 407: Isn't this logic exactly the same as what we have here? https://github.com/cloudnative-pg/cloudnative-pg/pull/4564/files#diff-7b2e81429f27e9e75a48e70527067a4d8a80a2450dab880b4b228d501359b7a8R688

If it is, better encapsulate and use it two times.

leonardoce (changes requested):

internal/management/controller/instance_controller.go
litaocdl: The one in instance mgr, we only refresh if pgcontrol is higher than user specific, make sure the standby has a higher value.  In restore, we use the higher value between use defined and pgcontrol.

leonardoce: I think it is the same, as you're always going to get a value from `controldataParams[X]`, and that value will always be greater than `0`.

litaocdl: you are right, I misunderstood

leonardoce: Yep, perhaps it was better for me to file this comment against the other occurrence of that cycle.

leonardoce: Let's leave them as they are, refactoring it now is way too much pain for the advantage we get.

leonardoce (approved):

leonardoce:
https://github.com/EnterpriseDB/cloudnative-pg/actions/runs/9315905824

PR #4567, opened 2024-05-17
ci: steps containing retriable code shouldn't terminate on a non-zero exit status

NiccoloFei:
Where possible, if the step is simple enough, I've started using `nick-fields/retry` instead of attempting new retries using the shell.

### DIFF
--- a/.github/workflows/continuous-delivery.yml
+++ b/.github/workflows/continuous-delivery.yml
@@ -1092,6 +1092,7 @@ jobs:
         name: Clean up
         if: always()
         run: |
+          set +e
           az extension add --allow-preview true --name monitor-control-service
           az account set --subscription ${{ secrets.AZURE_SUBSCRIPTION }}
           attempt=1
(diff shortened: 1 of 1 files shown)


.github/workflows/continuous-delivery.yml
```diff
@@ -1157,22 +1158,13 @@ jobs:
...
-                echo "Storage account deleted"
-                break
-              fi
-              echo "Failed deleting storage account ${{ env.AZURE_STORAGE_ACCOUNT }}, retrying"
-              sleep 5
-              attempt=$((attempt+1))
-          done
+        uses: nick-fields/retry@v3
+        with:
+          timeout_minutes: 1
```
mnencia, line 1163: Are you sure that this timeout is enough? The previous code did not constrain the execution time.

NiccoloFei: That code runs in about ~10 seconds (looking at some recent E2Es). I've increased the max timeout to 5 minutes though, which should be more than enough

sxd (approved):

mnencia (approved):

PR #4602, opened 2024-05-20
docs: add homebrew install method for kubectl plugin

Skaronator:
It's now possible to install the kubectl plugin with homebrew: https://formulae.brew.sh/formula/kubectl-cnpg

### DIFF
--- a/docs/src/kubectl-plugin.md
+++ b/docs/src/kubectl-plugin.md
@@ -98,6 +98,9 @@ kubectl krew upgrade cnpg
 
 ### Using Homebrew
 
+!!! Note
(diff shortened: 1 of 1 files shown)


Skaronator:
This should not be backported because only versions 1.23.1 and later are available on Homebrew.

mnencia:
Thank you for the PR. Having the plugin in Homebrew is a good thing. My only concern is that reading the docs, you could think that the CNPG developers are responsible for pushing new versions on Homebrew. Can you make it clear that publishing the plugin on Homebrew is not part of the CNPG release procedure?

Skaronator:
Certainly, I added a note block above it.

sxd (approved):

mnencia (approved):

PR #4609, opened 2024-05-20
test(fix): increase the short timeout

### LINKED ISSUE #4606: [Bug]: Tablespace e2e test sometimes goes timeout

### Contact Details

_No response_

### Version

1.23.1

1.30 (unsupported)

YAML manifest

### What happened?

Sometimes the tablespace tests (not all of them) will fail with a timeout like this:
```
[FAILED] Timed out after 120.000s.
  The function passed to Eventually failed at /home/runner/work/cloudnative-pg/cloudnative-pg/tests/e2e/tablespaces_test.go:921 with:
  Expected
      <[]string | len:7, cap:8>: [
          "/var/lib/postgresql/data",
          "/run",
          "/controller",
          "/dev/shm",
          "/etc/app-secret",
          "/var/lib/postgresql/wal",
          "/var/run/secrets/kubernetes.io/serviceaccount",
      ]
  to contain elements
      <[]string | len:1, cap:1>: [
          "/var/lib/postgresql/tablespaces/atablespace",
      ]
  the missing elements were
      <[]string | len:1, cap:1>: [
          "/var/lib/postgresql/tablespaces/atablespace",
      ]
  In [It] at: /home/runner/work/cloudnative-pg/cloudnative-pg/tests/e2e/tablespaces_test.go:943 @ 04/30/24 02:33:38.

```

### Cluster resource

_No response_

### Relevant log output

_No response_

### Code of Conduct

sxd:
The short timeout that is only used on tablespaces tests it was by default set to
5 seconds, this is a bit too short and it was producing some failures in some of
the cloud environment tests.

Also, we stop logging errors that aren't errors at all while getting the pod logs.

Closes #4606

### DIFF
--- a/.github/e2e_test_timeout.json
+++ b/.github/e2e_test_timeout.json
@@ -13,7 +13,8 @@
     "walsInMinio":               60,
     "minioInstallation":         300,
     "backupIsReady":             180,
-    "drainNode":                 900
+    "drainNode":                 900,
+    "short":                     5
   },
   "aks": {
     "failover":                  240,
--- a/.github/e2e_test_timeout.json
+++ b/.github/e2e_test_timeout.json
@@ -1,7 +1,7 @@
 {
   "local": {
-    "failover": 240,
-    "namespaceCreation": 30,
+    "failover":                  240,
+    "namespaceCreation":         30,
     "clusterIsReady":            600,
     "clusterIsReadyQuick":       300,
     "clusterIsReadySlow":        800,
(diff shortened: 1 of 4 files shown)


NiccoloFei (approved):

.github/e2e_test_timeout.json
```diff
@@ -16,8 +16,8 @@
     "drainNode":                 900
   },
```
mnencia, line 17:
```suggestion
    "drainNode":                 900,
    "short":                     5,
  },
```

mnencia: I think it's better to be explicit than rely on the default. As an alternative, we could use thedefault in all the "local" values.

sxd: I'll create a ticket to rebuild this entire logic, because currently it's something with little to no sense and repeated values =(

mnencia (approved):

PR #4647, opened 2024-05-23
test(fix): improve readability of some errors during upgrade E2E test

sxd:
When there is a failure extracting the `max_replication_slots` from the instance, the Atoi invocations will return an obscure "invalid syntax".
With this patch, we display a meaningful message and any message printed on standard error.

### DIFF
--- a/tests/e2e/upgrade_test.go
+++ b/tests/e2e/upgrade_test.go
@@ -188,18 +188,17 @@ var _ = Describe("Upgrade", Label(tests.LabelUpgrade, tests.LabelNoOpenshift), O
 			// Check that both parameters have been modified in each pod
 			for _, pod := range podList.Items {
 				pod := pod // pin the variable
-				Eventually(func() (int, error, error) {
-					stdout, _, err := env.ExecCommand(env.Ctx, pod, specs.PostgresContainerName, &commandTimeout,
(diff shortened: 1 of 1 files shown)


tests/e2e/upgrade_test.go
```diff
@@ -191,8 +191,15 @@ var _ = Describe("Upgrade", Label(tests.LabelUpgrade, tests.LabelNoOpenshift), O
 				Eventually(func() (int, error, error) {
 					stdout, _, err := env.ExecCommand(env.Ctx, pod, specs.PostgresContainerName, &commandTimeout,
 						"psql", "-U", "postgres", "-tAc", "show max_replication_slots")
-					value, atoiErr := strconv.Atoi(strings.Trim(stdout, "\n"))
-					return value, err, atoiErr
+					if err != nil {
+						return 0, err, nil
+					}
```
mnencia, line 196: The whole point of having multiple errors in the return is avoiding explicitly checking the errors.

If we handle the errors, we can return one.

leonardoce (approved):

mnencia:
There is no way `SHOW max_replication_slots` could return and empty result unless there is an error. I've changed the code to display the content of STDERR in case of empty result.

mnencia (approved):

PR #4648, opened 2024-05-23
fix: avoid reloading ident map when not needed

### LINKED ISSUE #4626: [Bug]: each times PostgreSQL starts, we SIGHUP it two times to change the pg_ident file

### Contact Details

leonardo.cecchi@enteprisedb.com

### Version

1.23.1

1.30 (unsupported)

YAML manifest

### What happened?

Everytime the instance manager starts PostgreSQL, it will SIGHUP it two times to install two slightly different versions of the pg_ident file.
This is easily reproducible by fencing on and off an instance.

### Cluster resource

_No response_

### Relevant log output

_No response_

### Code of Conduct

leonardoce:
Before this patch, the instance manager could write the ident map using two different formats even if the content was semantically the same.

This was triggering an unneeded configuration reload after the postmaster was started up.

Fixes: #4626

### DIFF
--- a/pkg/management/postgres/configuration.go
+++ b/pkg/management/postgres/configuration.go
@@ -198,16 +198,22 @@ func quoteHbaLiteral(literal string) string {
 	return fmt.Sprintf(`"%s"`, literal)
 }
 
-// GeneratePostgresqlIdent generates the pg_ident.conf content
-func (instance *Instance) GeneratePostgresqlIdent(cluster *apiv1.Cluster) (string, error) {
-	return postgres.CreateIdentRules(cluster.Spec.PostgresConfiguration.PgIdent,
-		getCurrentUserOrDefaultToInsecureMapping())
+// generatePostgresqlIdent generates the pg_ident.conf content given
+// a set of additional pg_ident lines that is usually taken from the
+// Cluster configuration
+func (instance *Instance) generatePostgresqlIdent(additionalLines []string) (string, error) {
+	return postgres.CreateIdentRules(
+		additionalLines,
+		getCurrentUserOrDefaultToInsecureMapping(),
+	)
 }
 
-// RefreshPGIdent generates and writes down the pg_ident.conf file
-func (instance *Instance) RefreshPGIdent(cluster *apiv1.Cluster) (postgresIdentChanged bool, err error) {
+// RefreshPGIdent generates and writes down the pg_ident.conf file given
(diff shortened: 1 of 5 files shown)


pkg/management/postgres/restore.go
```diff
@@ -715,36 +715,36 @@ func (info InitInfo) WriteInitialPostgresqlConf(cluster *apiv1.Cluster) error {
...
 	if err != nil {
-		return fmt.Errorf("while reading configuration files from ConfigMap: %w", err)
+		return fmt.Errorf("while generating pg_hba.conf: %w", err)
 	}
-	_, err = temporaryInstance.RefreshPGIdent(cluster)
+	_, err = temporaryInstance.RefreshPGIdent(cluster.Spec.PostgresConfiguration.PgIdent)
 	if err != nil {
-		return fmt.Errorf("while reading configuration files from ConfigMap: %w", err)
+		return fmt.Errorf("while generating pg_ident.conf: %w", err)
 	}
```
mnencia, line 723: This makes no sense; we generate these files, and then we don't install them.

We should either stop generating the pg_hba.conf and pg_ident.conf here or install them.

This deserves a separate issue.

mnencia (approved):

pkg/management/postgres/configuration.go
```diff
@@ -198,16 +198,22 @@ func quoteHbaLiteral(literal string) string {
...
+	)
 }
 
-// RefreshPGIdent generates and writes down the pg_ident.conf file
-func (instance *Instance) RefreshPGIdent(cluster *apiv1.Cluster) (postgresIdentChanged bool, err error) {
+// RefreshPGIdent generates and writes down the pg_ident.conf file given
+// a set of additional pg_ident lines that is usually taken from the
+// Cluster configuration
+func (instance *Instance) RefreshPGIdent(additionalLines []string) (postgresIdentChanged bool, err error) {
 	// Generate pg_hba.conf file
```
jsilvela, line 215:
```suggestion
	// Generate pg_ident.conf file
```

jsilvela (approved):

pkg/management/postgres/restore.go
leonardoce: Could you create it @mnencia ?

leonardoce:
E2e https://github.com/EnterpriseDB/cloudnative-pg/actions/runs/9266533929

PR #4650, opened 2024-05-23
test(e2e): fix backup failures for replica mode

### LINKED ISSUE #4646: [Bug]: Backup failure in Replica Mode E2Es

### Contact Details

_No response_

### Version

1.23.1

1.30 (unsupported)

YAML manifest

### What happened?

In `Context("can bootstrap a replica cluster from a backup")` we are creating two clusters with the same name, one for 
`It("using a Backup from the object store")` and another one for `It("using a Volume Snapshot")`, hence they end up pushing WALs to the same minio path.

Because the minio instance is shared between all E2Es, the test that would end up running as second would fail the `Empty archive check`.

### Cluster resource

_No response_

### Relevant log output

```shell
[FAILED] Timed out after 180.001s.
Expected
    <v1.BackupPhase>: running
to be equivalent to
    <string>: completed
```

### Code of Conduct

NiccoloFei:
1. Changed `Context("can bootstrap a replica cluster from a backup")` tests so that both object store and volume snapshot subtests can make use of the same source Cluster. This will save time and will fix the issue of Backups being stuck in "started" state.
2. Changed `AssertReplicaModeCluster()` to allow passing a name for the test table that will be created, so that it can be used multiple time against the same source Cluster.
3. Changed `AssertDetachReplicaModeCluster()` to add a new check to verify the non existence of the replica cluster's user. This assertion now also works standalone, and not only in conjunction with `AssertReplicaModeCluster()`.

Closes #4646

### DIFF
--- a/tests/e2e/asserts_test.go
+++ b/tests/e2e/asserts_test.go
@@ -977,8 +977,18 @@ func AssertDetachReplicaModeCluster(
 	})
 
 	By("verifying write operation on the replica cluster primary pod", func() {
-		AssertCreateTestDataWithDatabaseName(namespace, replicaClusterName, srcDatabaseName,
-			"replica_cluster_primary", psqlClientPod)
+		query := "CREATE TABLE IF NOT EXISTS replica_cluster_primary AS VALUES (1),(2);"
+		// Expect write operation to succeed
+		Eventually(func(g Gomega) {
+			var err error
+
+			// Get primary from replica cluster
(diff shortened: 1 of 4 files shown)


sxd (approved):

leonardoce (approved):

leonardoce:
/ok-to-merge e2e tests are fine on the test we fixed

Issue #4669, opened 2024-05-24
[Bug]: instance manager quitting without a proper shutdown


leonardoce:
### Is there an existing issue already for this bug?

- [X] I have searched for an existing issue, and could not find anything. I believe this is a new bug.

### I have read the troubleshooting guide

- [X] I have read the troubleshooting guide and I think this is a new bug.

### I am running a supported version of CloudNativePG

- [X] I have read the troubleshooting guide and I think this is a new bug.

### Contact Details

leonardo.cecchi@enterprisedb.com

### Version

1.23.1

### What version of Kubernetes are you using?

1.30 (unsupported)

### What is your Kubernetes environment?

Self-managed: kind (evaluation)

### How did you install the operator?

YAML manifest

### What happened?

After a PostgreSQL restart, if we can't connect to PostgreSQL after the programmed timeout, we immediately exit from the instance manager with exit status 1.
This prevents the instance manager from having a clean shutdown.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

mnencia:
Closed by #4670

PR #4670, opened 2024-05-24
fix: avoid shutting down the instance manager if it is not needed

leonardoce:
If we restart PostgreSQL but don't manage to connect to it, we should proceed with the instance manager's shutdown logic instead of quitting immediately with exit code 1.

This is especially important when PG shuts down because of missing disk space and when PG is up but trying to reach a consistent state.

Closes: #4063

### DIFF
--- a/internal/cmd/manager/instance/run/lifecycle/run.go
+++ b/internal/cmd/manager/instance/run/lifecycle/run.go
@@ -20,7 +20,6 @@ import (
 	"context"
 	"database/sql"
 	"fmt"
-	"os"
 	"sync"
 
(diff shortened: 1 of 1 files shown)


sxd (approved):

mnencia (approved):

leonardoce:
Tests: https://github.com/EnterpriseDB/cloudnative-pg/actions/runs/9224940814

jsilvela (approved):

mnencia:
The issue linked in the PR description is wrong. This should have closed #4669.

PR #4706, opened 2024-05-27
test(e2e): retry statements on failure

### LINKED ISSUE #4705: [Bug]: E2E statements creating test data are not retrying on failure

### Contact Details

_No response_

### Version

1.23.1

1.30 (unsupported)

YAML manifest

### What happened?

This issue affects only cloud providers. It mostly happens on GKE and I have been able to manually reproduce it only there, but it can also rarely happen in other cloud providers.

The scenario to reproduce it is to just create a one instance Cluster using the usual `AssertCreateCluster()`, and right after that use a function like `AssertCreateTestData()` which tries to create a table with some data inside (without retrying on failure).

This often errors out with:
```
command terminated with exit code 2 - psql: error: connection to server at "pg-basebackup-src-rw.cluster-pg-basebackup-6267.svc.cluster.local" (10.90.137.189), port 5432 failed: Connection refused
     Is the server running on that host and accepting TCP/IP connections?
```

We have a few places where this happens:
* E2E Drain Node with a single instance cluster when the PDB is disabled
* Bootstrap with pg_basebackup can bootstrap via pg_basebackup using basic authentication

What's going on is that, although the cluster is deemed ready by `AssertClusterIsReady()` (which is being used inside `AssertCreateCluster()`), pod to pod communication towards the pod of the newly created cluster is not yet ready.

So, since `AssertCreateTestData()` usually opens a connection from the `psqlClientPod` to the destination pod, it fails for the aforementioned reasons.

The reason this only happens for one instance Clusters is that, by the time replicas are ready, enough time passed so that communication with the primary pod is now ready (Pod to Pod networking is required to bootstrap the replicas from the primary, so surely we cannot reproduce this issue when we have Clusters with more than 1 instance) 

### Cluster resource

_No response_

### Relevant log output

_No response_

### Code of Conduct

NiccoloFei:
Closes #4705

### DIFF
--- a/tests/e2e/asserts_test.go
+++ b/tests/e2e/asserts_test.go
@@ -352,7 +352,7 @@ func AssertOperatorIsReady() {
 //
 // NOTE: even if we checked AssertClusterIsReady, a temporary DB connectivity issue would take
 // failureThreshold x periodSeconds to be detected
-func AssertDatabaseIsReady(namespace, clusterName string) {
+func AssertDatabaseIsReady(namespace, clusterName, dbName string) {
 	By(fmt.Sprintf("checking the database on %s is ready", clusterName), func() {
 		primary, err := env.GetClusterPrimary(namespace, clusterName)
 		Expect(err).ToNot(HaveOccurred())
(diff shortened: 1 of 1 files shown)


jsilvela (commented):
I agree on adding retry and Eventually.
But, two thoughts:
- we could/should have an Assert that checks that the DB is up. That would be a more explicit place to put the retry-timeout.
- Why are we (still) using the `psql` client for this. We have `ExecQueryInInstancePod`

BTW if we used the `ExecQueryInInstancePod` logic, we could have an `AssertDbReady` checking `db.Ping()` just like we do in the non-test code.

jsilvela (approved):

sxd (approved):

PR #4719, opened 2024-05-29
feat: liveness probe timeout configure

### LINKED ISSUE #4642: [Feature]: Ability to set timeouts for liveness probes on pods

The pods have a default value of 30 seconds for the liveness probes. It is hard coded.  This poses a problem whereby we are having maintainance on underlying storage or anything in compute that may take longer than 30 seconds then the operator will take action when in actual fact if only they have 5-10 extra seconds everything will be ok.

Have the ability in the cluster to set the liveness probe value with 30 seconds being the default

.I havent considered anything because nothing comes to mind.

### Additional context

_No response_

### Backport?

Yes

No

### Code of Conduct

litaocdl:
This patch provides the configuration for the allowed time before the 
liveness probe failure, the livenessProbeFailureThreshold is caculated
 by this configuration using:
FAILURE_THRESHOLD = ceil(livenessProbeTimeout / 10)

closes: #4642

### DIFF
--- a/api/v1/cluster_types.go
+++ b/api/v1/cluster_types.go
@@ -510,6 +510,14 @@ type ClusterSpec struct {
 	// The plugins configuration, containing
 	// any plugin to be loaded with the corresponding configuration
 	Plugins PluginConfigurationList `json:"plugins,omitempty"`
+
+	// LivenessProbeTimeout is the time in seconds that is allowed for a PostgreSQL instance
+	// to successfully respond to the liveness probe (default 30).
+	// The Liveness probe failure threshold is derived from this value using the formula:
+	// ceiling(livenessProbe / 10).
+	// +kubebuilder:default:=30
+	// +optional
+	LivenessProbeTimeout int32 `json:"livenessProbeTimeout,omitempty"`
 }
 
 // PluginConfigurationList represent a set of plugin with their
@@ -1330,6 +1338,12 @@ const (
 	// FailureThreshold of startupProbe, the formula is `FailureThreshold = ceiling(startDelay / periodSeconds)`,
 	// the minimum value is 1
 	DefaultStartupDelay = 3600
+
+	// DefaultLivenessProbeTimeout is the default value for livenessProbeTimeout, livenessProbeTimeout will be
+	// used to calculate the FailureThreshold of livenessProbe, the formula is
+	// `FailureThreshold = ceiling(livenessProbeTimeout / periodSeconds)`,
+	// the minimum value is 1
+	DefaultLivenessProbeTimeout = 30
 )
 
 // PostgresConfiguration defines the PostgreSQL configuration
@@ -2955,6 +2969,14 @@ func (cluster *Cluster) GetMaxSwitchoverDelay() int32 {
 		return cluster.Spec.MaxSwitchoverDelay
 	}
 	return DefaultMaxSwitchoverDelay
+}
+
+// GetLivenessProbeTimeout get the total allowed time before the liveness probe fail
+func (cluster *Cluster) GetLivenessProbeTimeout() int32 {
+	if cluster.Spec.LivenessProbeTimeout > 0 {
+		return cluster.Spec.LivenessProbeTimeout
+	}
+	return DefaultLivenessProbeTimeout
 }
 
 // GetPrimaryUpdateStrategy get the cluster primary update strategy,
--- a/api/v1/cluster_types.go
+++ b/api/v1/cluster_types.go
@@ -400,6 +400,14 @@ type ClusterSpec struct {
 	// +optional
 	FailoverDelay int32 `json:"failoverDelay,omitempty"`
 
+	// LivenessProbeTimeout is the time in seconds that is allowed for a PostgreSQL instance
+	// to successfully respond to the liveness probe (default 30).
+	// The Liveness probe failure threshold is derived from this value using the formula:
+	// ceiling(livenessProbe / 10).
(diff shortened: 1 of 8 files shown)


leonardoce (changes requested):
Besides the minor code suggestions, this patch is triggering a restart on every PG cluster once we update the operator.
This happens because the Pod specification we store in the annotations is different:

```
--- cluster-example-1.before.json       2024-05-30 13:40:13.871662145 +0200
+++ cluster-example-1.after.json        2024-05-30 13:40:19.168357874 +0200
@@ -149,17 +149,18 @@
         }
       ],
       "livenessProbe": {
         "httpGet": {
           "path": "/healthz",
           "port": 8000
         },
         "timeoutSeconds": 5,
-        "periodSeconds": 10
+        "periodSeconds": 10,
+        "failureThreshold": 3
       },
       "readinessProbe": {
         "httpGet": {
           "path": "/readyz",
           "port": 8000
         },
         "timeoutSeconds": 5,
         "periodSeconds": 10
```

Since the actual configuration didn't change, we should not trigger a restart.

Another point: this new parameter should be documented in the "Startup, liveness and readiness probes" section of the instance manager documentation.

api/v1/cluster_types.go
```diff
@@ -505,6 +505,14 @@ type ClusterSpec struct {
...
 	// any plugin to be loaded with the corresponding configuration
 	Plugins PluginConfigurationList `json:"plugins,omitempty"`
+
+	// LivenessProbeTimeout is the time in seconds that is allowed for a PostgreSQL instance
+	// to successfully respond to the liveness probe (default 30).
+	// The Liveness probe failure threshold is derived from this value using the formula:
+	// ceiling(livenessProbe / 10).
+	// +kubebuilder:default:=30
+	// +optional
+	LivenessProbeTimeout int32 `json:"livenessProbeTimeout,omitempty"`
```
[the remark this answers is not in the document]
litaocdl: That make sense!

litaocdl:
> Besides the minor code suggestions, this patch is triggering a restart on every PG cluster once we update the operator. This happens because the Pod specification we store in the annotations is different:
> 
> ```
> --- cluster-example-1.before.json       2024-05-30 13:40:13.871662145 +0200
> +++ cluster-example-1.after.json        2024-05-30 13:40:19.168357874 +0200
> @@ -149,17 +149,18 @@
>          }
>        ],
>        "livenessProbe": {
>          "httpGet": {
>            "path": "/healthz",
>            "port": 8000
>          },
>          "timeoutSeconds": 5,
> -        "periodSeconds": 10
> +        "periodSeconds": 10,
> +        "failureThreshold": 3
>        },
>        "readinessProbe": {
>          "httpGet": {
>            "path": "/readyz",
>            "port": 8000
>          },
>          "timeoutSeconds": 5,
>          "periodSeconds": 10
> ```
> 
> Since the actual configuration didn't change, we should not trigger a restart.
> 
> Another point: this new parameter should be documented in the "Startup, liveness and readiness probes" section of the instance manager documentation.



> Besides the minor code suggestions, this patch is triggering a restart on every PG cluster once we update the operator. This happens because the Pod specification we store in the annotations is different:
> 
> ```
> --- cluster-example-1.before.json       2024-05-30 13:40:13.871662145 +0200
> +++ cluster-example-1.after.json        2024-05-30 13:40:19.168357874 +0200
> @@ -149,17 +149,18 @@
>          }
>        ],
>        "livenessProbe": {
>          "httpGet": {
>            "path": "/healthz",
>            "port": 8000
>          },
>          "timeoutSeconds": 5,
> -        "periodSeconds": 10
> +        "periodSeconds": 10,
> +        "failureThreshold": 3
>        },
>        "readinessProbe": {
>          "httpGet": {
>            "path": "/readyz",
>            "port": 8000
>          },
>          "timeoutSeconds": 5,
>          "periodSeconds": 10
> ```
> 
> Since the actual configuration didn't change, we should not trigger a restart.
> 
> Another point: this new parameter should be documented in the "Startup, liveness and readiness probes" section of the instance manager documentation.

Thank you for this comments, this is real a good aspect to considering

pkg/specs/pods.go
```diff
@@ -252,6 +252,12 @@ func createPostgresContainers(cluster apiv1.Cluster, envConfig EnvConfig) []core
 
 	addManagerLoggingOptions(cluster, &containers[0])
 
+	// by default, the failureThreshold for liveness probe is 3, we do not want
+	// to change it if the user has not set the liveness probe timeout
+	if getLivenessProbeFailureThreshold(cluster.GetLivenessProbeTimeout()) != 3 {
```
leonardoce, line 257: I'm a bit scared of this "3" as a magic number. And I wouldn't like to have it even as a constant.
What would happen if a new API changed it? We'll we remember that?

I think it would be best just to use a `*int32` instead of an `int32` for this new parameter, and do this only if the user requested a custom configuration.

leonardoce (changes requested):

pkg/specs/pods.go
litaocdl: Thanks

litaocdl:
addressed comments and trigger a new e2e https://github.com/EnterpriseDB/cloudnative-pg/actions/runs/9348920588

leonardoce:
We should document that in the "Startup, liveness and readiness probes" section of the instance manager documentation.

leonardoce:
I did it @litaocdl

leonardoce (approved):

mnencia (approved):

PR #4765, opened 2024-06-03
fix: avoid to check backup container ID if pod container statuses are missing

### LINKED ISSUE #4767: [Bug]: Panic while checking backup container ID if pod container statuses are missing

### Contact Details

_No response_

### Version

1.23.1

1.30 (unsupported)

YAML manifest

### What happened?

{"level":"info","ts":"2024-05-28T07:09:22Z","msg":"Observed a panic in reconciler: runtime error: index out of range [0] with length 0","controller":"backup","controllerGroup":"postgresql.k8s.enterprisedb.io","controllerKind":"Backup","Backup":{"name":"p-r55dk9edcv-sch-20240515063800","namespace":"default"},"namespace":"default","name":"p-r55dk9edcv-sch-20240515063800","reconcileID":"62369a3b-fc02-48b2-8245-91d78ee3b571"}

panic: runtime error: index out of range [0] with length 0 [recovered]
	panic: runtime error: index out of range [0] with length 0

### Cluster resource

_No response_

### Relevant log output

_No response_

### Code of Conduct

gabriele-wolfox:
Closes #4767

### DIFF
--- a/api/v1/backup_types.go
+++ b/api/v1/backup_types.go
@@ -328,14 +328,10 @@ func (backupStatus *BackupStatus) SetAsCompleted() {
 // SetAsStarted marks a certain backup as started
 func (backupStatus *BackupStatus) SetAsStarted(targetPod *corev1.Pod, method BackupMethod) {
 	backupStatus.Phase = BackupPhaseStarted
-
-	if len(targetPod.Status.ContainerStatuses) > 0 {
-		backupStatus.InstanceID = &InstanceID{
-			PodName:     targetPod.Name,
-			ContainerID: targetPod.Status.ContainerStatuses[0].ContainerID,
-		}
+	backupStatus.InstanceID = &InstanceID{
+		PodName:     targetPod.Name,
+		ContainerID: targetPod.Status.ContainerStatuses[0].ContainerID,
 	}
-
 	backupStatus.Method = method
 }
 
--- a/api/v1/backup_types.go
+++ b/api/v1/backup_types.go
@@ -328,10 +328,14 @@ func (backupStatus *BackupStatus) SetAsCompleted() {
 // SetAsStarted marks a certain backup as started
 func (backupStatus *BackupStatus) SetAsStarted(targetPod *corev1.Pod, method BackupMethod) {
 	backupStatus.Phase = BackupPhaseStarted
-	backupStatus.InstanceID = &InstanceID{
-		PodName:     targetPod.Name,
-		ContainerID: targetPod.Status.ContainerStatuses[0].ContainerID,
+
+	if len(targetPod.Status.ContainerStatuses) > 0 {
+		backupStatus.InstanceID = &InstanceID{
+			PodName:     targetPod.Name,
+			ContainerID: targetPod.Status.ContainerStatuses[0].ContainerID,
+		}
 	}
+
 	backupStatus.Method = method
 }
 
--- a/api/v1/backup_types.go
+++ b/api/v1/backup_types.go
@@ -326,11 +326,11 @@ func (backupStatus *BackupStatus) SetAsCompleted() {
 }
 
 // SetAsStarted marks a certain backup as started
-func (backupStatus *BackupStatus) SetAsStarted(targetPod *corev1.Pod, method BackupMethod) {
+func (backupStatus *BackupStatus) SetAsStarted(podName, containerID string, method BackupMethod) {
 	backupStatus.Phase = BackupPhaseStarted
 	backupStatus.InstanceID = &InstanceID{
-		PodName:     targetPod.Name,
-		ContainerID: targetPod.Status.ContainerStatuses[0].ContainerID,
+		PodName:     podName,
+		ContainerID: containerID,
 	}
 	backupStatus.Method = method
 }
--- a/internal/controller/backup_controller.go
+++ b/internal/controller/backup_controller.go
@@ -197,7 +197,7 @@ func (r *BackupReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 		}
 		contextLogger.Debug("Found pod for backup", "pod", pod.Name)
 
-		if !utils.IsPodReady(*pod) || !utils.PodHasContainerStatuses(*pod) {
+		if !utils.IsPodReady(*pod) {
 			contextLogger.Info("Backup target is not ready, will retry in 30 seconds", "target", pod.Name)
 			backup.Status.Phase = apiv1.BackupPhasePending
 			r.Recorder.Eventf(&backup, "Warning", "BackupPending", "Backup target pod not ready: %s",
@@ -288,7 +288,7 @@ func (r *BackupReconciler) isValidBackupRunning(
 	}
 
 	var containerIsNotRestarted bool
-	if utils.PodHasContainerStatuses(pod) && backup.Status.InstanceID != nil {
+	if utils.PodHasContainerStatuses(pod) {
 		containerIsNotRestarted = backup.Status.InstanceID.ContainerID == pod.Status.ContainerStatuses[0].ContainerID
 	}
 	isPodActive := utils.IsPodActive(pod)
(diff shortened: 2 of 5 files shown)


jsilvela (changes requested):
I can see other occurrences of `ContainerStatuses[0]`, worth fixing while you're here.
And perhaps check for other `[0]`...

gabriele-wolfox:
> I can see other occurrences of `ContainerStatuses[0]`, worth fixing while you're here. And perhaps check for other `[0]`...

I put the focus on the error in the log, but It makes sense to fix them too.

tests/e2e/upgrade_test.go
```diff
@@ -535,7 +535,9 @@ var _ = Describe("Upgrade", Label(tests.LabelUpgrade, tests.LabelNoOpenshift), O
 			podList, err := env.GetClusterPodList(upgradeNamespace, clusterName1)
 			Expect(err).ToNot(HaveOccurred())
 			for _, pod := range podList.Items {
-				Expect(pod.Status.ContainerStatuses[0].RestartCount).To(BeEquivalentTo(0))
+				if len(pod.Status.ContainerStatuses) > 0 {
```
mnencia, line 538: I think that at this point, if `len(pod.Status.ContainerStatuses) == 0`, we should raise an error.

internal/controller/backup_controller.go
```diff
@@ -287,7 +287,10 @@ func (r *BackupReconciler) isValidBackupRunning(
 		return false, fmt.Errorf("unknown.spec.target received: %s", backup.Spec.Target)
 	}
 
-	containerIsNotRestarted := backup.Status.InstanceID.ContainerID == pod.Status.ContainerStatuses[0].ContainerID
+	var containerIsNotRestarted bool
+	if len(pod.Status.ContainerStatuses) > 0 {
+		containerIsNotRestarted = backup.Status.InstanceID.ContainerID == pod.Status.ContainerStatuses[0].ContainerID
```
mnencia, line 292: Is it possible that `backup.Status.InstanceID` is nil here? I think it could be after the change to the `SetAsStarted` function in this patch.

gabriele-wolfox: Adding a further check makes a lot of sense.

api/v1/backup_types.go
```diff
@@ -328,10 +328,14 @@ func (backupStatus *BackupStatus) SetAsCompleted() {
...
-		PodName:     targetPod.Name,
-		ContainerID: targetPod.Status.ContainerStatuses[0].ContainerID,
+
+	if len(targetPod.Status.ContainerStatuses) > 0 {
+		backupStatus.InstanceID = &InstanceID{
+			PodName:     targetPod.Name,
+			ContainerID: targetPod.Status.ContainerStatuses[0].ContainerID,
+		}
 	}
+
```
mnencia, line 338: I think it's incorrect to continue if the ContainedID is not detected correctly.

gabriele-wolfox: In order to apply this change, should I change SetAsStarted signature to check an error when it is invoked?
The error is returned if containerID is not detected properly.

jsilvela: I would not change SetAsStarted's signature.
But I would check there are ContainerStatuses before we call SetAsStarted.
Also, are you doing the changes @gabriele-wolfox , or am I?

internal/controller/backup_controller.go
```diff
@@ -560,6 +567,10 @@ func startInstanceManagerBackup(
 	pod *corev1.Pod,
 	cluster *apiv1.Cluster,
 ) error {
+	if len(pod.Status.ContainerStatuses) == 0 {
+		return fmt.Errorf("pod container statuses list is empty")
+	}
```
jsilvela, line 572: Here I'm not convinced.
If we emit this error, the calling function (`Reconcile`) will immediately mark the backup as failed, with no retry.
Whereas, a few lines previous, in a check for `utils.IsPodReady(*pod)` if it is not ready, it will retry.
I think retry is better.
Maybe we can take this as some sort of retriable condition.

gabriele-wolfox: Yeah, it could be retry is better, I have no strong opinion. 
We should decide and go for the behaviour we want.

internal/controller/backup_controller.go
```diff
@@ -366,6 +369,10 @@ func (r *BackupReconciler) reconcileSnapshotBackup(
 		return &ctrl.Result{RequeueAfter: 10 * time.Second}, nil
 	}
 
+	if len(targetPod.Status.ContainerStatuses) == 0 {
+		return nil, fmt.Errorf("pod container statuses list is empty")
+	}
```
jsilvela, line 374: here the logic does what we want. The overall `reconcileSnapshotBackup` fails with error,
and then the reconciliation gets re-looped.

internal/controller/backup_controller.go
jsilvela: Added the retry logic, please have a look to make sure i didn't go crazy

jsilvela:
manually took out the ok-to-merge flag since I only tested the backup tests in the slash-test

jsilvela (approved):

internal/controller/backup_controller.go
```diff
@@ -287,7 +287,10 @@ func (r *BackupReconciler) isValidBackupRunning(
 		return false, fmt.Errorf("unknown.spec.target received: %s", backup.Spec.Target)
 	}
 
-	containerIsNotRestarted := backup.Status.InstanceID.ContainerID == pod.Status.ContainerStatuses[0].ContainerID
+	var containerIsNotRestarted bool
+	if len(pod.Status.ContainerStatuses) > 0 && backup.Status.InstanceID != nil {
```
gabriele-wolfox, line 291: @jsilvela now that you introduced the utility function, you could invoke it also here, replacing with

`if !utils.IsPodReady(pod) && backup.Status.InstanceID != nil {`

internal/controller/backup_controller.go
```diff
@@ -197,7 +197,7 @@ func (r *BackupReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 		}
 		contextLogger.Debug("Found pod for backup", "pod", pod.Name)
 
-		if !utils.IsPodReady(*pod) {
+		if !utils.IsPodReady(*pod) || !utils.PodHasContainerStatuses(*pod) {
```
leonardoce, line 200: I think this added condition is not going to be useful.
How can a Pod be ready and not have any container statuses?

internal/controller/backup_controller.go
```diff
@@ -287,7 +287,10 @@ func (r *BackupReconciler) isValidBackupRunning(
 		return false, fmt.Errorf("unknown.spec.target received: %s", backup.Spec.Target)
 	}
 
-	containerIsNotRestarted := backup.Status.InstanceID.ContainerID == pod.Status.ContainerStatuses[0].ContainerID
```
leonardoce, line 290: This could be simplified. At this line, we're sure that `backup.Status.InstanceID != nil` otherwise, we returned a while ago.

If we find that this Pod has no container statuses, then it is new.
If it is new, it is not the same Pod that was taking the backup.
Therefore, this backup has been restarted.

leonardoce (changes requested):
While this patch is fixing the underlying issue, we could easily prevent this specific bug from happening in the future, too.

We could change the signature of [`SetAsStarted`](https://github.com/cloudnative-pg/cloudnative-pg/blob/0d27d7603739729d2e6cf3d178ea5d604ca5de6f/api/v1/backup_types.go#L329) to accept the pod name and the container ID, instead of accepting a pointer to the whole Pod. That would force us to actually check what is really happening.

internal/controller/backup_controller.go
```diff
@@ -287,7 +287,10 @@ func (r *BackupReconciler) isValidBackupRunning(
...
+	var containerIsNotRestarted bool
+	if utils.PodHasContainerStatuses(pod) {
+		containerIsNotRestarted = backup.Status.InstanceID.ContainerID == pod.Status.ContainerStatuses[0].ContainerID
+	}
```
leonardoce, line 293:
```suggestion
	containerIsNotRestarted := utils.PodHasContainerStatuses(pod) && backup.Status.InstanceID.ContainerID == pod.Status.ContainerStatuses[0].ContainerID
```

What do you think of this @gabriele-wolfox ? I find it more readable

leonardoce (changes requested):

leonardoce:
E2e tests: https://github.com/EnterpriseDB/cloudnative-pg/actions/runs/9369519768

leonardoce (approved):

PR #4769, opened 2024-06-03
feat: managed services support

### LINKED ISSUE #4339: [Feature]: introduce managed services

At the moment, CloudNativePG has a fixed number of services that are deployed with each Cluster:

- rw
- ro
- r

These services are all cluster IP.

There are use cases where:

- you don't want all the above services
- you want additional services with a different type (i.e., to expose the PostgreSQL service outside the Kubernetes cluster with a load balancer service)

We should improve the `Cluster` API with a new stanza called `managed.services`, providing capabilities like the ones shown below:

```yaml
kind: Cluster
metadata:
  name: cluster-example
spec:
  instances: 3
  storage:
    size: 1Gi
  managed:
    services:
      disabledDefaultServices: ['ro', 'r'] 
      additional:
      - type: rw
        serviceTemplate:
          metadata:
            name: cluster-example-lb-rw (REQUIRED)
          spec: (OPTIONAL)
            type: LoadBalancer
```

In the above example we disable the creation of the `cluster-example-ro` and `cluster-example-r` and we create an additional load balancer service.

In this case we use the Kubernetes `corev1.Service` resource API. We could have managed it ourselves, but our phylosophy is to reuse K8s resources.

### Additional context

Feel free to limit this issue to either the `additional` stanza or the `disabledDefaultServices` implementation (which I fear might be more difficult to implement as it might introduce regressions).

### Backport?

Yes

Yes

### Code of Conduct

armru:
By default, CloudNativePG creates three Kubernetes services for applications to access the cluster via the network:

- One pointing to the primary for read/write operations.
- One pointing to replicas for read-only queries.
- A generic one pointing to any instance for read operations.

This patch introduces managed services via the `managed.services` stanza, allowing you to:

- Disable the read-only and read services via configuration.
- Leverage the service template capability to create custom service resources, including load balancers, to access PostgreSQL outside Kubernetes (this is particularly useful for DBaaS purposes).

Closes #4339


internal/controller/cluster_create.go
```diff
@@ -302,45 +302,100 @@ func (r *ClusterReconciler) reconcilePoolerSecrets(ctx context.Context, cluster
...
-	return r.serviceReconciler(ctx, readWriteService)
+	if err := r.serviceReconciler(ctx, readWriteService, cluster.IsReadWriteServiceEnabled(), cluster); err != nil {
+		return err
+	}
+
+	return r.reconcileManagedServices(ctx, cluster)
 }
 
-func (r *ClusterReconciler) serviceReconciler(ctx context.Context, proposed *corev1.Service) error {
+func (r *ClusterReconciler) reconcileManagedServices(ctx context.Context, cluster *apiv1.Cluster) error {
```
armru, line 336: A default service could also be specified inside the managed services list. This would cause it to be reconciled with the template after the initial creation. 
This could help migrate the default services to the managed template.
I am looking forward to others' opinions on the matter.

api/v1/cluster_webhook.go
```diff
@@ -2338,6 +2339,51 @@ func (gcs *GoogleCredentials) validateGCSCredentials(path *field.Path) field.Err
 	return allErrors
 }
 
+func (r *Cluster) validateManagedServices() field.ErrorList {
```
armru, line 2342: we could potentially forbid the service names that could be used by the operator for the default services, see my other comment for more details

internal/controller/cluster_create.go
gabriele-wolfox: I agree with Armando's proposal to rely on a default to facilitate the migration to the managed services.

tests/e2e/fixtures/managed_services/cluster-managed-services-rw.yaml.template
```diff
@@ -0,0 +1,21 @@
...
+    services:
+      additional:
+        - selectorType: rw
+          serviceTemplate:
+            metadata:
+              name: "test-rw"
+              labels:
+                test-label: "true"
+              annotations:
+                test-annotation: "true"
```
gabriele-wolfox, line 21: Add missing empty line

tests/e2e/fixtures/managed_services/cluster-managed-services-no-default.yaml.template
```diff
@@ -0,0 +1,13 @@
...
+  name: cluster-disabled-default-services
+spec:
+  instances: 1
+  imageName: "${POSTGRES_IMG}"
+  storage:
+    size: 1Gi
+    storageClass: ${E2E_DEFAULT_STORAGE_CLASS}
+  managed:
+    services:
+      disabledDefaultServices: ["rw", "ro", "r"]
```
gabriele-wolfox, line 13: Add missing empty line

internal/controller/cluster_create.go
```diff
@@ -302,45 +302,105 @@ func (r *ClusterReconciler) reconcilePoolerSecrets(ctx context.Context, cluster
...
+			return nil
+		}
+		contextLogger.Info("creating service")
 		return r.Client.Create(ctx, proposed)
 	}
 	if err != nil {
 		return err
 	}
 
+	if clusterName, _ := IsOwnedByCluster(&livingService); clusterName != cluster.Name {
```
gabriele-wolfox, line 396: `clusterName` shadows the constant here https://github.com/cloudnative-pg/cloudnative-pg/blob/main/internal/controller/backup_controller.go#L61. I would rename it with `owner`.

internal/controller/cluster_create.go
```diff
@@ -302,45 +302,105 @@ func (r *ClusterReconciler) reconcilePoolerSecrets(ctx context.Context, cluster
...
+	if err != nil {
+		return err
+	}
+	for idx := range managedServices {
+		if err := r.serviceReconciler(ctx, cluster, &managedServices[idx], true); err != nil {
+			return err
+		}
+	}
+
+	containService := func(expected corev1.Service) func(iterated corev1.Service) bool {
```
gabriele-wolfox, line 347: `containService` declaration could be moved just before for-loop at line 362.

api/v1/cluster_webhook.go
gbartolini: I agree. Much simpler to manage

gbartolini:
> With new service templates, can we enforce the `cnpg.io/instanceRole` label in the selector instead of the deprecated `role` one? @armru

I have created this ticket #4894

litaocdl:
It seems we have a lot of places which leverage the old rw service name, like standby join, pgbench etc, if the old -rw service is disabled, we will face unresolved host name error.

armru:
I think that disabling the RW service could result in a lot of overhead and unintended situations, perhaps in the first shipped implementation of the managed services we could not list rw as a service that can be disabled, what do you think?
@litaocdl @mnencia @leonardoce @gbartolini

litaocdl:
> I think that disabling the RW service could result in a lot of overhead and unintended situations, perhaps in the first shipped implementation of the managed services we could not list rw as a service that can be disabled, what do you think? @litaocdl @mnencia @leonardoce @gbartolini

I agree with Armando.
The -rw service with cluster name is used in many places, some of those places like 

[restore.go ](https://github.com/cloudnative-pg/cloudnative-pg/blob/f200e185bd9ef078808c1686e2e52a3bdf4bdf18/pkg/management/postgres/restore.go#L894), [instances.go](https://github.com/cloudnative-pg/cloudnative-pg/blob/f200e185bd9ef078808c1686e2e52a3bdf4bdf18/pkg/management/postgres/instance.go#L1240) [pg_bench.go ](https://github.com/cloudnative-pg/cloudnative-pg/blob/f200e185bd9ef078808c1686e2e52a3bdf4bdf18/internal/cmd/plugin/pgbench/pgbench.go#L165) we do not have an easy way to get the managed rw service if default is disabled. 

so question is, should we treat the rw service as internal and not allow to disable ?

gbartolini:
> I think that disabling the RW service could result in a lot of overhead and unintended situations, perhaps in the first shipped implementation of the managed services we could not list rw as a service that can be disabled, what do you think? @litaocdl @mnencia @leonardoce @gbartolini

Yes, was talking about this right now with Leonardo and came to the same conclusion. We must prevent users from disabling the `rw` service.

armru:
we can then remove rw from the allowed disabled services list and we should be fine
@litaocdl

litaocdl (approved):

litaocdl:
default r and ro service is used in server certificate. 
```
  X509v3 Subject Alternative Name:
                DNS:cluster-example-managed-services-rw, DNS:cluster-example-managed-services-rw.cnpg, DNS:cluster-example-managed-services-rw.cnpg.svc, DNS:cluster-example-managed-services-r, DNS:cluster-example-managed-services-r.cnpg, DNS:cluster-example-managed-services-r.cnpg.svc, DNS:cluster-example-managed-services-ro, DNS:cluster-example-managed-services-ro.cnpg, DNS:cluster-example-managed-services-ro.cnpg.svc, DNS:cluster-example-managed-services-rw
```

litaocdl:
e2e: https://github.com/EnterpriseDB/cloudnative-pg/actions/runs/9577205589

litaocdl:
fix the failure and retrigger e2e https://github.com/EnterpriseDB/cloudnative-pg/actions/runs/9591311536

mnencia:
E2E tests on EDB fork https://github.com/EnterpriseDB/cloudnative-pg/actions/runs/9600158490

litaocdl:
trigger a run with ft=smoke depth=2 limit=aks,gke,eks
https://github.com/EnterpriseDB/cloudnative-pg/actions/runs/9611029976

mnencia:
Full E2E tests (including CSPs) running on EDB fork https://github.com/EnterpriseDB/cloudnative-pg/actions/runs/9645665394

mnencia:
Another full run of E2E tests on EDB fork https://github.com/EnterpriseDB/cloudnative-pg/actions/runs/9660551283

mnencia (approved):

gbartolini (approved):

PR #4775, opened 2024-06-04
feat(plugin): support color control via  flags and TTY detection

### LINKED ISSUE #4722: [Feature]: TTY detection for the plugin

When using the `cnpg` plugin, the output always uses ANSI colours, even when no terminal is attached (for example when piping to `less` or `cat`).

Implement `tty` detection (see https://github.com/logrusorgru/aurora/issues/2#issuecomment-299027757).

Also, introduce two options:

* `--colors/colours`: which forces ANSI colours
* `--no-colors/no-colours`: which disables ANSI colours

By default, `tty` detection is enforced.

We use the `aurora` library, and we should just use their guidelines.

### Additional context

_No response_

### Backport?

Yes

Yes

### Code of Conduct

YanniHu1996:
Introduced support for ANSI colors in the `cnpg` plugin through the `--color` option, which accepts `always`, `never`, and `auto` (default) as values. When `auto` is selected, the plugin automatically detects if the standard output channel is connected to a terminal and adds ANSI colors to the command output.

Closes: #4722

### DIFF
--- a/cmd/kubectl-cnpg/main.go
+++ b/cmd/kubectl-cnpg/main.go
@@ -66,6 +66,8 @@ func main() {
 		PersistentPreRunE: func(cmd *cobra.Command, _ []string) error {
 			logFlags.ConfigureLogging()
 
+			plugin.ConfigureColor(cmd)
+
 			// If we're invoking the completion command we shouldn't try to create
 			// a Kubernetes client and we just let the Cobra flow to continue
 			if cmd.Name() == "completion" || cmd.Name() == "version" ||
--- a/cmd/kubectl-cnpg/main.go
+++ b/cmd/kubectl-cnpg/main.go
@@ -81,28 +81,36 @@ func main() {
 
 	logFlags.AddFlags(rootCmd.PersistentFlags())
 	configFlags.AddFlags(rootCmd.PersistentFlags())
+	//	plugin.AddColorControlFlag(hibernateStatusCmd)
 
-	rootCmd.AddCommand(certificate.NewCmd())
-	rootCmd.AddCommand(destroy.NewCmd())
-	rootCmd.AddCommand(fence.NewCmd())
-	rootCmd.AddCommand(fio.NewCmd())
-	rootCmd.AddCommand(hibernate.NewCmd())
-	rootCmd.AddCommand(install.NewCmd())
-	rootCmd.AddCommand(maintenance.NewCmd())
-	rootCmd.AddCommand(pgbench.NewCmd())
-	rootCmd.AddCommand(promote.NewCmd())
-	rootCmd.AddCommand(reload.NewCmd())
-	rootCmd.AddCommand(report.NewCmd())
-	rootCmd.AddCommand(restart.NewCmd())
-	rootCmd.AddCommand(status.NewCmd())
-	rootCmd.AddCommand(versions.NewCmd())
-	rootCmd.AddCommand(backup.NewCmd())
-	rootCmd.AddCommand(psql.NewCmd())
-	rootCmd.AddCommand(snapshot.NewCmd())
-	rootCmd.AddCommand(logs.NewCmd())
-	rootCmd.AddCommand(pgadmin.NewCmd())
-	rootCmd.AddCommand(publication.NewCmd())
-	rootCmd.AddCommand(subscription.NewCmd())
+	subcommands := []*cobra.Command{
+		certificate.NewCmd(),
+		destroy.NewCmd(),
+		fence.NewCmd(),
+		fio.NewCmd(),
+		hibernate.NewCmd(),
+		install.NewCmd(),
+		maintenance.NewCmd(),
+		pgbench.NewCmd(),
+		promote.NewCmd(),
+		reload.NewCmd(),
+		report.NewCmd(),
+		restart.NewCmd(),
+		status.NewCmd(),
+		versions.NewCmd(),
+		backup.NewCmd(),
+		psql.NewCmd(),
+		snapshot.NewCmd(),
+		logs.NewCmd(),
+		pgadmin.NewCmd(),
+		publication.NewCmd(),
+		subscription.NewCmd(),
+	}
+
+	for _, cmd := range subcommands {
+		plugin.AddColorControlFlag(cmd)
+		rootCmd.AddCommand(cmd)
+	}
 
 	if err := rootCmd.Execute(); err != nil {
 		os.Exit(1)
--- a/cmd/kubectl-cnpg/main.go
+++ b/cmd/kubectl-cnpg/main.go
@@ -81,7 +81,6 @@ func main() {
 
 	logFlags.AddFlags(rootCmd.PersistentFlags())
 	configFlags.AddFlags(rootCmd.PersistentFlags())
-	//	plugin.AddColorControlFlag(hibernateStatusCmd)
 
 	subcommands := []*cobra.Command{
 		backup.NewCmd(),
--- a/cmd/kubectl-cnpg/main.go
+++ b/cmd/kubectl-cnpg/main.go
@@ -84,27 +84,27 @@ func main() {
 	//	plugin.AddColorControlFlag(hibernateStatusCmd)
 
 	subcommands := []*cobra.Command{
+		backup.NewCmd(),
 		certificate.NewCmd(),
 		destroy.NewCmd(),
 		fence.NewCmd(),
 		fio.NewCmd(),
 		hibernate.NewCmd(),
 		install.NewCmd(),
+		logs.NewCmd(),
 		maintenance.NewCmd(),
+		pgadmin.NewCmd(),
 		pgbench.NewCmd(),
 		promote.NewCmd(),
+		psql.NewCmd(),
+		publication.NewCmd(),
 		reload.NewCmd(),
 		report.NewCmd(),
 		restart.NewCmd(),
-		status.NewCmd(),
-		versions.NewCmd(),
-		backup.NewCmd(),
-		psql.NewCmd(),
 		snapshot.NewCmd(),
-		logs.NewCmd(),
-		pgadmin.NewCmd(),
-		publication.NewCmd(),
+		status.NewCmd(),
 		subscription.NewCmd(),
+		versions.NewCmd(),
 	}
 
 	for _, cmd := range subcommands {
--- a/docs/src/kubectl-plugin.md
+++ b/docs/src/kubectl-plugin.md
@@ -172,13 +172,9 @@ kubectl cnpg <command> <args...>
 ```
 
 !!! Note
-    The plugin uses colorization to help make the information easier to read at
-    a glance.
-    Some users may prefer to deactivate colorization, which can be done setting
-    the `--color` flag to `never`. By default the plugin will deactivate
-    colorization when its output is sent to a non-terminal (e.g. piping to a
-    pager or `grep`).
-
+    The plugin automatically detects if the standard output channel is connected to a terminal.
+    In such cases, it may add ANSI colors to the command output. To disable colors, use the
+    `--color=always` option with the command.
 ### Generation of installation manifests
 
 The `cnpg` plugin can be used to generate the YAML manifest for the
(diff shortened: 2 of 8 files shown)


go.mod
```diff
@@ -82,6 +82,7 @@ require (
 	github.com/json-iterator/go v1.1.12 // indirect
 	github.com/liggitt/tabwriter v0.0.0-20181228230101-89fcab3d43de // indirect
 	github.com/mailru/easyjson v0.7.7 // indirect
+	github.com/mattn/go-isatty v0.0.20 // indirect
```
leonardoce, line 85: Couldn't we just one `https://pkg.go.dev/golang.org/x/term` instead of adding an external dependency?

YanniHu1996:
the e2e from fork
https://github.com/EnterpriseDB/cloudnative-pg/actions/runs/9362999058

internal/cmd/plugin/color.go
```diff
@@ -0,0 +1,64 @@
...
+		aurora.WithColors(shouldColorize),
+		aurora.WithHyperlinks(true),
+	)
+	return nil
+}
+
+// AddColorControlFlags adds color control flags to the command
+func AddColorControlFlags(cmd *cobra.Command) {
+	cmd.Flags().Bool("colors", false, "Force colorized output even if no terminal is attached")
+	cmd.Flags().Bool("no-colors", false, "Disable colorized output")
```
leonardoce, line 62: Having multiple flags with opposite meanings could be misleading.
Why not just use a `color` output, whose value could be `always`, `auto`, or `never`?
This would be coherent with other Unix commands like `ls`, in both the GNU implementation and the BSD one.

leonardoce (changes requested):

internal/cmd/plugin/color.go
YanniHu1996: thanks Leo, it makes sense, And `auto` better represents that the plugin does a detect TTY automatically

internal/cmd/plugin/color.go
```diff
@@ -0,0 +1,66 @@
...
+}
+
+func configureColor(cmd *cobra.Command, isTTY bool) error {
+	colorFlag, err := cmd.Flags().GetString("color")
+	if err != nil {
+		return err
+	}
+
+	var shouldColorize bool
+	switch colorFlag {
```
leonardoce, line 40: In my opinion, `colorFlag` is not a string but a custom type representing an enumeration. 
A similar concept is implemented in the [snapshot owner reference](https://github.com/cloudnative-pg/cloudnative-pg/blob/0d27d7603739729d2e6cf3d178ea5d604ca5de6f/api/v1/cluster_types.go#L130) field.

Once we have a custom type with the relative constants, the following `switch` statement could use them instead of these strings.

The custom type could also implement `pflag.Value` encapsulating validation. 

[This comment](https://github.com/spf13/pflag/issues/236#issuecomment-931600452) contains a really nice example.

internal/cmd/plugin/color.go
```diff
@@ -0,0 +1,66 @@
...
+	aurora.DefaultColorizer = aurora.New(
+		aurora.WithColors(shouldColorize),
+		aurora.WithHyperlinks(true),
+	)
+	return nil
+}
+
+// AddColorControlFlag adds color control flags to the command
+func AddColorControlFlag(cmd *cobra.Command) {
+	cmd.Flags().String("color", "auto", "Control color output; options include 'always', 'auto', or 'never'")
```
leonardoce, line 60: Once the new type implements `pflag.Value`, this could be `Var` instead of `String`.

internal/cmd/plugin/hibernate/cmd.go
```diff
@@ -57,6 +57,7 @@ var (
 		},
 		RunE: func(cmd *cobra.Command, args []string) error {
 			clusterName := args[0]
+
```
leonardoce, line 60: We could remove this line to make this PR easier to review.

internal/cmd/plugin/status/cmd.go
```diff
@@ -41,6 +41,10 @@ func NewCmd() *cobra.Command {
 			ctx := cmd.Context()
 			clusterName := args[0]
 
+			if err := plugin.ConfigureColor(cmd); err != nil {
+				return err
+			}
+
```
leonardoce, line 47: I think we could actually do this at `PersistentRunE` level, so this will work in every `kubectl cnpg` subcommand.

leonardoce (changes requested):

internal/cmd/plugin/color.go
sxd: Hoo an Enum with the proper auto-complete :D just saying!

YanniHu1996: Thanks, Leo! I believe what you mean is to implement a custom type specifically for the Color flag, not a generic enumeration type.
Please check here https://github.com/cloudnative-pg/cloudnative-pg/pull/4775/files#diff-f6d12283f95d10d3d682413241cab651cec5e25f0c63d84ee42b292a717a5a53R28-R54

internal/cmd/plugin/status/cmd.go
YanniHu1996: Take into consideration that the color flag is only applicable to the hibernate status and status commands. Therefore, I have not designated it as a global flag, but rather assigned it at the command level.

internal/cmd/plugin/color.go
leonardoce: @YanniHu1996 That's exactly what I meant with: "custom type representing an enumeration"

internal/cmd/plugin/status/cmd.go
leonardoce: Yes, but everything will be ready if we implement another command requiring it.
With your implementation, we need to remember to set the colorizer in the new command, too.

internal/cmd/plugin/color.go
```diff
@@ -0,0 +1,97 @@
...
+}
+
+// ConfigureColor renews aurora.DefaultColorizer based on flags and TTY
+func ConfigureColor(cmd *cobra.Command) {
+	configureColor(cmd, term.IsTerminal(int(os.Stdout.Fd())))
+}
+
+func configureColor(cmd *cobra.Command, isTTY bool) {
+	colorFlag := cmd.Flag("color")
+	// skip if the command does not have the color flag
```
jsilvela, line 66: Shouldn't the default be considered to be `auto`?

internal/cmd/plugin/color.go
```diff
@@ -0,0 +1,97 @@
...
+}
+
+// Type implements pflag.Value interface
+func (e *colorValue) Type() string {
+	return "string"
+}
+
+// ConfigureColor renews aurora.DefaultColorizer based on flags and TTY
+func ConfigureColor(cmd *cobra.Command) {
+	configureColor(cmd, term.IsTerminal(int(os.Stdout.Fd())))
```
jsilvela, line 61: The library function used, `term.IsTerminal`, seems pretty coarse, and is returning true on code that cannot support colorizing sequences (eg the Plan9 tools).
That said, it discriminates when using pipes, so it’s somewhat useful.

internal/cmd/plugin/hibernate/cmd.go
```diff
@@ -111,6 +111,7 @@ func NewCmd() *cobra.Command {
 			"text",
 			"Output format. One of text, json, or yaml",
 		)
+	plugin.AddColorControlFlag(hibernateStatusCmd)
```
jsilvela, line 114: Adding the `--color` flag only in the hibernate or status subcommands is problematic IMO,
as users may find that adding it in other subcommands like fencing or backups, results in an exception.

jsilvela (approved):
Aside from a couple corrections: the code was intending to apply the `--color` flag only in the `hibernate` and `status` sub-commands.
If the user adds the `--color` option to other commands, it would error out, which is bad IMO. I've added the option to all subcommands, and will update the PR description which was specifying `hibernate` and `status` only
... FYI

docs/src/kubectl-plugin.md
```diff
@@ -171,6 +171,14 @@ Once the plugin was installed and deployed, you can start using it like this:
...
 ```
 
+!!! Note
+    The plugin uses colorization to help make the information easier to read at
+    a glance.
+    Some users may prefer to deactivate colorization, which can be done setting
+    the `--color` flag to `never`. By default the plugin will deactivate
+    colorization when its output is sent to a non-terminal (e.g. piping to a
+    pager or `grep`).
+
```
gbartolini, line 181:
```suggestion
!!! Note
    The plugin automatically detects if the standard output channel is connected to a terminal.
    In such cases, it may add ANSI colors to the command output. To disable colors, use the
    `--color=always` option with the command.
```

gbartolini (commented):
I would prefer to use `yes` instead of `always` and `no` instead of `never`. I made a small change in the docs, emphasizing on the automated detection of TTY. @YanniHu1996 can you please make the above changes? Otherwise, if you prefer this syntax, I am happy to get a broader opinion here (use the thumbs below).

cmd/kubectl-cnpg/main.go
```diff
@@ -79,28 +81,36 @@ func main() {
 
 	logFlags.AddFlags(rootCmd.PersistentFlags())
 	configFlags.AddFlags(rootCmd.PersistentFlags())
+	//	plugin.AddColorControlFlag(hibernateStatusCmd)
```
sxd, line 84: Why this line is commented?

cmd/kubectl-cnpg/main.go
```diff
@@ -79,28 +81,36 @@ func main() {
...
-	rootCmd.AddCommand(versions.NewCmd())
-	rootCmd.AddCommand(backup.NewCmd())
-	rootCmd.AddCommand(psql.NewCmd())
-	rootCmd.AddCommand(snapshot.NewCmd())
-	rootCmd.AddCommand(logs.NewCmd())
-	rootCmd.AddCommand(pgadmin.NewCmd())
-	rootCmd.AddCommand(publication.NewCmd())
-	rootCmd.AddCommand(subscription.NewCmd())
+	for _, cmd := range subcommands {
+		plugin.AddColorControlFlag(cmd)
```
sxd, line 111: This should be a persisten flag just like --namespace and everything so it will rule any color on any command

internal/cmd/plugin/color.go
```diff
@@ -0,0 +1,101 @@
...
+	// colorAlways configures the output to always be colorized
+	colorAlways colorConfiguration = "always"
+	// colorAuto configures the the output to be colorized only when attached to a terminal
+	colorAuto colorConfiguration = "auto"
+	// colorNever configures the the output never to be colorized
+	colorNever colorConfiguration = "never"
+)
+
+// String returns the strig representation
+func (e colorConfiguration) String() string {
```
sxd, line 42: I think this should be a pointer, right? to have the requested value of `e` updated

litaocdl: checked the invoker is only read the value, probably string(colorConfiguration) also works

YanniHu1996:
> I would prefer to use `yes` instead of `always` and `no` instead of `never`. I made a small change in the docs, emphasizing on the automated detection of TTY. @YanniHu1996 can you please make the above changes? Otherwise, if you prefer this syntax, I am happy to get a broader opinion here (use the thumbs below).

I prefer using 'always' and 'never', much like many commands do, which appears to be a convention. Adhering to such conventions simplifies the user experience. But regardless of convention, `yes` and' `no` are more straightforward and simple

litaocdl (approved):

mnencia:
The idea behind using "always", "auto", and "never" is to describe how the detection works:

* `always`: we always use colors in the output 
* `auto`: we use colors if we detect the presence of a TTY 
* `never`: colors are always disabled

I find it more descriptive than "yes", "no", and "auto", so my vote goes to always-auto-never.

gbartolini:
Ok, given that Yanni and Marco are so vocal on this, I suggest we keep always/never. Please proceed.

gbartolini (approved):

PR #4834, opened 2024-06-12
docs: Cluster-Wide Permissions and Operator Deployment

### LINKED ISSUE #4833: [Docs]: Cluster-Wide Permissions and Operator Deployment

This ticket includes the proposed new section on "Cluster-Wide Permissions and Operator Deployment" from PR #4638. I believe this section is out of place (located at the bottom of the page) and seems to have been written without fully understanding the content on the page.

I think this section should be aligned with the existing RBAC section, integrated with it, and most importantly, should not contradict it.

Describe cluster-wide permissions and operator deployments

security.md

none

### Additional context

see #4638 

### Backport?

Yes

### Code of Conduct

gbartolini:
Closes #4833

### DIFF
--- a/docs/src/security.md
+++ b/docs/src/security.md
@@ -112,7 +112,7 @@ cluster role which defines the set of rules/resources/verbs granted to the opera
 
 Below we provide some examples and, most importantly, the reasons why
 CloudNativePG requires full or partial management of standard Kubernetes
-namespaced or non-namespaced resources.
+namespaced or global resources.
 
 `configmaps`
 : The operator needs to create and manage default config maps for
--- a/docs/src/security.md
+++ b/docs/src/security.md
@@ -112,7 +112,7 @@ cluster role which defines the set of rules/resources/verbs granted to the opera
 
 Below we provide some examples and, most importantly, the reasons why
 CloudNativePG requires full or partial management of standard Kubernetes
-namespaced resources.
+namespaced or non-namespaced resources.
 
 `configmaps`
 : The operator needs to create and manage default config maps for
@@ -169,14 +169,13 @@ namespaced resources.
   decide in which nodes a pod can be scheduled preventing the replicas to be
   in the same node, specially if nodes are in different availability zones. This
   permission is also used to determine if a node is schedule or not, avoiding
-  the creation of pods that cannot be created at all.
+  the creation of pods that cannot be created at all and/or trigger a switch over if
+  the primary lives in a unscheduled node.
 
 To see all the permissions required by the operator, you can run `kubectl
 describe clusterrole cnpg-manager`.
 
-<!-- TODO: please integrate this or align it with the RBAC section -->
-
-### Cluster-Wide Permissions and Operator Deployment
+### ClusterRole Permissions and Operator Deployment
 
 #### Using the Manifest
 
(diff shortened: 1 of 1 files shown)


leonardoce (approved):

docs/src/security.md
```diff
@@ -169,11 +169,59 @@ namespaced resources.
   decide in which nodes a pod can be scheduled preventing the replicas to be
   in the same node, specially if nodes are in different availability zones. This
   permission is also used to determine if a node is schedule or not, avoiding
-  the creation of pods that cannot be created at all.
+  the creation of pods that cannot be created at all and/or trigger a switch over if
```
jsilvela, line 172: hmm, this is not very good. Going to try some changes

```suggestion
  the creation of pods that cannot be created at all and/or trigger a switch over if
```

docs/src/security.md
```diff
@@ -165,15 +165,66 @@ namespaced resources.
...
+
+Even with these permissions, if someone gains access to the `ServiceAccount`,
+they will only have `get`, `list`, and `watch` permissions, which are limited
+to viewing resources. If an unauthorized user already has access to the
+`ServiceAccount`, it indicates a more significant security issue.
+
+It's crucial to prevent users from accessing the operator's `ServiceAccount`
+and any other `ServiceAccount` with elevated permissions.
+
+#### Can This Issue Be Mitigated?
```
jsilvela, line 216: Don't like this section. Firstly, what "Issue" is it making reference to?
Also, it looks way too apologetic.

docs/src/security.md
```diff
@@ -169,14 +169,13 @@ namespaced resources.
...
   in the same node, specially if nodes are in different availability zones. This
   permission is also used to determine if a node is schedule or not, avoiding
-  the creation of pods that cannot be created at all.
+  the creation of pods that cannot be created at all and/or trigger a switch over if
+  the primary lives in a unscheduled node.
 
 To see all the permissions required by the operator, you can run `kubectl
 describe clusterrole cnpg-manager`.
 
-<!-- TODO: please integrate this or align it with the RBAC section -->
```
gbartolini: Why was this not considered? @sxd

sxd: It was

gbartolini:
@sxd @jsilvela can you please verify this is fine?

docs/src/security.md
```diff
@@ -165,14 +171,63 @@ namespaced resources.
...
+```sh
+kubectl describe clusterrole cnpg-manager
+```
+
+!!! Info
+    You can customize the YAML file to align with your security requirements by
+    using a `RoleBinding` instead of a `ClusterRoleBinding`. This approach enables
+    deployment of the operator in a dedicated namespace with restricted access from
+    administrators. This ensures that standard users cannot access the operator's
+    `ServiceAccount`.
```
jsilvela, line 204: I don't think we explain anywhere how to do this, and it's not obvious.
Should we have this info box?
Or maybe add a section on how to customize the YAML?

jsilvela (approved):

jsilvela:
Aside from my comment on the info box, looks good.

sxd (approved):

docs/src/security.md
sxd: Probably we can do that on the plugin side, but that will be out of the scope of this documentation for sure

gbartolini: I found this awkward too, but I understand the need to outline this possibility. I have tried several times to write a sentence that explains that CNPG doesn't support this yet, but could not come with an easy and clear way to do it. Hence, less is more and decided to omit it.

gbartolini: Honestly, I was expecting the whole `cnpg generate` command to already take into account all of this, and was puzzled when I discovered it wasn't actually working. So I'd rather work on a solution than documenting the current state (which is not a feature we provide).

PR #4836, opened 2024-06-12
chore: centralize `HTTPClient` creation

armru:
This patch aims to avoid code duplication for the HTTPClient initialization and ensure that we always initialize a client capable of both HTTP and HTTPS requests.

### DIFF
--- a/pkg/management/postgres/webserver/backup_client.go
+++ b/pkg/management/postgres/webserver/backup_client.go
@@ -19,19 +19,17 @@ package webserver
 import (
 	"bytes"
 	"context"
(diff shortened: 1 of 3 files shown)


armru:
unsure if we should backport or not

mnencia:
This should follow the destiny of the TLS communication patch, so only main for now.

mnencia (approved):

leonardoce:
E2e: https://github.com/EnterpriseDB/cloudnative-pg/actions/runs/9593196636

leonardoce (approved):

PR #4882, opened 2024-06-17
fix(plugin): handle errors during the `status` command execution

### LINKED ISSUE #4847: [Bug]: plugin command status fails getting PDB

### Contact Details

_No response_

### Version

1.23.2

1.30 (unsupported)

YAML manifest

### What happened?

When the user doesn't have any permission at all over the `poddisruptionbudgets` the status command will fail with the following error:
```
 Error: while extracting PodDisruptionBudgetList: poddisruptionbudgets.policy is forbidden: User "sxd" cannot list resource "poddisruptionbudgets" in API group "policy" in the namespace "jg"
```
And it doesn't show anything even if I have access to the cluster.

### Cluster resource

_No response_

### Relevant log output

_No response_

### Code of Conduct

armru:
This patch ensures that we continue the `status` command execution even if we encounter an error while trying to fetch the status of the instances.

Closes #4847

### DIFF
--- a/internal/cmd/plugin/status/status.go
+++ b/internal/cmd/plugin/status/status.go
@@ -19,7 +19,6 @@ package status
 import (
 	"context"
 	"fmt"
-	apierrs "k8s.io/apimachinery/pkg/api/errors"
 	"os"
 	"path"
 	"sort"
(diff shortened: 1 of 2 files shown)


sxd (approved):

internal/cmd/plugin/status/status.go
```diff
@@ -19,6 +19,7 @@ package status
 import (
 	"context"
 	"fmt"
+	apierrs "k8s.io/apimachinery/pkg/api/errors"
```
gabriele-wolfox, line 22: Linter issue

gabriele-wolfox (approved):

gabriele-wolfox (approved):

mnencia (approved):

PR #4892, opened 2024-06-18
fix: improve error handling in `EnsureDirectoryExists` function

### LINKED ISSUE #4891: [Bug]:  `EnsureDirectoryExists`  always returns nil if error isn't  `the file is not exist`

### Contact Details

huyantian0@gmail.com

### Version

1.23.2

1.30 (unsupported)

YAML manifest

### What happened?

https://github.com/cloudnative-pg/cloudnative-pg/blob/8d0b05a895744443909c0d8522a524a816a39ac2/pkg/fileutils/fileutils.go#L299-L308

 `EnsureDirectoryExists`  always returns nil if error isn't  `the file is not exist` 

### Cluster resource

_No response_

### Relevant log output

_No response_

### Code of Conduct

YanniHu1996:
This update modifies the `EnsureDirectoryExists` function to propagate previously ignored unexpected errors.
The function also verifies that the target is a directory when it already exists.

Closes #4891

### DIFF
--- a/pkg/fileutils/fileutils.go
+++ b/pkg/fileutils/fileutils.go
@@ -298,16 +298,21 @@ func EnsureParentDirectoryExist(fileName string) error {
 
 // EnsureDirectoryExists check if the passed directory exists or not, and if
 // it doesn't exist, create it using 0700 as permissions bits.
-// If the directory exists and its permissions are not 0700, that's fine
-func EnsureDirectoryExists(destinationDir string) (err error) {
-	if _, err = os.Stat(destinationDir); errors.Is(err, fs.ErrNotExist) {
-		err = os.MkdirAll(destinationDir, 0o700)
-		if err != nil {
-			return err
+// If the directory exists it doesn't change the permissions.
+func EnsureDirectoryExists(destinationDir string) error {
+	stat, err := os.Stat(destinationDir)
+	if err != nil {
+		if errors.Is(err, fs.ErrNotExist) {
+			return os.MkdirAll(destinationDir, 0o700)
 		}
+		return err
 	}
 
-	return err
+	if !stat.IsDir() {
+		return fs.ErrInvalid
+	}
+
+	return nil
 }
 
 // MoveFile moves a file from a source path to its destination by copying
--- a/pkg/fileutils/fileutils.go
+++ b/pkg/fileutils/fileutils.go
@@ -22,8 +22,10 @@ import (
 	"bufio"
 	"bytes"
 	"context"
+	"errors"
 	"fmt"
 	"io"
+	"io/fs"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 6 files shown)


YanniHu1996:
e2e from fork 
https://github.com/EnterpriseDB/cloudnative-pg/actions/runs/9560725211

mnencia (approved):

pkg/fileutils/fileutils.go
```diff
@@ -296,15 +296,15 @@ func EnsureParentDirectoryExist(fileName string) error {
 
 // EnsureDirectoryExists check if the passed directory exists or not, and if
 // it doesn't exist, create it using 0700 as permissions bits
-func EnsureDirectoryExists(destinationDir string) error {
-	if _, err := os.Stat(destinationDir); os.IsNotExist(err) {
```
jsilvela, line 300: while we're modifying this function, I think we should ensure that if the directory exists, it should have the right permissions. Going to try adding tests and modify.

pkg/fileutils/fileutils.go
```diff
@@ -296,15 +296,15 @@ func EnsureParentDirectoryExist(fileName string) error {
...
+func EnsureDirectoryExists(destinationDir string) (err error) {
```
jsilvela, line 299: while we're modifying this function, I think we should ensure that if the directory exists, it should have the right permissions. Going to try adding tests and modify.

jsilvela:
I added some testing and realized if given an existing dir with the wrong permissions,
the function would do nothing and report success.
Changed that to chmod if it finds wrong permissions.

pkg/fileutils/fileutils.go
```diff
@@ -294,14 +295,18 @@ func EnsureParentDirectoryExist(fileName string) error {
...
-			return err
-		}
+	fileInfo, err := os.Stat(destinationDir)
+	if os.IsNotExist(err) {
+		return os.MkdirAll(destinationDir, 0o700)
+	} else if err != nil {
+		return err
+	}
+	if fileInfo.Mode().Perm() != fs.FileMode(0o700) {
+		return os.Chmod(destinationDir, 0o700) // nolint:gosec
```
mnencia, line 309: Please don't do that. It could have unintended consequences, and it goes beyond ensuring that the directory exists.

jsilvela: yep, found out the tests were failing. Rolled back.

jsilvela (approved):

jsilvela:
Green tests https://github.com/EnterpriseDB/cloudnative-pg/actions/runs/9649606706

mnencia:
Full E2E running on EDB fork https://github.com/EnterpriseDB/cloudnative-pg/actions/runs/9659652078

PR #4895, opened 2024-06-18
fix: ignore ServerName mismatch in internal TLS

### LINKED ISSUE #4885: [Bug]: TLS verification will fail with user provided server certificate

### Contact Details

_No response_

### Version

trunk (main)

1.30 (unsupported)

YAML manifest

### What happened?

When you provide valid global certificates for the servers, it is not guaranteed that they contain the <cluster-name>-rw DNS name, as usually, the certificate emitter cannot verify it; thus, the name is not included in the final certificate. That's why the [streaming connection to the primary](https://github.com/cloudnative-pg/cloudnative-pg/blob/bb8fc8439b0f3c8701feab8ff33eecdaf97a014f/pkg/management/postgres/conninfo.go#L38) is configured with `sslmode=verify-ca` instead of full certification.

The operator must use the same verification level when speaking with the instance.

In the future, we can consider configuring this behavior in the Cluster CR.

### Cluster resource

_No response_

### Relevant log output

_No response_

### Code of Conduct

mnencia:
Make the operator verify server certificates against the CA only, disregarding the DNS name.
This is to match what PostgreSQL does for communication within the cluster, avoiding errors in the operator communications where Postgres accepts the certificates. 

Closes #4885


armru (approved):

mnencia:
E2E running on EDB fork https://github.com/EnterpriseDB/cloudnative-pg/actions/runs/9611995929

sxd (approved):

mnencia:
This is related to #4936

gbartolini (approved):

PR #4897, opened 2024-06-18
chore: deprecate `role` label

### LINKED ISSUE #4894: [Feature]: adopt `cnpg.io/instanceRole` in service selector

We still use the deprecated `role` label in the service selector. Given that we are adding managed services, we should use this opportunity to switch entirely the new service selectors to the `cnpg.io/instanceRole` label.

Replace `role` with `cnpg.io/instanceRole` in the service selector. Given that pods will still have both labels, the change will not break any existing cluster.

N/A

### Additional context

_No response_

### Backport?

No

Yes

### Code of Conduct

armru:
This patch removes the old `role` label from the Service and PodDistruptionBudget selector generation.

Closes #4894 

Note for reviewers: I don't think we should backport the label deprecation.

### DIFF
--- a/internal/cmd/plugin/psql/psql_test.go
+++ b/internal/cmd/plugin/psql/psql_test.go
@@ -132,7 +132,7 @@ func fakePod(name, role string) corev1.Pod {
 			Name:      name,
 			Namespace: "default",
 			Labels: map[string]string{
-				utils.ClusterRoleLabelName: role,
+				utils.ClusterInstanceRoleLabelName: role,
 			},
 		},
 	}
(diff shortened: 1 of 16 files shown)


armru:
I think we should not backport this

YanniHu1996 (approved):

NiccoloFei (approved):

mnencia:
The results from the previous run look abnormally high on failures, but the failures seem unrelated. I started another E2E full run on the EDB fork to exclude environment issues https://github.com/EnterpriseDB/cloudnative-pg/actions/runs/9744215344

PR #4911, opened 2024-06-19
chore(security): removed non required permissions

### LINKED ISSUE #1640: Reassess role assigned to the service account of the cluster
Please verify each rule associated to the role that the service account requires for the instance manager to talk to the API server, starting from: https://github.com/cloudnative-pg/cloudnative-pg/blob/main/pkg/specs/roles.go#L57

Try and see if we can reduce and optimize that list, by following POLA.

sxd:
Some permissions were there because of legacy reasons and are not needed
anymore, this aims to reduce the amount of permissions required by the
operator and keep as short as possible.

Closes #1640

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -127,18 +127,14 @@ rules:
   - mutatingwebhookconfigurations
   verbs:
   - get
-  - list
   - patch
-  - update
 - apiGroups:
   - admissionregistration.k8s.io
   resources:
   - validatingwebhookconfigurations
   verbs:
   - get
-  - list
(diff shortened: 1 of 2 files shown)


leonardoce:
Started another E2e test run: https://github.com/EnterpriseDB/cloudnative-pg/actions/runs/9611272212

leonardoce (approved):

leonardoce:
E2e: https://github.com/EnterpriseDB/cloudnative-pg/actions/runs/9615944656

mnencia (approved):

PR #4927, opened 2024-06-21
feat: add optional TLS communication for the metrics exporter on the instances

### LINKED ISSUE #4710: [Feature]: the metrics port on instance pods should support HTTPS

As part of increased security for CloudNativePG, the prometheus metrics port
served in port 9187 should support TLS

Use basic auth and TLS which is supported by Prometheus.
https://prometheus.io/docs/prometheus/latest/configuration/https/

none

### Additional context

_No response_

### Backport?

Yes

No

### Code of Conduct

jsilvela:
Enable optional TLS communication on the Cluster's metrics port.
When TLS is enabled, the metrics exporter uses the same server certificate
used by PostgreSQL.

Closes #4710


pkg/specs/podmonitor.go
```diff
@@ -41,8 +44,41 @@ func (c ClusterPodMonitorManager) BuildPodMonitor() *monitoringv1.PodMonitor {
...
+					},
+					Key: certs.CACertKey,
+				},
+			},
+			KeySecret: &corev1.SecretKeySelector{
+				LocalObjectReference: corev1.LocalObjectReference{
+					Name: c.cluster.GetServerCASecretName(),
+				},
+				Key: certs.CAPrivateKeyKey,
+			},
```
jsilvela, line 73: Per Marco, look into leaving empty

jsilvela:
@gbartolini I was just looking at that. You're right.

jsilvela:
After discussion with @mnencia we decided to leave this PR as is,
and create a new issue for the manager metrics.
https://github.com/cloudnative-pg/cloudnative-pg/issues/5070
cc/ @gbartolini

mnencia (approved):

armru (approved):

mnencia:
While reviewing the documentation, I realized that our API here is incomplete. If someone is implementing TLS security in metrics scraping, they probably want more versatility in the TLS configuration, something like the following:

```
type TLSConfig struct {
	// Secret containing the CA cert to use to validate the client.
	CASecret *v1.SecretKeySelector `json:"caSecret,omitempty"`
	// Secret containing the server cert file.
	CertSecret *v1.SecretKeySelector `json:"certSecret,omitempty"`
	// Secret containing the server key file.
	KeySecret *v1.SecretKeySelector `json:"keySecret,omitempty"`
}
```

I don't say we should do it now, but we should be prepared to add it. So before merging, I want to check this idea and have a future-proof API.

mnencia:
After a brief discussion, we concluded that the optimal interface to implement is as follows:

```yaml
spec:
  monitoring:
    tls:            
      enable: true 
```

This approach provides more room for future enhancements without cluttering the monitoring section.

mnencia (approved):

mnencia:
Waiting for the E2E result before merging.

docs/src/monitoring.md
```diff
@@ -90,6 +91,49 @@ spec:
...
 
+### Enabling TLS on the Metrics Port
+
+To enable TLS communication on the metrics port, configure the `spec.monitoring.tls.enable`
+setting to `true`. This setup ensures that the metrics exporter uses the same
+server certificates as PostgreSQL to secure communication on port 5432.
+
+!!! important
+    Changing the `spec.monitoring.tls.enable` value will cause all instances to restart.
+    Plan accordingly to minimize downtime.
```
jsilvela, line 102: "All instances to restart" seems very chaotic, where this is our well known operation of a rolling restart.
And "minimize downtime" ... with a well performed rolling restart, there would be none, right?

jsilvela, line 102:
```suggestion
    Changing the `spec.monitoring.tls.enable` value will initiate a rolling restart of the cluster.
```

mnencia: What about?

```suggestion
    Changing the `spec.monitoring.tls.enable` value will cause a rollout of all instances in the cluster.
```

config/crd/bases/postgresql.cnpg.io_clusters.yaml
```diff
@@ -3909,6 +3909,18 @@ spec:
                           type: string
                       type: object
                     type: array
+                  tls:
+                    description: |-
+                      Configure TLS communication for the metrics endpoint.
+                      Changing tls.enable option will force a rollout of all instances.
+                    properties:
+                      enable:
```
gbartolini, line 3917: Why 'enable` when we already have `enabled` in other places?

mnencia: Because you are right, and this is a typo, nobody noticed.

mnencia: fix committed

config/crd/bases/postgresql.cnpg.io_clusters.yaml
```diff
@@ -3909,6 +3909,18 @@ spec:
                           type: string
                       type: object
                     type: array
+                  tls:
+                    description: |-
+                      Configure TLS communication for the metrics endpoint.
+                      Changing tls.enable option will force a rollout of all instances.
```
gbartolini, line 3915: `enable` -> `enabled`

mnencia: fixed

NiccoloFei (approved):

PR #4939, opened 2024-06-24
feat: leverage pgbouncer's auth_dbname

### LINKED ISSUE #1383: Introduce support for auth_dbname in pgbouncer
PgBouncer has introduced the patch that we submitted some time ago about the `auth_dbname`:
https://github.com/pgbouncer/pgbouncer/commit/0ba795f063da121ad7aa3d63699058e85fd3c382

The introduction of its support in the operator might simplify the integration with pgbouncer as well as reduce the workload on the reconciler.

phisco:
Fixes #1383.

An initial draft implementation to discuss

### DIFF
--- a/api/v1/cluster_types.go
+++ b/api/v1/cluster_types.go
@@ -1097,6 +1097,9 @@ type PoolerIntegrations struct {
 type PgBouncerIntegrationStatus struct {
 	// +optional
 	Secrets []string `json:"secrets,omitempty"`
+
+	// +optional
+	AuthDBName string `json:"authDBName,omitempty"`
 }
 
 // ReplicaClusterConfiguration encapsulates the configuration of a replica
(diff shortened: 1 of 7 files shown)


leonardoce:
#8671 is an alternative and simpler implementation of this concept, that doesn't allow the user to change the name of the database where the `user_search` function is created. It may be used as a starting point or just to encourage discussion.

gbartolini:
> #8671 is an alternative and simpler implementation of this concept, that doesn't allow the user to change the name of the database where the `user_search` function is created. It may be used as a starting point or just to encourage discussion.

I really like it. We should not need that feature. The goal of the `auth_dbname` option in PgBouncer that we both developed, was finalised to use it against the `postgres` database only in CloudNativePG, in order to reduce the maintenance burden of the reconciliation loop.

This is already a big improvement.

gbartolini:
Closing this as replaced by #8671

PR #4952, opened 2024-06-26
feat(managed-services): add support for `updateStrategy`

### LINKED ISSUE #4922: [Feature]: specify an update strategy for managed services

If you are using managed services to implement load-balancers for your Cluster, it could happen that a change in the template cannot be applied with a patch to the service, and the service must be recreated instead.

For example, the [AWS docs](https://docs.aws.amazon.com/eks/latest/userguide/network-load-balancing.html#:~:text=Network%20Load%20Balancers%20are%20created,when%20you%20created%20your%20cluster.) about network load-balancers contains this:

> Network Load Balancers are created with the internal aws-load-balancer-scheme, by default.
> Do not edit the annotations after creating your service. If you need to modify it, delete the service object and create it again with the desired value for this annotation.

We need a declarative way to ask the operator to drop and recreate the services.

I would add a `updateStrategy` field to the `ManagedService` object with two possible values:

* patch (default): update the existing service
* replace: delete and recreate the service

N/A

### Additional context

_No response_

### Backport?

No

Yes

### Code of Conduct

armru:
Introduce the `updateStrategy` field to control how the operator updates a service definition. By default, the operator uses the `patch` strategy, applying changes directly to the service. Alternatively, the `recreate` strategy deletes the existing service and recreates it from the template.

This is particularly useful to handle changes in cloud-based load balancers for DBaaS scenarios.

Closes #4922

### DIFF
--- a/tests/e2e/managed_services_test.go
+++ b/tests/e2e/managed_services_test.go
@@ -21,6 +21,7 @@ import (
 
 	corev1 "k8s.io/api/core/v1"
 	apierrs "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 
 	apiv1 "github.com/cloudnative-pg/cloudnative-pg/api/v1"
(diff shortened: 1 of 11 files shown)


tests/e2e/managed_services_test.go
```diff
@@ -163,4 +164,60 @@ var _ = Describe("Managed services tests", Label(tests.LabelSmoke, tests.LabelBa
 			}, testTimeouts[utils.ManagedServices]).Should(Succeed())
 		})
 	})
+
+	It("should properly handle replace update strategy", func(ctx SpecContext) {
```
armru, line 168: Perhaps we could add another E2E that runs in cloud environments and tries to create and update a load balancer.

@mnencia

mnencia: It's a good idea, but the test could be very flaky (load balancers are very slow to provision) and may require specific per-CSP tests to be meaningful.

@fcanovai, what do you think?

mnencia (approved):

leonardoce (approved):

gbartolini (approved):

PR #4960, opened 2024-06-26
docs: remove strflocaltime from troubleshooting.md

### LINKED ISSUE #4957: [Docs]: remove strflocaltime from troubleshooting

The command:
```
kubectl logs -n <NAMESPACE> <CLUSTER>-<N> | \
  jq -r 'select(.logger=="postgres") | [(.ts|strflocaltime("%Y-%m-%dT%H:%M:%S %Z")), .record.message] | @csv'
```

is not working
`jq: error (at <stdin>:5487): strftime/1 requires parsed datetime inputs
`

from what @jsilvela told me, this was needed when the kubectl logs where reporting .ts in epoch.
Since this is not the case any more I would remove the strflocaltime.

Then the logs .ts are reportet in UTC and I didn't find a way to change the output to use the local timezone.
Let me know if you have some ideas otherwise I would just remove the strflocaltime.

remove strflocaltime

troubleshooting.md

### Additional context

please assign this issue to me

### Backport?

Yes

### Code of Conduct

smiyc:
closes #4957

Signed-off-by: Daniel Chambre smiyc@pm.me

### DIFF
--- a/docs/src/troubleshooting.md
+++ b/docs/src/troubleshooting.md
@@ -345,7 +345,7 @@ kubectl logs -n <NAMESPACE> <CLUSTER>-<N> | \
   jq -r 'select(.logger=="postgres") | [.ts, .record.message] | @csv'
 ```
 
-If the timestamp is displayed in Unix Epoch time, you can convert it to a user-friedgly format:
+If the timestamp is displayed in Unix Epoch time, you can convert it to a user-friendly format:
 
 ```shell
 kubectl logs -n <NAMESPACE> <CLUSTER>-<N> | \
--- a/docs/src/troubleshooting.md
+++ b/docs/src/troubleshooting.md
@@ -338,11 +338,18 @@ kubectl logs -n <NAMESPACE> <CLUSTER>-<N> | \
   jq 'select(.logger=="postgres") | .record.message'
 ```
 
-The following example also adds the timestamp in a user-friendly format:
+The following example also adds the timestamp:
 
 ```shell
 kubectl logs -n <NAMESPACE> <CLUSTER>-<N> | \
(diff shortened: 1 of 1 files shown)


jsilvela (changes requested):
The change should work ... however ... sometimes we do see UNIX-style timestamps in the logs.
I think it would be even better if you added an extra paragraph with the original command in case you're getting UNIX timestamps.

jsilvela:
BTW we should not see UNIX timestamps, but the K8s logging is very messy, so there may be paths that don't do RFC3339...

smiyc (commented):
> BTW we should not see UNIX timestamps, but the K8s logging is very messy, so there may be paths that don't do RFC3339... 

like so?

docs/src/troubleshooting.md
```diff
@@ -338,7 +338,14 @@ kubectl logs -n <NAMESPACE> <CLUSTER>-<N> | \
...
 
-The following example also adds the timestamp in a user-friendly format:
+The following example also adds the timestamp:
+
+```shell
+kubectl logs -n <NAMESPACE> <CLUSTER>-<N> | \
+  jq -r 'select(.logger=="postgres") | [.ts, .record.message] | @csv'
+```
+
+If the timestamp is displayed in Unix Epoch time, you can convert it to a user-friedgly format:
```
jsilvela, line 348:
```suggestion
If the timestamp is displayed in Unix Epoch time, you can convert it to a user-friendly format:
```

jsilvela:
Exactly like you did (without the typo though ;) )

jsilvela (approved):

mnencia (approved):

PR #4974, opened 2024-06-28
chore: use a proxy to the pod instead of exec a command inside the pod

### LINKED ISSUE #4941: [Bug]: Avoid the need of `pods/exec` permission in the plugin

### Contact Details

_No response_

### Version

1.23.2

1.30 (unsupported)

YAML manifest

### What happened?

The plugin requires `pods/exec` permissions to execute the `status` command, we should avoid this permission for full and switch to `pods/proxy` permission since this will allow us to get the status of the pods without allowing the user to execute bash inside the pods and running commands.

### Cluster resource

_No response_

### Relevant log output

_No response_

### Code of Conduct

sxd:
This patch employs a `pods/proxy` call to read the instance status,
eliminating the need to execute any command inside the pod.
This reduces the required permissions for the status plugin command.
The update ensures backward compatibility with previous versions of
the operator.

Closes #4941

### DIFF
--- a/internal/cmd/plugin/status/status.go
+++ b/internal/cmd/plugin/status/status.go
@@ -104,7 +104,7 @@ func Status(ctx context.Context, clusterName string, verbose bool, format plugin
 		return err
 	}
 
-	status := ExtractPostgresqlStatus(ctx, cluster)
+	status := extractPostgresqlStatus(ctx, cluster)
 	err = plugin.Print(status, format, os.Stdout)
 	if err != nil || format != plugin.OutputFormatText {
 		return err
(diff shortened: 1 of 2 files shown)


armru (approved):

NiccoloFei (approved):

mnencia (changes requested):
This patch breaks the compatibility with old versions of the operator. We need to fix this before merging.

mnencia:
We need to implement something like we did here https://github.com/cloudnative-pg/cloudnative-pg/commit/692f373da60e4fce0cbd74fe041d5d19ee8b4013#diff-d90e33a455c7ba5834af26c2c0ccf13ecdf069e264f5d1f34d7ac4818bfa7c01R75-R78

mnencia (approved):

PR #4982, opened 2024-07-01
fix: immediate backup not retried

### LINKED ISSUE #4981: [Bug]: ScheduledBackup - Immediate backup not retried

### Contact Details

gabriele.fedi@enterprisedb.com

### Version

1.23.2

1.30 (unsupported)

Other

### What happened?

When a `ScheduledBackup` is created with `immediate: true`, the operator tries to create a backup resource immediately. If this operation fails for any reason, the immediate backup creation won't be retried.

The issue is easy to reproduce creating a `ScheduledBackup` first, targeting a not existing cluster and setting the `backupOwnerReference` field as `cluster`. On the immediate backup creation, the operator will notice that the target cluster doesn't exist and will log the error. Creating the targeted cluster afterwards, should unblock the `ScheduledBackup` from creating the immediate backup, but this doesn't happen. 

Since the controller manager's clients mostly rely on caches by default, if both the `Cluster` and the `ScheduledBackup` are created together (or within a small amount of time), the `ScheduledBackup` controller's client could return a `notFound` error while fetching the `Cluster` and the immediate `Backup` will never be created. 

### Cluster resource

```shell
---
kind: ScheduledBackup
metadata:
  name: backup-example
spec:
  immediate: true
  schedule: "0 0 0 * * *"
  backupOwnerReference: cluster
  cluster:
    name: pg-backup
---

kind: Cluster
metadata:
  name: pg-backup
spec:
  instances: 1
  storage:
    size: 1Gi
```

### Relevant log output

```shell
cnpg-controller-manager-5c79c78fb8-zz7sq manager {"level":"error","ts":"2024-07-01T08:02:36Z","msg":"Reconciler error","controller":"scheduledbackup","controllerGroup":"postgresql.cnpg.io","controllerKind":"ScheduledBackup","ScheduledBackup":{"name":"backup-example","namespace":"default"},"namespace":"default","name":"backup-example","reconcileID":"0ae23107-cd62-46d8-92a6-93f4e63d74b5","error":"Cluster.postgresql.cnpg.io \"pg-backup\" not found","stacktrace":"sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler\n\tvendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:324\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem\n\tvendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:261\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2\n\tvendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:222"}
```

### Code of Conduct

GabriFedi97:
Avoid setting the `lastCheckTime` on `ScheduledBackup` before creating the immediate `Backup`. If the operation fails, the operator will skip it in the next, re-queued, reconciliation loop if it finds the `lastCheckTime` set in the status.

Moving the creation of the immediate `Backup` before the `lastCheckTime` status update, solves the issue. 

Closes #4981

### DIFF
--- a/internal/controller/scheduledbackup_controller.go
+++ b/internal/controller/scheduledbackup_controller.go
@@ -174,11 +174,13 @@ func ReconcileScheduledBackup(
 	now := time.Now()
 	origScheduled := scheduledBackup.DeepCopy()
 
+	if scheduledBackup.Status.LastCheckTime == nil && scheduledBackup.IsImmediate() {
+		// we populate the status (lastCheckTime...) by following the same rules of the scheduled backup
+		event.Eventf(scheduledBackup, "Normal", "BackupSchedule", "Scheduled immediate backup now: %v", now)
+		return createBackup(ctx, event, cli, scheduledBackup, now, now, schedule, true)
+	}
+
 	if scheduledBackup.Status.LastCheckTime == nil {
-		if scheduledBackup.IsImmediate() {
-			event.Eventf(scheduledBackup, "Normal", "BackupSchedule", "Scheduled immediate backup now: %v", now)
-			return createBackup(ctx, event, cli, scheduledBackup, now, now, schedule, true)
-		}
 		// This is the first time we check this schedule,
 		// let's wait until the first job will be actually
 		// scheduled
--- a/internal/controller/scheduledbackup_controller.go
+++ b/internal/controller/scheduledbackup_controller.go
@@ -172,7 +172,6 @@ func ReconcileScheduledBackup(
 	}
 
 	now := time.Now()
-	origScheduled := scheduledBackup.DeepCopy()
 
 	if scheduledBackup.Status.LastCheckTime == nil && scheduledBackup.IsImmediate() {
 		// we populate the status (lastCheckTime...) by following the same rules of the scheduled backup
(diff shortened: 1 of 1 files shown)


armru:
Hey, @GabriFedi97. While I understand what we are trying to fix, I don't think this is the correct way to do it. 
I think a better way to fix this issue would take into account the fact that the immediate backup can fail instead of having a workaround with lastchecktime

armru (approved):

armru:
I made the code more explicit and added the risk tag. I strongly suggest for another maintainer to review this

GabriFedi97:
> A better way to fix this issue would be to consider the fact that the immediate backup can fail instead of having a workaround with lastchecktime.

Do you mean by having an additional field in the status?

mnencia (approved):

mnencia:
I agree with Armando that we should improve the handling of backup failures. However, the proposed change looks correct and clearer than the previous code. I'm approving this PR and asking @leonardoce for the final review.

leonardoce:
E2e test run: https://github.com/EnterpriseDB/cloudnative-pg/actions/runs/9954655019

leonardoce (approved):

leonardoce:
The code for the backup controller can be largely improved, but this patch makes it a bit clearer.

PR #5006, opened 2024-07-02
fix(pooler): enhance rollout control during operator upgrade

### LINKED ISSUE #4999: [Bug]: pgbouncer pod unexpectedly restarted

### Contact Details

_No response_

### Version

1.23.2

1.30 (unsupported)

YAML manifest

### What happened?

In the pull reqeust https://github.com/cloudnative-pg/cloudnative-pg/commit/19589259221a668c1b1bcc9f0f6c47787ec981f3
We added a new field called `name` in the struct `Metadata` which is referenced by PoolerSpec. And the logic to compute object hash includes  type definition info 
```go
func DeepHashObject(hasher hash.Hash, objectToWrite interface{}) error {
    hasher.Reset()
    printer := spew.ConfigState{
        Indent:         " ",
        SortKeys:       true,
        DisableMethods: true,
        SpewKeys:       true,
    }

    _, err := printer.Fprintf(hasher, "%#v", objectToWrite)
    return err
}
```

So when upgrading the operator, the deployment of the Pooler will be patched as well and the underlying pod will restarted

We should avoid this case

### Cluster resource

_No response_

### Relevant log output

_No response_

### Code of Conduct

mnencia:
This update clarifies Pooler pods' restart behavior. Previously, the operator image name was excluded from the configuration hash calculation, so an operator upgrade would not have generated a rollout until anything else in the spec changed, even if that change did not require a rollout. This behavior was surprising for the user.

To remove surprises, we now include the operator image in the configuration hash, so every upgrade triggers a rollout.

Additionally, for users who prefer more control over rollouts, we've introduced the annotation `cnpg.io/reconcilePodSpec: disabled`. This annotation prevents any updates to the Pooler deployment, except for changes to the number of replicas.

Closes #4999

### DIFF
--- a/pkg/specs/pgbouncer/deployments.go
+++ b/pkg/specs/pgbouncer/deployments.go
@@ -43,7 +43,9 @@ const (
 // Deployment create the deployment of pgbouncer, given
 // the configurations we have in the pooler specifications
 func Deployment(pooler *apiv1.Pooler, cluster *apiv1.Cluster) (*appsv1.Deployment, error) {
-	poolerHash, err := hash.ComputeVersionedHash(pooler.Spec, 3)
+	operatorImageName := config.Current.OperatorImageName
+
+	poolerHash, err := computeTemplateHash(pooler, operatorImageName)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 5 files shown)


YanniHu1996 (approved):
It's best to verify whether the pgbouncer pod is restarted such as what we have already done against instance pod

armru (approved):

litaocdl (approved):

litaocdl:
waiting for pr https://github.com/cloudnative-pg/cloudnative-pg/pull/5044 merged

pkg/specs/pgbouncer/deployments.go
```diff
@@ -150,6 +152,20 @@ func Deployment(pooler *apiv1.Pooler, cluster *apiv1.Cluster) (*appsv1.Deploymen
 	}, nil
 }
 
+func computeTemplateHash(pooler *apiv1.Pooler, operatorImageName string) (string, error) {
+	type deploymentHash struct {
```
YanniHu1996, line 156: The object used to calculate the hash is changed, so the pgbouncer pod will be restarted after upgrading the operator

leonardoce: Yes, agreed

PR #5031, opened 2024-07-04
chore: improve OLM environment detection

### LINKED ISSUE #3764: [Feature]: Improve how we detect that we're running on a OLM environment

Currently we relay on SCC to assume we're running under an OLM compatible environment, we should relay on looking for operators.coreos.com/(v1|v2) resources

Look the operators.coreos.com kind API to detect that we have an OLM compatible environment

There's non

### Additional context

_No response_

### Backport?

Yes

Yes

### Code of Conduct

sxd:
As far as now, we were relying on the detection of the SecurityContextConstraint to know that we were running on an OLM environment but this is not a secure way since an OLM environment may or may not have a SCC inside, that's why we should look at the `operators.coreos.com/v1` resources and check that  it exists, if this is the case, we're in an OLM environment.

This helps us to improve the plugin report and properly retrieve more useful information.

Closes #3764

### DIFF
--- a/internal/cmd/plugin/report/olm.go
+++ b/internal/cmd/plugin/report/olm.go
@@ -23,6 +23,7 @@ import (
 	"path/filepath"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/client-go/dynamic"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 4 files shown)


internal/cmd/plugin/report/olm.go
```diff
@@ -0,0 +1,120 @@
...
+	}
+
+	for idx := range listFromLabelOpenshift.Items {
+		item := listFromLabelOpenshift.Items[idx]
+		if !isContained(item, listFromLabelNamespace.Items) {
+			listFromLabelOpenshift.Items = append(listFromLabelOpenshift.Items, item)
+		}
+	}
+
+	return listFromLabelOpenshift, err
```
fcanovai, line 75: This won't merge the lists. If the operator isn't in `openshift-operators`, but we have an operator in a namespace, we return empty list.

internal/cmd/plugin/report/olm.go
```diff
@@ -0,0 +1,120 @@
...
+}
+
+// olmOperatorReport contains the operator data in Oolm
```
fcanovai, line 78:
```suggestion
// olmOperatorReport contains the operator data in olm
```

fcanovai (changes requested):
Beside the broken piece of code, I'm not convinced about the entire logic.

This PR is checking the `openshift-operators` namespace, and the namespace defined in the plugin call (or `default`). However, it first checks for the operator deployment in the plugin-defined namespace, exiting if the operator is not there. So looking into openshift-operator is not useful, you have to get the operator location right. I think we could just check the namespace asked for in the plugin.

armru (approved):

internal/cmd/plugin/report/olm.go
```diff
@@ -0,0 +1,98 @@
...
+	"github.com/cloudnative-pg/cloudnative-pg/internal/cmd/plugin"
+)
+
+// getOlmResourceList gets the desired resource using the Dynamic Client
+// to avoid code dependencies on the OLM libraries
+func getOlmResourceList(
+	ctx context.Context,
+	dynamicClient dynamic.Interface,
+	namespace, resource string,
+) (client.ObjectList, error) {
```
armru, line 39: Wouldn't make it better to return a []client.object instead of an ObjectList?

fcanovai (approved):

PR #5036, opened 2024-07-06
docs: considerations on block storage solutions

weiyentan:
This adds to the documentation for storage to document the feature of data-locality highlighting the scenarios where instances of the data could be on the same node. This can add a risk because if the node is destroyed (as per treating nodes like cattle) the cluster can be corrupted. The documentation highlights this and suggests certain possible solutions that can be investigated.

### DIFF
--- a/docs/src/storage.md
+++ b/docs/src/storage.md
@@ -467,7 +467,7 @@ To use a pre-provisioned volume in CloudNativePG:
     cluster.) Make sure you check for any pods stuck in `Pending` after you deploy
     the cluster. If the condition persists, investigate why it's happening.
 
-## Block storage considerations (Ceph/ Longhorn)
+## Block storage considerations (Ceph/Longhorn)
 
 Most block storage solutions in Kubernetes recommend having multiple replicas
 of a volume to improve resiliency. This works well for workloads that don't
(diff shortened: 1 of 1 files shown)


docs/src/storage.md
```diff
@@ -467,16 +467,43 @@ To use a pre-provisioned volume in CloudNativePG:
...
+approach works well for workloads that lack built-in resiliency.
+
+However, CloudNativePG integrates this resiliency directly into the Postgres
+`Cluster` through the number of instances and the persistent volumes attached
+to them, as explained in ["Synchronizing the state"](architecture.md#synchronizing-the-state).
+
+As a result, defining additional replicas at the storage level can lead to
+write amplification, unnecessarily increasing disk I/O and space usage.
+
+In such cases, consider reducing the number of replicas at the block storage
```
jsilvela, line 483: I don't love "in such cases" here. Do we mean "cases where the application handles resiliency above the block level"? Or do we mean, "when using CloudNativePG"?
```suggestion
For CloudNativePG usage, consider reducing the number of replicas at the block storage
```

jsilvela (commented):
This is a very good document @weiyentan , thanks.
I left an inline suggestion.
And, for completeness, perhaps we could discuss the case where a single instance is used? Obviously not recommended to use a single instance, but if there were a need? Should block-level redundacy be used?

weiyentan:
Happy to talk about this. @jsilvela

docs/src/storage.md
gbartolini: I suggest you directly make the change @jsilvela so that we can have this doc in the next release. Thanks.

gbartolini (approved):

jsilvela (approved):

jsilvela:
Made the change and approved.
@gbartolini I don't have auth to merge this, can you?
Oh, on @weiyentan 's commits, the signoff is missing. See the DCO failure report:
> Commit sha: [a62fbe9], Author: Wei-Yen Tan, Committer: Gabriele Bartolini; The sign-off is missing.
Commit sha: [45d8fb4], Author: Wei-Yen Tan, Committer: Gabriele Bartolini; The sign-off is missing.
Commit sha: [1b86b2c], Author: Wei-Yen Tan, Committer: Gabriele Bartolini; The sign-off is missing.

PR #5081, opened 2024-07-12
fix: avoid running pg_rewind when creating a replica from a snapshot

### LINKED ISSUE #5080: [Bug]: replication doesn't work for replicas bootstrapped from a snapshot backup of a primary

### Contact Details

leonardo.cecchi@gmail.com

### Version

1.23.2

1.30 (unsupported)

YAML manifest

### What happened?

Steps to reproduce the issue:

1. configure a local MinIO installation
2. apply the attached cluster definition
3. take a volume snapshot backup with `kubectl-cnpg backup cluster-example-with-volume-snapshot -m volumeSnapshot`
4. wait for the backup to finish
5. scale the cluster up to `2`

The new Pod will be in crash loop backoff, or the new replica will not ever attach to its primary via streaming replication.

That happens because the new PVC set is primed from the volume snapshots taken from a primary server.
When the new PostgreSQL instance starts up, it believes to be an old primary and run `pg_rewind` to align itself with the new primary.

`pg_rewind` will detect a non-clean shutdown (online backup) and start a postmaster in single-user mode.
The postmaster will do a checkpoint, diverging from the primary history while retaining the same timeline.

This happens only when a new replica is bootstrapped from an online volume snapshot backup taken from the primary node.

### Cluster resource

```shell
kind: Cluster
metadata:
  name: cluster-example-with-volume-snapshot
spec:
  instances: 1

  # Persistent storage configuration
  storage:
    storageClass: csi-hostpath-sc
    size: 1Gi
  walStorage:
    storageClass: csi-hostpath-sc
    size: 1Gi

  # Backup properties
  backup:
    volumeSnapshot:
       className: csi-hostpath-snapclass
    barmanObjectStore:
      destinationPath: s3://backups/
      endpointURL: http://minio:9000
      s3Credentials:
        accessKeyId:
          name: minio
          key: ACCESS_KEY_ID
        secretAccessKey:
          name: minio
          key: ACCESS_SECRET_KEY
      wal:
        compression: gzip
```

### Relevant log output

_No response_

### Code of Conduct

leonardoce:
The instance manager was starting `pg_rewind` against new replicas created from an online volume snapshot backup of the primary node. This happens because the primary PGDATA was snapshotted.

Since the PGDATA was not shutdown correctly (online backup), `pg_rewind` will start a postmaster process that will write a shutdown checkpoint record. By doing that, the history will diverge from the primary one while retaining the same timeline.

Fixes: #5080


litaocdl (approved):

leonardoce:
@litaocdl Your review on the E2e tests is not fine. You'll not trigger this bug just by scaling up a cluster.
You need a volume snapshot backup to be taken from the primary node, after the primary node have been created.

We need a separate E2e tests for this, and we need to be sure it works.
Can you please revert it back?
And, by the way, [here](https://github.com/cloudnative-pg/cloudnative-pg/pull/5081/files#diff-983baa1150278dc85dbfb8947b64abc72159029a15ddd251ab10a81c2d462668R731) we're scaling a cluster and not a Pod.

I'm adding a "do not merge" label. With the reviewed E2e tests we won't be sure to have pinpointed the issue.

litaocdl:
Hello @leonardoce, the e2e change is adding the case in the existing ordered test case, the snapshot is taken [here](https://github.com/cloudnative-pg/cloudnative-pg/pull/5081/files?diff=unified&w=0#diff-983baa1150278dc85dbfb8947b64abc72159029a15ddd251ab10a81c2d462668L639),  based on this snapshot we do a restore and later do a scaleup, so it is the scaleup using the snapshot taken from primary pod, we save time for creating another cluster. 

The label for scaledup pod is need to be changed, I added a section to validate the data exists in new pod is because, some times the pod is successfully joined but streaming is not started, so verify the data added after the snapshot is taken in the new pod of scaled cluster

fcanovai (approved):

PR #5097, opened 2024-07-16
wip: support `recovery_min_apply_delay`

### LINKED ISSUE #5079: [Bug]: recovery_min_apply_delay rejected by admissions webhook

Found the following issue: https://github.com/cloudnative-pg/cloudnative-pg/issues/4021

This implies that recovery_min_apply_delay is accepted as postgresql parameter, however, when attempting to apply this parameter via clusters yaml, we get an admissions webhook rejection "Invalid value: "1h" Can't set fixed configuration parameter."

The docs also state [the following](https://cloudnative-pg.io/documentation/1.23/postgresql_conf/#fixed-parameters):

```
Some PostgreSQL configuration parameters should be managed exclusively by the operator. The operator prevents the user from setting them using a webhook.

Users are not allowed to set the following configuration parameters in the postgresql section:

<snip>
recovery_min_apply_delay
```

So it seems we're at an impasse, where to implement a delayed replica we need to set a postgresql parameter, but the operator does not allow the postgresql parameter to be set by the user even though the documentation suggests this should be possible.

Is it possible to set recovery_min_apply_delay via another means that I've overlooked?

_No response_

_No response_

_No response_

### Additional context

_No response_

### Backport?

Yes

### Code of Conduct

armru:
Closes #5079


gbartolini:
Superseded by #5181

PR #5110, opened 2024-07-19
chore: reduce `current.Configuration` global variable usage

armru:


leonardoce (approved):

leonardoce:
#5134 is an addition to this PR that removes nearly all the usages, but it is complex to review and probably not worth the inconvenience.

mnencia (approved):

PR #5111, opened 2024-07-19
feat: add support for CNPG-I `SetStatusInCluster` capability

armru:
This patch adds support for the CNPG-I capability [SetStatusInCluster](https://github.com/cloudnative-pg/cnpg-i/blob/5844b833f4eb1104110cd856d13f45d42d7828a6/proto/operator.proto#L22), allowing the plugins that implement this new capability to store their state as a raw JSON string inside the `Cluster` status.

Note for the reviewers:
- No issue has opened for this PR, given that CNPG-I is in active development and not yet v1.0.
- No backport.

## Release notes

CNPG Operator now supports the CNPG-I capability [SetStatusInCluster](https://github.com/cloudnative-pg/cnpg-i/blob/5844b833f4eb1104110cd856d13f45d42d7828a6/proto/operator.proto#L22)

### DIFF
--- a/internal/cnpi/plugin/client/cluster.go
+++ b/internal/cnpi/plugin/client/cluster.go
@@ -19,6 +19,7 @@ package client
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"slices"
 
@@ -92,6 +93,11 @@ func (data *data) MutateCluster(ctx context.Context, object client.Object, mutat
 	return nil
 }
 
+var (
+	errInvalidJSON      = errors.New("invalid json")
+	errSetClusterStatus = errors.New("set cluster status invokation failed")
+)
+
 func (data *data) SetClusterStatus(ctx context.Context, cluster client.Object) (map[string]string, error) {
 	contextLogger := log.FromContext(ctx)
 	serializedObject, err := json.Marshal(cluster)
@@ -120,7 +126,7 @@ func (data *data) SetClusterStatus(ctx context.Context, cluster client.Object) (
 		response, err := plugin.OperatorClient().SetClusterStatus(ctx, &request)
 		if err != nil {
 			pluginLogger.Error(err, "Error while calling SetClusterStatus")
-			return nil, err
+			return nil, fmt.Errorf("%w: %w", errSetClusterStatus, err)
 		}
 
 		if len(response.JsonStatus) == 0 {
@@ -130,7 +136,7 @@ func (data *data) SetClusterStatus(ctx context.Context, cluster client.Object) (
 		if err := json.Unmarshal(response.JsonStatus, &json.RawMessage{}); err != nil {
 			contextLogger.Error(err, "found a malformed json while evaluating SetClusterStatus response",
 				"pluginName", plugin.Name())
-			return nil, err
+			return nil, fmt.Errorf("%w: %w", errInvalidJSON, err)
 		}
 
 		pluginStatuses[plugin.Name()] = string(response.JsonStatus)
--- a/internal/cnpi/plugin/client/cluster.go
+++ b/internal/cnpi/plugin/client/cluster.go
@@ -124,7 +124,13 @@ func (data *data) SetClusterStatus(ctx context.Context, cluster client.Object) (
 		}
 
 		if len(response.JsonStatus) == 0 {
+			contextLogger.Trace("json status is empty, skipping it", "pluginName", plugin.Name())
 			continue
+		}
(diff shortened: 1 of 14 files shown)


internal/cnpi/plugin/client/cluster.go
```diff
@@ -92,6 +93,58 @@ func (data *data) MutateCluster(ctx context.Context, object client.Object, mutat
...
+			err,
+		)
+	}
+
+	pluginStatuses := make(map[string]string)
+	for idx := range data.plugins {
+		plugin := data.plugins[idx]
+
+		if !slices.Contains(plugin.OperatorCapabilities(), operator.OperatorCapability_RPC_TYPE_SET_CLUSTER_STATUS) {
+			continue
```
ringerc, line 117: Cool, so it's an optional new capability. Good.

internal/cnpi/plugin/client/cluster.go
```diff
@@ -92,6 +93,58 @@ func (data *data) MutateCluster(ctx context.Context, object client.Object, mutat
...
+		response, err := plugin.OperatorClient().SetClusterStatus(ctx, &request)
+		if err != nil {
+			pluginLogger.Error(err, "Error while calling SetClusterStatus")
+			return nil, fmt.Errorf("%w: %w", errSetClusterStatus, err)
+		}
+
+		if len(response.JsonStatus) == 0 {
+			contextLogger.Trace("json status is empty, skipping it", "pluginName", plugin.Name())
+			continue
+		}
```
ringerc, line 135: Does this prevent the plugin from clearing a previously set status?

ringerc:
Overall looks good to me.

internal/cnpi/plugin/client/cluster.go
leonardoce: An empty string does not represent a valid JSON object.
The status can be cleared by passing `{}`.
Having merged that, we're going to validate it in `cnpg-i-machinery` and in the operator too.

leonardoce (approved):

leonardoce:
https://github.com/EnterpriseDB/cloudnative-pg/actions/runs/10215314666

gabriele-wolfox (approved):

internal/cnpi/plugin/client/contracts.go
```diff
@@ -67,9 +67,12 @@ type ClusterCapabilities interface {
...
 	ValidateClusterUpdate(
 		ctx context.Context,
-		oldObject client.Object,
-		newObject client.Object,
+		cluster client.Object,
+		mutatedCluster client.Object,
 	) (field.ErrorList, error)
+
+	// SetClusterStatus returns a map of [pluginName]: statuses to be assigned to the cluster
+	SetClusterStatus(ctx context.Context, cluster client.Object) (map[string]string, error)
```
sxd, line 75: This name is really weird since it's not setting the status of a cluster but the status of a plugin, wouldn't be better to have something like `SetPluginStatus` ? what do you think @armru @leonardoce

armru: The GRPC was defined while trying to maintain coherence with the other `operator` capabilities.

Given that the plugin is invoking that capability and is setting its status inside the cluster is not really that misleading, but I'm open for eventual changes

internal/cnpi/plugin/client/cluster.go
armru: This PR contains the validation, line 136. In the original draft it was not very accurate but I improved it.

internal/cnpi/plugin/client/contracts.go
sxd: The status is inside the cluster so you understand is the status regarding to that plugin in that status, I don't see that like the issue, the issue is that "SetClusterStatus" implies that will set the status of the cluster, not the plugin in that cluster, but I see that this is already defined in the cnpg-i so there's nothing to do here

armru: Would SetStatusInCluster be better?
We can still modify cnpg-i given that this gprc is not used yet

mnencia: I think that `SetClusterStatus` is a coherent name. If we want to make it more explicit, we could add `Plugin` to the name and use `SetClusterPluginStatus`.

mnencia: After a brief chat with @sxd we agreed that the best name to avoid ambiguity is `SetStatusInCluster` as proposed by @armru.

sxd: Yup! came to update too and was already updated :D

mnencia (approved):

PR #5134, opened 2024-07-23
chore(refactor): remove configuration global variable

leonardoce:


leonardoce:
Sounds like too much inconvenience for what we gain. I'm closing it.

PR #5206, opened 2024-07-31
feat: stricter readiness probe for replicas

leonardoce:
Enhance the readiness probe functionality to avoid marking streaming replicas as ready before establishing a successful connection with the corresponding primary.

This improvement allows users to quickly identify and address inconsistencies between replicas and primary instances.

# Release Notes

The readiness probe now fails for streaming replicas that were never connected to the primary instance, allowing incoherent replicas to be discovered promptly.


leonardoce:
https://github.com/EnterpriseDB/cloudnative-pg/actions/runs/10182207522

leonardoce:
https://github.com/EnterpriseDB/cloudnative-pg/actions/runs/10183490630

gbartolini:
I think this is a fix, and I vote for inclusion in 1.24 directly. We should not declare a replica ready if it's never been able to establish a replication connection.

jsilvela (approved):

armru (approved):

sxd (approved):

sxd (approved):

PR #5207, opened 2024-07-31
test(e2e): prevent failures in the corruption test

leonardoce:
`rm -fr` fails when a file is created meanwhile the target directory is removed.

This patch replaces it with `find` that is not prone to that error.

# Release notes

NONE


sxd (approved):

jsilvela (approved):

leonardoce:
E2e: https://github.com/EnterpriseDB/cloudnative-pg/actions/runs/10178293086

PR #5209, opened 2024-07-31
fix: drop stale replica cluster connections

### LINKED ISSUE #5197: [Bug]: after promotion of a new designated primary, streaming is inconsistent

### Contact Details

jaime.silvela@mailfence.com

### Version

1.23.2

1.29

YAML manifest

### What happened?

This happens when we do a promotion to a new designated primary, in a replica cluster.
The old primary will correctly stream from the new designated primary.
However, any other replica will continue to stream from the old primary.

Old designated primary: cluster-replica-tls-1
New designated primary: cluster-replica-tls-2

Let's see the streaming connections reading from the new designated primary.
The old primary is now streaming from the new one:

``` console
# select application_name, state from pg_stat_replication;
   application_name    |   state   
-----------------------+-----------
 cluster-replica-tls-1 | streaming
(1 row)
```

Let's look at the streaming connections reading from the old primary.
The other replica instance never switched to the new designated primary:

``` console
# select application_name, state from pg_stat_replication;
   application_name    |   state   
-----------------------+-----------
 cluster-replica-tls-3 | streaming
(1 row)
```

The kubectl plugin does not distinguish the situation here, and shows all streaming connections OK:

``` console
Instances status
Name                   Database Size  Current LSN  Replication role              Status  QoS         Manager Version  Node
----                   -------------  -----------  ----------------              ------  ---         ---------------  ----
cluster-replica-tls-1  29 MB          0/C000000    Standby (in Replica Cluster)  OK      BestEffort  1.23.2           pg-operator-e2e-v1-30-2-worker3
cluster-replica-tls-2  29 MB          0/C000000    Designated primary            OK      BestEffort  1.23.2           pg-operator-e2e-v1-30-2-worker2
cluster-replica-tls-3  29 MB          0/C000000    Standby (in Replica Cluster)  OK      BestEffort  1.23.2           pg-operator-e2e-v1-30-2-worker
```

### Cluster resource

```shell
This is the plain cluster-replica-tls.yaml from docs/src/samples:

 yaml
kind: Cluster
metadata:
  name: cluster-replica-tls
spec:
  instances: 3

  bootstrap:
    pg_basebackup:
      source: cluster-example

  replica:
    primary: cluster-example
    source: cluster-example

  storage:
    size: 1Gi

  externalClusters:
  - name: cluster-example
    connectionParameters:
      host: cluster-example-rw.default.svc
      user: streaming_replica
      sslmode: verify-full
      dbname: postgres
    sslKey:
      name: cluster-example-replication
      key: tls.key
    sslCert:
      name: cluster-example-replication
      key: tls.crt
    sslRootCert:
      name: cluster-example-ca
      key: ca.crt
```
```


### Relevant log output

_No response_

### Code of Conduct

### PR DESCRIPTION
**Pull request by leonardoce**
This patch makes the instance manager terminate all existing operator-related replication connections following a role change in a replica cluster.

For context, demoting a PostgreSQL instance involves shutting it down, adjusting the necessary signal files and configuration, and then restarting it, which inherently disconnects all existing connections.

In a replica cluster, demotion is unnecessary since it comprises replicas only. In this scenario, only the primary_conninfo parameter needs to be modified, which doesn't require a shutdown. However, this also implies that replicas receiving data from the old primary won't have their connections terminated.

Consequently, high-availability replicas connected to the previous primary will remain connected, necessitating manual intervention to terminate those connections and re-establish them with the new endpoint.

Closes: #5197 

## Release notes

The instance manager will now terminate all existing operator-related replication connections following a role change in a replica cluster.

### DIFF
--- a/tests/e2e/asserts_test.go
+++ b/tests/e2e/asserts_test.go
@@ -179,6 +179,50 @@ func AssertSwitchoverWithHistory(
 			}, timeout).ShouldNot(HaveOccurred())
 		})
 	}
+
+	if isReplica {
+		By("verifying that the all standbys are streaming from the new primary", func() {
+			timeout := 120
+			commandTimeout := time.Second * 10
+			Eventually(func(g Gomega) {
+				podList, err := env.GetClusterPodList(namespace, clusterName)
+				g.Expect(err).ToNot(HaveOccurred())
+
+				var standbys []interface{}
+				for _, pod := range podList.Items {
+					if specs.IsPodStandby(pod) {
+						standbys = append(standbys, pod.Name)
+					}
+				}
+
+				for _, pod := range podList.Items {
+					stdout, _, err := env.ExecCommand(
+						env.Ctx,
+						pod,
+						specs.PostgresContainerName,
+						&commandTimeout,
+						"psql", "-U", "postgres", "-tAc",
+						"select string_agg(application_name, ',')  from pg_stat_replication;",
+					)
+					g.Expect(err).ToNot(HaveOccurred())
+
+					if specs.IsPrimary(pod.ObjectMeta) {
+						appNames := strings.Split(strings.TrimSpace(stdout), ",")
+						g.Expect(appNames).To(
+							ContainElements(standbys...),
+							"not all standbys are streaming from the new primary "+pod.Name,
+						)
+					} else {
+						g.Expect(strings.TrimSpace(stdout)).To(
+							BeEmpty(),
+							fmt.Sprintf("the standby %s should not stream to any other instance", pod.Name),
+						)
+					}
+				}
+
+			}, timeout).ShouldNot(HaveOccurred())
+		})
+	}
 }
 
 // AssertCreateCluster creates the cluster and verifies that the ready pods
--- a/tests/e2e/asserts_test.go
+++ b/tests/e2e/asserts_test.go
@@ -178,16 +178,14 @@ func AssertSwitchoverWithHistory(
 				return nil
 			}, timeout).ShouldNot(HaveOccurred())
 		})
-	}
-
-	if isReplica {
+	} else {
 		By("verifying that the all standbys are streaming from the new primary", func() {
 			timeout := 120
 			commandTimeout := time.Second * 10
 			Eventually(func(g Gomega) {
 				podList, err := env.GetClusterPodList(namespace, clusterName)
 				g.Expect(err).ToNot(HaveOccurred())
-
+				g.Expect(podList.Items).To(HaveLen(oldPodListLength))
 				var standbys []interface{}
 				for _, pod := range podList.Items {
 					if specs.IsPodStandby(pod) {
@@ -219,7 +217,6 @@ func AssertSwitchoverWithHistory(
 						)
 					}
 				}
-
 			}, timeout).ShouldNot(HaveOccurred())
 		})
 	}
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by leonardoce** [comment]
https://github.com/EnterpriseDB/cloudnative-pg/actions/runs/10183524076

**Review by YanniHu1996 — APPROVED** [review]

File: tests/e2e/asserts_test.go [hunk 1]
```diff
@@ -178,6 +178,47 @@ func AssertSwitchoverWithHistory(
...
+						)
+					} else {
+						g.Expect(strings.TrimSpace(stdout)).To(
+							BeEmpty(),
+							fmt.Sprintf("the standby %s should not stream to any other instance", pod.Name),
+						)
+					}
+				}
+			}, timeout).ShouldNot(HaveOccurred())
+		})
```
armru, line 221: Hello @YanniHu1996, what are we trying to achieve here? 

I've added line 188, because the list could have returned a different number of instances, and the test would have still succeeded. 

cc @leonardoce if you can help me understand given that you wrote the patch and have better insights

tests/e2e/asserts_test.go
```diff
@@ -178,6 +178,47 @@ func AssertSwitchoverWithHistory(
...
 		})
+	} else {
+		By("verifying that the all standbys are streaming from the new primary", func() {
+			timeout := 120
+			commandTimeout := time.Second * 10
+			Eventually(func(g Gomega) {
+				podList, err := env.GetClusterPodList(namespace, clusterName)
+				g.Expect(err).ToNot(HaveOccurred())
+				g.Expect(podList.Items).To(HaveLen(oldPodListLength))
+				var standbys []interface{}
```
leonardoce, line 189: I believe this is a `[]string`

tests/e2e/asserts_test.go
leonardoce: I think Yanni is trying to assess that:

1. the HA replicas are connected to the designed primary and streaming WALs from there
2. the HA replicas are not streaming to anyone else
3. the designed primary is streaming from its source

We don't actually test the actual topology as we don't cover the third point.
BTW: why don't we use the cluster definition? We really need just the pod names and the name of the current and designated primary.

tests/e2e/asserts_test.go
YanniHu1996: This adheres to the interface of `ContainElements`

https://github.com/onsi/gomega/blob/master/matchers.go#L420-L424

YanniHu1996: oh, the `ContainElements` flattens the slice internally, please ignore the above comment 

https://github.com/onsi/gomega/blob/master/matchers/consist_of.go#L62-L72

YanniHu1996:
the e2e from fork 
https://github.com/EnterpriseDB/cloudnative-pg/actions/runs/10346879509/job/28636233232

tests/e2e/asserts_test.go
YanniHu1996: Thanks for your comments, I submitted the commit based on them

https://github.com/cloudnative-pg/cloudnative-pg/pull/5209/commits/8c657c743598816b9a048cfc0b510d06c2200683

mnencia (approved):

fcanovai (approved):

fcanovai:
We do have a few failures due to https://github.com/cloudnative-pg/cloudnative-pg/pull/5236, but they are not tied to the PR here.

PR #5226, opened 2024-08-05
fix(docs): increase `xid_age` alerting limit

TillHein:
The default xid_age value for a forced auto-vacuum is set to 200.000.000. The new values reflect these defaults.

```
#autovacuum_freeze_max_age = 200000000  # maximum XID age before forced vacuum
```

Issues with xid occur at an age of ~2.000.000.000 so there is plenty of time to react to the alert with the increased value.


docs/src/samples/monitoring/prometheusrule.yaml
```diff
@@ -29,7 +29,7 @@ spec:
         description: Over 150,000,000 transactions from frozen xid on pod {{ $labels.pod  }}
         summary: Number of transactions from the frozen XID to the current one
       expr: |-
-        cnpg_pg_database_xid_age > 150000000
+        cnpg_pg_database_xid_age > 220000000
```
gbartolini, line 32:
```suggestion
        cnpg_pg_database_xid_age > 300000000
```

gbartolini (approved):

leonardoce (approved):

PR #5232, opened 2024-08-06
Webhook only warns about nodeMaintenance window if used in PDB-disabling mode

AlbinoDrought:
nodeMaintenanceWindow is also used with reusePVC: false which does not disable PDBs, leading to an irrelevant warning emitted here.

Originally Introduced: bf4248a33d3ffcec2a99c9bb5b5e7f94bc80ce91 
Ref: https://github.com/cloudnative-pg/cloudnative-pg/issues/2570#issuecomment-1884962321


sxd:
Closing due to inactivity and cannot be easy rebased

PR #5236, opened 2024-08-07
test: make the test for replica detach more reliable

jsilvela:
The replica detach test is flaky, failing about 5-10% of the time.
E.g. https://github.com/cloudnative-pg/cloudnative-pg/actions/runs/10269046932/job/28413809548

This PR rewrites the test to consider the cluster Phase, rather than
relying on the transition time of the Ready condition.

### DIFF
--- a/tests/e2e/asserts_test.go
+++ b/tests/e2e/asserts_test.go
@@ -1093,6 +1093,8 @@ func AssertDetachReplicaModeCluster(
 		Eventually(func(g Gomega) {
 			cluster, err := env.GetCluster(namespace, replicaClusterName)
 			g.Expect(err).ToNot(HaveOccurred())
+			g.Expect(cluster.Spec.ReplicaCluster.Enabled).NotTo(BeNil(), "replica.enabled was nil")
+			g.Expect(*cluster.Spec.ReplicaCluster.Enabled).To(BeFalse(), "replica.enabled still true")
 			condition, err := testsUtils.GetConditionsInClusterStatus(namespace, cluster.Name, env,
 				apiv1.ConditionClusterReady)
 			g.Expect(err).ToNot(HaveOccurred())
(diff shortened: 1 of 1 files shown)


jsilvela:
Initial approach is naive. Discarding.

In the failed CI/CD run we can see that the cluster got a restart after the detach (due to the app-secret being created and changing the PodSpec)

``` yaml
{"level":"info","ts":"2024-08-06T16:18:29Z","msg":"Restarting primary instance without a switchover first","controller":"cluster","controllerGroup":"postgresql.cnpg.io","controllerKind":"Cluster","Cluster":{"name":"cluster-replica-basicauth","namespace":"replica-mode-basic-auth-3854"},"namespace":"replica-mode-basic-auth-3854","name":"cluster-replica-basicauth","reconcileID":"e885266a-a583-4727-a498-8984ec1ed3ee","primaryPod":"cluster-replica-basicauth-1","reason":"original and target PodSpec differ in volumes: element app-secret has been added"}
```

This actually registers a non-healthy cluster phase, which should have a non-ready transition to the Ready condition.
``` go
contextLogger.Info("Restarting primary instance without a switchover first",
			"reason", reason)
		err := r.RegisterPhase(ctx, cluster, apiv1.PhaseInplaceDeletePrimaryRestart, reason)
		if err != nil {
			return false, err
		}
```

fcanovai:
The failing tests has highlighted two bugs. One was related on the operator getting stale information and patching the status with those (https://github.com/cloudnative-pg/cloudnative-pg/pull/5387).

Then we seem to have additional failures because of https://github.com/cloudnative-pg/cloudnative-pg/issues/5401.

So, for now the tests seems to work correctly and actually get bugs. I'd avoid changing it for now and re-evaluate after the second fix is in place.

fcanovai:
The fixes in the above mentioned issues seem to have solved the failures we had. Closing this PR.

Issue #13, opened 2022-04-11
Logical replication (`Publication` and `Subscription` CRD)


gbartolini:
Declarative way to publish changes to one or more tables using logical replication (publications), as well as to subscribe to any Postgres 10+ publication (subscriptions) and consume those events.

- [ ] [split from `#13`] plugin command to set up all the objects needed for logical replication

---

POC: https://github.com/cloudnative-pg/cloudnative-pg/pull/5329

The POC assumes the Database CRD/controller are there already.
Like the Database code, the ObservedGeneration field in the Status allows skipping reconciliation if there’s nothing new.
Like the Database code, there is a ReclaimPolicy field determining whether objects are deleted or retained for End-of-life

API’s:

``` yaml
apiVersion: postgresql.cnpg.io/v1
 kind: Publication
 metadata:
   name: publication-sample
 spec:
   name: pub
   dbname: app
   cluster:
     name: cluster-example
   target:
     allTables: {}
```

``` yaml
apiVersion: postgresql.cnpg.io/v1
 kind: Subscription
 metadata:
   name: subscription-sample
 spec:
   name: sub
   dbname: app
   publicationName: pub
   cluster:
     name: cluster-example-dest
   externalClusterName: cluster-example

```

### Questions

Do we want to detect changes on the side of Postgres? At the moment we use the ObservedGeneration to decide if we may skip reconciliation. Changes made within Postgres are ignored, then

^ the consensus about the above question is not to look for changes on the side of Postgres

### References

https://www.postgresql.org/docs/current/sql-createpublication.html
https://www.postgresql.org/docs/current/sql-alterpublication.html
https://www.postgresql.org/docs/current/sql-createsubscription.html
https://www.postgresql.org/docs/current/sql-altersubscription.html

ns-gzhang:
Any plan to support Debezium for CDC, which requires logical replication I believe?

abratanich:
Greetings.  is it planned to add such support at the HA level in the near future?

ns-jhu:
do we know when this feature is available ? without this , software such as debezium doesn't work well when postgres failover happens.  In Zalando, we can  config the replication_slot in the secondary, which is  called permanent replication_slot, to enable debezium to work with postgres change event streaming seamlessly when failover happens. that is a highly desirable feature. 

I checked version 1.18, and see this:
```	rows, err := db.QueryContext(
2		ctx,
[...]

PR #5254, opened 2024-08-08
feat: declarative database management, logical subscriptions and publications

### LINKED ISSUE #4679: Declarative database management (`Database` CRD)
Provide a way to manage Postgres databases in a declarative manner, by looking at the options in the [CREATE DATABASE](https://www.postgresql.org/docs/14/sql-createdatabase.html) DDL command. This can be done in multiple steps, as it is linked with role management, bootstrapping (application database, import), and, potentially, tablespaces too. If we extend functionalities to the [ALTER DATABASE](https://www.postgresql.org/docs/14/sql-alterdatabase.html) command, we might also cover setting configuration parameters for a specific database.

*NOTE: the list of sub-issues below was written before GitHub started offering sub-issues. The remaining sub-issues are tracked now in the Sub-issues heading in this GH page.*

- [x] #5501
- [x] #5502
- [x] #5521
- [x] #5522
- [x] #5523
- [x] #5593
- [x] #5587
- [x] #5530
- [x] #5680
- [x] #5681
- [x] #5544
- [x] #5627
- [x] #5629
- [x] #5692
- [x] #5709
- [x] #5722
- [x] #5773
---

## Declarative Database Management

Provide a way to manage Postgres databases declaratively, by looking at the options in the [CREATE DATABASE](https://www.postgresql.org/docs/14/sql-createdatabase.html) DDL command. This can be done in multiple steps, as it is linked with role management, bootstrapping (application database, import), and, potentially, tablespaces too. If we extend functionalities to the [ALTER DATABASE](https://www.postgresql.org/docs/14/sql-alterdatabase.html) command, we might also cover setting configuration parameters for a specific database.

Initial POC from Leonardo: https://github.com/cloudnative-pg/cloudnative-pg/pull/5254

There is a sibling issue / PR for declarative Publication and Subscription https://github.com/cloudnative-pg/cloudnative-pg/pull/5329
… for another epic and another document. Here for reference and limit-setting.

### Proposed approach from Leonardo’s POC

Have a dedicated CRD, with the following UI:

``` yaml
 kind: Database
 metadata:
   name: db-one
 spec:
   name: one
   owner: app
   cluster:
     name: cluster-example
```

And a dedicated controller.
A field called `ObservedGeneration` in the DatabaseStatus struct helps prevent reconciliation if the CRD in its current Generation has already been applied in the PostgreSQL instance.

### Questions

- In Leonardo’s POC, the Name of the Database CRD is immutable.

``` go
// The name inside PostgreSQL
// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="name is immutable"
Name string `json:"name"`
```

This makes good sense from a Kubernetes perspective, but raises the question, how are we going to support ALTER DATABASE name RENAME TO new_name?

- The POC includes ALTER DATABASE name SET TABLESPACE new_tablespace. If the tablespace does not yet exist, do we patch the cluster adding the new tablespace, wait for that to succeed, then proceed with the ALTER?

- This feature opens the door for GRANT PRIVILEGES on roles for databases. Is that something we’ll want in the future as part of the CNPG CRD’s?
- Should we do anything about schemas?
- Should we support “adopting” existing databases? This seems to be already possible with the current implementation. In case a database already exists, you can create its related resource and the DB will become “managed”. 
- Do we want to detect changes on the side of Postgres? At the moment we use the ObservedGeneration to decide if we may skip reconciliation. Changes made within Postgres are ignored, then

### Reference

Options from CREATE DATABASE
https://www.postgresql.org/docs/current/sql-createdatabase.html

[ OWNER [=] user_name ]
[ TEMPLATE [=] template ]
[ ENCODING [=] encoding ]
[ STRATEGY [=] strategy ]
[ LOCALE [=] locale ]
[ LC_COLLATE [=] lc_collate ]
[ LC_CTYPE [=] lc_ctype ]
[ ICU_LOCALE [=] icu_locale ]
[ ICU_RULES [=] icu_rules ]
[ LOCALE_PROVIDER [=] locale_provider ]
[ COLLATION_VERSION = collation_version ]
[ TABLESPACE [=] tablespace_name ]
[ ALLOW_CONNECTIONS [=] allowconn ]
[ CONNECTION LIMIT [=] connlimit ]
[ IS_TEMPLATE [=] istemplate ]
[ OID [=] oid ]

Options from ALTER DATABASE

https://www.postgresql.org/docs/current/sql-alterdatabase.html
ALTER DATABASE name [ [ WITH ] option [ ... ] ]

where option can be:

    ALLOW_CONNECTIONS allowconn
    CONNECTION LIMIT connlimit
    IS_TEMPLATE istemplate

ALTER DATABASE name RENAME TO new_name

ALTER DATABASE name OWNER TO { new_owner | CURRENT_ROLE | CURRENT_USER | SESSION_USER }

ALTER DATABASE name SET TABLESPACE new_tablespace

ALTER DATABASE name REFRESH COLLATION VERSION

ALTER DATABASE name SET configuration_parameter { TO | = } { value | DEFAULT }
ALTER DATABASE name SET configuration_parameter FROM CURRENT
ALTER DATABASE name RESET configuration_parameter
ALTER DATABASE name RESET ALL

leonardoce:
This patch adds a prototype of the declarative database management feature. The `.spec.managed.databases` section allows the user to specify a number of databases to be reconciled and created into the cluster.

The `.status.databasesStatus` section contains information about the reconciliation process.

Closes: #4679


farodin91:
I will like to know, if it would be better to move outside of the cluster resource. 
Just an idea. A bit like the concept of strimzi with the resource for topics.

api/v1/cluster_types.go
```diff
@@ -2804,6 +2811,59 @@ func (roleConfiguration *RoleConfiguration) GetRoleInherit() bool {
 	return true
 }
 
+// DatabaseConfiguration is the representation in Kubernetes of a PostgreSQL Database
```
jsilvela, line 2814: I know this patch is a prototype.
Just mentioning here, perhaps for a separate ticket, that we're missing the locale and collate settings.

pkg/management/postgres/webserver/database.go
```diff
@@ -0,0 +1,293 @@
...
+	if err != nil {
+		log.Debug(
+			"Error while getting DB connection",
+			"dbname", dbname,
+			"err", err.Error())
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+
+	sqlCreateDatabase := fmt.Sprintf("CREATE DATABASE %s ", dbname)
```
jsilvela, line 266: this being a PUT, we might get it for an existing database, so we'd also need to have ALTER statements if the DB already exists, no?
I know the PATCH is the right verb for that.
So maybe if we try to PUT a DB that already exists, we could reject it.

jsilvela (commented):
This is very nice. I like the design a lot.
I know it's meant to be the first of several PR's, so left a few questions inline.
Code looks great, only the general suggestion that we could reuse the constants from the Go http package. E.g.
200 -> http.StatusOK
"GET" -> http.MethodGet

leonardoce:
> I will like to know, if it would be better to move outside of the cluster resource. Just an idea. A bit like the concept of strimzi with the resource for topics.

Yes, I thought about that, too. This is like a mini-operator inside CNPG, reconciling a section of the Spec inside the relative section of the status.

I used this design because we've already `cluster.spec.managed.services` and `cluster.spec.managed.roles`, and it felt weird to put the database in a separate resource. Given that, a new user would just look for databases under `cluster.spec.managed.databases`, and it would be surprising for him to know that they're in a different resource.

Anyway, both designs have merits. 

If I were to start from scratch, I'd probably use a separate resource.

api/v1/cluster_types.go
leonardoce: Agree. I did that to avoid having too many parameters that are usable only when creating a database.
You can't change them after you create your DB.

The same applies to **encoding**.

pkg/management/postgres/webserver/database.go
leonardoce: PATCH is already implemented, too. We could keep the one we like best between PATCH and PUT and use just that.

farodin91:
> > I will like to know, if it would be better to move outside of the cluster resource. Just an idea. A bit like the concept of strimzi with the resource for topics.
> 
> 
> 
> Yes, I thought about that, too. This is like a mini-operator inside CNPG, reconciling a section of the Spec inside the relative section of the status.
> 
> 
> 
> I used this design because we've already `cluster.spec.managed.services` and `cluster.spec.managed.roles`, and it felt weird to put the database in a separate resource. Given that, a new user would just look for databases under `cluster.spec.managed.databases`, and it would be surprising for him to know that they're in a different resource.
> 
> 
> 
> Anyway, both designs have merits. 
> 
> 
> 
> If I were to start from scratch, I'd probably use a separate resource.


I understand. I think for tenant use case it would be great, if you manage the database with per tenant based resource instead of all have to access the main cluster resource. Roles and user could be managed per database in the future.

leonardoce:
> I understand. I think for tenant use case it would be great, if you manage the database with per tenant based resource instead of all have to access the main cluster resource.

I got it. Unfortunately it's not easy to split PostgreSQL in tenants many resources and configurations are shared. 

Anyway, I really like having a separate resource and I may come up with an implementation of it. I find the design of Strimzi really elegant. 

> Roles and user could be managed per database in the future.

Yes, but PostgreSQL users are global objects. This has pros and cons.

leonardoce:
What I really like about this approach is the database controller being really neatly separated from the rest of the operator codebase.

farodin91:
> This is what I came with:
> 
> 
> 
> ```
> 
> $ k get database db-one -o yaml 
> 
> apiVersion: postgresql.cnpg.io/v1
> 
> kind: Database
> 
> metadata:
> 
>   [...]
> 
>   generation: 1
> 
>   name: db-one
> 
>   namespace: default
> 
> spec:
> 
>   cluster:
> 
>     name: cluster-example
> 
>   name: one
> 
>   owner: app
> 
> status:
> 
>   observedGeneration: 1
> 
>   ready: true
> 
> 
> 
> $ k get cluster
> 
> NAME              AGE   INSTANCES   READY   STATUS                     PRIMARY
> 
> cluster-example   36m   3           3       Cluster in healthy state   cluster-example-1
> 
> 
> 
> $ k get database
> 
> NAME     CLUSTER           PG NAME   READY   ERROR
> 
> db-one   cluster-example   one       true    
> 
> db-two   cluster-example   two               ERROR: role "app-two" does not exist (SQLSTATE 42704)...
> 
> ```
> 
> 
> 
> It's now up to us to choose with API we like best.
> 
> 
> 
> cc: @farodin91 

I really like the new variant. 

Ideas for future:
- Init Script per database
- Backup per database
- User managed per database
- Plugin Support
- new metrics

I never used tablespace, but something like to creation of tablespace per resource. 

I'm just thinking loud.

pkg/management/postgres/webserver/database.go
leonardoce: Resolving this as this PR is not using the webserver anymore.

gbartolini:
I know we have the `managed` stanza and have already implemented `roles` underneath that, but moving forward, the direction of CloudNativePG is to learn from our experience and modularise the API, with the goal to delegate responsibilities to different CRDs.

Do you see this as a foreseeable path:

- implement declarative databases (and contained objects such as publications, subscriptions, foreign data wrappers and potentially extensions?) as separate CRDs
- in a second phase, implement roles support with CRDs, allowing users to keep using the existing method, in a mutually exclusive way

In the future v2 API (not yet on the horizon) converge on the CRD model.

If possible, I would like to see this PoC in multiple PRs, at least for the database CRD. It goes without saying that we might make changes to the roadmap, given that there's already code for pubs/subs.

Thoughts?

gbartolini:
> If possible, I would like to see this PoC in multiple PRs, at least for the database CRD.

Could you please split the work in two PRs? That is:

- keep databases in this PR
- move pubs/subs in a new PR referencing #13 

I have started to play with it, and it looks great. It needs a lot of documentation work, and we clearly need to address:

- relationship with default application database (disjointed)
- deletion of `Database` objects vs deletion in the database (add an option such as `reclaimPolicy: retain/delete`?)
- relationship with roles

Thanks.

sxd:
One of the nice concept of having the database management and creation out of the cluster, or what I think at least, it's that we can have the proper permissions in that object to create a kubernetes dba that will manage only databases users and permissions allowing the possibility to reduce the amount of permissions per users when you only need to add users or create a database.

On the other hands, this could lead us to enable an external entity to manage the users with access to a database, there are many IdP out in the world that can be easily integrated while we separated the roles.

The only things that worries me here is the restore process and how this could be affected having these two objects separated, cluster and database I mean, but I think that the solution here will go more on the way users manage the Kubernetes objects, since they will need to restore the objects anyway, but what if we want to restore only one database in that cluster? that will create some complications in the process.

My two cents :)

philippemnoel:
Will this have any impact on the ability to easily install extensions in the databases? We notably install different extensions in different databases

leonardoce:
> Will this have any impact on the ability to easily install extensions in the databases? We notably install different extensions in different databases

No. This allows you to create/drop/alter databases, each one being represented by its own CR. 

There's nothing reconciling the database content.

mnencia:
> [Gabriele]
> I know we have the `managed` stanza and have already implemented `roles` underneath that, but moving forward, the direction of CloudNativePG is to learn from our experience and modularise the API, with the goal to delegate responsibilities to different CRDs.
> 
> Do you see this as a foreseeable path:
> 
> * implement declarative databases (and contained objects such as publications, subscriptions, foreign data wrappers and potentially extensions?) as separate CRDs
> * in a second phase, implement roles support with CRDs, allowing users to keep using the existing method, in a mutually exclusive way
> 

This looks like a good plan. Our Cluster CRD is already complex enough.

> [Leonardo]
> What I really like about this approach is the database controller being really neatly separated from the rest of the operator codebase.

I agree. Managing the databases in a separate CR is easier to understand bot both the user and the developer.

Issue #4423, opened 2024-05-02
[Bug]: Operator restarting due to DetectAvailableArchitectures()


sxd:
### Is there an existing issue already for this bug?

- [X] I have searched for an existing issue, and could not find anything. I believe this is a new bug.

### I have read the troubleshooting guide

- [X] I have read the troubleshooting guide and I think this is a new bug.

### I am running a supported version of CloudNativePG

- [X] I have read the troubleshooting guide and I think this is a new bug.

### Version

1.23.0

### What version of Kubernetes are you using?

1.30 (unsuppprted)

### What is your Kubernetes environment?

Self-managed: kind (evaluation)

### How did you install the operator?

YAML manifest

### What happened?

This issue seems to be present overall, but so far I’ve only been able to reproduce it in certain cloud providers (mainly EKS, but also AKS).

What’s happening is that utils.DetectAvailableArchitectures() is slowing down RunController() enough so that the ReadinessProbePeriod 10 seconds are not respected anymore, and the pod gets killed by the Kubelet (and thus restart).

DetectAvailableArchitectures() should be calculating each architecture’s sha256 hash asynchronously, so it shouldn’t lock the startup of the manager. 
Needs more investigation to understand if the function is not working properly or if we are just hitting the timeout. 

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

mnencia:
Closed by 2be105b0fa4871b8f423e06d41bbbc3dcb5c8cda

PR #5297, opened 2024-08-15
test: improve upgrade E2E test traceability

### LINKED ISSUE #5331: [Feature]: smarter E2E test for online upgrades

We've been seeing some problems with the E2E test for online upgrade:

- In `main`, the "latest operator" used for the upgrade test is 1.24.0-rc1. This is useless as most people will be updating to 1.24.0 from 1.23.3 or previous releases
- The online upgrade test may fail for valid reasons if the operator upgrade would restart the pods. This happens for example when upgrading from previous releases to 1.24.0-rc1. This is valuable information to get from our CI/CD

1. in the online upgrade test, rc, alpha, beta manifest should not be considered "the latest" in the `InstallLatestCNPGOperator` function (`/tests/utils/upgrade.go`)
2. for cases when the online upgrade test must fail (because pod gets restarted) the test should run all the way to see if the cluster gets updated properly. It could then report that the pods were restarted BUT the upgrade worked ... i.e. it was a rolling upgrade

Previously, a workaround had been applied to avoid the failing E2E when upgrading to 1.24.
This workaround was later rolled back: https://github.com/cloudnative-pg/cloudnative-pg/pull/5204/files
Indeed, the failing E2E test for online upgrade is valuable information for us, and may prompt us to include the restart of pods in the upgrade documentation.

### Additional context

https://github.com/cloudnative-pg/cloudnative-pg/pull/5297

### Backport?

Yes

Yes

### Code of Conduct

jsilvela:
Enhanced logging and traceability in the end-to-end upgrade tests.

Closes #5331


internal/cmd/manager/controller/controller.go
```diff
@@ -102,9 +103,18 @@ func RunController(
 ) error {
 	ctx := context.Background()
 
+	var operatorHash string
```
gabriele-wolfox, line 106: I would replace lines from 106 to 112 with

```
	operatorHash, err := executablehash.Get()
	if err != nil {
		setupLog.Error(err, "could not get executable hash")
		return err
	}
```

jsilvela: 👍

jsilvela: I would not `return err`. IMO this is a "best effort" thing.

gabriele-wolfox: If we have an error here, something really bad is happening and I would exit.
But yes, I would say return or not is debatable :)

tests/e2e/upgrade_test.go
```diff
@@ -385,6 +418,16 @@ var _ = Describe("Upgrade", Label(tests.LabelUpgrade, tests.LabelNoOpenshift), O
...
 		})
+		By("getting the operator info", func() {
+			pod, err := env.GetOperatorPod()
+			Expect(err).NotTo(HaveOccurred())
+			GinkgoWriter.Println("image used for operator", pod.Spec.Containers[0].Image)
+			lgs, err := env.GetPodLogs(pod.Namespace, pod.Name)
+			Expect(err).NotTo(HaveOccurred())
+			lines := strings.Split(lgs, "\n")
+			Expect(lines).ShouldNot(BeEmpty())
+			GinkgoWriter.Println(lines[0])
```
gabriele-wolfox, line 429: I would add here a caption of what we are printing.
Something like
`GinkgoWriter.Println("pod logs", lines[0])`

jsilvela: Agree

gabriele-wolfox, line 429: I would add here a caption of what we are printing.
Something like
`GinkgoWriter.Println("pod logs", lines[0])`

internal/cmd/manager/controller/controller.go
gabriele-wolfox, line 106: I would replace lines from 106 to 112 with

```
	operatorHash, err := executablehash.Get()
	if err != nil {
		setupLog.Error(err, "could not get executable hash")
		return err
	}
```

gabriele-wolfox (approved):

tests/e2e/upgrade_test.go
gabriele-wolfox, line 429:
```suggestion
			GinkgoWriter.Println("pod logs", lines[0])
```

internal/cmd/manager/controller/controller.go
```diff
@@ -102,9 +103,18 @@ func RunController(
...
+	h, err := executablehash.Get()
+	if err != nil {
+		setupLog.Error(err, "could not get executable hash")
+	} else {
+		operatorHash = h
+	}
```
gabriele-wolfox, line 112:
```suggestion
	operatorHash, err := executablehash.Get()
	if err != nil {
		setupLog.Error(err, "could not get executable hash")
		return err
	}
```

gabriele-wolfox (approved):

tests/e2e/upgrade_test.go
```diff
@@ -282,12 +305,17 @@ var _ = Describe("Upgrade", Label(tests.LabelUpgrade, tests.LabelNoOpenshift), O
 				return err
 			}
 
-			var count int
+			var count, nonUpgradeCount int
+			By("checking cluster events for manager")
```
YanniHu1996, line 309: Hi @jsilvela ,  the message `By("checking cluster events for manager")` is inside the `retry.OnError`. It may be printed multiple times. Can it be moved out of the `retry.OnError`?

YanniHu1996 (approved):

tests/e2e/upgrade_test.go
jsilvela: In my opinion, no, but not a strong feeling.

mnencia (changes requested):
This patch combines both code and test changes. I've provided specific comments regarding the code modifications, as they alter the operator's behavior and therefore cannot be accepted as part of a test patch.

pkg/utils/discovery.go
```diff
@@ -64,8 +64,6 @@ func (arch *AvailableArchitecture) GetHash() string {
 	if arch.hash == "" {
 		arch.calculateHash()
 	}
-	arch.mx.Lock()
-	defer arch.mx.Unlock()
 	return arch.hash
```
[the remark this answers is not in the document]
jsilvela: Hmm, so we want read protection, ok

internal/cmd/manager/controller/controller.go
```diff
@@ -102,9 +103,16 @@ func RunController(
 ) error {
 	ctx := context.Background()
 
+	operatorHash, err := executablehash.Get()
```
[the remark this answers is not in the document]
jsilvela: do you remember why?
Just thinking, for the "smart" upgrade test we want, checking the hash in the pods is the same as the new operator, seems like a very solid test...

internal/controller/cluster_upgrade.go
```diff
@@ -697,39 +697,51 @@ func (r *ClusterReconciler) upgradeInstanceManager(
...
+				"(operator hash: %s — instance manager hash: %s)",
+				postgresqlStatus.Pod.Name,
+				instanceManagerIsUpgrading,
+				operatorHash[:6],
+				instanceManagerHash[:6])
+
+			r.Recorder.Event(cluster, "Normal", "SkippedInstanceManagerUpgrade", message)
+			contextLogger.Info(message)
+			continue
+		}
```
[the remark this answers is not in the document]
jsilvela: Maybe I can try to reduce. But, it's very difficult to understand what's going on with online upgrades without a bit of logging like this one.

mnencia:
This issue is related to https://github.com/cloudnative-pg/cloudnative-pg/issues/4423 for the hash calculation.

mnencia:
> - Adds the executable hash on the operator binary and displays in
the init log
> - Computes the executable hash from the instances before and after
an upgrade
> - Add events to explain why an upgrade may have been skipped

pkg/utils/discovery.go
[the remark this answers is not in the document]
mnencia: The issue is that writing on a sting variable is not guaranteed to be atomic (even if it is in many architectures). Since this may be modified in another thread, we must protect the read operation.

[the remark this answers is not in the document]
jsilvela: then fixing so `calculateHash` returns a value.

jsilvela:
There. That's better:
Screenshot 2024-08-22 at 13 22 09

Asking, once again, for review.
Addressed your comments @mnencia , and also applied the changes discussed in https://github.com/cloudnative-pg/cloudnative-pg/issues/5331

jsilvela:
If I may say so myself ... nice

Screenshot 2024-08-22 at 13 33 03

mnencia:
The code is okay, but this pull request does too many things simultaneously. I'm going to split it into multiple PRs.

NiccoloFei (approved):

mnencia (approved):

PR #5325, opened 2024-08-19
feat: declarative database management

### LINKED ISSUE #5501: Add CRD and controller for Database objects

leonardoce:
Add support for declarative database management by introducing a new Custom Resource called `Database`.

This new CR is controlled by the instance manager of the primary instance.

Database deletion is controlled via a finalizer, with a default retention policy of `retain`. As a result, deleting a Database object in Kubernetes will not remove the underlying PostgreSQL database by default. The user can change this behavior by explicitly setting the retention policy to `delete`.

Closes #5501


leonardoce:
This PR have been split from #5454 and contains only the code related to the declarative database management.

internal/management/controller/database_controller.go
```diff
@@ -0,0 +1,393 @@
...
+	}
+
+	return nil
+}
+
+func (r *DatabaseReconciler) createDatabase(
+	ctx context.Context,
+	db *sql.DB,
+	obj *apiv1.Database,
+) error {
```
jsilvela, line 294: This and the other functions that interact with Postgres, IMO should go in a dedicate file perhaps in `pkg/postgres`

internal/management/controller/database_controller.go
```diff
@@ -0,0 +1,393 @@
...
+
+	// This is not for me!
+	if database.Spec.ClusterRef.Name != r.instance.ClusterName {
+		return ctrl.Result{}, nil
+	}
+
+	// If everything is reconciled, we're done here
+	if database.Generation == database.Status.ObservedGeneration {
+		return ctrl.Result{}, nil
+	}
```
jsilvela, line 96: What if there's drift on the side of the database because a user made changes there?
If I'm understanding right, as long as the CR didn't change, this would be left alone.

leonardoce:
This PR has been split. I'm closing it in favor of #5329 and #5325.

gabriele-wolfox:
E2E test run on  EDB fork https://github.com/EnterpriseDB/cloudnative-pg/actions/runs/10814905112

gabriele-wolfox (approved):

NiccoloFei (approved):

jsilvela (approved):

pschichtel:
Just two quick questions without having looked at the changes beyond the surface:

* will it be possible to "adopt" existing databases into operator management?
* the db-operator project, that I'm currently using for this task, has a mechanism to prevent accidental database deletion (if enabled they simple doesn't delete the underlying database), which seems useful in conjuction with CD tools. Is something like this planned?

Thanks for the work!

mnencia (approved):

mnencia:
> * will it be possible to "adopt" existing databases into operator management?

If you create a `Database` object and the database already exists, the database is "adopted" and reconciled to the definition in the object.

> * the db-operator project, that I'm currently using for this task, has a mechanism to prevent accidental database deletion (if enabled they simple doesn't delete the underlying database), which seems useful in conjuction with CD tools. Is something like this planned?

The `databaseReclaimPolicy` property controls exactly that behavior.

docs/src/declarative_database_management.md
```diff
@@ -0,0 +1,60 @@
+# Declarative Database Management
+
+Declarative database management enables users to control the lifecycle of
+databases using a new Custom Resource Definition (CRD) called `Database`.
```
jsilvela, line 4:
```suggestion
databases via a new Custom Resource Definition (CRD) called `Database`.
```

PR #5329, opened 2024-08-20
feat: add declarative management of PostgreSQL logical replication

### LINKED ISSUE #5567: [split from `#13`] add CRDs and controllers for Publication and Subscription

leonardoce:
This pull request adds a feature for managing PostgreSQL logical replication through a declarative method by introducing two new Kubernetes CRDs: Publications and Subscriptions.

## Release notes

Introduced declarative management of PostgreSQL publications and subscriptions via Kubernetes CRDs

Closes #5567 

## Notes for reviewers

Based on #5325. To be merged after that.


jsilvela:
After discussion, we agree that we should fix OLM in this PR.

gabriele-wolfox (approved):

NiccoloFei (approved):

config/olm-samples/postgresql_v1_publication.yaml
```diff
@@ -0,0 +1,11 @@
...
+kind: Publication
+metadata:
+  name: publication-sample
+spec:
+  name: pub
+  dbname: app
+  cluster:
+    name: cluster-sample
+  target:
+    allTables: {}
```
jsilvela, line 11:
```suggestion
    allTables: true
```

docs/src/e2e.md
```diff
@@ -127,3 +127,6 @@ and the following suite of E2E tests are performed on that cluster:
      * Declarative creation of temporary tablespaces
      * Backup / recovery from object storage
      * Backup / recovery from volume snapshots
+
+* **Publication and Subscription:**
```
jsilvela, line 131: IMO this belongs in the Replication grouping, we don't need a top-level grouping for it

docs/src/samples/cluster-example-logical-source.yaml
```diff
@@ -30,3 +30,15 @@ spec:
...
+kind: Publication
+metadata:
+  name: cluster-example-pub
+spec:
+  name: pub
+  dbname: app
+  cluster:
+    name: cluster-example
+  target:
+    allTables: {}
```
jsilvela, line 44:
```suggestion
    allTables: true
```

jsilvela (approved):
Nice. One comment in the test. In the fixture directory, we're making files be `.template` even when they don't have any variables to substitute. It works but IMO we can use plain `.yaml`

internal/management/controller/common.go
```diff
@@ -71,3 +75,24 @@ func getClusterFromInstance(
...
+		keys = append(keys, k)
+	}
+
+	// sort the slice by keys
+	sort.Strings(keys)
+
+	b := new(bytes.Buffer)
+	for _, key := range keys {
+		// TODO(armru): should we sanitize the key?
+		// TODO(armru): any alternative to pg.QuoteLiteral?
```
armru, line 92: @mnencia

armru:
Given that I've made extensive changes I feel like I'm not impartial enough to do a review to approve the patch, I will leave the mantainers approvals to the others

api/v1/publication_types.go
```diff
@@ -0,0 +1,128 @@
...
+	DBName string `json:"dbname"`
+
+	// The owner
+	Owner string `json:"owner,omitempty"`
+
+	// Parameters
+	Parameters map[string]string `json:"parameters,omitempty"`
+
+	// Publication target
+	Target PublicationTarget `json:"target,omitempty"`
```
armru, line 58: Is `omitempty` really needed?

api/v1/publication_types.go
```diff
@@ -0,0 +1,158 @@
...
+// +kubebuilder:validation:XValidation:rule="(has(self.allTables) && !has(self.objects)) || (!has(self.allTables) && has(self.objects))",message="allTables and objects are mutually exclusive"
+type PublicationTarget struct {
+	// All tables should be published
+	// +optional
+	AllTables bool `json:"allTables,omitempty"`
+
+	// Just the following schema objects
+	// +optional
+	// +kubebuilder:validation:XValidation:rule="!(self.exists(o, has(o.table) && has(o.table.columns)) && self.exists(o, has(o.tablesInSchema)))",message="specifying a column list when the publication also publishes tablesInSchema is not supported"
+	// +kubebuilder:validation:MaxItems=100000
```
mnencia, line 79: This may look weird, but it is required otherwise the CRD cannot be loaded in the API server. The error would be
```
The CustomResourceDefinition "publications.postgresql.cnpg.io" is invalid: spec.validation.openAPIV3Schema.properties[spec].properties[target].properties[objects].x-kubernetes-validations[0].rule: Forbidden: estimated rule cost exceeds budget by factor of 1.153433x (try simplifying the rule, or adding maxItems, maxProperties, and maxLength where arrays, maps, and strings are declared)
```

api/v1/publication_types.go
mnencia: It's not. I've removed it.

docs/src/samples/publication-example-objects.yaml
```diff
@@ -9,10 +9,8 @@ spec:
   dbname: app
   target:
     objects:
+      - tablesInSchema: public
       - table:
```
jsilvela: OK, so this will now fail, since tablesInSchema is incompatible with `table`, right? If so should the sample be renamed to reflect that it will fail?

mnencia: The example works perfectly. If you want a failing example here I have one:

```
kind: Publication
metadata:
  name: publication-example-objects-fail
spec:
  cluster:
    name: cluster-example
  name: pub-objects-fail
  dbname: app
  target:
    objects:
      - tablesInSchema: public
      - table:
          schema: another_schema
          name: numbers_three
          only: true
          columns:
            - i
            - m
```

armru (approved):

mnencia (approved):

gbartolini:
I was trying this patch, specifically applied the following examples:

- docs/src/samples/cluster-example-logical-source.yaml
- docs/src/samples/cluster-example-logical-destination.yaml

When I delete the two clusters, I find it odd that publications and subscriptions are left (I understand that we created the object, but still ... I find it odd). This is their status:

```
kind: Subscription
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"postgresql.cnpg.io/v1","kind":"Subscription","metadata":{"annotations":{},"name":"cluster-example-dest-sub","namespace":"default"},"spec":{"cluster":{"name":"cluster-example-dest"},"dbname":"app","externalClusterName":"cluster-example","name":"sub","publicationName":"pub"}}
  creationTimestamp: "2024-11-26T14:39:52Z"
  generation: 1
  name: cluster-example-dest-sub
  namespace: default
  resourceVersion: "80870"
  uid: 45bdbf14-aef0-4c2b-95d1-28b11976e031
spec:
  cluster:
    name: cluster-example-dest
  dbname: app
  externalClusterName: cluster-example
  name: sub
  publicationName: pub
  subscriptionReclaimPolicy: retain
status:
  applied: true
  observedGeneration: 1
---
kind: Publication
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"postgresql.cnpg.io/v1","kind":"Publication","metadata":{"annotations":{},"name":"cluster-example-pub","namespace":"default"},"spec":{"cluster":{"name":"cluster-example"},"dbname":"app","name":"pub","target":{"allTables":true}}}
  creationTimestamp: "2024-11-26T14:29:12Z"
  generation: 1
  name: cluster-example-pub
  namespace: default
  resourceVersion: "80808"
  uid: cfee4a78-a333-490d-9935-3f2a97da6822
spec:
  cluster:
    name: cluster-example
  dbname: app
  name: pub
  publicationReclaimPolicy: retain
  target:
    allTables: true
status:
  applied: true
  observedGeneration: 1
```

At least, if we keep them, is it possible to say that it is not applied and throw an error?

jsilvela:
Interesting case. Had not thought about it.
So, right now the Pub/Sub CRD's (and same with the Database CRD) work in a "reconcile-and-forget" model. See how `Generation == ObservedGeneration`.
By similarity with the case of deleting the Pub/Sub or Database from Postgres when we delete the CRD's, I can think of a finalizer being installed in the Cluster object, if there are Pub/Sub pointing to it.
This is going to require some design discussion.

jsilvela:
Actually, really easy to fix.
We just need to check the cluster for existence before we get to the comparison of ObservedGeneration with Generation.

jsilvela:
After discussion we agree on an approach.
We have opened a sub-issue to track: https://github.com/cloudnative-pg/cloudnative-pg/issues/6172

gbartolini (approved):

docs/src/logical_replication.md
```diff
@@ -0,0 +1,444 @@
...
+is the source of data changes. It acts as a logical container for the change
+sets (also known as *replication sets*) generated from one or more tables within
+a database. Publications can be defined on any PostgreSQL 10+ instance acting
+as the *publisher*, including instances managed by popular DBaaS solutions in the
+public cloud. Each publication is tied to a single database and provides
+fine-grained control over which tables and changes are replicated.
+
+For publishers outside Kubernetes, you can [create publications using SQL](https://www.postgresql.org/docs/current/sql-createpublication.html)
+or leverage the [`cnpg publication create` plugin command](kubectl-plugin.md#logical-replication-publications).
+
```
jsilvela, line 45: I'm not sure I'm getting this. In this paragraph we're on the topic of publishers outside Kubernetes. So does it make sense to offer the possibility of using the kubectl plugin command `publication create`? Is that better in a separate paragraph?

docs/src/logical_replication.md
```diff
@@ -0,0 +1,444 @@
...
+- `metadata.name`: A unique name for the Kubernetes `Subscription` object
+  within its namespace.
+- `spec.cluster.name`: The name of the PostgreSQL cluster where the
+  subscription will be created.
+- `spec.dbname`: The name of the database in which the subscription will be
+  created.
+- `spec.name`: The name of the subscription as it will appear in PostgreSQL.
+- `spec.externalClusterName`: The name of the external cluster, as defined in
+  the `spec.cluster.name` cluster's configuration. This references the
+  publisher database.
```
jsilvela, line 225: "references the publisher database" IMO this could be confusing.
The externalClusterName references the publisher cluster, while the publisher database is really implicit in the choice of publicationName, no?

jsilvela: Answering my own comment: the database name is in the connection parameters.

docs/src/logical_replication.md
mnencia: `publication create` works on source databases outside Kubernetes as long as the destination Cluster has a usable definition for the source in the external clusters.

PR #5360, opened 2024-08-23
Use single database for pgbouncer authentication queries

usernamedt:
I propose to add the `AuthQueryDb` setting that enables the use of PgBouncer `auth_dbname` setting:

> Database name in the [database] section to be used for authentication purposes. This option can be either global or overridden in the connection string if this parameter is specified.

The issue with the current approach is that users might accidentally delete or revoke access to the `public.user_search()` function, which could lock them out of a specific database. By placing this access function in a designated database (e.g., 'postgres'), it becomes easier to manage this behavior and reduce the risk of being locked out.

However, I am unsure where exactly this setting should belong, as it influences both the Pooler and Postgres behavior. I have uploaded a proof-of-concept code with a fixed database ('postgres'). If the community finds this feature valuable, I am willing to update the PR with the required logic.

### DIFF
--- a/internal/management/controller/instance_controller.go
+++ b/internal/management/controller/instance_controller.go
@@ -574,6 +574,12 @@ func (r *InstanceReconciler) reconcileDatabases(ctx context.Context, cluster *ap
 					fmt.Errorf("could not reconcile extensions for database %s: %w", databaseName, err))
 			}
 		}
+		if databaseName != "postgres" {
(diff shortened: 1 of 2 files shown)


gbartolini:
Did you check this issue first? #1383

usernamedt:
Oops, looks like mine is a duplicate:) I'll look into it, thank you.

PR #5368, opened 2024-08-24
✨ Add support for ICU locales in InitDB flags

elafarge:
Postgres 15 introduced support for ICU locales.

This PR makes it possible to configure such locales via the initdb field of the cluster spec.


mnencia:
Superseded by #6220

PR #5387, opened 2024-08-27
fix(webserver): obtain the cluster from the informer for the`/cache/` endpoint

### LINKED ISSUE #5388: [Bug]: the `/cache/` endpoint should use the informer instead of `LoadUnsafeCluster`

### Contact Details

_No response_

### Version

1.24.0

1.31

YAML manifest

### What happened?

The `/cache` endpoint uses the `LoadUnsafeCluster` method, which could result in several issues:
- The returned cluster could not be the most recent one.
- There is no real performance benefit given that we have instantiated a cluster with the informer

We should proceed to obtain the cluster from the informer, ensuring that it is the most recent version available.

### Cluster resource

_No response_

### Relevant log output

_No response_

### Code of Conduct

armru:
This patch removes the `LoadUnsafeCluster` from the `/cache/` endpoint, ensuring that the returned cluster is always the most recent one.

The cached cluster had no real performance benefits over using the already instantiated client with an informer.

Closes #5388 

## Release Notes

The localwebserver  endpoint `/cache/` now obtains the cluster resource from the client informer, ensuring that it is always up to date.

### DIFF
--- a/pkg/management/postgres/webserver/local.go
+++ b/pkg/management/postgres/webserver/local.go
@@ -24,6 +24,7 @@ import (
 	"net/http"
 	"strings"
 
+	apierrs "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/client-go/tools/record"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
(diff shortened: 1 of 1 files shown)


armru:
the commit title needs to be improved

mnencia (approved):
The implementation is ok. Other tickets to completely remove `LoadClusterUnsafe()` will follow.

fcanovai (approved):

PR #5389, opened 2024-08-28
fix(instance): increase timeout to get cluster from Kubernetes API server

### LINKED ISSUE #5382: [Bug]: Bootstrap job failed while istio sidecar inject in

### Contact Details

_No response_

### Version

1.23.4

1.31

YAML manifest

### What happened?

This probably because we have quite short time configured for retry and waiting  of api server, and while istio sidecar stop the network in bootstrap job pod  [cloudnative-pg/pkg/management/client.go at main · cloudnative-pg/cloudnative-pg](https://github.com/cloudnative-pg/cloudnative-pg/blob/main/pkg/management/client.go#L50)

### Cluster resource

_No response_

### Relevant log output

_No response_

### Code of Conduct

hh24k:
Increase the timeout for retrieving the cluster resource from the Kubernetes API server, improving reliability in cases of longer startup times, such as when Istio sidecar injection is enabled. This adjustment improves robustness and allows more time for the sidecar to become ready.

Closes #5382 

# Release notes

Increased timeout during instance startup for retrieving cluster resources, improving reliability in environments with longer startup times, such as with Istio sidecar injection.

### DIFF
--- a/pkg/management/client.go
+++ b/pkg/management/client.go
@@ -147,7 +147,6 @@ func WaitKubernetesAPIServer(ctx context.Context, clusterObjectKey client.Object
 		}
 		return nil
 	})
-
 	if err != nil {
 		const message = "error while waiting for the API server to be reachable"
 		logger.Error(err, message)
--- a/pkg/management/client.go
+++ b/pkg/management/client.go
@@ -132,7 +132,7 @@ func NewEventRecorder() (record.EventRecorder, error) {
 // WaitKubernetesAPIServer will wait for the kubernetes API server to by ready.
 // Returns any error if it can't be reached.
 func WaitKubernetesAPIServer(ctx context.Context, clusterObjectKey client.ObjectKey) error {
-	logger := log.FromContext(ctx).WithName("kube-api-server-check")
+	logger := log.FromContext(ctx).WithName("wait-kubernetes-api-server")
 
 	cli, err := NewControllerRuntimeClient()
 	if err != nil {
@@ -142,7 +142,7 @@ func WaitKubernetesAPIServer(ctx context.Context, clusterObjectKey client.Object
 
 	err = retry.OnError(readinessCheckRetry, resources.RetryAlways, func() error {
 		if err := cli.Get(ctx, clusterObjectKey, &apiv1.Cluster{}); err != nil {
-			logger.Warning("The API server is currently unreachable. Will wait and retry")
+			logger.Warning("Encountered an error while executing get cluster. Will wait and retry", "error", err.Error())
 			return err
 		}
 		return nil
--- a/pkg/management/client.go
+++ b/pkg/management/client.go
@@ -129,10 +129,10 @@ func NewEventRecorder() (record.EventRecorder, error) {
 	return recorder, nil
 }
 
-// WaitKubernetesAPIServer will wait for the kubernetes API server to by ready.
-// Returns any error if it can't be reached.
-func WaitKubernetesAPIServer(ctx context.Context, clusterObjectKey client.ObjectKey) error {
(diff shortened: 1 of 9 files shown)


litaocdl (approved):

pkg/management/client.go
```diff
@@ -49,8 +49,8 @@ var (
 	// readinessCheckRetry is used to wait until the API server is reachable
 	readinessCheckRetry = wait.Backoff{
 		Steps:    5,
-		Duration: 10 * time.Millisecond,
-		Factor:   5.0,
+		Duration: 1 * time.Second,
+		Factor:   3.0,
```
jsilvela, line 53: So, in the last step we'd wait for 3^(5-1) = 81 seconds, right?
Step 1 - 1 sec
...
Step 4 - 27 sec
Step 5 - 81 sec

jsilvela: All in all, accumulates to about 6 minutes. Seems reasonable.

pkg/management/client.go
```diff
@@ -132,17 +132,22 @@ func NewEventRecorder() (record.EventRecorder, error) {
...
 		logger.Error(err, "error while creating a standalone Kubernetes client")
 		return err
 	}
 
-	if err := retry.OnError(readinessCheckRetry, resources.RetryAlways, func() (err error) {
-		return cli.Get(ctx, clusterObjectKey, &apiv1.Cluster{})
-	}); err != nil {
+	err = retry.OnError(readinessCheckRetry, resources.RetryAlways, func() (err error) {
+		if err = cli.Get(ctx, clusterObjectKey, &apiv1.Cluster{}); err != nil {
+			logger.Info("The api server is currently unreachable, will wait and retry later")
```
jsilvela, line 145: This may deserve to be a Warning

pkg/management/client.go
jsilvela: In the previous implementation, the waiting time would accumulate to about 78 seconds.
Not that bad, no?

jsilvela (approved):

armru (approved):

pkg/management/client.go
```diff
@@ -132,17 +132,22 @@ func NewEventRecorder() (record.EventRecorder, error) {
...
 		return err
 	}
 
-	if err := retry.OnError(readinessCheckRetry, resources.RetryAlways, func() (err error) {
-		return cli.Get(ctx, clusterObjectKey, &apiv1.Cluster{})
-	}); err != nil {
+	err = retry.OnError(readinessCheckRetry, resources.RetryAlways, func() error {
+		if err := cli.Get(ctx, clusterObjectKey, &apiv1.Cluster{}); err != nil {
+			logger.Warning("Encountered an error while executing get cluster. Will wait and retry", "error", err.Error())
+			return err
```
armru, line 146: I modified the log message because it could end up being misleading, we are exiting in case of any error, which could be a lack of permission or cluster not existing anymore, for example (even if unlikely)

armru: perhaps WaitForGetCluster would be more meaningful?

mnencia (approved):

pkg/management/client.go
mnencia: I think the actual times are a lot shorter. My calculation is the following:

1st retry: 1 second (base duration)
2nd retry: 3 seconds (1 * 3)
3rd retry: 9 seconds (3 * 3)
4th retry: 27 seconds (9 * 3)
5th retry: 81 seconds (27 * 3)

Summing all together, it gives `1 + 3 + 9 + 27 + 81 = 121` seconds total wait time, excluding the effect of jitter. 

Previously it was around `10 ms + 50 ms + 250 ms + 1,250 ms + 6,250 ms = 7,810 milliseconds` (or ~ 7.81 seconds).

PR #5403, opened 2024-08-30
fix: remove unused superuser and app secret volumes

mnencia:
This commit removes the unused superuser secret and app secret volumes, simplifying the pod definition. These volumes were never referenced in the code, and their removal prevents unnecessary instance restarts when enabling or disabling superuser access and when promoting a replica cluster.

# Release notes

Fix:
* Removed unused superuser secret and app secret volumes from Pods to prevent unnecessary instance restarts when enabling/disabling superuser access or promoting a replica cluster.

### DIFF
--- a/pkg/specs/podspec_diff.go
+++ b/pkg/specs/podspec_diff.go
@@ -116,7 +116,12 @@ func compareMaps[V comparable](current, target map[string]V) (bool, string) {
 	return true, ""
 }
 
-func isCurrentVolumeToBeIgnored(name string) bool {
+// shoudlIgnoreCurrentVolume checks if a volume or mount is the superuser or app
+// mount, which had been added, superflously, in previous versions. If so, ignores
+// it for the PodSpec drift detector, to avoid unnecessary restarts
+//
+// TODO: delete this function after minor version 1.24 is discontinued
+func shoudlIgnoreCurrentVolume(name string) bool {
 	return name == "superuser-secret" || name == "app-secret"
 }
 
@@ -124,7 +129,7 @@ func compareVolumes(currentVolumes, targetVolumes []corev1.Volume) (bool, string
 	current := make(map[string]corev1.Volume)
 	target := make(map[string]corev1.Volume)
 	for _, vol := range currentVolumes {
-		if isCurrentVolumeToBeIgnored(vol.Name) {
+		if shoudlIgnoreCurrentVolume(vol.Name) {
 			continue
 		}
 		current[vol.Name] = vol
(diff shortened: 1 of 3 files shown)


litaocdl (approved):

litaocdl:
@mnencia I also removed the volume, seems to me they are not been used too.

mnencia:
> @mnencia I also removed the volume, seems to me they are not been used too.

You are 100% correct.

mnencia:
Given that we managed to avoid the rollout on the upgrade, I've added back the backport request.

pkg/specs/podspec_diff_test.go
```diff
@@ -0,0 +1,40 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package specs
+
+import (
```
jsilvela, line 19: bit of a nit here, but we already have existing PodSpec tests, in `pods_test.go`. We might consider moving this there, or possibly even better, bringing over the PodSpec tests to this file.

pkg/specs/podspec_diff.go
```diff
@@ -116,10 +116,17 @@ func compareMaps[V comparable](current, target map[string]V) (bool, string) {
 	return true, ""
 }
 
+func isCurrentVolumeToBeIgnored(name string) bool {
```
jsilvela, line 119: This is very nice English, perhaps a bit too poetic for code :)

```suggestion
func shouldVolumeBeIgnored(name string) bool {
```

jsilvela (commented):
Makes sense, but ... given we're eliminating the unnecessary app/superuser volumes and mounts ... do we need to ignore them in the PodSpec diff?
The extra starts are fixed with the deletion, so the "ignore" here is only in case someone adds those volumes with exactly those names...
Hmm, I guess this is useful if we have online upgrade enabled, to prevent the rolling update.
But then we have this code looking for volumes/mounts by name for the future...

jsilvela:
After discussion with @mnencia on that function just there to avoid another restart,
agreed that we can have a //TODO annotation in that function, mentioning 1.25 release.

jsilvela (approved):

pkg/specs/podspec_diff_test.go
mnencia: I agree it would be better. Let's do it in another PR.

fcanovai (approved):

PR #5422, opened 2024-09-02
chore: improve inplace updates observability

### LINKED ISSUE #5331: [Feature]: smarter E2E test for online upgrades

mnencia:
Enhance logging and tracing for in-place updates, providing better visibility into the upgrade process.

Partially closes #5331 

Split from #5297

### DIFF
--- a/internal/controller/cluster_upgrade.go
+++ b/internal/controller/cluster_upgrade.go
@@ -701,39 +701,50 @@ func (r *ClusterReconciler) upgradeInstanceManager(
 		}
 		operatorHash := targetManager.GetHash()
 
-		if instanceManagerHash != "" && instanceManagerHash != operatorHash && !instanceManagerIsUpgrading {
-			// We need to upgrade this Pod
-			contextLogger.Info("Upgrading instance manager",
-				"pod", postgresqlStatus.Pod.Name,
-				"oldVersion", postgresqlStatus.ExecutableHash)
-
-			if cluster.Status.Phase != apiv1.PhaseOnlineUpgrading {
-				err := r.RegisterPhase(ctx, cluster, apiv1.PhaseOnlineUpgrading, "")
-				if err != nil {
-					return err
-				}
-			}
-
-			err = r.InstanceClient.UpgradeInstanceManager(ctx, postgresqlStatus.Pod, targetManager)
-			if err != nil {
-				enrichedError := fmt.Errorf("while upgrading instance manager on %s (hash: %s): %w",
-					postgresqlStatus.Pod.Name,
-					operatorHash[:6],
-					err)
-
-				r.Recorder.Event(cluster, "Warning", "InstanceManagerUpgradeFailed",
-					fmt.Sprintf("Error %s", enrichedError))
-				return enrichedError
-			}
-
-			message := fmt.Sprintf("Instance manager has been upgraded on %s (hash: %s — previous hash: %s)",
+		if instanceManagerIsUpgrading || instanceManagerHash == "" || instanceManagerHash == operatorHash {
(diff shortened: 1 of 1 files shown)


internal/controller/cluster_upgrade.go
```diff
@@ -697,39 +697,50 @@ func (r *ClusterReconciler) upgradeInstanceManager(
...
-				}
-			}
+		if instanceManagerIsUpgrading || instanceManagerHash == "" || instanceManagerHash == operatorHash {
+			message := fmt.Sprintf("Instance manager will skip upgrade on %s (upgrading: %t) "+
+				"(operator hash: %s — instance manager hash: %s)",
+				postgresqlStatus.Pod.Name,
+				instanceManagerIsUpgrading,
+				operatorHash[:6],
+				instanceManagerHash[:6])
+			contextLogger.Trace(message)
```
sxd, line 707: @mnencia why this is a Trace and not Debug? the next message is Info level and it's completely in context with this one, so I think this is Debug or Info level but not Trace for sure, what do you think ?

internal/controller/cluster_upgrade.go
```diff
@@ -697,39 +697,50 @@ func (r *ClusterReconciler) upgradeInstanceManager(
...
 		}
+
+		message := fmt.Sprintf("Instance manager has been upgraded on %s "+
+			"(oldHash: %s — newHash: %s)",
+			postgresqlStatus.Pod.Name,
+			instanceManagerHash[:6],
+			operatorHash[:6])
+
+		r.Recorder.Event(cluster, "Normal", "InstanceManagerUpgraded", message)
+		contextLogger.Info(message)
```
sxd, line 743: @mnencia and this one is Info, I think that the one indicating skip should be Info for sure, from a user perspective

internal/controller/cluster_upgrade.go
```diff
@@ -704,7 +704,7 @@ func (r *ClusterReconciler) upgradeInstanceManager(
 				instanceManagerIsUpgrading,
 				operatorHash[:6],
 				instanceManagerHash[:6])
-			contextLogger.Trace(message)
+			contextLogger.Info(message)
```
mnencia, line 707: This would emit an info line for every pod on every reconciliation cycle because this block is executed when there is nothing to do.

sxd: Then Trace XD

sxd (approved):

NiccoloFei (approved):

PR #5429, opened 2024-09-03
fix: avoid skipping log lines longer than 64k

mnencia:
This patch increases the line buffer limit of every `bufio.Scanner` we use from 64k to 1MB. It also ensures the starting buffer is 4k to avoid wasting memory.

# Release Notes

Fixed an issue where log lines longer than 64k were skipped by increasing the buffer limit to 1MB.

### DIFF
--- a/tests/e2e/suite_test.go
+++ b/tests/e2e/suite_test.go
@@ -150,7 +150,7 @@ var _ = SynchronizedAfterSuite(func() {
 // of output are not legal JSON
 func saveLogs(buf *bytes.Buffer, logsType, specName string, output io.Writer, capLines int) {
 	scanner := bufio.NewScanner(buf)
-	scanner.Buffer(make([]byte, 0, 64*1024), 1024*1024)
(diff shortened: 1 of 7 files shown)


fcanovai (approved):

tests/e2e/suite_test.go
```diff
@@ -150,7 +150,7 @@ var _ = SynchronizedAfterSuite(func() {
 // of output are not legal JSON
 func saveLogs(buf *bytes.Buffer, logsType, specName string, output io.Writer, capLines int) {
 	scanner := bufio.NewScanner(buf)
-	scanner.Buffer(make([]byte, 0, 64*1024), 1024*1024)
+	scanner.Buffer(make([]byte, 0, 4096), 1024*1024)
```
sxd, line 153: Here we're reducing the size not increasing :S

mnencia: >  It also ensures the starting buffer is 4k to avoid wasting memory.

sxd (approved):

PR #5215, opened 2024-08-01
chore: improve operator logs in E2E tests

jsilvela:
This patch improves the collection and display of operator logs on test failure.
Now the operator logs will be filtered to collect only those logs that were
affecting the namespace of the test.

# Release notes

NONE

PR #5457, opened 2024-09-09
tests(e2e): catch operator logs during the tests run

### LINKED ISSUE #5453: [Chore]: Capture operator logs

### Contact Details

_No response_

### Version

1.24.0

1.31

YAML manifest

### What happened?

With the new way to catch the logs we left the operator logs outside, we should:
* Use stern to stream the operator logs into a file, always, using the podname to keep different files so we can detect when it was restarted and changed in time
* Delete the directory if there wasn't a failed test, otherwise kept for debugging 

### Cluster resource

_No response_

### Relevant log output

_No response_

### Code of Conduct

sxd:
Since we included the catch all for the logs of all the pods and show that it works,
now we can do the same for the operator logs using the same functions and model.

We gather the operator pods by matching the labels and we split them by different
pod name, meaning that if the operator is destroyed and recreated we will have a
different log file for the new pod.

Closes #5453


mnencia:
Manual run to use the workflow definition on the branch https://github.com/cloudnative-pg/cloudnative-pg/actions/runs/10812091268

mnencia (approved):

tests/utils/namespace.go
```diff
@@ -175,38 +51,23 @@ func (env TestingEnvironment) GetOperatorLogs(buf *bytes.Buffer) error {
 	return streamPodLog.Stream(env.Ctx, buf)
 }
 
-// DumpNamespaceOperatorLogs writes the operator logs related to a namespace into a writer
-// and also into a file
-func (env TestingEnvironment) DumpNamespaceOperatorLogs(namespace, testName string, output io.Writer) {
```
jsilvela, line 180: IMO this is a loss. This code was ensuring we saved the operator logs for the test's namespace only. And would also show on Ginkgo operator logs for this namespace that were Error or Warn.
Is this functionality captured in a different way in this PR?

jsilvela: I have no problem with the general operator log tailing being converted to stern.
It's the namespace-filtered operator logs that I think were an improvement over a single file with all the operator logs.

sxd: @jsilvela after months of having this I haven't found any one using the operator logs in the Ginkgo namespace, instead, is more noise than help in there because start to fill up the Ginkgo buffer and one start to miss the focus on the error, about the namespace captured log, that's right, we may improve that in the future but for now an operator log that you can grep with the namespace looks fair enough, this is a long process of improving the e2e tests, for sure if we found needed we can incorporate that functionality in the next few improvements but for now I think it's ok to keep this out, there's other PRs that I'll be opening to improve the e2e

jsilvela: > after months of having this I haven't found any one using the operator logs in the Ginkgo namespace
not been months.
This was merged 20 days ago. https://github.com/cloudnative-pg/cloudnative-pg/pull/5215
As to "not found anyone" -- right, so, I did use it, it helped for an investigation, which is why I decided to make the functionality available.

> we may improve that in the future but for now an operator log that you can grep with the namespace looks fair enough
This is not about adding code, but about deleting less of it.

that said, @sxd , I'm not going to ask for changes, go ahead. But I'm also not giving approve.

fcanovai (approved):

PR #5486, opened 2024-09-10
fix: primary is not switchover when adding wal storage

### LINKED ISSUE #5484: [Bug]: primary pod is not switchover when add new wal volume

### Contact Details

_No response_

### Version

1.24.0

1.31

YAML manifest

### What happened?

when primaryUpdateMethod=switchover is set, update the cluster yaml file and add a wal storage, noticed that primary pod is doing restart instead of switchover 

### Cluster resource

_No response_

### Relevant log output

_No response_

### Code of Conduct

litaocdl:
Under `primaryUpdateMethod=switchover` and add wal volume in cluster, 
after standby is restarted, the old primary pod is not doing a switchover but 
restart itself with primary role not changed.  The old primary should switchover first
then let podSpec decide if a recreate pod is needed. 

closes: #5484

### DIFF
--- a/internal/controller/cluster_upgrade.go
+++ b/internal/controller/cluster_upgrade.go
@@ -233,8 +233,10 @@ func (r *ClusterReconciler) updateRestartAnnotation(
 // rollout describes whether a rollout should happen, and if so whether it can
 // be done in-place, and what the reason for the rollout is
 type rollout struct {
-	required             bool
-	canBeInPlace         bool
+	required     bool
+	canBeInPlace bool
+	// if primaryForceRecreate is true, the old pod will be
+	// recreated no matter what primaryUpdateMethod is set
 	primaryForceRecreate bool
 	reason               string
 }
@@ -495,10 +497,12 @@ func checkPodInitContainerIsOutdated(pod *corev1.Pod, _ *apiv1.Cluster) (rollout
 }
 
 func checkHasMissingPVCs(pod *corev1.Pod, cluster *apiv1.Cluster) (rollout, error) {
-	if persistentvolumeclaim.InstanceHasMissingMounts(cluster, pod) {
+	missingPVC, missingTbsPVC := persistentvolumeclaim.InstanceHasMissingMounts(cluster, pod)
+	if missingPVC {
 		return rollout{
-			required:             true,
-			primaryForceRecreate: true,
+			required: true,
+			// we force the recreation of the primary pod if the tablespace PVC is missing
+			primaryForceRecreate: missingTbsPVC,
 			reason:               "attaching a new PVC to the instance Pod",
 		}, nil
 	}
--- a/internal/controller/cluster_upgrade.go
+++ b/internal/controller/cluster_upgrade.go
@@ -235,8 +235,8 @@ func (r *ClusterReconciler) updateRestartAnnotation(
 type rollout struct {
 	required     bool
 	canBeInPlace bool
-	// if primaryForceRecreate is true, the old pod will be
-	// recreated no matter what primaryUpdateMethod is set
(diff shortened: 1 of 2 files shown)


internal/controller/cluster_upgrade.go
```diff
@@ -495,10 +497,14 @@ func checkPodInitContainerIsOutdated(pod *corev1.Pod, _ *apiv1.Cluster) (rollout
...
+	if missingPVC, missingTbsPVC := persistentvolumeclaim.InstanceHasMissingMounts(
+		cluster,
+		pod,
+	); missingPVC {
 		return rollout{
-			required:             true,
-			primaryForceRecreate: true,
+			required: true,
+			// we force the recreation of the primary pod if the tablespace PVC is missing
+			primaryForceRecreate: missingTbsPVC,
```
jsilvela, line 507: I'm a bit confused here.
So, if there is a missing tablespace, we force an in-place primary upgrade, and otherwise we go with the primaryUpdate strategy.
But, when looking for `InstanceHasMissingMounts`, we iterate over mounts. We could find that there is both a PG WAL and a table space needing creation. Why will the TBS "win"?

jsilvela (commented):
Before proceeding I'd like to understand this.
Also, bigger picture: if we want to honor the primary update strategy, should we even have the `primaryForceRecreate` as an option?
Nothing wrong with doing a switchover because we have a new table space, no?

internal/controller/cluster_upgrade.go
litaocdl: @jsilvela For a missing wal volume, the old primary is able to restart postgres and demote itself without issue, because wal folder can locates under pg_data. For missing tablespace volume, the old primary can not restart postgres (crashloop )as new primary (through switchover) already have tablespace created.

so for missing wal, old primary restart and demoting, and then podSpec recreate the pod (pod restart)
for missing tablespace, old primary should recreate pod directly without a switchover, otherwize, switchover will failed on postgres restart.

jsilvela:
The logic for the upgrade is more complicated than it needs to be. Not the fault of this PR, but this PR adds more complexity. Had a chat with Marco about the special logic for upgrades, given the Tablespaces feature.
And his idea was that the tablespace controller should simply not do the `CREATE TABLESPACE` until all the instance pods have the required mount.
If we had that, the tablespaces could follow the same logic as the WAL vol, and the upgrade method would not matter.

I've opened https://github.com/cloudnative-pg/cloudnative-pg/issues/5564
for that.
Tao, can the present PR wait? Or better to fix, and eventually simplify once #5564 is done?

litaocdl:
> The logic for the upgrade is more complicated than it needs to be. Not the fault of this PR, but this PR adds more complexity. Had a chat with @mnencia about the special logic for upgrades, given the Tablespaces feature. And his idea was that the tablespace controller should simply not do the `CREATE TABLESPACE` until all the instance pods have the required mount. If we had that, the tablespaces could follow the same logic as the WAL vol, and the upgrade method would not matter.
> 
> I've opened #5564 for that. Tao, can the present PR wait? Or better to fix, and eventually simplify once #5564 is done?

Hello @jsilvela 

Mount first is a good idea. I think this PR can wait, it is a issue found internal and we just need an explaination why expected switchover is not happened. Let use the alternate way.

jmealo:
I just hit this today. Curious what the correct behavior should be. I wasn't expecting downtime.

mnencia:
I'm closing this PR as there's agreement that this is not the solution we want to implement.

jsilvela:
@jmealo long time I know, but I have a fix for this. #9720 
And @litaocdl would like to get your take on that PR if you have time.

PR #5496, opened 2024-09-11
chore(tests): remove ignore-fails tests

sxd:
The ignore-fails test was created a long time ago to ensure that the `ignore-test` test labels were working as intended. However, we never used it, which indicates that it is not necessary.

Moreover, the new GitHub output from ginkgo links every error, but the ignore-fail test causes too many annotations in the GitHub Summary, which reduces the feature's usefulness.

For this reason, we decided to remove the ignore-fails feature and the corresponding E2E test.


mnencia (approved):

jsilvela (approved):

jsilvela:
@sxd @mnencia I went a bit further and deleted all the ignorable-failures logic we have in the hack scripts.
Also, have a clean run and a run with a forced code error (see above) to check.

NiccoloFei (approved):

PR #5499, opened 2024-09-11
feat: declarative databases

gabriele-wolfox:
Signed-off-by: Gabriele Quaresima <gabriele.quaresima@enterprisedb.com>


gabriele-wolfox:
Opened by mistake while testing Leonardo's one

PR #5516, opened 2024-09-16
feat: enable configuration of `full_page_writes`

### LINKED ISSUE #5506: [Feature]: make full_page_writes a configurable parmeters (currently it's fixed)

Hi,

According to https://cloudnative-pg.io/documentation/1.23/postgresql_conf/#dynamic-shared-memory-settings, `full_page_writes` is a "fixed parameter" (i.e. read-only).

I understand that in most common cases this is a dangerous setting to mess with BUT that's not true when using ZFS for PersistentVolume storage : in that case turning full_page_writes `off` is a well-known optimisation. (see [https://wiki.postgresql.org/wiki/Full_page_writes#When_can_you_turn_it_off](https://wiki.postgresql.org/wiki/Full_page_writes#When_can_you_turn_it_off.3F))

Chances are that this setting was made read-only in the first place because it was more convenient to handle the [https://cloudnative-pg.io/documentation/1.23/database_import/#import-optimizations](https://cloudnative-pg.io/documentation/1.23/database_import/#import-optimizations) that temporarily toggle this setting `off` during import then back `on` afterwards and would require a bit of logic/code in the operator to handle the case where this should not be turned back `on` after import in case the user wants it `off` for good.

Remove `full_page_writes` from the list of fixed parmeters and have the operator not turn this setting `off` after initial database import if it's `on` in the Cluster configuration.

They are none :-) 

### Additional context

_No response_

### Backport?

Yes

No

### Code of Conduct

zekiyeaydemir:
This patch enables users to customize the `full_page_writes` GUC in PostgreSQL.
By default, it is set to `on` as per PostgreSQL recommendation.

Closes #5506


sxd:
Hello @zekiyeaydemir !

You have missing the documentation about this, can you make those changes please?

Regards!

.github/pg_versions.json
```diff
@@ -1,7 +1,7 @@
 {
   "17": [
     "17rc1",
-    "17rc1-5"
+    "17rc1-3"
   ],
```
sxd, line 5: @zekiyeaydemir this change shouldn't be here, is downgrading a pg version =)

leonardoce (approved):

gbartolini (approved):

PR #5553, opened 2024-09-19
chore: use cloudnativepg/machinery

leonardoce:


mnencia (approved):

api/v1/backup_webhook.go
```diff
@@ -17,6 +17,7 @@ limitations under the License.
 package v1
 
 import (
+	"github.com/cloudnative-pg/machinery/pkg/log"
```
phisco, line 20: 🤔 maybe we should update golangci-lint's config to group these below too?

mnencia: I don't know. The last group is for imports from the same repo. The `machinery` is part of the same project but in a different repo.

PR #5570, opened 2024-09-22
fix(docs): example `xid_age` alert

ardentperf:
Because autovacuum_freeze_max_age defaults to 200 million, autovacuum generally does not advance the oldest xid on tables and databases until they reach this age. Accordingly, any alert less than 200 million is not useful. A more reasonable default alert would be something like 300 million. Note that vacuum_failsafe_age doesn't start until 1.6 billion by default. If someone is decreasing the failsafe age (for example to under a billion) then it might even make sense to raise the alarming threshold above the failsafe age.

### DIFF
--- a/docs/src/samples/monitoring/alerts.yaml
+++ b/docs/src/samples/monitoring/alerts.yaml
@@ -21,10 +21,10 @@ groups:
       severity: warning
   - alert: PGDatabase
     annotations:
-      description: Over 150,000,000 transactions from frozen xid on pod {{ $labels.pod  }}
+      description: Over 300,000,000 transactions from frozen xid on pod {{ $labels.pod  }}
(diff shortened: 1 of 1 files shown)


ardentperf:
FYI - this is the source of the metric we're talking about

https://github.com/cloudnative-pg/charts/blob/0bfb29772e9544f366e097dce739000cfc091bf3/charts/cloudnative-pg/values.yaml#L255

ardentperf:
it probably should also be changed at https://github.com/cloudnative-pg/cloudnative-pg/blob/2e9cf42abdb8595faf1a354bd08342f1b784bf0c/docs/src/samples/monitoring/prometheusrule.yaml#L32

gbartolini (approved):

leonardoce (approved):

PR #5574, opened 2024-09-23
chore: abort certificate creation upon fatal error detection

### LINKED ISSUE #5575: [Bug]:  leaf certificates should not be generated when encountering a fatal error

### Contact Details

_No response_

### Version

1.24.0

1.31

YAML manifest

### What happened?

In the main branch we wrongly try to create leaf certificate when any kind of error is returned but this is wrong given that every error other than NotFound will be fatal in this scenario. We should proceed to fix this.

```
// ensureLeafCertificate check if we have a certificate for PostgreSQL and generate/renew it
func (r *ClusterReconciler) ensureLeafCertificate(
	ctx context.Context,
	cluster *apiv1.Cluster,
	secretName client.ObjectKey,
	commonName string,
	caSecret *v1.Secret,
	usage certs.CertType,
	altDNSNames []string,
	additionalLabels map[string]string,
) error {
	var secret v1.Secret
	err := r.Get(ctx, secretName, &secret)
	if err == nil {
		return r.renewAndUpdateCertificate(ctx, caSecret, &secret, altDNSNames)
	}

	serverSecret, err := generateCertificateFromCA(caSecret, commonName, usage, altDNSNames, secretName)
	if err != nil {
		return err
	}

...rest of the code
```

### Cluster resource

_No response_

### Relevant log output

_No response_

### Code of Conduct

armru:
Closes #5575 

## Release notes
Fixed an issue that resulted in the operator trying  to generate CA certificates when encountering fatal errors.

## Note for reviewers
- This PR could potentially be split in two
- This PR simplifies and uniforms the leaf generation code
- This PR is labeled as fix due to ensureLeafCertificate proceeding even when the error was fatal

```
// ensureLeafCertificate check if we have a certificate for PostgreSQL and generate/renew it
func (r *ClusterReconciler) ensureLeafCertificate(
	ctx context.Context,
	cluster *apiv1.Cluster,
	secretName client.ObjectKey,
	commonName string,
	caSecret *v1.Secret,
	usage certs.CertType,
	altDNSNames []string,
	additionalLabels map[string]string,
) error {
	var secret v1.Secret
	err := r.Get(ctx, secretName, &secret)
	if err == nil {
		return r.renewAndUpdateCertificate(ctx, caSecret, &secret, altDNSNames)
	}

	serverSecret, err := generateCertificateFromCA(caSecret, commonName, usage, altDNSNames, secretName)
	if err != nil {
		return err
	}

...rest of the code
```

### DIFF
--- a/internal/controller/cluster_pki.go
+++ b/internal/controller/cluster_pki.go
@@ -43,23 +43,7 @@ func (r *ClusterReconciler) setupPostgresPKI(ctx context.Context, cluster *apiv1
 		return fmt.Errorf("generating server CA certificate: %w", err)
 	}
 
-	// This is the certificate for the server
-	serverCertificateName := client.ObjectKey{Namespace: cluster.GetNamespace(), Name: cluster.GetServerTLSSecretName()}
-	opts := x509.VerifyOptions{KeyUsages: []x509.ExtKeyUsage{x509.ExtKeyUsageServerAuth}}
-	err = r.ensureServerLeafCertificate(
-		ctx,
-		cluster,
-		serverCertificateName,
-		cluster.GetServiceReadWriteName(),
-		serverCaSecret,
-		certs.CertTypeServer,
-		cluster.GetClusterAltDNSNames(),
-		&opts)
-	if err != nil {
-		if apierrors.IsNotFound(err) {
-			return fmt.Errorf("missing specified server TLS secret %s: %w",
-				cluster.Status.Certificates.ServerTLSSecret, err)
-		}
+	if err = r.ensureServerLeafCertificate(ctx, cluster, serverCaSecret); err != nil {
 		return fmt.Errorf("generating server TLS certificate: %w", err)
 	}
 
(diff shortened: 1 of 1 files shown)


internal/controller/cluster_pki.go
```diff
@@ -329,23 +323,26 @@ func (r *ClusterReconciler) ensureLeafCertificate(
...
+				serverSecret.Labels = make(map[string]string)
+			}
+			serverSecret.Labels[k] = v
 		}
-		serverSecret.Labels[k] = v
+		return r.Create(ctx, serverSecret)
+	default:
+		return err
 	}
-	return r.Create(ctx, serverSecret)
```
armru, line 348: the only case where this logic should be executed is when we get a not found, in the other case we should bubble up the error

armru: I meant to refer the old code alone but I couldn't find a way

mnencia (approved):

leonardoce:
E2e: https://github.com/EnterpriseDB/cloudnative-pg/actions/runs/11232732833

leonardoce (approved):

PR #5576, opened 2024-09-23
test(e2e): prevent VolumeSnapshot E2E to fail because of "tar"

leonardoce:
E2e tests are using the CSI-hostpath implementation of VolumeSnapshots, which is based on `tar`.

When taking a snapshot of a running PostgreSQL instance, `tar` may download the content of a file that was present at the start of the process but removed later.

With this patch, the setup script is using the `ignoreFailedRead` VolumeSnapshotClass to prevent the VolumeSnapshot from failing because of that.

See: https://github.com/kubernetes-csi/csi-driver-host-path/pull/543


fcanovai (approved):

leonardoce:
E2e: https://github.com/EnterpriseDB/cloudnative-pg/actions/runs/10997238853

mnencia (approved):

sxd (approved):

PR #5581, opened 2024-09-23
feat: set finalizers on the PVC-initialization jobs

leonardoce:
The operator is using a few jobs to initialize the PVC before starting PostgreSQL on them.

To ensure that the final status of the Jobs can be collected by the operator, the code is now setting a finalizer on them and removing it before deleting the Job.

### DIFF
--- a/internal/controller/cluster_cleanup.go
+++ b/internal/controller/cluster_cleanup.go
@@ -23,6 +23,7 @@ import (
 	batchv1 "k8s.io/api/batch/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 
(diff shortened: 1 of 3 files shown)


leonardoce:
E2e: https://github.com/EnterpriseDB/cloudnative-pg/actions/runs/10997959386

leonardoce:
E2e tests are all green.

gbartolini (approved):

mnencia (approved):

internal/controller/cluster_cleanup.go
```diff
@@ -44,8 +45,23 @@ func (r *ClusterReconciler) cleanupCompletedJobs(
 			continue
 		}
 
+		contextLogger.Debug("Removing finalizer from job", "job", job.Name)
```
gabriele-wolfox, line 48: I would move the logger just before the patch, otherwise if `removeFinalizer` returns false you are logging anyway.

leonardoce:
This is not useful anymore, we discovered that our E2e tests were deleting the Job by mistake and fixed the actual issue.

PR #5582, opened 2024-09-23
chore: use builtin multi namespace watcher

### LINKED ISSUE #5688: [Chore]: use builtin multi namespace watcher

The operator uses a custom-made code to watch multiple namespaces

The operator should use the multi namespace watcher offered by controller-runtime.

NA

### Additional context

We should carefully test this to avoid introducing new bugs

### Backport?

No

Yes

### Code of Conduct

armru:
Closes #5688


leonardoce:
I think this should be thoroughly tested on OpenShift, installing the operator in multiple namespaces but not in the whole cluster.

internal/cmd/manager/controller/controller.go
```diff
@@ -314,6 +307,20 @@ func RunController(
...
+	if len(conf.WatchNamespace) == 0 {
+		setupLog.Info("Listening for changes on all namespaces")
+		return nil
+	}
+	namespaces := map[string]cache.Config{}
+	for _, namespace := range conf.WatchedNamespaces() {
+		namespaces[namespace] = cache.Config{}
+	}
+
+	setupLog.Info("Listening for changes", "watchNamespaces", namespaces)
```
fcanovai, line 320: We'll get an error in the logs this way:
```
{"level":"info","ts":"2024-10-01T12:33:02Z","logger":"setup","msg":"Listening for changes","watchNamespacesError":"json: unsupported type: cache.TransformFunc"}
```

```suggestion
	setupLog.Info("Listening for changes", "watchNamespaces", 
		append(conf.WatchedNamespaces(), conf.OperatorNamespace))

```

internal/cmd/manager/controller/controller.go
```diff
@@ -314,6 +307,20 @@ func RunController(
...
 
+func getNamespacesToWatch(conf *configuration.Data) map[string]cache.Config {
+	if len(conf.WatchNamespace) == 0 {
+		setupLog.Info("Listening for changes on all namespaces")
+		return nil
+	}
+	namespaces := map[string]cache.Config{}
+	for _, namespace := range conf.WatchedNamespaces() {
+		namespaces[namespace] = cache.Config{}
+	}
```
fcanovai, line 318: This won't watch any more the namespace in which the operator is installed unless explicitly asked by the user, in case of a multi-namespace setup in openshift. Cluster creation will fail because the operator will try to copy secrets from its own namespace to the cluster's.


We could use something like
```suggestion
	for _, namespace := range conf.WatchedNamespaces() {
		namespaces[namespace] = cache.Config{}
	}
	namespaces[conf.OperatorNamespace] = cache.Config{}
```
implicitly adding the operator namespace to the watched ones.

fcanovai (changes requested):

internal/cmd/manager/controller/controller.go
armru: It wont because the function will exit if 
```
	if len(conf.WatchNamespace) == 0 {
		setupLog.Info("Listening for changes on all namespaces")
		return nil
	}
```

armru:
@fcanovai @sxd made the requested changes

armru:
this should be merged eventually

mnencia:
Moved to 1.28.0 because we want to test this change for some time before releasing it.

armru:
closed in favor of #9412

PR #5643, opened 2024-09-26
Update troubleshooting documentation for Linkerd integration

99hops:


gbartolini:
Can someone with Linkerd experience review this?

docs/src/troubleshooting.md
```diff
@@ -800,3 +800,68 @@ API. Please check your networking.
...
+
+If your Cluster's initialization job is stuck in the `Running` status, it may be due to the Linkerd sidecar remaining active after the job has completed. Linkerd's proxy `/shutdown` endpoint is **disabled by default**, which prevents the sidecar from gracefully terminating, leaving the job in a stalled state.
+
+When the CNPG operator completes its `import` or `bootstrap` task, it attempts to shut down the sidecars to complete the job. However, without the `/shutdown` endpoint being enabled, Linkerd's proxy remains active, causing the job to never fully exit.
+
+#### Debugging the Issue
+
+To diagnose this issue, you can:
+
+1. **Port-forward** to the `-join1` pod to access the Linkerd proxy.
```
sxd, line 814: the format includes a `-join-1` there's a missing dash, also, do you think it's possible to add the same for the initdb, which have the same problem

sxd: default              cluster-example-1-initdb-rlsrs                                  1/2     NotReady   0          2m30s
that's the sample I got

docs/src/troubleshooting.md
```diff
@@ -800,3 +800,68 @@ API. Please check your networking.
...
+
+    ```bash
+    kubectl port-forward pod/postgresql-join1 4191:4191 -n <namespace>
```
sxd, line 817: in all the documentation we use cluster-example as the name of the cluster, we need to keep the same name

docs/src/troubleshooting.md
```diff
@@ -800,3 +800,68 @@ API. Please check your networking.
...
+
+   If you receive the response `shutdown endpoint is not enabled`, this confirms that the shutdown endpoint is disabled by default.
+
+#### Enabling the `/shutdown` Endpoint
+
+There are two ways to resolve this:
+
+1. **Adding an Annotation to the CNPG Cluster Resource**:
+
+   You can add the following annotation to your CNPG `Cluster` manifest to enable the shutdown endpoint in the Linkerd proxy:
```
sxd, line 834: I think it's important to clarify that this needs to be done before the cluster is created since it will not be propagated once the cluster is created, meaning the pods will get stuck

docs/src/troubleshooting.md
```diff
@@ -800,3 +800,68 @@ API. Please check your networking.
...
+    metadata:
+      name: postgresql
+      namespace: dev
+    spec:
+      inheritedMetadata:
+        annotations:
+          config.linkerd.io/proxy-admin-shutdown: enabled  # Enable Linkerd's shutdown endpoint
+    ```
+
+2. **Enabling the Shutdown Endpoint via Helm in ArgoCD**:
```
sxd, line 848: This only covers this case, we should be more generic like "add the following configurations to .. when deploying... " make it more generic

docs/src/troubleshooting.md
```diff
@@ -800,3 +800,68 @@ API. Please check your networking.
...
+          enableShutdownEndpoint: true  # Enable the shutdown endpoint globally
+    ```
+
+   This will configure Linkerd to allow the CNPG operator to successfully shut down the proxy when needed.
+
+#### Summary
+
+If the Linkerd `/shutdown` endpoint is not enabled, it will prevent the CNPG bootstrap job from completing successfully, leaving the job in a `Running` state indefinitely. To fix this:
+- Enable the shutdown endpoint via an annotation (`config.linkerd.io/proxy-admin-shutdown: enabled`) or
+- Modify your Helm installation (`proxy.enableShutdownEndpoint: true`) if using ArgoCD and Helm.
```
sxd, line 865: again we should be generic here

Kakadus:
In case anyone works on this: I needed to add `config.linkerd.io/shutdown-grace-period: "30s"`. Otherwise, the linkerd-proxy container errors and the job always restarts the pod and does not complete. Seems to be related to https://github.com/linkerd/linkerd2/issues/8033.

PR #5647, opened 2024-09-27
docs: complete the database feature documentation

### LINKED ISSUE #5587: [split from `#4679`]: Improve and finalize documentation
- In addition to the feature documentation, update other existing documentation, such as the operator capability levels and the index page
- mention that changing the name of the database is not supported
- mention that the names `postgres`, `template0`, `template1` are reserved
- what happens when two Database objects deal with the same Postgres database?
- add a section on finalizer cleanup -- perhaps not in this document, as the finalizer cleanup is becoming more widespread
- explain what happens if the CRD contains fields that are not supported by the operand version
-

jsilvela:
Closes #5587

### DIFF
--- a/docs/src/declarative_database_management.md
+++ b/docs/src/declarative_database_management.md
@@ -24,9 +24,7 @@ spec:
 Note that the Database references a Cluster, on which the database will be
 created.
 The Database object is managed by the instance manager of the cluster's
-primary instance. Declarative database management is not supported in replica
-clusters, as replica clusters lack a primary instance to manage the Database
-object.
+primary instance.
 
 In the CRD, the `metadata.name` field represents the name the object
 will have in Kubernetes, which is guaranteed to be unique per namespace.
@@ -45,6 +43,13 @@ the database created in Postgres.
     and [ALTER](https://www.postgresql.org/docs/current/sql-alterdatabase.html)
     commands, it does not support renaming of databases. Changing the
     `spec.name` in a Database object will be rejected at the Kubernetes level.
+
+### Database objects defined on  replica clusters
+
+Database objects declared on a replica cluster cannot be enforced, since the
+replica does not have write privileges.
+Instead, a database object defined on a replica cluster will be periodically
+re-queued, and will be enforced once the cluster is promoted.
 
 ### Reserved names
 
--- a/docs/src/declarative_database_management.md
+++ b/docs/src/declarative_database_management.md
@@ -202,12 +202,13 @@ Database object, directly on Postgres, for example by issuing commands on
 The fields `observedGeneration` and `generation` described above will ensure
 that once a Database has been reconciled to its defined `generation`, it will
 not be re-applied by the instance manager. Therefore, your manual changes will
-not be rolled back accidentally.
+not be rolled back inadvertently.
 
 !!! Note
     A Database manifest is applied to the fullest. A field included in a
     manifest may override a value that had been written previously on
     the database in Postgres
 
-The choice is yours to make alterations on your databases via Database manifests
-or using other mechanisms.
+CloudNativePG gives you the flexibility to make alterations on your databases
+via Database manifests, via direct changes, or mixing matching to fit your
+use case.
--- a/docs/src/declarative_database_management.md
+++ b/docs/src/declarative_database_management.md
@@ -44,6 +44,31 @@ the database created in Postgres.
     commands, it does not support renaming of databases. Changing the
     `spec.name` in a Database object will be rejected at the Kubernetes level.
 
+### Managing an existing database via a Database manifest
+
+It is possible to declare a Database object that will handle an existing
+database. In such case, the Database's fields will be applied using `ALTER`
+statements, rather than `CREATE`. There are differences between these two
+Postgres commands. In particular, the options accepted by `ALTER` are a subset
+of those accepted by `CREATE`.
+
+The database reconciler will transparently handle this on behalf of the user,
+making it easy to honor a Database manifest no matter the previous history
+of the cluster.
+
+There is however a difference regarding handling of "immutable" fields: on an
+existing Database object, any modification that alters the immutable fields will
+be rejected at the Kubernetes level.
+On a newly declared Database manifest that references an existing database, the
+immutable fields will simply be ignored, as they are not valid options in the
+`ALTER DATABASE` command.
+
+!!! Warning
+    If a Database manifest references an existing database, any fields in the
+    manifest that cannot be set in `ALTER DATABASE` will be dropped silently.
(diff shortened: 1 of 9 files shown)


docs/src/operator_capability_levels.md
```diff
@@ -287,6 +293,7 @@ upgrades.
 
 ### Operator Upgrade
 
+
```
jsilvela, line 296:
```suggestion
```

docs/src/declarative_database_management.md
```diff
@@ -23,25 +21,81 @@ spec:
     name: cluster-example
 ```
 
-Once the reconciliation cycle is completed successfully, the `Database` 
-status will show a `ready` field set to `true` and an empty `error` field.
+Note that the Database references a Cluster, on which the database will be
+created.
+The Database object is managed by the instance manager of the cluster's
+primary instance. Declarative database management is not supported in replica
```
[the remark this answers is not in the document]
jsilvela: Ah, interesting. OK on one hand, the Database object reconciliation would fail, and would be re-queued, so if eventually we promoted the cluster, the database reconciliation would proceed.
Now, in a different direction: this feature is not so much doing "management" as doing declarative creation. Once a manifest is successfully applied, it's over and done until we alter it (and the `generation` is incremented).
So, to understand your use case better: if we create Database Foo in the primary, then it will be replicated in the replica cluster, no? So having a copy of the Database manifest in the replica cluster would not be of any use.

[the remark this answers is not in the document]
jsilvela: I've rewritten this, I hope it's now clearer.

docs/src/declarative_database_management.md
```diff
@@ -57,4 +116,104 @@ spec:
...
+The fields `observedGeneration` and `generation` described above will ensure
+that once a Database has been reconciled to its defined `generation`, it will
+not be re-applied by the instance manager. Therefore, your manual changes will
+not be rolled back inadvertently.
+
+!!! Note
+    A Database manifest is applied to the fullest. A field included in a
+    manifest may override a value that had been written previously on
+    the database in Postgres
+
```
jsilvela, line 216:
```suggestion
```

docs/src/declarative_database_management.md
```diff
@@ -23,25 +21,111 @@ spec:
...
+statements, rather than `CREATE`. There are differences between these two
+Postgres commands. In particular, the options accepted by `ALTER` are a subset
+of those accepted by `CREATE`.
+
+The database reconciler will transparently handle this on behalf of the user,
+making it easy to honor a Database manifest no matter the previous history
+of the cluster.
 
-### Database Deletion and Reclaim Policies
+There is however a difference regarding handling of "immutable" fields: on an
```
jsilvela, line 59:
```suggestion
There is however a difference regarding the handling of "immutable" fields: on an
```

jsilvela, line 59:
```suggestion
There is however a difference regarding the handling of "immutable" fields: on an
```

docs/src/declarative_database_management.md
```diff
@@ -23,25 +21,111 @@ spec:
...
+There is however a difference regarding the handling of "immutable" fields: on an
+existing Database object, any modification of the immutable fields will
+be rejected at the Kubernetes level.
+On a newly declared Database manifest that references an existing database, the
+immutable fields will simply be ignored, as they are not valid options in the
+`ALTER DATABASE` command.
+
+!!! Warning
+    If a Database manifest references an existing database, any fields in the
+    manifest that cannot be set in `ALTER DATABASE` will be dropped silently.
```
jsilvela, line 68:
```suggestion
    manifest that cannot be set in `ALTER DATABASE` will be ignored.
```

docs/src/declarative_database_management.md
```diff
@@ -23,25 +21,111 @@ spec:
...
+    While declarative database management in CloudNativePG adheres to the
+    PostgreSQL database
+    [CREATE](https://www.postgresql.org/docs/current/sql-createdatabase.html)
+    and [ALTER](https://www.postgresql.org/docs/current/sql-alterdatabase.html)
+    commands, it does not support renaming of databases. Changing the
+    `spec.name` in a Database object will be rejected at the Kubernetes level.
+
+### Managing an existing database via a Database manifest
+
+It is possible to declare a Database object that will handle an existing
```
jsilvela, line 49:
```suggestion
It is possible to declare a Database object that references an existing
```

gabriele-wolfox:
LGTM

gabriele-wolfox (approved):

gbartolini:
Can you please rebase this @jsilvela @gabriele-wolfox? I have got conflicts and I am afraid I might make a mistake. Thanks.

gbartolini (approved):

docs/src/declarative_database_management.md
```diff
@@ -1,54 +1,130 @@
-# Declarative Database Management
+# PostgreSQL Database Management
 
-Declarative database management enables users to control the lifecycle of
-databases via a new Custom Resource Definition (CRD) called `Database`.
+CloudNativePG simplifies PostgreSQL database provisioning by automatically
```
jsilvela, line 3: I think this makes more sense if we speak of cluster.
```suggestion
CloudNativePG simplifies PostgreSQL cluster provisioning by automatically
```

docs/src/declarative_database_management.md
```diff
@@ -1,54 +1,130 @@
...
-Below is an example of a basic `Database` configuration:
+## Key Concepts
+
+### Scope of Management
+
+!!! Important
+    CloudNativePG manages **global objects** in PostgreSQL clusters, such as
+    databases, roles, and tablespaces. However, it does **not** manage the content
+    of databases (e.g., schemas and tables). For database content, specialized
+    tools or the applications themselves should be used.
```
jsilvela, line 24: Not sure about "application themselves". Which applications? I can make that you're talking about the user payloads, but IMO this is not obvious.
```suggestion
    tools, or the user-defined payloads should be used.
```

mnencia (approved):

jsilvela (commented):
two nits

docs/src/declarative_database_management.md
gbartolini: I actually meant the database.

jsilvela: So, we say CNPG simplifies database provisioning by creating the `app` database automatically.
An the rest of the document talks about a completely different way of creating databases.

mnencia: The idea is to explain why we always create the app database.

jsilvela: 👍

PR #5651, opened 2024-09-27
fix(cnpg-i): ensure instance manager invokes only the available plugins

### LINKED ISSUE #5648: [Bug]: CNPG-I instance manager plugins invocation

### Contact Details

gabriele.fedi@enterprisedb.com

### Version

1.24.0

1.31

YAML manifest

### What happened?

When plugins are declared in the Cluster manifest, the instance manager tries to iterate over them to delegate the WAL archiving and restore, trying to find a plugin having such capability. Once having invoked WAL archiving on capable plugins, then the task is also delegated to barman-cloud. 
Some plugins though, might be running only as sidecars in the operator's pod and not in the operand ones, as they might support capabilities meaningful at operator level only, ie `LifecycleCapabilities`.  What happens is that the instance manager tries to load those plugins, which are not available in the operand, complaining that it can't find them.

### Cluster resource

```shell
kind: Cluster
metadata:
  name: cnpg-plugin-test
spec:
  instances: 3

  plugins:
  - enabled: true
    name: lifecycle-only.plugin.com

  storage:
    size: 1Gi
```

### Relevant log output

```shell
{"level":"error","ts":"2024-09-26T12:15:45Z","logger":"wal-archive","msg":"Error while loading required plug
ins","logging_pod":"p-qmpfrd13ck-1","error":"Unknown plugin: lifecycle-only.plugin.com","stacktrace":"github.com/cloudnative-pg/cloudnative-pg/pkg/management/log.(*logger).Error\n\tcloudnative-pg/pkg/management/log/log.go:125\ngithub.com/cloudnative-pg/cloudnative-pg/internal/cmd/ma
nager/walarchive.archiveWALViaPlugins\n\tcloudnative-pg/internal/cmd/manager/walarchive/cmd.go:264\ng
ithub.com/cloudnative-pg/cloudnative-pg/internal/cmd/manager/walarchive.run\n\tcloudnative-pg/in
ternal/cmd/manager/walarchive/cmd.go:166\ngithub.com/cloudnative-pg/cloudnative-pg/internal/cmd/manager
/walarchive.NewCmd.func1\n\tcloudnative-pg/internal/cmd/manager/walarchive/cmd.go:92\ngithub.com/spf1
3/cobra.(*Command).execute\n\tcloudnative-pg/vendor/github.com/spf13/cobra/command.go:985\ngithub.com
/spf13/cobra.(*Command).ExecuteC\n\tcloudnative-pg/vendor/github.com/spf13/cobra/command.go:1117\ngit
hub.com/spf13/cobra.(*Command).Execute\n\tcloudnative-pg/vendor/github.com/spf13/cobra/command.go:104
1\nmain.main\n\tcloud-native-postgres/cmd/manager/main.go:70\nruntime.main\n\tpkg/mod/golang.org/toolchain@v
0.0.1-go1.23.1.linux-amd64/src/runtime/proc.go:272"}
```

### Code of Conduct

GabriFedi97:
The instance manager should try to load only available plugins, as some of them declared in the Cluster spec might be available only to the operator.

closes #5648

### DIFF
--- a/internal/cnpi/plugin/client/client.go
+++ b/internal/cnpi/plugin/client/client.go
@@ -18,6 +18,7 @@ package client
 
 import (
 	"context"
+	"errors"
 
 	"github.com/cloudnative-pg/machinery/pkg/log"
 
@@ -42,9 +43,27 @@ func (data *data) getPlugin(pluginName string) (connection.Interface, error) {
 	return nil, ErrPluginNotLoaded
 }
 
+// load loads the plugins with the specified name
 func (data *data) load(ctx context.Context, names ...string) error {
 	for _, name := range names {
 		pluginData, err := data.repository.GetConnection(ctx, name)
+		if err != nil {
+			return err
+		}
+
+		data.plugins = append(data.plugins, pluginData)
+	}
+	return nil
+}
+
+// loadAvailable loads only the plugins that are actually available, matching the names
+func (data *data) loadAvailable(ctx context.Context, names ...string) error {
+	for _, name := range names {
+		pluginData, err := data.repository.GetConnection(ctx, name)
+		var unknownPluginErr *repository.ErrUnknownPlugin
+		if errors.As(err, &unknownPluginErr) {
+			continue
+		}
 		if err != nil {
 			return err
 		}
@@ -84,6 +103,20 @@ func WithPlugins(ctx context.Context, repository repository.Interface, names ...
 		repository: repository,
 	}
 	if err := result.load(ctx, names...); err != nil {
+		return nil, err
+	}
+
+	return result, nil
+}
+
+// WithAvailablePlugins creates a new CNP-I client for plugins in a certain repository,
+// loading the available plugins with the specified name. If a plugin is not available,
+// it will be skipped.
+func WithAvailablePlugins(ctx context.Context, repository repository.Interface, names ...string) (Client, error) {
+	result := &data{
+		repository: repository,
+	}
+	if err := result.loadAvailable(ctx, names...); err != nil {
 		return nil, err
 	}
 
--- a/internal/cnpi/plugin/client/client.go
+++ b/internal/cnpi/plugin/client/client.go
@@ -18,7 +18,6 @@ package client
 
 import (
 	"context"
-	"errors"
 
 	"github.com/cloudnative-pg/machinery/pkg/log"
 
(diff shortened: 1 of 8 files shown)


internal/cnpi/plugin/client/client.go
```diff
@@ -54,6 +56,23 @@ func (data *data) load(ctx context.Context, names ...string) error {
 	return nil
 }
 
+// loadAvailable loads only the plugins that are actually available, matching the names
+func (data *data) loadAvailable(ctx context.Context, names ...string) error {
+	for _, name := range names {
+		pluginData, err := data.repository.GetConnection(ctx, name)
+		var unknownPluginErr *repository.ErrUnknownPlugin
+		if errors.As(err, &unknownPluginErr) {
```
GabriFedi97, line 64: May it worth to log a warning here?

leonardoce:
I simplified your implementation @GabriFedi97, give it a look

GabriFedi97:
Hey @leonardoce, thanks for the review. I agree, your implementation is certainly cleaner. I believe your implementation could behave slightly different though and there is an edge case which is worth to mention IMO. 
Assuming to have a plugin running on both operator side, for registering itself as a operand sidecar, and operand side to handle WAL related hooks, with my implementation I tried to keep relying on the Cluster spec, instead of just relying on the available sockets in plugin path. I saw that we are not restarting/recreating the pg pods if the plugins change or get disabled in the Cluster spec, which means that if a plugin socket is present on the plugins path, with the current implementation it would be kept considered for WAL archiving invocation (or other related capabilities). In that case the pods should be deleted manually, to let the operator recreating it without the disabled plugin sidecar.

leonardoce:
The main problem is that plugins are not being uninstalled, and this happens regardless of whether this patch is applied.
We certainly need to address that. Let me think about it.

GabriFedi97:
Yeah right, they would just be ignored as the `GetEnabledPluginNames` [function](https://github.com/cloudnative-pg/cloudnative-pg/blob/0c61f4cee71e3c7b4fd5632c9835848bb07a431b/api/v1/cluster_funcs.go#L72) would not return them and the plugin would get uninstalled during the next pod recreation.

leonardoce:
Yes, but with your implementation that was implied and buried inside the error handling, and not explicit.
I'm trying to make this clearer.

leonardoce:
https://github.com/EnterpriseDB/cloudnative-pg/actions/runs/11073055607

leonardoce (approved):

armru (approved):

PR #5677, opened 2024-09-28
fix: get the latest major from DefaultImageName for tests

sxd:
We were using a fixed number for the latest PostgreSQL version, now we
get that from the DefaultImageName to keep up with further upgrades of
PostgreSQL major versions.


mnencia (approved):

api/v1/cluster_webhook_test.go
```diff
@@ -1440,6 +1441,8 @@ var _ = Describe("validate image name change", func() {
 			Expect(clusterNew.validateImageChange(&clusterOld)).To(HaveLen(1))
 		})
 		It("doesn't complain going from default imageName to same major imageCatalogRef", func() {
+			majorDefaultVersion, err := strconv.ParseFloat(utils.GetImageTag(versions.DefaultImageName), 32)
+			Expect(err).ToNot(HaveOccurred())
```
jsilvela, line 1445: I think that to unit test this function it's better if we isolate ourselves from the DefultImage.
The parsing here could fail, and that would tell us nothing about the capacity to upgrade
from imageName to imageCatalogRef.

jsilvela (commented):
IMO we should not be parsing the DefaultImage in this test.
This could fail and it would tell us nothing about the webhooks.
How about setting a dummy major version and making sure the old cluster has it:

```
			majorVersion := 16
			clusterOld := Cluster{
				Spec: ClusterSpec{
					ImageName: fmt.Sprintf("ghcr.io/cloudnative-pg/postgresql:%d.0", majorVersion),
				},
			}
			clusterNew := Cluster{
				Spec: ClusterSpec{
					ImageCatalogRef: &ImageCatalogRef{
						TypedLocalObjectReference: corev1.TypedLocalObjectReference{
							Name: "test",
							Kind: "ImageCatalog",
						},
						Major: majorVersion,
```

gbartolini:
Abandon

PR #5689, opened 2024-09-30
fix: remove `pg_database_size` from status probe

### LINKED ISSUE #5686: [chore]: remove `pg_database_size` from the instance manager probe

The instance manager probe currently asks for the size of all the databases in an instance. See https://github.com/cloudnative-pg/cloudnative-pg/blob/25281f568abe7eed3f3f161e479bdfdf391ffad2/pkg/management/postgres/probes.go#L97

In large deployments this might take a long time and fail.

Remove it from the instance manager status and move this in the plugin directly for the `cnpg status` command.

N/A

### Additional context

_No response_

### Backport?

Yes

Yes

### Code of Conduct

leonardoce:
The instance manager status probe was using `pg_database_size` which scans the whole PGDATA to compute the current size of every database.

Instead of doing that in the status probe, causing a scan in each reconciliation loop, we now do that only when requested by the `kubectl cnpg status` plugin.

Closes: #5686

### DIFF
--- a/internal/cmd/plugin/status/status.go
+++ b/internal/cmd/plugin/status/status.go
@@ -66,6 +66,9 @@ type PostgresqlStatus struct {
 
 	// ErrorList store the possible errors while getting the PostgreSQL status
 	ErrorList []error
+
+	// The size of the cluster
+	TotalClusterSize string
 }
 
 func (fullStatus *PostgresqlStatus) getReplicationSlotList() postgres.PgReplicationSlotList {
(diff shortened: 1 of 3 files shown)


fcanovai (approved):

leonardoce:
E2e: https://github.com/EnterpriseDB/cloudnative-pg/actions/runs/11109311647

gbartolini (approved):

PR #5694, opened 2024-10-02
Add readiness/liveness probe timeout options as well as startup initial delay

### LINKED ISSUE #4852: [Feature]: Make pod liveness/readiness probes timeout configurable

Current databaze pods have 5 seconds timeout to liveness/readiness probes and this timeout is hardcoded in `pkg/specs/pods.go`. It would be nice, if this could be configurable. In our setup, we need timeout about 20 seconds, it seems that DNS is involved in the probe and if DNS reply is lost, the Pod is restarted. With increased timeout, all is running just fine. I understand, it should not be default, but configurable.

Add new field to cluster spec so that Pod probes timeout can be changed. 

I do not know about alternative as the timeout is currently hardcoded.

### Additional context

_No response_

### Backport?

No

Yes

### Code of Conduct

bentastic27:
resolves https://github.com/cloudnative-pg/cloudnative-pg/issues/4852

Tested with this:

```
kind: Cluster
metadata:
  name: somecluster
spec:
  instances: 1
  startupProbeInitialDelaySeconds: 30
  startupProbeTimeoutSeconds: 10
  livenessProbeTimeoutSeconds: 10
  readinessProbeTimeoutSeconds: 10
  storage:
    size: 1Gi
```

### DIFF
--- a/api/v1/cluster_types.go
+++ b/api/v1/cluster_types.go
@@ -364,6 +364,26 @@ type ClusterSpec struct {
 	// +optional
 	LivenessProbeTimeout *int32 `json:"livenessProbeTimeout,omitempty"`
 
+	// Number of seconds after the container has started before startup probes are initiated.
+	// More info: https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle#container-probes
+	// +optional
+	StartupProbeInitialDelaySeconds int32 `json:"startupProbeInitialDelaySeconds,omitempty"`
+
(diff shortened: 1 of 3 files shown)


gbartolini:
Superseded by #6266

gbartolini:
@bentastic27 could you please check the PR #6266 and make sure it is OK with you?

bentastic27:
Looks good @gbartolini , thanks!

PR #5697, opened 2024-10-02
fix(backup): add `BackupStartingCondition` condition to cluster while using the plugin method

armru:
## Release Notes

The operator now correctly adds the `BackupStartingCondition` to the cluster resource while taking a backup with a plugin.

## Note for reviewers

Previously while taking a backup we only set the backup failed condition but we never communicated when the backup was starting, this patch aims to fix that


leonardoce (approved):

leonardoce:
E2e: https://github.com/EnterpriseDB/cloudnative-pg/actions/runs/11164985246

leonardoce:
E2e tests are green!

gbartolini (approved):

PR #5735, opened 2024-10-06
chore(tests): split backup tests into file per backend

### LINKED ISSUE #5632: [Bug]: Split backup & restore tests into different object store technology

### Contact Details

_No response_

### Version

1.24.0

1.31

YAML manifest

### What happened?

The current backup&restore tests is all in one file, this make a bit difficult to debug and add new ones, since there's only two Describe() with similar name but inside we can find tests that use MinIO, Azurite and Azure under the same topic., currently is kind of difficult to modify one tests related to MinIO because we need to skip between unrelated tests and is really hard to see what is actually needed to change without breaking other tests.

I propose to split these tests into three files, one for MinIO one for Azurite and another for Azure so we can make a clear distinction between them when add/fix tests and when adding other backup&restore tests with different technologies in the future.

### Cluster resource

_No response_

### Relevant log output

_No response_

### Code of Conduct

sxd:
The backup and restore tests were living in the same file for all the
different backend (minio, azurite, azure) therefore we had a big
confusing file, now, every backend has is own file with is own
functions to work with.

Closes #5632


fcanovai (approved):

NiccoloFei:
E2E run here: https://github.com/EnterpriseDB/cloudnative-pg/actions/runs/11443886312

NiccoloFei (approved):

PR #5774, opened 2024-10-09
test(database): add e2e case of declarative database with delete reclaim policy

### LINKED ISSUE #5680: [split from `#4679`]: Improve and finalize E2E coverage
- E2E test also missing a test with retain policy of Delete:
  - we can create a new CRD to re-adopt the existing DB, but with retain: delete
  - then delete the CRD
- E2E new feature_type for declarative databases
- E2E test what happens when we delete the namespace or the CNPG cluster

gabriele-wolfox:
Closes #5680

### DIFF
--- a/tests/e2e/declarative_database_management_test.go
+++ b/tests/e2e/declarative_database_management_test.go
@@ -69,27 +69,9 @@ var _ = Describe("Declarative database management", Label(tests.LabelSmoke, test
 			})
 		})
 
-		assertDatabaseExists := func(namespace, primaryPod, dbname string, shouldContain bool) {
-			Eventually(func(g Gomega) {
-				stdout, _, err := env.ExecQueryInInstancePod(
-					utils.PodLocator{
-						Namespace: namespace,
-						PodName:   primaryPod,
-					},
-					"postgres",
-					"\\l")
-				g.Expect(err).ToNot(HaveOccurred())
-				if shouldContain {
-					g.Expect(stdout).Should(ContainSubstring(dbname))
-				} else {
-					g.Expect(stdout).ShouldNot(ContainSubstring(dbname))
-				}
-			}, 300).Should(Succeed())
-		}
-
 		assertDatabaseHasExpectedFields := func(namespace, primaryPod string, db apiv1.Database) {
 			query := fmt.Sprintf("select count(*) from pg_database where datname = '%s' "+
-				"and encoding = %s and datctype = '%s' and datcollate = '%s'",
+				"and encoding = pg_char_to_encoding('%s') and datctype = '%s' and datcollate = '%s'",
 				db.Spec.Name, db.Spec.Encoding, db.Spec.LcCtype, db.Spec.LcCollate)
 			Eventually(func(g Gomega) {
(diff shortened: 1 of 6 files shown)


tests/e2e/declarative_database_management_test.go
```diff
@@ -48,18 +51,20 @@ var _ = Describe("Declarative databases management test", Label(tests.LabelSmoke
...
+			databaseWithDeleteRetainPolicy  *apiv1.Database
+			err                             error
 		)
 
 		BeforeAll(func() {
-			// Create a cluster in a namespace we'll delete after the test
-			namespace, err = env.CreateUniqueTestNamespace(namespacePrefix)
+			// We set the namespace name as a constant, as we are going to test the namespace deletion inside
+			// the suite (not after), in order to check the finalizers removal
+			err = env.CreateNamespace(namespace)
```
jsilvela, line 67: why not use the CreateUnique function?

tests/e2e/declarative_database_management_test.go
```diff
@@ -154,13 +159,83 @@ var _ = Describe("Declarative databases management test", Label(tests.LabelSmoke
...
+
+				By("verifying the db exists", func() {
+					primaryPodInfo, err := env.GetClusterPrimary(namespace, clusterName)
+					Expect(err).ToNot(HaveOccurred())
+
+					assertDatabaseExists(namespace, primaryPodInfo.Name, dbname, true)
+				})
+			})
+
+			It("maintains the db when Database CRD is removed", func() {
```
jsilvela, line 200: wait ... this is exactly the opposite. Inside it's checking the Postgres DB does NOT exist, line 209

gabriele-wolfox: You are right, it should be "does not maintain..."

tests/e2e/declarative_database_management_test.go
```diff
@@ -101,65 +100,147 @@ var _ = Describe("Declarative databases management test", Label(tests.LabelSmoke
...
-					assertDatabaseExists(namespace, primaryPodInfo.Name, dbname, true)
-				})
+				Eventually(func(g Gomega) {
+					err := env.Client.Get(env.Ctx, databaseNamespacedName, dbObj)
+					g.Expect(err).ToNot(HaveOccurred())
+					g.Expect(dbObj.Status.Ready).Should(BeTrue())
+				}, 300).WithPolling(10 * time.Second).Should(Succeed())
+			})
+			By("deleting the namespace and making sure it succeeds before timeout", func() {
+				err := env.DeleteNamespaceAndWait(namespace, 30)
```
jsilvela, line 242:
```suggestion
				err := env.DeleteNamespaceAndWait(namespace, 60)
```

jsilvela (approved):

tests/e2e/declarative_database_management_test.go
gabriele-wolfox: Yep, I started with 60 seconds, but after exercising the suite I guess it is fine to reduce it.

NiccoloFei (approved):

mnencia (approved):

PR #5775, opened 2024-10-09
fix(plugin): avoid jamming JSON logs with `logs` command

### LINKED ISSUE #5769: [Bug]: `kubectl cnpg logs cluster` does not produce a JSON stream

### Contact Details

leonardo.cecchi@enteprisedb.com

### Version

1.24.0

1.31

YAML manifest

### What happened?

If you create a simple cluster like the one in `docs/src/samples/cluster-example.yaml` and then use `kubectl cnpg logs cluster` to extract the logs, the result is not parsable as a JSON stream.

```
$ kubectl cnpg logs cluster cluster-example | jq
[...]
{
  "level": "info",
  "ts": "2024-10-09T07:05:41Z",
  "msg": "Starting workers",
  "controller": "instance-tablespaces",
  "controllerGroup": "postgresql.cnpg.io",
  "controllerKind": "Cluster",
  "worker count": 1
}
jq: parse error: Expected separator between values at line 24, column 110
```

### Cluster resource

```shell
See docs/src/samples/cluster-example.yaml
```

### Relevant log output

```shell
This is a example of a broken JSON object:

{"level":"info","ts":"2024-10-08T15:25:21Z","msg":"Refreshed configuration file","controller":"instance-clust{"level":"info","ts":"2024-10-08T15:25:00Z","logger":"setup","msg":"Starting CloudNativePG Instance Manager","logging_pod":"cluster-example-1","version":"1.24.0","build":{"Version":"1.24.0-dev122","Commit":"f18665bac","Date":"2024-10-08"}}
```

### Code of Conduct

leonardoce:
`kubectl cnpg logs cluster` collects the log stream of all the Pods belonging to the selected cluster and forwards their output to a final destination stream with no coordination besides basic locking at the byte level.

This resulted in multiple JSON objects being written at the same time, which made the output garbled. Tools relying on the JSON-stream format cannot parse that too.

With this patch, we coordinate the reading process, scan the stream line by line, and write every separate line for the final stream.

Closes: #5769

### DIFF
--- a/pkg/utils/logs/cluster_logs.go
+++ b/pkg/utils/logs/cluster_logs.go
@@ -17,6 +17,7 @@ limitations under the License.
 package logs
 
 import (
+	"bufio"
 	"context"
 	"io"
 	"log"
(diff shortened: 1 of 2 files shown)


pkg/utils/logs/cluster_logs.go
```diff
@@ -237,9 +238,25 @@ func (csr *ClusterStreamingRequest) streamInGoroutine(
 		}
 	}()
 
-	_, err = io.Copy(output, logStream)
-	if err != nil {
-		log.Printf("error sending logs to writer, pod %s: %v", podName, err)
-		return
+	scanner := bufio.NewScanner(logStream)
```
mnencia, line 241: We should increase the maximum line buffer as we do in all the other places.
```suggestion
	scanner := bufio.NewScanner(logStream)
	scanner.Buffer(make([]byte, 0, 4096), 1024*1024)
```

mnencia (approved):

gbartolini (approved):

fcanovai (approved):

PR #5779, opened 2024-10-09
feat(plugin): add `logs pretty` command

### LINKED ISSUE #5770: [Feature]: `kubectl cnpg logs pretty`

The output of the CNPG Pods is a machine-readable JSON stream, but unfortunately is hard on the human eyes.
As an example:

```
$ kubectl cnpg logs cluster cluster-example | head
{"level":"info","ts":"2024-10-09T07:05:41Z","logger":"setup","msg":"Starting CloudNativePG Instance Manager","logging_pod":"cluster-example-3","version":"1.24.0","build":{"Version":"1.24.0-dev122","Commit":"f18665bac","Date":"2024-10-08"}}
{"level":"info","ts":"2024-10-09T07:05:41Z","logger":"setup","msg":"Checking for free disk space for WALs before starting PostgreSQL","logging_pod":"cluster-example-3"}
{"level":"info","ts":"2024-10-09T07:05:41Z","logger":"setup","msg":"starting tablespace manager","logging_pod":"cluster-example-3"}
{"level":"info","ts":"2024-10-09T07:05:41Z","logger":"setup","msg":"starting external server manager","logging_pod":"cluster-example-3"}
{"level":"info","ts":"2024-10-09T07:05:41Z","logger":"setup","msg":"starting controller-runtime manager","logging_pod":"cluster-example-3"}
{"level":"info","ts":"2024-10-09T07:05:41Z","msg":"Starting EventSource","controller":"instance-cluster","controllerGroup":"postgresql.cnpg.io","controllerKind":"Cluster","source":"kind source: *v1.Cluster"}
{"level":"info","ts":"2024-10-09T07:05:41Z","msg":"Starting Controller","controller":"instance-cluster","controllerGroup":"postgresql.cnpg.io","controllerKind":"Cluster"}
{"level":"info","ts":"2024-10-09T07:05:41Z","msg":"Starting EventSource","controller":"instance-database","controllerGroup":"postgresql.cnpg.io","controllerKind":"Database","source":"kind source: *v1.Database"}
{"level":"info","ts":"2024-10-09T07:05:41Z","msg":"Starting Controller","controller":"instance-database","controllerGroup":"postgresql.cnpg.io","controllerKind":"Database"}
{"level":"info","ts":"2024-10-09T07:05:41Z","msg":"Starting webserver","logging_pod":"cluster-example-3","address":"localhost:8010","hasTLS":false}
```

It would be great to have a tool allowing the user to get human-readable logs, possibly familiar to the eyes of a PostgreSQL's DBA.

I'd like to have a `kubectl cnpg logs decode` command allowing me to decode the JSON stream and get something more readable.
That command could also allow me to filter only for the loggers I'm interested in. Like in:

```
$ kubectl logs cluster-example-1 | kubectl cnpg logs decode --filter postgresql --level info
```

Or even:

```
$ kubectl cnpg  logs cluster cluster-example -f | kubectl cnpg logs decode --filter postgresql --level info
```

Manually decoding the logs with some magic helper tool, like `jq`.
That is indeed possible and would cover this use case, too, but would it be nice to have something ready to use?

### Additional context

_No response_

### Backport?

No

Yes

### Code of Conduct

leonardoce:
This commit introduces the `logs pretty` command to the `cnpg` plugin. The command reads a log stream from standard input and outputs a human-readable format to standard output.

Key features:

- Allows filtering logs by specific pods or loggers.
- Supports excluding logs below a specified severity level.

This enhancement provides users with better log readability and more granular control over the logs they wish to view.

Closes: #5770 

# Release notes

```
Add the `logs pretty` command to the `cnpg` plugin to read a log stream from standard input and output a human-readable format, with options to filter log entries.
```

### DIFF
--- a/docs/src/kubectl-plugin.md
+++ b/docs/src/kubectl-plugin.md
@@ -798,7 +798,7 @@ $ kubectl cnpg logs cluster cluster-example | kubectl cnpg logs pretty
 [...]
 ```
 
-Alternatively, it can be used in combination with other commands producing
+Alternatively, it can be used in combination with other commands that produce
 CNPG logs in JSON format, such as `stern`, or `kubectl logs`, as in the
 following example:
 
--- a/docs/src/kubectl-plugin.md
+++ b/docs/src/kubectl-plugin.md
@@ -175,6 +175,7 @@ kubectl cnpg <command> <args...>
     The plugin automatically detects if the standard output channel is connected to a terminal.
     In such cases, it may add ANSI colors to the command output. To disable colors, use the
     `--color=never` option with the command.
+
 ### Generation of installation manifests
 
 The `cnpg` plugin can be used to generate the YAML manifest for the
@@ -779,6 +780,41 @@ kubectl cnpg logs cluster cluster-example --output my-cluster.log
 
 Successfully written logs to "my-cluster.log"
 ```
+
+#### Pretty
+
+The `pretty` sub-command reads a log stream from the process standard input
+and decodes and formats it in a human-readable format.
+
+It is designed to be used in a pipeline with `kubectl cnpg logs cluster` like in
+the following example:
+
+```
+$ kubectl cnpg logs cluster cluster-example | kubectl cnpg logs pretty
+2024-10-09T07:05:41Z cluster-example-1 setup Starting CloudNativePG Instance Manager
+2024-10-09T07:05:41Z cluster-example-1 setup Checking for free disk space for WALs before starting PostgreSQL
+2024-10-09T07:05:41Z cluster-example-1 setup starting tablespace manager
+[...]
+```
+
+The `pretty` sub-commands also support log selection, allowing the user to
+filter out the logs of the selected pods, of the selected loggers, and less
+important than a specified level. The following examples gives a glimpse of how
+to achieve that:
+
+```
+$ kubectl cnpg logs cluster cluster-example | kubectl cnpg logs pretty --pods cluster-example-1 --loggers postgres --log-level info
+2024-10-09T07:05:41Z cluster-example-1 postgres 2024-10-09 07:05:41.458 UTC [31] LOG:  redirecting log output to logging collector process
+2024-10-09T07:05:41Z cluster-example-1 postgres 2024-10-09 07:05:41.458 UTC [31] HINT:  Future log output will appear in directory "/controller/log".
+2024-10-09T07:05:41Z cluster-example-1 postgres LOG: ending log output to stderr
+2024-10-09T07:05:41Z cluster-example-1 postgres LOG: starting PostgreSQL 17.0 (Debian 17.0-1.pgdg110+1) on x86_64-pc-linux-gnu, compiled by gcc (Debian 10.2.1-6) 10.2.1 20210110, 64-bit
+2024-10-09T07:05:41Z cluster-example-1 postgres 2024-10-09 07:05:41.458 UTC [31] LOG:  ending log output to stderr
+2024-10-09T07:05:41Z cluster-example-1 postgres 2024-10-09 07:05:41.458 UTC [31] HINT:  Future log output will go to log destination "csvlog".
+[...]
+```
+
+The `-h` option can be used to get an explanation of all the available flags and
+their semantics.
 
 ### Destroy
 
--- a/docs/src/kubectl-plugin.md
+++ b/docs/src/kubectl-plugin.md
@@ -791,10 +791,10 @@ shown in the following example:
 
 ``` sh
 $ kubectl cnpg logs cluster cluster-example | kubectl cnpg logs pretty
-2024-10-10T16:36:18.668100649Z INFO cluster-example-1 instance-manager Starting CloudNativePG Instance Manager {"build":{"Commit":"953f0b2cf","Date":"2024-10-09","Version":"1.24.0-dev128"},"version":"1.24.0"}
-2024-10-10T16:36:18.668268413Z INFO cluster-example-1 instance-manager Checking for free disk space for WALs before starting PostgreSQL
-2024-10-10T16:36:18.686309469Z INFO cluster-example-1 instance-manager starting tablespace manager
-2024-10-10T16:36:18.686375141Z INFO cluster-example-1 instance-manager starting external server manager
+2024-10-15T17:35:00.336 INFO     cluster-example-1 instance-manager Starting CloudNativePG Instance Manager
+2024-10-15T17:35:00.336 INFO     cluster-example-1 instance-manager Checking for free disk space for WALs before starting PostgreSQL
+2024-10-15T17:35:00.347 INFO     cluster-example-1 instance-manager starting tablespace manager
+2024-10-15T17:35:00.347 INFO     cluster-example-1 instance-manager starting external server manager
 [...]
 ```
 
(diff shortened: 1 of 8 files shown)


docs/src/kubectl-plugin.md
```diff
@@ -780,6 +781,39 @@ kubectl cnpg logs cluster cluster-example --output my-cluster.log
 Successfully written logs to "my-cluster.log"
 ```
 
+#### Pretty
+
+The `pretty` sub-command reads a log stream from standard input and formats it
+into a human-readable output, and attempts to sort the entries by timestamp.
+
+It is intended to be used in combination with `kubectl cnpg logs cluster`, as
+shown in the following example:
```
jsilvela, line 790: last week chatting with Gabriele, I had the impression we wanted to get rid of `Kubectl CNPG logs cluster` in favor of `stern`.
Not the plan anymore?
Or, to be done in a future PR?

docs/src/kubectl-plugin.md
```diff
@@ -780,6 +781,39 @@ kubectl cnpg logs cluster cluster-example --output my-cluster.log
...
+
+```
+$ kubectl cnpg logs cluster cluster-example | kubectl cnpg logs pretty
```
jsilvela, line 793: if we don't get rid of `Kubectl cnpg logs`, wouldn't `pretty` make better sense as an option?
`Kubectl cnpg logs --pretty cluster cluster-example`

leonardoce: In that case, you won't be able to use it with the logs you collected before on your own.

jsilvela: No. Responding here for traceability, rather than deleting.
Seeing all the options supported by `pretty`, it would be unwieldy.

fcanovai (approved):

docs/src/kubectl-plugin.md
leonardoce: Yes, #5780 is about that.
I agree with recommending `stern`, but we should fix what we have if possible, too.

fcanovai (approved):

internal/cmd/plugin/logs/pretty/pretty.go
```diff
@@ -0,0 +1,242 @@
...
+		"The list of pods to receive from. Defaults to all.")
+	cmd.Flags().Var(&bf.minLevel, "min-level",
+		`Hides the messages whose log level is less important than the specified one.
+Should be empty or one of error, warning, info, debug, or trace.`)
+
+	return cmd
+}
+
+// decode progressively decode the logs
+func (bf *prettyCmd) decode(ctx context.Context, reader io.Reader, recordChannel chan<- logRecord) {
```
jsilvela, line 99: Ah, lovely concurrent implementation.
Inspired by Rob Pike? https://go.dev/talks/2011/lex.slide#1

internal/cmd/plugin/logs/pretty/pretty.go
```diff
@@ -0,0 +1,242 @@
...
+		}
+
+		recordChannel <- *record
+	}
+
+	close(recordChannel)
+}
+
+// group transforms a stream of logs in a stream of log groups
+func (bf *prettyCmd) group(ctx context.Context, logChannel <-chan logRecord, groupChannel chan<- []logRecord) {
```
jsilvela, line 131: I'm guessing the main reason to have this is to be able to sort by timestamp within  each group of 200 logs.
Is it worth it? I mean, is it a bad assumption that the process producing the logs, be that `stern` or `Kubectl logs` or `Kubectl cnpg logs`, will return logs in chronological order?

internal/cmd/plugin/logs/pretty/pretty.go
```diff
@@ -0,0 +1,242 @@
...
+}
+
+// write write the logs on the output
+func (bf *prettyCmd) write(ctx context.Context, recordGroupChannel <-chan []logRecord, writer io.Writer) {
+	logRecordComparison := func(l1, l2 logRecord) int {
+		if l1.TS < l2.TS {
+			return -1
+		} else if l1.TS > l2.TS {
+			return 1
+		}
```
jsilvela, line 184: In line with my earlier comment about the grouping of logs into groups to make them sortable:
I have seen some processes producing timestamps in Unix epoch, and some processes producing timestamps in RFC3390 format.
If we had heterogeneous TS types here, this sorting would be unreliable.

internal/cmd/plugin/logs/pretty/pretty.go
jsilvela: The fixed 200 is not my worry here, but the complexity added.
I guess we could `cat` several files with logs, and then this would be a great feature.
But if the intent is to pipe after `stern` or `Kubectl logs`, I think this is overkill, no?

internal/cmd/plugin/logs/pretty/pretty.go
leonardoce: No, it came out of my head! So it deserves to be reviewed carefully :)

internal/cmd/plugin/logs/pretty/pretty.go
gbartolini: I think that, without sorting, logs are very difficult to interpret. The differentiation brought by CNPG is the concept of cluster, not single pods.

This is an attempt to go in that direction and it is a good enough implementation for now, enabling users to test it widely, before investing more in the feature.

My 2 cents here.

internal/cmd/plugin/logs/pretty/pretty.go
leonardoce: Which processes are you referring to? I think that with #5783 and https://github.com/cloudnative-pg/machinery/pull/38 we should have all the timestamps in the same format.

If this is not true, we should fix it.

jsilvela: Ah, OK, fixed there. That's it, `zap` by default uses RFC3390, so I'd see differences sometimes.

internal/cmd/plugin/logs/pretty/pretty.go
jsilvela: Makes sense. Thanks

leonardoce: Adding a command line flag for the group size is a good idea. I'm adding it.

leonardoce: Done!

internal/cmd/plugin/logs/pretty/pretty.go
```diff
@@ -0,0 +1,250 @@
...
+				wait.Done()
+			}()
+
+			wait.Wait()
+			return nil
+		},
+	}
+
+	cmd.Flags().IntVar(&groupSize, "group-size", 200,
+		"The maximum size of the log section to be sorted")
```
jsilvela, line 93: "group" will be very extraneous to the end user.
How about "sorting window size" or something like that?
```suggestion
	cmd.Flags().IntVar(&groupSize, "sorting-window-size", 200,
		"The maximum size of the window where logs are collected for sorting")
```

jsilvela: Probably better with "sorting-group-size"

leonardoce: done

jsilvela: @leonardoce I tested this picking the `sorting-window-size` as 0.
It actually works, but the effect is not to do grouping at all. Do we want to accept this?

jsilvela (approved):
Manual testing, looks great.
I've added a bunch of documentation that merits a check.
And, one pending question about the possibility of setting the sorting-group-size to 0. It works, so maybe it's OK.

internal/cmd/plugin/logs/pretty/pretty.go
```diff
@@ -0,0 +1,251 @@
...
+				bf.write(cmd.Context(), recordGroupsChannel, os.Stdout)
+				wait.Done()
+			}()
+
+			wait.Wait()
+			return nil
+		},
+	}
+
+	cmd.Flags().IntVar(&sortingGroupSize, "sorting-group-size", 200,
```
gbartolini, line 92:
```suggestion
	cmd.Flags().IntVar(&sortingGroupSize, "sorting-group-size", 1000,
```

docs/src/kubectl-plugin.md
```diff
@@ -780,6 +781,72 @@ kubectl cnpg logs cluster cluster-example --output my-cluster.log
...
+[...]
+```
+
+The `pretty` sub-command will try to sort the log stream,
+to make logs easier to reason about. In order to achieve this, it gathers the
+logs into groups, and within groups it sorts by timestamp. This is the only
+way to sort interactively, as `pretty` may be piped from a command in "follow"
+mode. The sub-command will add a group separator line, `---`, at the end of
+each sorted group. The size of the grouping can be configured via the
+`--sorting-group-size` flag, as illustrated in the following example:
```
gbartolini, line 834:
```suggestion
`--sorting-group-size` flag (default: 1000), as illustrated in the following example:
```

armru (approved):

PR #5790, opened 2024-10-10
docs: refactor logging page, mentioning `stern`

### LINKED ISSUE #5780: [Docs]: Recommend `stern` for log tailing

[CloudNativePG containers log by default in JSON format to stdout](https://cloudnative-pg.io/documentation/current/logging/), expecting downstream integration with log management tools. However, working with DBAs who approach to Kubernetes, one of the things they miss is to have access to logs, even if just limited to an educational environment.

We tried to provide the `cnpg logs cluster` command which however will require complete rewriting to work with multiple containers inside a pod (as per the new CNPG-I initiative).

Given that our mission is to integrate with tools that exist in Kubernetes and we have already adopted it for our E2E infrastructure, I suggest we refine the current docs and add a section in which we recommend using [stern](https://github.com/stern/stern).

https://cloudnative-pg.io/documentation/current/logging/
https://cloudnative-pg.io/documentation/current/troubleshooting/#logs
https://cloudnative-pg.io/documentation/current/kubectl-plugin/#logs

This section should become deprecated IMHO unless we decide to fix it and enhance it: https://cloudnative-pg.io/documentation/current/kubectl-plugin/#logs

### Additional context

We should make sure that we don't stick with a single tool and we highlight the existence of other tools.

### Backport?

Yes

### Code of Conduct

gbartolini:
Closes #5780


docs/src/logging.md
```diff
@@ -1,43 +1,50 @@
 # Logging
 
-The operator is designed to log in JSON format directly to standard output,
-including PostgreSQL logs.
+The operator outputs logs in JSON format directly to standard output, including
+PostgreSQL logs, without persisting them to storage for security reasons. This
```
jsilvela, line 4: I think the more important bit here is this is standard in K8s, and more broadly it's a 12-factor app recommendation.

docs/src/logging.md
```diff
@@ -1,43 +1,50 @@
...
+    the `level` and `ts` fields using the `log-field-level` and
+    `log-field-timestamp` flags in the operator controller. This can be configured
+    by editing the `Deployment` definition of the `cloudnative-pg` operator.
 
-## Operator log
+## Operator Logs
 
-You can specify a log level in the cluster spec with the option `logLevel`.
-You can set it to `error`, `warning`, `info`(default), `debug`, or `trace`.
+You can configure the log level for the operator in the cluster specification
```
jsilvela, line 35: wait are we not mixing cluster log level with operator log level?

docs/src/logging.md
```diff
@@ -1,43 +1,50 @@
...
+using the `logLevel` option. Available log levels are: `error`, `warning`,
+`info` (default), `debug`, and `trace`.
 
-Currently, you can set the log level only when an instance starts. You can't
-change it at runtime. If you change the value in the cluster spec after the cluster
-was started, it takes effect only in the new pods and not the old ones.
+!!! Important
+   Currently, the log level can only be set at the time the instance starts.
```
jsilvela, line 40: Looking at the change, and the original is already confusing.
We're talking about clusters and instances in the Operator Logs section.
Going to propose a change here...

jsilvela (commented):
Re-reading this, I see that in the "Operator Logs" section we've always been mixing operator and instance logs.
Perhaps this is on purpose to be less pedantic?
I do think it's important to distinguish between the instance logs configurable in the cluster YAML, and the operator logs configurable through the ConfigMap for the deployment.
If you agree, @gbartolini , I'd split this.

docs/src/logging.md
jsilvela: yes, that's exactly what we say in the Troubleshooting page

docs/src/logging.md
```diff
@@ -1,43 +1,50 @@
...
-    If your log ingestion system requires it, you can rename the `level` and `ts` field names using the `log-field-level` and
-    `log-field-timestamp` flags of the operator controller. Edit the `Deployment` definition of the
-    `cloudnative-pg` operator.
+    If your log ingestion system requires custom field names, you can rename
+    the `level` and `ts` fields using the `log-field-level` and
+    `log-field-timestamp` flags in the operator controller. This can be configured
+    by editing the `Deployment` definition of the `cloudnative-pg` operator.
 
-## Operator log
+## Operator Logs
```
jsilvela, line 33: Reading the code, and indeed, the Operator logLevel is not easily configurable. It is not supported in the ConfigMap, and the only real way to configure it is to `Kubectl edit` the deployment and specify the level in the `--log_level` invocation.
In the CI/CD we do configure the level by directly messing with the `--log-level` flag.

docs/src/logging.md
```diff
@@ -1,43 +1,50 @@
...
+    operator and should be handled at the Kubernetes infrastructure level.
+    For more information, see the
+    [Kubernetes Logging Architecture](https://kubernetes.io/docs/concepts/cluster-administration/logging/)
+    documentation.
 
-- `level` – Log level (`info`, `notice`, ...).
+Each log entry includes the following fields:
+
+- `level` – The log level (e.g., `info`, `notice`).
 - `ts` – The timestamp (epoch with microseconds).
```
jsilvela, line 19: I've been trying to get a definitive answer on the timestamp format. I cannot find one. Different loggers use unix epoch or RFC3390, for example `zap` uses the latter. I'd probably remove the parenthesized part here.

jsilvela (approved):
Optional comment on mentioning security as the rationale for logging to stdout.
Also, there was a rebase conflict I fixed, adding `instance manager` in the  "Other logs" section. Worth a check.

docs/src/logging.md
jsilvela: Huh, so the timestamp has recently been fixed to unix epoch here.
https://github.com/cloudnative-pg/machinery/pull/38/files
I will put this back in.

jsilvela: Ah, no! We're fixing to nanoseconds!
You know what, I think we don't need to specify the format in this doc.

leonardoce (approved):

docs/src/logging.md
gbartolini: CloudNativePG is fine.

gbartolini:
> Re-reading this, I see that in the "Operator Logs" section we've always been mixing operator and instance logs. Perhaps this is on purpose to be less pedantic? I do think it's important to distinguish between the instance logs configurable in the cluster YAML, and the operator logs configurable through the ConfigMap for the deployment. If you agree, @gbartolini , I'd split this.

If you think it is worth it @jsilvela please open a new issue for 1.24.2. Thanks!

jsilvela:
@gbartolini that comment was from yesterday. I already added the split between cluster logs and operator logs ... just so you know, seems to have been merged.

Issue #5366, opened 2024-08-23
[Bug]: wal-archive instance manager subcommand should not connect to the kubernetes-api server


mnencia:
### Is there an existing issue already for this bug?

- [X] I have searched for an existing issue, and could not find anything. I believe this is a new bug.

### I have read the troubleshooting guide

- [X] I have read the troubleshooting guide and I think this is a new bug.

### I am running a supported version of CloudNativePG

- [X] I have read the troubleshooting guide and I think this is a new bug.

### Version

trunk (main)

### What version of Kubernetes are you using?

1.31

### What is your Kubernetes environment?

Self-managed: kind (evaluation)

### How did you install the operator?

YAML manifest

### What happened?

The `wal-archive` instance manager subcommand creates a ControllerRuntimeClient. This is a costly operation, and we should avoid it.

https://github.com/cloudnative-pg/cloudnative-pg/blob/24437b6879d8ef015f97bd856052c30eb5931b80/internal/cmd/manager/walarchive/cmd.go#L81-L85

The client is used later to patch a condition, using the Cluster from the cache as the source state. This should be moved inside the main instance manager code to avoid the cost of opening a client and potential consistency issues.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

gbartolini:
This was introduced before the open source release in commit:dcd0d0ba89a5a6cfc4838a0761730be39c9d78bd

PR #5503, opened 2024-09-12
fix: define a `TMPDIR` environment

brouberol:
Tools like `barman` crash when `tempfile.mkdtemp` cannot find any writeable directory amongst the list of "usual suspects" (`/tmp`, `/var/tmp`, `/usr/tmp`, etc), as I've described in #5420 

```
File "/usr/lib/python3.11/tempfile.py", line 454, in gettempdir
    return _os.fsdecode(_gettempdir())
                        ^^^^^^^^^^^^^
  File "/usr/lib/python3.11/tempfile.py", line 447, in _gettempdir
    tempdir = _get_default_tempdir()
              ^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/tempfile.py", line 362, in _get_default_tempdir
    raise FileNotFoundError(_errno.ENOENT,
FileNotFoundError: [Errno 2] No usable temporary directory found in ['/tmp', '/var/tmp', '/usr/tmp', '/srv/app']
```

As the operator creates the PG pods with `ReadOnlyRootFilesystem` set to `true`, setting this environment variable will allow barman to write its temporary files under `/controller` (we could argue it could be `/run`, as the same volume is mounted to both mountpoints). This is due to CPyhon considering on this [environment variable](https://github.com/python/cpython/blob/main/Lib/tempfile.py#L163-L165) to add potential temporary directories to try when running `mkdtemp` or `mktemp`.

```python
postgres@postgresql-test-2:/srv/app$ python3
Python 3.11.2 (main, May  2 2024, 11:59:08) [GCC 12.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import os
>>> os.getenv('TMPDIR')
>>> import tempfile
>>> tempfile.mkdtemp()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.11/tempfile.py", line 512, in mkdtemp
    prefix, suffix, dir, output_type = _sanitize_params(prefix, suffix, dir)
                                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/tempfile.py", line 265, in _sanitize_params
    dir = gettempdir()
          ^^^^^^^^^^^^
  File "/usr/lib/python3.11/tempfile.py", line 454, in gettempdir
    return _os.fsdecode(_gettempdir())
                        ^^^^^^^^^^^^^
[...]

PR #5833, opened 2024-10-15
chore: avoid uselessly patching the ContinuousArchivingSuccess condition

leonardoce:
The `archive_command` callback is keeping updated the `ContinuousArchivingSuccess` condition using the result of `barman-cloud-wal-archive` or the corresponding CNPG-i plugin.

This change makes the instance manager deterministically hit the API server when the condition changes instead of using `reflect.DeepEqual`, avoiding false positives.

See also: #5366

### DIFF
--- a/pkg/conditions/conditions.go
+++ b/pkg/conditions/conditions.go
@@ -18,7 +18,6 @@ package conditions
 
 import (
 	"context"
-	"reflect"
 
 	"k8s.io/apimachinery/pkg/api/meta"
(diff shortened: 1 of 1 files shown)


fcanovai (approved):

fcanovai:
E2e: https://github.com/EnterpriseDB/cloudnative-pg/actions/runs/11346340514

mnencia (approved):

fcanovai:
/ok-to-merge Test are failing for upgrades due to https://github.com/cloudnative-pg/cloudnative-pg/pull/5503, not for this PR

PR #5878, opened 2024-10-17
feat: make data durability configurable for sync replication

### LINKED ISSUE #5793: [Feature]: flexible data durability enforcement in synchronous replication

With the introduction of the new synchronous replication interface in CloudNativePG version 1.24 (see #5095), the system now prioritizes **data durability**, as typically expected by PostgreSQL DBAs. However, this approach contrasts with the previously established (now deprecated) method of managing synchronous replication using `minSyncReplicas` and `maxSyncReplicas`, which places an emphasis on **self-healing** and system availability over strict data durability.

This creates a potential conflict for users who rely on the older method's flexibility and ability to maintain availability in the face of replica failure, especially in cloud-native environments where resilience and self-healing are often prioritized. The current behavior in version 1.24 may lead to scenarios where availability is sacrificed in favor of ensuring synchronous data replication, which could be unexpected for users accustomed to the previous balance of these two factors.

To provide more flexibility in managing data durability, we propose adding a method to relax the strict enforcement of data durability in the new synchronous replication interface.

In particular, the proposal is to introduce a new option in the `.spec.postgresql.synchronous` section, called `dataDurability`, with two configurable values:

1. `required` (default):  
   - Synchronous replication will strictly enforce data durability.
   - If there are no healthy replicas, write operations with synchronous commit set to `on`, `remote_write`, or `remote_apply` will hang until a replica becomes available.
   - This option ensures data consistency at all times, as expected by traditional DBAs, and it is the default

2. `preferred`:  
   - Data durability will be enforced whenever healthy replicas are available.
   - If not enough replicas are available, the system will disable synchronous replication, allowing write operations to continue. This prioritizes system availability and self-healing over strict durability.
   - This method is designed for cloud-native environments where availability and resilience are key concerns, and restores the default pre-1.24 behavior.

The `preferred` method can only be applied if both `preStandbyNames` and `postStandbyNames` are empty, ensuring that it does not interfere with replica preferences that may be set.

N/A

### Additional context

This feature request was raised by Damien Glisse in the [Developer's meeting (Oct 2, 2024 - 48)](https://docs.google.com/document/d/1Bmf2AZG5WLKAyESJbYk7MbsfiuD3jgdIDQrDkNuKT9w/edit#heading=h.wevss7y071v6).

### Backport?

No

Yes

### Code of Conduct

leonardoce:
This patch introduces an option to relax strict data durability requirements within the new synchronous replication interface. By setting `.spec.postgresql.synchronous.dataDurability` to `preferred`, the operator limits the required number of synchronous standby replicas to match the number of available instances, prioritizing self-healing over strict data durability.

This configuration requires both `standbyNamesPre` and `standbyNamesPost` to be empty.

Closes: #5793

### DIFF
--- a/api/v1/cluster_types.go
+++ b/api/v1/cluster_types.go
@@ -1179,11 +1179,11 @@ type DataDurabilityMethod string
 
 const (
 	// DataDurabilityMethodRequired means that data durability is strictly enforced
-	DataDurabilityMethodRequired DataDurabilityMethod = DataDurabilityMethod("required")
+	DataDurabilityMethodRequired DataDurabilityMethod = "required"
 
 	// DataDurabilityMethodPreferred means that data durability is enforced
 	// only when healthy replicas are available
-	DataDurabilityMethodPreferred DataDurabilityMethod = DataDurabilityMethod("preferred")
+	DataDurabilityMethodPreferred DataDurabilityMethod = "preferred"
 )
 
 // SynchronousReplicaConfiguration contains the configuration of the
--- a/api/v1/cluster_types.go
+++ b/api/v1/cluster_types.go
@@ -1229,6 +1229,7 @@ type SynchronousReplicaConfiguration struct {
 	// This can only be set if both `standbyNamesPre` and `standbyNamesPost` are empty.
 	// +kubebuilder:validation:Enum=required;preferred
 	// +kubebuilder:default:=required
+	// +optional
 	DataDurability DataDurabilityMethod `json:"dataDurability,omitempty"`
 }
 
--- a/api/v1/cluster_types.go
+++ b/api/v1/cluster_types.go
@@ -1173,17 +1173,17 @@ const (
 	SynchronousReplicaConfigurationMethodAny = SynchronousReplicaConfigurationMethod("any")
 )
 
-// DataDurabilityMethod specifies how strictly to enforce synchronous replication
+// DataDurabilityLevel specifies how strictly to enforce synchronous replication
 // when cluster instances are unavailable. Options are `required` or `preferred`.
-type DataDurabilityMethod string
+type DataDurabilityLevel string
 
 const (
-	// DataDurabilityMethodRequired means that data durability is strictly enforced
-	DataDurabilityMethodRequired DataDurabilityMethod = "required"
+	// DataDurabilityLevelRequired means that data durability is strictly enforced
+	DataDurabilityLevelRequired DataDurabilityLevel = "required"
 
-	// DataDurabilityMethodPreferred means that data durability is enforced
+	// DataDurabilityLevelPreferred means that data durability is enforced
 	// only when healthy replicas are available
-	DataDurabilityMethodPreferred DataDurabilityMethod = "preferred"
+	DataDurabilityLevelPreferred DataDurabilityLevel = "preferred"
 )
 
 // SynchronousReplicaConfiguration contains the configuration of the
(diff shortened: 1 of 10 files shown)


leonardoce:
Can you help me with documentation & the E2e tests?

fcanovai:
e2e: https://github.com/EnterpriseDB/cloudnative-pg/actions/runs/11388357993

api/v1/cluster_types.go
```diff
@@ -1162,10 +1162,24 @@ const (
 	SynchronousReplicaConfigurationMethodAny = SynchronousReplicaConfigurationMethod("any")
 )
 
+// DataDurabilityMethod specifies how strictly to enforce synchronous replication
+// when cluster instances are unavailable. Options are `required` or `preferred`.
+type DataDurabilityMethod string
```
jsilvela, line 1167: "method" seems to say, this is the way data durability is achieved, IMO not exactly what we mean.
IMO, perhaps "strictness" or "level" would be clearer
```suggestion
type DataDurabilityLevel string
```

api/v1/cluster_types.go
```diff
@@ -1162,10 +1162,24 @@ const (
...
+
+const (
+	// DataDurabilityMethodRequired means that data durability is strictly enforced
+	DataDurabilityMethodRequired DataDurabilityMethod = DataDurabilityMethod("required")
```
jsilvela, line 1171: I don't think we need to cast the constants ... this line looks like Java (ha)
```suggestion
	DataDurabilityMethodRequired DataDurabilityMethod = "required"
```

api/v1/cluster_types.go
leonardoce: There's more info in the issue https://github.com/cloudnative-pg/cloudnative-pg/issues/5793

jsilvela: Yes, I read that. But, the "method" we use to get data durability is synchronous replication, no? Here we're just choosing how strict to be in the face of available replicas.

jsilvela: I.e. I'd agree with "method" if the alternatives were "strict-synchronous" and "loose-synchronous"

fcanovai: Makes sense, I've changed it

fcanovai:
new replication e2e run: https://github.com/EnterpriseDB/cloudnative-pg/actions/runs/11406184671

fcanovai (approved):

jsilvela:
I was looking at the documentation. Both for this PR and for the existing documentation on synchronous replication, I don't think we say anything about configuring the number of instances for quorum that is too high.
For instance, what if we do
```
spec:
  instances: 3
  postgresql:
    synchronous:
      method: any
      number: 5
      dataDurability: required
```
From inspection, it's obvious that this cluster will basically be useless.
Shouldn't we have some measure of defense?
And, in addition, shouldn't we have a Prometheus alert?
- in the case of dataDurability `preferred`, to alert how often the sync replication is being disabled
- in the case of `required`, to alert how often the number of available instances is too low for quorum

gbartolini:
> ```
> spec:
>   instances: 3
>   postgresql:
>     synchronous:
>       method: any
>       number: 5
>       dataDurability: required
> ```

So basically the webhook should check `number` to be < (instances + count(preSyncStandbyNames) + count(postSyncStandbyNames))?

gbartolini (approved):

armru (approved):

mnencia (approved):

PR #6056, opened 2024-11-11
feat: managed role should only grant role for missing in inRoles

### LINKED ISSUE #6024: [Feature]: Skip revoke role from user if not declared in inRoles of managed role

 In manage role,the user role is followed by the rule that add by declare and revoke when non-declare.  

But, since pg 16, this feature change the behavor of non-super user to create role , see this feature 

Allow roles that create other roles to automatically inherit the new role's rights or the ability to  SET ROLE to the new role

The behavor now when a non-super user `admin` going to create a role Bob, Bob will be added as the parent role of `admin`. 

create role Bob  is equals to `create role Bob role `admin` 

If the `admin` is non-super user and declared in managed role , Bob will be grant to `admin` and later will be revoked from admin. 

As revoke role has more restriction, and usually a role is added will be used, how about

when a role is defined in inRoles or managed role, we grant the role for the user. 

ignore the role which is missing in managed role but exists in database, NO revoke. 

n/a

### Additional context

_No response_

### Backport?

Yes

No

### Code of Conduct

litaocdl:
This match ensure that managed role only grant missing roles in database
for the user. If roles are missing in InRoles, managed role will not revoke the roles
from the declared user. This is for the changes in postgres 16
about [role enhancement](https://github.com/postgres/postgres/commit/e5b8a4c098ad6add39626a14475148872cd687e0)
When a non-super user Alice create a new role bob, Alice will be added as the member of
Bob.

Closes: #6024

### DIFF
--- a/internal/management/controller/roles/contract.go
+++ b/internal/management/controller/roles/contract.go
@@ -20,10 +20,10 @@ import (
 	"database/sql"
 	"reflect"
 
+	"github.com/cloudnative-pg/machinery/pkg/stringset"
 	"github.com/jackc/pgx/v5/pgtype"
 
 	apiv1 "github.com/cloudnative-pg/cloudnative-pg/api/v1"
-	"github.com/cloudnative-pg/machinery/pkg/stringset"
 )
 
 // DatabaseRole represents the role information read from / written to the Database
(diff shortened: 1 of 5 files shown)


litaocdl:
e2e rune: https://github.com/EnterpriseDB/cloudnative-pg/actions/runs/11774679765

PR #6066, opened 2024-11-12
perf(wal-archive): remove typed client from cmd execution

### LINKED ISSUE #5366: [Bug]: wal-archive instance manager subcommand should not connect to the kubernetes-api server

armru:
Closes #5366 

# Release notes

Performance Improvement. The `wal-archive` command in the instance manager now uses the local webserver `controller-runtime` client instead of creating a new one.

# Note for reviewers

- Given that it is a performance improvement, I'm not adding the backport tag


mnencia (approved):

gabriele-wolfox (approved):

gabriele-wolfox:
LGTM

leonardoce (approved):

Issue #5969, opened 2024-10-29
[Bug]: Former primary doesn't flush the WAL archive before being resynced


gbartolini:
### Is there an existing issue already for this bug?

- [x] I have searched for an existing issue, and could not find anything. I believe this is a new bug.

### I have read the troubleshooting guide

- [x] I have read the troubleshooting guide and I think this is a new bug.

### I am running a supported version of CloudNativePG

- [x] I have read the troubleshooting guide and I think this is a new bug.

### Version

1.24.0

### What version of Kubernetes are you using?

1.31

### What is your Kubernetes environment?

Self-managed: kind (evaluation)

### How did you install the operator?

YAML manifest

### What happened?

An old primary that comes back online after a failover should ensure that the WAL archiver duty has been fulfilled before running pg_rewind. While this is accentuated in the presence of a lagging archiver, it should be systematically ensured in order to avoid gaps in the PITR continuum.

This is potentially a severe bug IMHO as far as PITR is concerned.

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

mnencia:
Closed in https://github.com/cloudnative-pg/cloudnative-pg/pull/6141

PR #6141, opened 2024-11-20
fix: ensure former primary WALs are flushed before resyncing

armru:
This patch fixes an issue where the former primary node may fail to archive its WAL files before resynchronizing as a replica after a switchover event. This gap in the WAL stream could prevent point-in-time recovery (PITR) to positions between the current primary node’s promotion and the next backup. The fix ensures the former primary flushes and archives all pending WAL files before rejoining the cluster, maintaining a complete and consistent WAL stream.

Closes #5959

## Release Notes

Ensure former primary nodes flush and archive all WAL files before resynchronizing as replicas to prevent gaps in the WAL stream and enable reliable point-in-time recovery.


gbartolini:
Is this related to #5969?

leonardoce:
E2e: https://github.com/EnterpriseDB/cloudnative-pg/actions/runs/11969473222

leonardoce (approved):

mnencia (approved):

Issue #6043, opened 2024-11-06
[Feature]: allow cross-namespace Database and Role configuration


lennartack:
### Is there an existing issue already for this feature request/idea?

- [x] I have searched for an existing issue, and could not find anything. I believe this is a new feature request to be evaluated.

### What problem is this feature going to solve? Why should it be added?

The new CRD for Database and the upcoming Role CRD use `LocalObjectReference` for the cluster setting. This makes it awkward to share a Postgres cluster among applications in different namespaces.

### Describe the solution you'd like

It would be nice if a Cluster in another namespace can be referenced in Database and Role objects. This should be disabled by default for security reasons, but could be enabled using an option on the Cluster resource (separately for Database and Role resources).

### Describe alternatives you've considered

Deploy all resources in the same namespaces. This is awkward for us since we have a large number of applications in separate namespace. They should be kept separate and use separate deployment pipelines.

### Backport?

No

### Are you willing to actively contribute to this feature?

No

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

ghormoon:
I'd also like to see this, especially in usecase where i'll have hundreds few MB big databases for dynamically created instances of application.
Having one "central" cluster would be way easier in terms of backups management etc.

only downside i can think of is that currently there's an issue to restore a single DB from backup. it would be very helpful to be able to restore it per DB instead of per cluster.

Inrixia:
+1 Currently having to do some jank to work around this.

Being able to have a single cluster and then use crd's in seperate namespaces to deploy databases would be ideal. With role creation also being namespace independant re https://github.com/cloudnative-pg/cloudnative-pg/issues/6498 so that access can be isolated and the entire db/role lifecycle can be contained within each namespaces individual deployment.

olivergregorius:
I would like this feature as well!

Inrixia:
Any update?

Inrixia:
Afik this is still intended for a upcoming release?

aries1980:
> Afik this is still intended for a upcoming release?

Christmas is coming early this year?

mrharpo:
This feature would be very helpful for us. I would like to allow some or all namespaces to create a `Database` yaml, which would provision a database in the appropriate cluster.

To achieve this currently, I would have to deploy a cluster per namespace, or use something like kubernetes-reflector to manage objects between namespaces.

JCL38-ORANGE:
Hello,
[...]

PR #6155, opened 2024-11-22
feat: introduce `DatabaseRole` CRD for declarative role management

### LINKED ISSUE #5341: Declarative role management (`Role` CRD)
With the implementation of database management via the `Database` CRD, it’s important to offer an alternative approach to role management, currently handled through the `.spec.managed.roles` stanza. We propose introducing a new CRD called `Role`, which will eventually replace the existing implementation. For now, we will maintain both interfaces, ensuring that they are mutually exclusive if coexistence is not feasible.

### LINKED ISSUE #5940: [Feat]: add CRD and basic controller

### LINKED ISSUE #5967: [Docs]: update the managed-roles documentation with the CRD based roles

There is already documentation for managed roles.
We should explain the differences between the existing cluster-based managed roles and the new CRD based roles and reconciler.

_No response_

docs/src/declarative_role_management.md

_No response_

### Additional context

_No response_

### Backport?

No

### Code of Conduct

jsilvela:
Introduce a `DatabaseRole` custom resource to manage a PostgreSQL role as a standalone Kubernetes object, instead of declaring it inline in the Cluster's `.spec.managed.roles` stanza. Each role gets its own lifecycle, status, and RBAC, which suits GitOps workflows and lets role definitions live next to the applications that own them. The spec reuses the same `RoleConfiguration` structure as the inline method, so migrating a role is a matter of moving the stanza into its own manifest.

Reconciliation follows the split already used by the `Database`, `Publication`, and `Subscription` resources: the instance manager applies the role to PostgreSQL and reports the result in the resource status, while the operator watches the password secrets and relays their changes through a status condition, so the instance manager never needs to watch secrets across the namespace.

The `databaseRoleReclaimPolicy` field controls what happens to the role when the resource is deleted, `retain` (default) or `delete`, mirroring persistent volumes. A role is dropped only by the object that actually reconciled it, and `ensure: absent` is rejected in favor of deleting the resource. When the same role name is managed both inline and through a `DatabaseRole`, the Cluster spec takes precedence and the resource reports the conflict instead of fighting over it.

The inline `managed.roles` stanza remains fully supported; the `DatabaseRole` resource is the recommended method going forward.

Closes #5940
Closes #5341
Closes #5967

Assisted-by: Claude Opus 4.8


internal/management/controller/roles/contract.go
```diff
@@ -120,3 +123,37 @@ func (d *DatabaseRole) isEquivalentTo(inSpec apiv1.RoleConfiguration) bool {
 
 	return reflect.DeepEqual(role, spec) && d.hasSameValidUntilAs(inSpec)
 }
+
+// ApplyPassword updates a database role with the password located in the Secret
+// it returns the resource version of the Secret
```
jsilvela, line 128:
```suggestion
// ApplyPassword updates a database role with the password located in the Secret,
// and it returns the resource version of the Secret
```

internal/management/controller/roles/contract.go
```diff
@@ -120,3 +123,37 @@ func (d *DatabaseRole) isEquivalentTo(inSpec apiv1.RoleConfiguration) bool {
...
+func (d *DatabaseRole) ApplyPassword(
+	ctx context.Context,
+	cl client.Client,
+	rolePassword passwordManager,
+	namespace string,
+) (string, error) {
+	var passVersion string
+	switch {
```
jsilvela, line 136:
```suggestion
	switch {
```

internal/management/controller/roles/contract.go
```diff
@@ -120,3 +123,37 @@ func (d *DatabaseRole) isEquivalentTo(inSpec apiv1.RoleConfiguration) bool {
...
+				rolePassword.GetRoleSecretName())
+	default: // role.PasswordSecret != nil && !rolePassword.ShouldDisablePassword():
+		passwordSecret, err := getPassword(ctx, cl, rolePassword, namespace)
+		if err != nil {
+			return "", err
+		}
+
+		d.Password = sql.NullString{Valid: true, String: passwordSecret.password}
+		passVersion = passwordSecret.version
+		return passVersion, nil
```
jsilvela, line 157:
```suggestion
		return passwordSecret.version, nil
```

internal/management/controller/roles/contract.go
```diff
@@ -120,3 +123,37 @@ func (d *DatabaseRole) isEquivalentTo(inSpec apiv1.RoleConfiguration) bool {
...
+	case rolePassword.GetRoleSecretName() == "" && rolePassword.ShouldDisablePassword():
+		d.Password = sql.NullString{}
+		return "", nil
+	case rolePassword.GetRoleSecretName() != "" && rolePassword.ShouldDisablePassword():
+		// this case should be prevented by the validation webhook,
+		// and is an error
+		return "",
+			fmt.Errorf("cannot reconcile: password both provided and disabled: %s",
+				rolePassword.GetRoleSecretName())
+	default: // role.PasswordSecret != nil && !rolePassword.ShouldDisablePassword():
```
jsilvela, line 149:
```suggestion
	default:
```

jsilvela (commented):
suggestions from the other PR https://github.com/cloudnative-pg/cloudnative-pg/pull/5973

jsilvela:
Thanks @sxd for reopening. Whew!

jsilvela:
@sxd ??? why has this been deleted again?

.wordlist-en-custom.txt
```diff
@@ -285,6 +285,7 @@ PGAudit
 PGDATA
 PGDG
 PGData
+PGRole
```
gbartolini: I would prefer to refer to it simply as `Role` for consistency with the current naming conventions. Otherwise, we should apply `PG` to other resources too (why are we using `Cluster` instead of `PGCluster`?).

While I understand that `role` already corresponds to `roles.rbac.authorization.k8s.io`, meaning we will need to specify the full name of the custom resource (CR), I still believe this approach is preferable.

jsilvela: I don't disagree. I'll reply on the top issue to keep it centralized.

gbartolini:
In here we decided to call the CRD `DatabaseRole` instead of `Role`: https://github.com/cloudnative-pg/cloudnative-pg/issues/5341#issuecomment-3766909464

I am sorry about that @leonardoce, is it possible to change it now or do you prefer to keep `Role`?

gbartolini:
I have been playing with the feature and I love it. I have updated the docs (at the moment with the `Role` CRD name, which @leonardoce will replace with `DatabaseRole` once he's around). There is only one case that doesn't resonate with me.

If I create this role:

```yaml
kind: Role
metadata:
  name: role-app
spec:
  cluster:
    name: cluster-example
  name: app
  createdb: true
  login: true
  roleReclaimPolicy: delete
```

Then, if I try to delete the role, the operation hangs (the finalizer is not able to successfully complete), the `status` of the role doesn't mention anything (and the role is not `Terminating`), the Postgres logs show:

```console
2026-04-04T19:43:30.792 ERROR    cluster-example-1 postgres         role "app" cannot be dropped because some objects depend on it
2026-04-04T19:43:30.792 ERROR    cluster-example-1 instance-manager Reconciler error
2026-04-04T19:48:58.479 ERROR    cluster-example-1 postgres         role "app" cannot be dropped because some objects depend on it
2026-04-04T19:48:58.479 ERROR    cluster-example-1 instance-manager Reconciler error
```

Ideally we should be able to see that a role is pending deletion and its status reflects the error (what we see in the Postgres logs). Is it possible to implement it like this @leonardoce? Thanks.

internal/management/controller/databaserole_controller.go
```diff
@@ -0,0 +1,362 @@
...
+	}
+
+	dbRole := roleAdapter{
+		RoleConfiguration: role.Spec.RoleConfiguration,
+	}.toDatabaseRole()
+	passwordVersion, err := dbRole.ApplyPassword(ctx, r.Client, &role.Spec, r.instance.GetNamespaceName())
+	if err != nil {
+		return "", fmt.Errorf("while getting the role password: %w", err)
+	}
+	for _, r := range rolesInDB {
```
GellertBabel: I think we are missing an if like
```
if role.Spec.Ensure == apiv1.EnsureAbsent {
				return "", roles.Delete(ctx, db, dbRole)
			}
```

internal/management/controller/databaserole_controller.go
```diff
@@ -0,0 +1,362 @@
...
+				return "", fmt.Errorf("while getting the membership updates required: %w", err)
+			}
+			err = roles.UpdateMembership(ctx, db, dbRole, toGrant, toRevoke)
+			if err != nil {
+				return "", fmt.Errorf("while getting the membership updates required: %w", err)
+			}
+			return passwordVersion, roles.Update(ctx, db, dbRole)
+		}
+	}
+
```
GellertBabel, line 445: Also when absent is true we should return nil?

```
if role.Spec.Ensure == apiv1.EnsureAbsent {
		return "", nil
	}
```

GellertBabel (dismissed):
Is ensure absent left out intentionally?

config/crd/bases/postgresql.cnpg.io_databaseroles.yaml
```diff
@@ -0,0 +1,291 @@
...
+                - present
+                - absent
+                type: string
+              inRoles:
+                description: |-
+                  List of one or more existing roles to which this role will be
+                  immediately added as a new member. Default empty.
+                items:
+                  type: string
+                type: array
```
jonasbadstuebner, line 118: I still hope for the feature I proposed in #10007 to become a reality.
And would open a PR when the introduction of the `DatabaseRole` CRD is done.

But I want to prevent introducing a deprecated field to the CRD shortly after its release.

To allow for more flexibility with the grants, I suggest the following change to the CRD:
```suggestion
              roleGrants:
                description: |-
                  List of one or more existing roles to which this role will be
                  added as a new member. Default empty.
                items:
                  properties:
                    name:
                      description: The name of the role this role will be granted.
                      type: string
                    required:
                    - name
                  type: object
                type: array
```
This would allow me to reuse the `roleGrants` field and introduce `inherit`, `admin` and `set` as new properties to it, instead of creating a separate field where another migration would be necessary.  
This would make my PR less invasive and prevent a new code path from being created.

For users that don't want or need the feature from #10007 the change is trivial. They have to migrate to the `DatabaseRole` CRD anyway, so making small changes to the list of desired roles should not be a problem.
And they would not be impacted by my changes later, if the CRD structure already supports it.

I renamed the `inRoles` field to `roleGrants`, because I think this is more descriptive already, especially if existing roles are reconciled. Because if you know the PostgreSQL syntax for creating a role, `inRoles` and the description of the field imply that it only applies on creating the Role. But there is also reconciliation logic for this field that does not make use of `IN ROLES` at all, but uses `GRANT`.

Drawbacks:
1. The code has to be adopted to pick the `name` from the list instead of using it as-is.  
This change is the most invasive one for your PR.
2. The migration of `inRoles` isn't a plain copy-paste (which I think is fine)

jmealo:
Let me know if you need help testing this. Looking forward to this! 🥳  (If we add #6043 self-service databases for app teams can be a thing out of box with some light VAP!).

internal/management/controller/databaserole_controller.go
mnencia: Since this comment was written, the handling changed direction: `ensure: absent` is now rejected for `DatabaseRole`, both by a CEL rule on the CRD and by a runtime guard in the reconciler, with a message pointing the user to `roleReclaimPolicy: delete`. For a standalone resource, the reclaim policy is the natural way to express "drop the role when the object is removed", mirroring the PV/PVC behavior. The `ensure` field is still present in the spec only because the structure is shared with the inline `managed.roles` stanza.

internal/management/controller/databaserole_controller.go
mnencia: Same as the thread above: `ensure: absent` is rejected before reaching this code, so neither branch is needed here.

config/crd/bases/postgresql.cnpg.io_databaseroles.yaml
mnencia: The `DatabaseRole` spec deliberately inlines the same `RoleConfiguration` structure used by `.spec.managed.roles`: migrating a role is copying the stanza into a resource, and both reconcilers share the same adapter code. Changing `inRoles` into a different shape only here would fork the two APIs and create the parallel code path we both want to avoid. Also, inline roles remain supported, so there is no forced migration to lean on.

This does not block #10007. A `roleGrants` list with `admin`, `inherit` and `set` per membership can be added later as a new optional field next to `inRoles`, mutually exclusive via a validation rule, and it could land for inline and standalone roles at the same time. Deprecating `inRoles` afterwards remains an option rather than a requirement.

You have a point that the field is not only used at creation time: membership changes on an existing role are applied with `GRANT` and `REVOKE`. We amended the field description accordingly.

mnencia (approved):

config/crd/bases/postgresql.cnpg.io_databaseroles.yaml
jonasbadstuebner: Wouldn’t implementing the feature I suggested also benefit from reusing the adapter code anyway?
If inline roles remain supported, having two code paths is unavoidable.
Does your code for the `DatabaseRole` CRD handle resolving the order in which roles have to be created so you don’t run into a race condition with `IN ROLES`?

armru (approved):

leonardoce (approved):

gbartolini (approved):

PR #6157, opened 2024-11-22
chore: drop support for `pg_rewind` in PostgreSQL 12

### LINKED ISSUE #6156: [Chore]: Remove the logic to perform a crash recovery before running `pg_rewind`

Starting with PostgreSQL 13, it now automatically handles crash recovery before the rewind process (you can check out the [related commit](https://git.postgresql.org/gitweb/?p=postgresql.git;a=commitdiff;h=5adafaf17) for details). Since PostgreSQL 12 has reached its end-of-life, I think it's a good idea to remove the code that, in case of `pg_rewind` failure, starts and immediately shuts down PostgreSQL before trying to execute `pg_rewind` again.

Remove the code.

Doing nothing but keeping that logic might add unnecessary noise to the logs if `pg_rewind` fails.

### Additional context

_No response_

### Backport?

No

Yes

### Code of Conduct

leonardoce:
Starting with PostgreSQL 13 pg_rewind automatically handles crash recovery before starting, so it is not needed to manually the postmaster after the first invocation of pg_rewind failed.

Leveraging this feature, this patch removes the code that was manually handling the crash recovery process.

Closes: #6156

### DIFF
--- a/internal/management/controller/instance_startup.go
+++ b/internal/management/controller/instance_startup.go
@@ -270,23 +270,7 @@ func (r *InstanceReconciler) verifyPgDataCoherenceForPrimary(ctx context.Context
 		// retrying after having started up the instance.
 		err = r.instance.Rewind(ctx, pgVersion)
 		if err != nil {
-			contextLogger.Info(
-				"pg_rewind failed, starting the server to complete the crash recovery",
-				"err", err)
-
-			// pg_rewind requires a clean shutdown of the old primary to work.
-			// The only way to do that is to start the server again
-			// and wait for it to be available again.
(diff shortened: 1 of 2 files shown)


leonardoce:
E2e: https://github.com/EnterpriseDB/cloudnative-pg/actions/runs/11974812961

mnencia (approved):

sxd (approved):

sxd:
@leonardoce @gbartolini this should go only on main right? I think we shouldn't backport this to 1.22 1.23 and 1.24

gbartolini (approved):

PR #6216, opened 2024-11-29
feat(pooler): expand configurable options for PgBouncer

### LINKED ISSUE #5276: [Feature]: Allow setting `dns_max_ttl` in PgBouncer

[This article](https://blog.digitalis.io/using-pgbouncer-with-consul-for-postgresql-high-availability-686049c232fc) outlines the motivation, but in essence this will allow reducing downtime during failover when using the Pooler (PgBouncer).

Even though the article is for Consul-based DNS, I believe the architecture and problem is almost identical for Kubernetes-based DNS.

`dns_max_ttl` is not an allowable option in the Pooler settings for PgBouncer ([docs](https://cloudnative-pg.io/documentation/1.23/connection_pooling/)). However, it does not appear to be set at all in the configuration file that CNPG puts onto the PgBouncer pods. As a result, it seems like this should be an allowable option for user configuration especially as it helps achieve the goal outlined above.

None

### Additional context

None

### Backport?

No

No

### Code of Conduct

fullykubed:
This commit extends the `Pooler` API by supporting additional PgBouncer configuration parameters. The newly supported parameters are:  

- `cancel_wait_timeout`  
- `dns_max_ttl`  
- `dns_nxdomain_ttl`  
- `listen_backlog`  
- `max_packet_size`  
- `pkt_buf`  
- `sbuf_loopcnt`  
- `server_tls_ciphers`  
- `server_tls_protocols`  
- `suspend_timeout`  
- `tcp_defer_accept`  
- `tcp_socket_buffer`  

These additions provide greater flexibility, control and responsibility over PgBouncer's behaviour, catering to a broader range of use cases and deployment scenarios.  

Closes #5276  

## Release Notes

Extend the `Pooler` API by supporting additional PgBouncer configuration parameters.


gbartolini:
I have also added server_tls_protocols and server_tls_ciphers. Although the default values are OK, we should also for example enable to the `tlsv1.3` (an EDB customer brought this to my attention and I forgot to comment here).

gbartolini (approved):

mnencia (approved):

bensonyerima:
Bouncer doesn't allow to set server_tls_protocols and server_tls_ciphers as part of the k8s CR. As this solution does not support setting the ciphers of TLS 1.3,
Is it possible to add something like this?

PR #6220, opened 2024-11-29
feat(initdb): add support for ICU and built-in locale providers

### LINKED ISSUE #5386: [Feature]: Support ICU configuration options with `initdb`

PostgreSQL 15 introduced support for the ICU library to provide locale services at cluster initialization, through two new options for `initdb`:

- `--icu-locale`
- `--icu-locale-provider`

See: https://www.postgresql.org/docs/current/app-initdb.html

This is related to #5163 

Add two new options in the `.spec.bootstrap.initdb` stanza:

- `localeProvider`, accepting two values: `libc` and `icu` (empty by default)
- `icuLocale`, accepting the locale ID as a string

A valued `icuLocale` requires `localeProvider` set to `icu`. Prevent setting these options with PostgreSQL < 15.

N/A

### Additional context

At the moment we will not consider the `--icu-rules` option.

### Backport?

No

Yes

### Code of Conduct

mnencia:
This patch enhances the PostgreSQL database initialization process (`initdb`) by introducing support for the ICU locale provider (available since PostgreSQL 16) and the built-in locale provider (available since PostgreSQL 17). Users can now specify the desired locale provider when initializing a new PostgreSQL cluster, offering improved localization flexibility.

Closes #5386 

## Release Notes

Add support for ICU and built-in locale providers during cluster initialization.

### DIFF
--- a/.wordlist-en-custom.txt
+++ b/.wordlist-en-custom.txt
@@ -601,6 +601,7 @@ bootstrapinitdb
 bootstraprecovery
 br
 bs
+builtinLocale
 bw
 byStatus
 bypassrls
@@ -833,6 +834,8 @@ httpGet
 https
 hugepages
 icu
+icuLocale
+icuRules
 ident
 imageCatalogRef
 imageName
@@ -921,6 +924,7 @@ livenessProbeTimeout
 lm
 localeCType
 localeCollate
+localeProvider
 localhost
 localobjectreference
 locktype
--- a/api/v1/cluster_types.go
+++ b/api/v1/cluster_types.go
@@ -1471,6 +1471,10 @@ type BootstrapInitDB struct {
 	// +optional
 	LocaleCType string `json:"localeCType,omitempty"`
 
+	// Sets the default collation order and character classification in the new database.
+	// +optional
+	Locale string `json:"locale,omitempty"`
+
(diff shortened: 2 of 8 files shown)


gbartolini:
@mnencia Shall we ensure in the webhook that `builtinLocale` requires `localeProvider` to be set to `builtin` and `icuLocale` requires `localeProvider` to be set to `icu`?

I tried, it work. Obviously, the documentation can/should be improved, but for now we can send readers to the Postgres docs.

gbartolini (approved):

mnencia:
> @mnencia Shall we ensure in the webhook that `builtinLocale` requires `localeProvider` to be set to `builtin` and `icuLocale` requires `localeProvider` to be set to `icu`?
> 
> I tried, it work. Obviously, the documentation can/should be improved, but for now we can send readers to the Postgres docs.

I've added a cel validation for the fields. I didn't add an enum validation for the localeProvider field because it changes across the versions. However, if we want to add it, it's a one-line change. Let me know what you think.

mnencia:
I think we should evaluate exposing the `--locale` property, which acts as a default for all providers.

gbartolini (approved):

litaocdl (approved):

PR #6246, opened 2024-12-03
chore: set object status as failed  when referred cluster has been deleted

### LINKED ISSUE #6211: on cluster deletion, ensure the database objects are marked as not reconciled

gabriele-wolfox:
Closes #6211


internal/controller/finalizers_delete.go
```diff
@@ -107,6 +115,18 @@ func (r *ClusterReconciler) deleteFinalizersForResource(
 				)
 				return err
 			}
+
+			// set status as failed, as the orphan resource is not reconciled
+			obj.(StatusSetter).SetAsFailed(fmt.Errorf("orphan resource is not reconciled"))
```
armru, line 120: This line is not type-safe, I will make a commit to improve it

Issue #4332, opened 2024-04-18
[Bug]: kubectl cnpg psql doesn't pass `--context` to kubectl


redbaron:
### Is there an existing issue already for this bug?

- [X] I have searched for an existing issue, and could not find anything. I believe this is a new bug.

### I have read the troubleshooting guide

- [X] I have read the troubleshooting guide and I think this is a new bug.

### I am running a supported version of CloudNativePG

- [X] I have read the troubleshooting guide and I think this is a new bug.

### Version

1.22.2

### What version of Kubernetes are you using?

1.29

### What is your Kubernetes environment?

Cloud: Google GKE

### How did you install the operator?

Helm

### What happened?

`kubectl cnpg psql --context X`  doesn't pass `--context X` to underneath kubectl invocation , as a result psql doesn't work unless context X is selected as default in $KUBECONFIG

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

sxd:
Hi @redbaron 

The flag `--context` is a global flag, can you test passing this flag before the `psql` command ?

Regards,

redbaron:
> Hi @redbaron
> 
> The flag `--context` is a global flag, can you test passing this flag before the `psql` command ?
> 
> Regards,

Tried multiple variants, same result.

```
$ kubectl  cnpg --context "X" -n s-foobar psql s-foobar-3cbd
Error from server (NotFound): namespaces "s-foobar" not found
$ kubectl ctx X
Switched to context "X".
$ kubectl  cnpg --context "X" -n s-foobar psql s-foobar-3cbd
psql (15.6 (Debian 15.6-1.pgdg110+2))
Type "help" for help.

postgres=# 

```

sxd:
@redbaron 

Thanks for reporting this, I can confirm it's a bug! I'll add some more labels asking for help on a PR

Best Regards!

armru:
Unable to reproduce on the latest version

redbaron:
Still reproducible for me on 

```
$ kubectl cnpg version                                                           
Build: {Version:1.23.2 Commit:4bef8412 Date:2024-06-12}
```

gbartolini:
This is a duplicate of #6227 - closing it.

PR #6257, opened 2024-12-03
fix(plugin): pass context in `psql` command

### LINKED ISSUE #6227: [Bug]: cnpg plugin not handling command line context correctly

### Contact Details

lbruun@apache.org

### Version

1.24.0

1.28

Self-managed: RKE

YAML manifest

### What happened?

The cnpg plugin command for `psql` does not seem to propagate the `--context` option correctly.

### Reproduce

Ensure that your kube config defines at least 2 contexts. In this example we call them `c1`and `c2`. For the sake of the example it helps if `c2` is unreachable.

If your current set context is `c2` and you want to execute a `kubectl cnpg` command using `c1` as the context then you would do

```bash
kubectl cnpg --context c1 --namespace myns  psql  my-pg-cluster 
```

However, this will not work as `--context c1` is not taken into account. The command will instead operate on the set context, in our example that would be `c2`.
If `c2` in unreachable then this is quite easy to see.

### Workaround

Change the set context before executing the command, i.e.:

```bash
kubectl config use-context c1
kubectl cnpg --namespace myns  psql  my-pg-cluster 
```

### Cluster resource

```shell

```

### Relevant log output

```shell

```

### Code of Conduct

sxd:
We were not passing the context from the kubectl cnpg plugin call to
the kubectl command executed when calling psql, this PR fix that by
adding a new variable to the plugin that has the Kubernetes context
used to call the plugin.

Closes #6227 #4332

### DIFF
--- a/internal/cmd/plugin/plugin.go
+++ b/internal/cmd/plugin/plugin.go
@@ -43,6 +43,9 @@ var (
 	// Namespace to operate in
 	Namespace string
 
+	// KubeContext to operate with
+	KubeContext string
+
 	// NamespaceExplicitlyPassed indicates if the namespace was passed manually
 	NamespaceExplicitlyPassed bool
 
@@ -95,6 +98,8 @@ func SetupKubernetesClient(configFlags *genericclioptions.ConfigFlags) error {
 	if err != nil {
 		return err
 	}
+
+	KubeContext = *configFlags.Context
 
 	ClientInterface = kubernetes.NewForConfigOrDie(Config)
 
--- a/internal/cmd/plugin/psql/cmd.go
+++ b/internal/cmd/plugin/psql/cmd.go
@@ -45,6 +45,7 @@ func NewCmd() *cobra.Command {
 			psqlOptions := CommandOptions{
 				Replica:     replica,
 				Namespace:   plugin.Namespace,
+				Context:     plugin.KubeContext,
 				AllocateTTY: allocateTTY,
 				PassStdin:   passStdin,
 				Args:        psqlArgs,
--- a/internal/cmd/plugin/psql/psql.go
+++ b/internal/cmd/plugin/psql/psql.go
@@ -59,6 +59,9 @@ type CommandOptions struct {
 	// The Namespace where we're working in
 	Namespace string
 
+	// The Context to execute the command
+	Context string
+
 	// Whether we should we allocate a TTY for psql
 	AllocateTTY bool
 
@@ -105,6 +108,10 @@ func NewCommand(
 func (psql *Command) getKubectlInvocation() ([]string, error) {
 	result := make([]string, 0, 13+len(psql.Args))
 	result = append(result, "kubectl", "exec")
+
+	if psql.Context != "" {
+		result = append(result, "--context", psql.Context)
+	}
 
 	if psql.AllocateTTY {
 		result = append(result, "-t")


YanniHu1996:
In addition to the `context`, there are several Kubernetes configuration flags that need to be included when calling the `kubectl exec` command. These flags are:
```sh
 kubectl options
The following options can be passed to any command:

    --as='':
        Username to impersonate for the operation. User could be a regular user or a service account in a namespace.

    --as-group=[]:
        Group to impersonate for the operation, this flag can be repeated to specify multiple groups.

    --as-uid='':
        UID to impersonate for the operation.

    --cache-dir='/Users/huyantian/.kube/cache':
        Default cache directory

    --certificate-authority='':
        Path to a cert file for the certificate authority

    --client-certificate='':
        Path to a client certificate file for TLS

    --client-key='':
        Path to a client key file for TLS

    --cluster='':
        The name of the kubeconfig cluster to use

    --context='':
        The name of the kubeconfig context to use

    --disable-compression=false:
        If true, opt-out of response compression for all requests to the server

    --insecure-skip-tls-verify=false:
        If true, the server's certificate will not be checked for validity. This will make your HTTPS connections
        insecure

    --kubeconfig='':
        Path to the kubeconfig file to use for CLI requests.

    --log-flush-frequency=5s:
        Maximum number of seconds between log flushes

    --match-server-version=false:
        Require server version to match client version

    -n, --namespace='':
        If present, the namespace scope for this CLI request

    --password='':
        Password for basic authentication to the API server

    --profile='none':
        Name of profile to capture. One of (none|cpu|heap|goroutine|threadcreate|block|mutex)

    --profile-output='profile.pprof':
        Name of the file to write the profile to

    --request-timeout='0':
        The length of time to wait before giving up on a single server request. Non-zero values should contain a
        corresponding time unit (e.g. 1s, 2m, 3h). A value of zero means don't timeout requests.

    -s, --server='':
        The address and port of the Kubernetes API server

    --tls-server-name='':
        Server name to use for server certificate validation. If it is not provided, the hostname used to contact the
        server is used

    --token='':
        Bearer token for authentication to the API server

    --user='':
        The name of the kubeconfig user to use

    --username='':
        Username for basic authentication to the API server

    -v, --v=0:
        number for the log level verbosity

    --vmodule=:
        comma-separated list of pattern=N settings for file-filtered logging (only works for the default text log
        format)

    --warnings-as-errors=false:
        Treat warnings received from the server as errors and exit with a non-zero exit code
```

@mnencia Do you think we should address this in this pull request, or a separate one?

---

More info

Global Flags for the CNPG Plugin Related to Kubernetes

```sh 
--as string                      Username to impersonate for the operation. User could be a regular user or a service account in a namespace.
--as-group stringArray           Group to impersonate for the operation, this flag can be repeated to specify multiple groups.
--as-uid string                  UID to impersonate for the operation.
--cache-dir string               Default cache directory (default "/Users/huyantian/.kube/cache")
--certificate-authority string   Path to a cert file for the certificate authority
--client-certificate string      Path to a client certificate file for TLS
--client-key string              Path to a client key file for TLS
--cluster string                 The name of the kubeconfig cluster to use
--context string                 The name of the kubeconfig context to use
--disable-compression            If true, opt-out of response compression for all requests to the server
--insecure-skip-tls-verify       If true, the server's certificate will not be checked for validity. This will make your HTTPS connections insecure
--kubeconfig string              Path to the kubeconfig file to use for CLI requests.
-n, --namespace string               If present, the namespace scope for this CLI request
--request-timeout string         The length of time to wait before giving up on a single server request. Non-zero values should contain a corresponding time unit (e.g. 1s, 2m, 3h). A value of zero means don't timeout requests. (default "0")
-s, --server string                  The address and port of the Kubernetes API server
--tls-server-name string         Server name to use for server certificate validation. If it is not provided, the hostname used to contact the server is used
--token string                   Bearer token for authentication to the API server
--user string                    The name of the kubeconfig user to use
 ```

mnencia (approved):

mnencia:
> @mnencia Do you think we should address this in this pull request, or a separate one?

Let's open a new ticket for the other arguments.

YanniHu1996 (approved):

NiccoloFei (approved):

PR #6258, opened 2024-12-03
feat: ensure unique manager for declarative PostgreSQL resources

### LINKED ISSUE #5922: Detect if a publication / subscription in Postgres is already managed by another K8s Publication/Subscription

jsilvela:
This patch extends the existing logic that ensures only one manager exists for a Postgres resource in the Database support code to also cover Publications and Subscriptions. 

Close #5922


jsilvela:
This is working:

``` sh
NAME                        AGE   CLUSTER           PG NAME   APPLIED   MESSAGE
publication-example         29s   cluster-example   pub-all   true      
publication-example-other   5s    cluster-example   pub-all   false     "pub-all" is already managed by object "publication-example"
```

``` sh
NAME                        AGE   CLUSTER                PG NAME   APPLIED   MESSAGE
subscription-sample         35s   cluster-example-dest   sub       true      
subscription-sample-other   2s    cluster-example-dest   sub       false     "sub" is already managed by object "subscription-sample"
```

However, the feature generally is badly lacking unit tests.
Given enough time I'll add unit tests in this PR or a following one.

internal/management/controller/common.go
```diff
@@ -25,12 +25,51 @@ import (
...
+	GetName() string
+	GetManagedObjectName() string
+	GetClusterRef() corev1.LocalObjectReference
+	HasReconciliations() bool
+}
+
+// detectTooManyManagers checks if a certain `manager` wants to manage the same
+// underlying object as some previously existing manager
+//
+//	also known as turf war
```
jsilvela, line 45:
```suggestion
```

armru (approved):

gabriele-wolfox (approved):

mnencia (approved):

gbartolini:
This should have been a chore IMHO, not a feature commit.

PR #6266, opened 2024-12-04
feat(probes): enable customization of startup, liveness, and readiness probes

### LINKED ISSUE #4852: [Feature]: Make pod liveness/readiness probes timeout configurable

leonardoce:
This patch enables users to customize the default behavior of readiness, liveness, and startup probes implemented by CloudNativePG by introducing the `.spec.probes` section.

Users are responsible for ensuring that any custom probe settings align with the operational requirements of the cluster to avoid unintended disruptions.

Special thanks to @bentastic27 for the contribution that inspired this improvement.

Closes: #4852


leonardoce:
E2e: https://github.com/EnterpriseDB/cloudnative-pg/actions/runs/12157706881

mnencia (approved):

leonardoce:
The E2e tests are green!

gbartolini (approved):

gbartolini:
This patch should be backported.

bentastic27 (approved):

PR #6303, opened 2024-12-09
docs: Remove deprecated Pod Security Policies from docs

### LINKED ISSUE #5996: [Bug]: Documentation still reports pod security policies

### Contact Details

_No response_

### Version

1.24.0

1.31

YAML manifest

### What happened?

The security page still reports pod security policies - which have been deprecated. See: https://cloudnative-pg.io/documentation/current/security/#pod-security-policies

We should change it into security contexts.

I exported the ones for the postgres container:

```yaml
    securityContext:
      allowPrivilegeEscalation: false
      capabilities:
        drop:
        - ALL
      privileged: false
      readOnlyRootFilesystem: true
      runAsNonRoot: true
      seccompProfile:
        type: RuntimeDefault
```

### Cluster resource

_No response_

### Relevant log output

_No response_

### Code of Conduct

sharifmshaker:
Removing deprecated Pod Security Policies information from security docs. 

Closes #5996

### DIFF
--- a/docs/src/security.md
+++ b/docs/src/security.md
@@ -293,7 +293,7 @@ CloudNativePG.
 : The instance manager requires to `update` and `patch` the status of any
   `Backup` resource in the namespace
 
-### Security Contexts
+### Pod and Container Security Contexts
 
 A [Security Context](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/)
 defines privilege and access control settings for a pod or container.
@@ -325,11 +325,10 @@ securityContext:
     type: RuntimeDefault
 ```
 
-#### RedHat OpenShift SCC
+#### Security Context Constraints
 
-When running in a RedHat OpenShift environment the operator does not explicitly set the security context of the
-PostgreSQL cluster pods, but rather allows the pods to inherit the restricted [Security Context Constraints (SCC)](https://docs.openshift.com/container-platform/4.17/authentication/managing-security-context-constraints.html) 
-that are defined within OpenShift.
+When running in an environment that is utilizing [Security Context Constraints (SCC)](https://docs.openshift.com/container-platform/4.17/authentication/managing-security-context-constraints.html)
+the operator does not explicitly set the security context of the PostgreSQL cluster pods, but rather allows the pods to inherit the restricted Security Context Constraints that are already defined.
 
 ### Restricting Pod access using AppArmor
 
--- a/docs/src/security.md
+++ b/docs/src/security.md
@@ -299,19 +299,23 @@ A [Security Context](https://kubernetes.io/docs/tasks/configure-pod-container/se
 defines privilege and access control settings for a pod or container.
 
 CloudNativePG does not require *privileged* mode for container execution.
-The PostgreSQL containers run as the `postgres` system user. No component whatsoever requires running as `root`.
+The PostgreSQL containers run as the `postgres` system user. No component
+whatsoever requires running as `root`.
 
-Likewise, Volume access does not require *privileged* mode nor `root` privileges.
-Proper permissions must be assigned by the Kubernetes platform and/or administrators.
-The PostgreSQL containers run with a read-only root filesystem (i.e. no writable layer).
+Likewise, Volume access does not require *privileged* mode nor `root`
+privileges. Proper permissions must be assigned by the Kubernetes platform
+and/or administrators. The PostgreSQL containers run with a read-only root
+filesystem (i.e. no writable layer).
 
-The operator manages the setting of security contexts for all pods and containers of a PostgreSQL cluster.
-The [Seccomp Profile](https://kubernetes.io/docs/tutorials/security/seccomp/) to be used for the PostgreSQL
-containers can be configured with the `spec.seccompProfile` section of the `Cluster` resource. If this 
-section is left blank, the containers will use a seccompProfile `Type` of `RuntimeDefault`, that is, the
-container runtime default.
+The operator manages the setting of security contexts for all pods and 
+containers of a PostgreSQL cluster. The [Seccomp Profile](https://kubernetes.io/docs/tutorials/security/seccomp/) 
+to be used for the PostgreSQL containers can be configured with the 
+`spec.seccompProfile` section of the `Cluster` resource. If this section is left
+blank, the containers will use a seccompProfile `Type` of `RuntimeDefault`, that
+is, the container runtime default.
 
-The security context of PostgreSQL containers using the default `seccompProfile` will look like this:
+The security context of PostgreSQL containers using the default `seccompProfile`
+will look like this:
 ```
 securityContext:
   allowPrivilegeEscalation: false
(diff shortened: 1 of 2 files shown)


docs/src/security.md
```diff
@@ -293,28 +293,39 @@ CloudNativePG.
...
+The PostgreSQL containers run as the `postgres` system user. No component whatsoever requires running as `root`.
 
-Likewise, Volumes access does not require *privileges* mode or `root` privileges either.
+Likewise, Volume access does not require *privileges* mode nor `root` privileges.
 Proper permissions must be properly assigned by the Kubernetes platform and/or administrators.
 The PostgreSQL containers run with a read-only root filesystem (i.e. no writable layer).
 
-The operator explicitly sets the required security contexts.
+The operator explicitly sets the required security contexts for all pods and
+containers in the cluster. However, the
```
jsilvela, line 309: The "However" here ... not sure. Configurability of the seccompProfile is not "against" the setting of the security context.

jsilvela (approved):
Nice work. Took the chance to fix a couple of nits in the section.
One optional thought left as an inline comment.

docs/src/security.md
```diff
@@ -314,7 +305,24 @@ Likewise, Volumes access does not require *privileges* mode or `root` privileges
 Proper permissions must be properly assigned by the Kubernetes platform and/or administrators.
 The PostgreSQL containers run with a read-only root filesystem (i.e. no writable layer).
 
-The operator explicitly sets the required security contexts.
+The operator explicitly sets the required security contexts for all pods and containers in the cluster.
+The [Seccomp Profile](https://kubernetes.io/docs/tutorials/security/seccomp/) to be used for the security 
+context of the PostgreSQL containers can be configured with the `spec.seccompProfile` of the `Cluster` resource.
+If no `seccompProfile` is requested, the `Cluster` containers will use a profile `Type` of `RuntimeDefault`.
```
sxd, line 311: This is forgetting one important point, the OLM case

sharifmshaker: Can you elaborate a little on what you are referring to with the OLM case or point me in the direction of relevant information? I am not so familiar with the details there. From what I have read it is already documented in the security doc that when deployed with OLM the operator can be configured with permissions to manage clusters in all or specific namespaces. I didn't see how this specifically changed how the PostgreSQL security context is modified on PG pods. 

Something I did find was that in OpenShift the PG pods inherit the security context constraints from OpenShift as opposed to be explicitly set on the pods.  I added an additional explanatory note about this to the doc in the PR.

docs/src/security.md
```diff
@@ -293,19 +293,10 @@ CloudNativePG.
...
-    this deprecation, we acknowledge that the operator is currently undergoing
-    testing in older and unsupported versions of Kubernetes. Therefore, this
-    section is retained for those specific scenarios.
-
-A [Pod Security Policy](https://kubernetes.io/docs/concepts/policy/pod-security-policy/)
-is the Kubernetes way to define security rules and specifications that a pod needs to meet
-to run in a cluster.
-For InfoSec reasons, every Kubernetes platform should implement them.
+A [Security Context](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/) is a
+Kubernetes way to defines privilege and access control settings for a Pod or Container.
```
jsilvela, line 299: Non-grammatical.

jsilvela (changes requested):
Hi @sharifmshaker , I had made some changes a few days ago.
You seem to have forced-pushed and eliminated them.
There were a few optional changes.
But the one I point out below is an outright grammar error and should be fixed.

docs/src/security.md
sharifmshaker: Apologies! Changed `Kubernetes way to define...`

sxd:
This shouldn't be backported

docs/src/security.md
sxd: https://github.com/cloudnative-pg/cloudnative-pg/blob/main/pkg/specs/pods.go#L426

sharifmshaker: The function to which you are referring is why I added an additional note stating that in the openshift case we inherit the security context based on the security context constraints provided by openshift. Is this sufficient or is more explanation required?

sharifmshaker:
> This shouldn't be backported

Is that just the label? It doesn't look like I have permissions to add or remove labels on these PRs. Is there another way to prevent the backport?

jsilvela (approved):

docs/src/security.md
```diff
@@ -293,28 +293,43 @@ CloudNativePG.
...
+    drop:
+    - ALL
+  privileged: false
+  readOnlyRootFilesystem: true
+  runAsNonRoot: true
+  seccompProfile:
+    type: RuntimeDefault
+```
+
+#### RedHat OpenShift SCC
```
sxd, line 328: CNP doesn't support RedHat OpenShift, mealning that it shouldn't be mentioned in the documentation

docs/src/security.md
```diff
@@ -293,28 +293,43 @@ CloudNativePG.
...
+
+When running in a RedHat OpenShift environment the operator does not explicitly set the security context of the
+PostgreSQL cluster pods, but rather allows the pods to inherit the restricted [Security Context Constraints (SCC)](https://docs.openshift.com/container-platform/4.17/authentication/managing-security-context-constraints.html) 
+that are defined within OpenShift.
```
sxd, line 332: Mentioning explicitly OpenShift here is not a good idea, probably is better to refer to with something "like the security context constraints" or something like that to avoid confusions

docs/src/security.md
```diff
@@ -293,28 +293,43 @@ CloudNativePG.
 : The instance manager requires to `update` and `patch` the status of any
   `Backup` resource in the namespace
 
-### Pod Security Policies
+### Security Contexts
```
sxd, line 296: This is for the pods and containers right? I think we should mention that

docs/src/security.md
sharifmshaker: Thanks @sxd for reviewing. So based my searching around, all references to Security Context Constraints that I found were in relation to openshift. It was my understanding that it was an openshift specific feature and I couldn't find documentation to reference it that was not also talking about openshift. Am I looking at the wrong thing for SCC? I would like to link documentation here as part of the explanation, but if we can't mention openshift I am wondering if you know what documentation I can link here?

docs/src/security.md
sharifmshaker: Something like `Pod and Container Security Contexts`. The fact that security contexts apply to both is mentioned in the body of the doc
```
The operator manages the setting of security contexts for all pods and containers of a PostgreSQL cluster.
```
so assuming you also want it in the title?

docs/src/security.md
sharifmshaker: I removed the explicit references to openshift, but I left the link to the openshift doc since that seems to be the main source of truth for SCC

jsilvela:
We should be at the point of wrapping this up. Are there any outstanding objections?
I don't want this to slip the next release...
@sharifmshaker please make sure Jonathan is happy with your changes.
Also, please can you make sure the lines fit in 80 columns? It's a rule of sorts for our docs.

jsilvela:
I think this is only waiting for your comments @sxd , can you please take a look?

sxd (approved):

PR #6455, opened 2024-12-24
test(e2e): raise `AssertClusterStandbysAreStreaming` timeout

### LINKED ISSUE #6398: [Bug]: insufficient timeout for reconnections in e2e

### Contact Details

_No response_

### Version

trunk (main)

1.31

YAML manifest

### What happened?

Whenever `AssertClusterStandbysAreStreaming`  is called, a 120 seconds timeout is used. However is is possible that the moment the connection is started the kube proxies are not up to date yet with their routing and the connections goes nowhere. Since the `tcp_syn_retries` kernel value is 6 by default, the timeout to retry a connection if the first one fails is 127 seconds.

From https://docs.kernel.org/networking/ip-sysctl.html
```
tcp_syn_retries - INTEGER

    Number of times initial SYNs for an active TCP connection attempt will be retransmitted. Should not be higher than 127. Default value is 6, which corresponds to 67seconds (with tcp_syn_linear_timeouts = 4) till the last retransmission with the current initial RTO of 1second. With this the final timeout for an active TCP connection attempt will happen after 131seconds.
```

Therefore we need to raise the timeout to allow at least a retry.

### Cluster resource

```shell

```

### Relevant log output

```shell

```

### Code of Conduct

armru:
Closes #6398

### DIFF
--- a/tests/e2e/asserts_test.go
+++ b/tests/e2e/asserts_test.go
@@ -1575,7 +1575,7 @@ func AssertClusterRestoreWithApplicationDB(namespace, restoreClusterFile, tableN
 	})
 
 	// Restored standby should be attached to restored primary
-	AssertClusterStandbysAreStreaming(namespace, restoredClusterName, 120)
+	AssertClusterStandbysAreStreaming(namespace, restoredClusterName, 140)
 
 	// Gather Credentials
 	appUser, appUserPass, err := testsUtils.GetCredentials(restoredClusterName, namespace,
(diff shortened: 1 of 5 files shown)


fcanovai (approved):

gabriele-wolfox (approved):

gbartolini (approved):

gbartolini:
Hopefully with the new refactoring we will be able to have an inventory of all timeouts we have, in a centralized file or similar.

PR #6546, opened 2025-01-09
feat: add support for configuring PostgreSQL extensions via Image Volume

NiccoloFei:
Add support to allow setting up PostgreSQL extensions via container images, taking advance of the `ImageVolume` VolumeSource of Kubernetes and the new proposed GUC of PostgreSQL `extension_control_path`.

Introduce a new API field `Extensions` in the `cluster.Spec.PostgresConfiguration` stanza, which takes an array of PostgreSQL extensions which should be added to the Cluster.


gbartolini:
The `extension_control_path` patch has been merged in PostgreSQL. The `postgres-trunk-containers` project now ships that GUC and we can use it to test in the future. See: https://github.com/postgres/postgres/commit/4f7f7b0375854e2f89876473405a8f21c95012af

pkg/postgres/configuration.go
```diff
@@ -693,6 +702,14 @@ func CreatePostgresqlConfiguration(info ConfigurationInfo) *PgConfiguration {
 		configuration.OverwriteConfig("temp_tablespaces", strings.Join(info.TemporaryTablespaces, ","))
 	}
 
+	if len(info.ImageVolumeExtensions) > 0 {
+		// Set all ExtensionControlPaths
+		setExtensionControlPath(info, configuration)
+
+		// Set all DynamicLibraryPaths
+		setDynamicLibraryPath(info, configuration)
```
hossain-rayhan, line 710: Is it guaranteed that we'll always have both `extensionControlPath` and `dynamicLibraryPath` if we are inside the block?

NiccoloFei: @hossain-rayhan I think so, given that we want to adhere to the PGXS standard declared in https://www.postgresql.org/docs/devel/extend-pgxs.html, for which every extension (even the most basic ones) will have extension control and SQL files in `/share` and the shared library in `/lib`. So we'll have to set both those GUCs to allow PG to find the extension's files. Although I'm wondering how the build infra would behave in case of pure-SQL extensions. I'll try to test/investigate this specific use case, but I think the worst that could happen is that we'd end up setting a useless path inside `dynamic_library_path`.
Were you thinking of a specific scenario, or was it just a generic question?

NiccoloFei:
Closing this PR as its purpose was to serve as exploration work for the new feature.
Superseded by https://github.com/cloudnative-pg/cloudnative-pg/pull/7991

PR #6600, opened 2025-01-14
Add cnpg suffix of cnpg resources

### LINKED ISSUE #6308: [Docs]: Prefix the cnpg resource type in kubectl examples

The resource type in Kubernetes, for example `cluster`, need only be unique when combined with the API group. This is a problem because several Kubernetes CRDs from various projects seems to use the name `cluster`  (and you may find some that would use the name `backup` too, etc).

If a Kubernetes cluster has several resource types with the same name, then the `kubectl get` command will list (only) the one that comes first in the retrieved API discovery document. This may be a surprising result.

To be on the safe side the resource type can be qualified when using it in a `kubectl` command.

In others words:

This may be ambiguous:

```bash
kubectl get cluster     # Show all Pg Clusters (or?)
```
.. but this is precise (and shows intent);

```bash
kubectl get cluster.postgresql.cnpg.io   # Show all Pg Clusters
```

Bottom line: I believe the qualified form should be evangelized as a kind of a best practice somewhere in the documentation .. to make sure scripts are future proof. 

Put a comment somewhere about using _qualified_ resource type names in `kubectl`  commands. .. if you want to be safe.

_No response_

_No response_

### Additional context

_No response_

### Backport?

Yes

### Code of Conduct

pchovelon:
fix #6308

### DIFF
--- a/contribute/e2e_testing_environment/README.md
+++ b/contribute/e2e_testing_environment/README.md
@@ -91,7 +91,7 @@ will create a deployment, and add two services on ports 6060 and 4040
 respectively, in the same namespace as the operator:
 
 ``` console
-kubectl get svc -n cnpg-system        
+kubectl get service -n cnpg-system        
 
 NAME                   TYPE        CLUSTER-IP     EXTERNAL-IP   PORT(S)    AGE
 cnpg-pprof             ClusterIP   10.96.17.58    <none>        6060/TCP   9m41s
--- a/docs/src/declarative_hibernation.md
+++ b/docs/src/declarative_hibernation.md
@@ -39,7 +39,7 @@ The hibernation status can be monitored by looking for the `cnpg.io/hibernation`
 condition:
 
 ``` sh
-$ kubectl get cluster <cluster-name> -o "jsonpath={.status.conditions[?(.type==\"cnpg.io/hibernation\")]}" 
+$ kubectl get cluster.postgresql.cnpg.io <cluster-name> -o "jsonpath={.status.conditions[?(.type==\"cnpg.io/hibernation\")]}" 
 
 {
         "lastTransitionTime":"2023-03-05T16:43:35Z",
--- a/docs/src/kubectl-plugin.md
+++ b/docs/src/kubectl-plugin.md
@@ -638,7 +638,7 @@ metadata:
 
 The `cluster` sub-command gathers the following:
 
-* **cluster resources**: the cluster information, same as `kubectl get cluster -o yaml`
+* **cluster resources**: the cluster information, same as `kubectl get cluster.postgresql.cnpg.io -o yaml`
 * **cluster pods**: pods in the cluster namespace matching the cluster name
 * **cluster jobs**: jobs, if any, in the cluster namespace matching the cluster name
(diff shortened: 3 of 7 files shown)


contribute/e2e_testing_environment/README.md
```diff
@@ -91,7 +91,7 @@ will create a deployment, and add two services on ports 6060 and 4040
 respectively, in the same namespace as the operator:
 
 ``` console
-kubectl get svc -n cnpg-system        
+kubectl get service -n cnpg-system
```
sxd, line 94: why this is necessary ?

docs/src/declarative_hibernation.md
```diff
@@ -39,7 +39,7 @@ The hibernation status can be monitored by looking for the `cnpg.io/hibernation`
 condition:
 
 ``` sh
-$ kubectl get cluster <cluster-name> -o "jsonpath={.status.conditions[?(.type==\"cnpg.io/hibernation\")]}" 
+$ kubectl get cluster.postgresql.cnpg.io <cluster-name> -o "jsonpath={.status.conditions[?(.type==\"cnpg.io/hibernation\")]}"
```
sxd, line 42: I don't think this is required, why you want to add it?

docs/src/kubectl-plugin.md
```diff
@@ -638,7 +638,7 @@ metadata:
 
 The `cluster` sub-command gathers the following:
 
-* **cluster resources**: the cluster information, same as `kubectl get cluster -o yaml`
+* **cluster resources**: the cluster information, same as `kubectl get cluster.postgresql.cnpg.io -o yaml`
```
sxd, line 641: same as before

sxd (commented):
Why is this needed? besides the fact that none of them are prefix but suffix and change the service call

contribute/e2e_testing_environment/README.md
pchovelon: To be more consistent through all the documentation (see docs/src/quickstart.md)

docs/src/declarative_hibernation.md
pchovelon: To be more consistent through all the documentation :)

docs/src/kubectl-plugin.md
pchovelon: To be more consistent through all the documentation :)

docs/src/declarative_hibernation.md
thomasboussekey: Agree with @pchovelon , in our kubernetes clusters, we have both [crossplane](https://www.crossplane.io/) and cloudnativepg operators declared.
Crossplane "gets" the results for `kubectl get cluster` by default, since we have installed it.
So, we need to use the full name `cluster.postgresql.cnpg.io`, in order to view the CNPG cluster, now.

lbruun:
An alternative to this PR is in [PR-1704](https://github.com/kubernetes/kubectl/issues/1704)

lbruun:
[PR-7443](https://github.com/cloudnative-pg/cloudnative-pg/pull/7443) is an alternative to this PR.

PR #6603, opened 2025-01-15
chore: add SBOMs for the operator container images and archive builds

### LINKED ISSUE #4199: [Feature]: Add SBOMs for operator images and kubectl plugin

Currently we don't generate any kind of SBOMS

We should generate SBOMs for the operator images and the kubectl plugin, for both of them it should be during release time. These SBOMs should be published alongside with the other packages published per release

We could also generate these SBOMs during the CI/CD process

### Additional context

_No response_

### Backport?

Yes

No

### Code of Conduct

litaocdl:
Closes #4199

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -633,7 +633,7 @@ jobs:
       - name: Run Snyk to check Docker image for vulnerabilities
         uses: snyk/actions/docker@master
         if: |
-          !github.event.repository.fork && 
+          !github.event.repository.fork &&
           !github.event.pull_request.head.repo.fork
         continue-on-error: true
         env:
@@ -645,7 +645,7 @@ jobs:
       - name: Upload result to GitHub Code Scanning
         uses: github/codeql-action/upload-sarif@v3
         if: |
-          !github.event.repository.fork && 
+          !github.event.repository.fork &&
           !github.event.pull_request.head.repo.fork
         continue-on-error: true
         with:
@@ -666,7 +666,7 @@ jobs:
           cache-to: ${{ env.BUILD_PUSH_CACHE_TO }}
 
       - name: Generate SBOM for image
-        uses: anchore/sbom-action@v0.15.10
+        uses: anchore/sbom-action@v0.17.9
         with:
           image: ${{ steps.docker-meta.outputs.tags }}
           artifact-name: cloudnative-pg-${{ env.VERSION }}.spdx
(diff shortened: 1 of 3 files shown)


litaocdl:
0.30.0 is the release sample
https://github.com/EnterpriseDB/cloudnative-pg/releases/tag/v0.30.0

.github/workflows/continuous-integration.yml
```diff
@@ -687,6 +703,14 @@ jobs:
             VERSION=${{ env.VERSION }}
           tags: ${{ steps.docker-meta-ubi9.outputs.tags }}
 
+      - name: Generate SBOM for image
+        uses: anchore/sbom-action@v0.17.9
+        with:
+          image: ${{ steps.docker-meta-ubi9.outputs.tags }}
+          artifact-name: cloudnative-pg-${{ env.VERSION }}-ubi9.spdx
+          upload-artifact: true
+          upload-artifact-retention: 7d
```
sxd, line 712: We shouldn't have a retention for these artifacts, for any artifact that it's an SBOM

.goreleaser.yml
```diff
@@ -113,6 +113,9 @@ signs:
   - artifacts: checksum
     args: ["--batch", "-u", "{{ .Env.GPG_FINGERPRINT }}", "--output", "${signature}", "--detach-sign", "${artifact}"]
 
+sboms:
+  - artifacts: archive
+
```
sxd, line 118: can we just make the goreleaser the single point to generate sboms so we can reproduce it locally with no other command execution?

litaocdl: SBOM in goreleaser can only work with `goreleaser release` command. So in ci workflow, if we want to generate SBOM report, we can not use goreleaser (as we are using goreleaser build).

.github/workflows/continuous-integration.yml
litaocdl: This is the SBOM in CI workflow, which will be trigger and generate SBOM report for every ci run, probably a longer retention date ?

sxd (approved):

sxd:
I think we're generating the SBOM only for one architecture and we have at least two architectures there, can you check if that it's the case @litaocdl  ?

litaocdl:
@sxd you are right, syft only generate reports on the architecture it is running, unless extra params is given. let me try.

litaocdl:
@sxd I update the pr with both arm64 and amd64 scaned, syft can only scan one by one, so we have to call the sbom action multiple times for multiple architecture

sxd:
Superseeded by https://github.com/cloudnative-pg/cloudnative-pg/pull/6806 
Closing

PR #6607, opened 2025-01-15
fix: ensure cluster dependant resources reconcile after cluster rehydratation

### LINKED ISSUE #6550: [Bug]: Imperative Hibernation breaks declarative databases

### Contact Details

_No response_

### Version

1.25 (latest patch)

other (unsupported)

Other

Helm

### What happened?

These are the steps to reproduce:
1. Create a `clusters.postgresql.cnpg.io` resource, let's call it `mycluster`
2. Create a database `mydb` via `databases.postgresql.cnpg.io` targeting `mycluster`
3. Hibernate the cluster using the `kubectl` plugin: `kubectl cnpg hibernate on mycluster`

The status of `mydb` will then read:
```
status:
  applied: false
  message: cluster resource has been deleted, skipping reconciliation
```
This state will continue after hibernation is turned off again, i.e. `mydb` is no longer reconciled from then on.

### Cluster resource

```shell

```

### Relevant log output

```shell

```

### Code of Conduct

armru:
Closes #6550 

# Release notes

```
After the cluster rehydration, the `Database`, `Publication`, and `Subscription` CRDs now properly reconcile instead of being stuck at `cluster resource has been deleted, skipping reconciliation. This is done by forcing `status.observedGeneration` to zero.
```

### DIFF
--- a/internal/controller/finalizers_delete.go
+++ b/internal/controller/finalizers_delete.go
@@ -86,6 +86,7 @@ type clusterOwnedResourceWithStatus interface {
 	GetClusterRef() corev1.LocalObjectReference
 	GetStatusMessage() string
 	SetAsFailed(err error)
+	SetStatusObservedGeneration(obsGeneration int64)
 }
 
 func toSliceWithPointers[T any](items []T) []*T {
@@ -121,8 +122,9 @@ func notifyOwnedResourceDeletion[T clusterOwnedResourceWithStatus](
 
 		if obj.GetStatusMessage() != statusMessage {
 			obj.SetAsFailed(errors.New(statusMessage))
-			if err := cli.Status().Patch(ctx, obj, client.MergeFrom(origObj)); err != nil {
-				itemLogger.Error(err, "error while setting failed status for cluster deletion")
+			obj.SetStatusObservedGeneration(0)
+			if err := cli.Status().Update(ctx, obj); err != nil {
+				itemLogger.Error(err, "error while updating failed status for cluster deletion")
 				return err
 			}
 		}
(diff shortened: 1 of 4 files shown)


internal/controller/finalizers_delete.go
```diff
@@ -121,8 +122,9 @@ func notifyOwnedResourceDeletion[T clusterOwnedResourceWithStatus](
 
 		if obj.GetStatusMessage() != statusMessage {
 			obj.SetAsFailed(errors.New(statusMessage))
-			if err := cli.Status().Patch(ctx, obj, client.MergeFrom(origObj)); err != nil {
-				itemLogger.Error(err, "error while setting failed status for cluster deletion")
+			obj.SetStatusObservedGeneration(0)
+			if err := cli.Status().Update(ctx, obj); err != nil {
```
mnencia, line 126: Note for reviewers: an update is needed here because the observed generation value would be removed by the `omitempty` serialization otherwise.

mnencia:
```suggestion
			obj.SetStatusObservedGeneration(0)
			// We need to use an update here because of the observed generation set to 0
			// that would be ignored with the patch method.
			if err := cli.Status().Update(ctx, obj); err != nil {
```

mnencia (approved):

gabriele-wolfox (approved):

jsilvela:
I'm not convinced that this PR is the right strategy:
Since it sets the ObservedGeneration to 0, this reconciles the DB object again once the cluster is hydrated.
This means re-execute.
And this goes against the model of the Database/Pub/Sub CRDs, where they don't get re-executed.

Example:
we create a DB with owner `foo`
we make changes in the DB, perhaps with migrations or by a db admin, and change the owner to `bar`
we hibernate, then rehydrate the cluster
the DB object re-executes and changes the owner back to `foo`

sxd (approved):

leonardoce (approved):

PR #6609, opened 2025-01-15
6599: /restart endpoint and additional trace logging to remediate "not-ready" pod issue

jmealo:
This PR adds a `/restart` endpoint to the instance manager and additional logging and error handling to help with #6599.

**Problem:**
**- Instance Manager:**
- On Port 8000 (over SSL/TLS): `/readyz`, `/pg/status`, `/update`  (and the `/restart` in this PR) log that they've sent headers and response, however, nothing is received on the socket by the client (from inside the container, over `localhost` -- as well as externally)
- On Port 9187 (over HTTP): `/metrics` no headers or body received by client (hangs indefinitely)
**- Instance Manager:** `/healthz` respond fine
- `pg_isready` runs fine
- The Postgres instance is accepting connections and working fine

**Observations:**
- Sometimes the `defer` statements do not run, for example, if you run:
`curl -X PUT -T /controller/manager -k https://localhost:8000/update -vvvvvv`
That'll force an in-place upgrade of an instance in good health, however, when the instance manager is in a bad state, it hangs indefinitely. If you Ctrl + C `curl` and try the command again, it'll complain that an upgrade is already in process. The code has a defer statement to unset that the instance is being upgraded, however, it never executes.
- This indicates that the request handles never complete, or there's a panic not being handled somewhere.

**What I've tried:**
- Connecting from inside and outside the container
- Leaving the connections open overnight
- Using HTTP 1.1 and HTTP/2
- ARM64 and AMD64 arch
- Release versions of CNPG 1.24.0, 1.25.0 and builds of this branch**

** This branch has only been tested on AMD64 because the Makefile doesn't support multi-arch builds right now.

I've added enough trace logging to help sort out what's going on.


pkg/management/postgres/webserver/webserver.go
```diff
@@ -118,12 +118,23 @@ func (ws *Webserver) Start(ctx context.Context) error {
 	return nil
 }
 
+func sendTextResponse(w http.ResponseWriter, statusCode int, message string) {
+	w.Header().Set("Content-Type", "text/plain")
+	w.WriteHeader(statusCode)
```
jmealo, line 123: I'm interested if explicitly sending the headers will help us here (as my debugging indicates we get nothing after the TLS handshake).

It's curious that we get `succeeding` logs for `/pg/status` despite the hang.

pkg/management/postgres/webserver/remote.go
```diff
@@ -170,8 +196,7 @@ func (ws *remoteWebserverEndpoints) pgStatus(w http.ResponseWriter, _ *http.Requ
 		return
 	}
 
-	w.Header().Set("Content-Type", "application/json")
-	_, _ = w.Write(js)
+	sendJSONResponseWithData(w, http.StatusOK, js)
```
jmealo, line 199: I'm hoping this will provide us some insight into getting no response, despite seeing the `"Instance status probe succeeding"` trace.

jmealo:
@sxd Does this look like a good start? What we have right now isn't sufficient to troubleshoot the issue.
If a real gopher could attach more error context to the context and make sure that gets logged, that'd be even better.

pkg/management/postgres/webserver/remote.go
```diff
@@ -132,19 +133,44 @@ func (ws *remoteWebserverEndpoints) isServerHealthy(w http.ResponseWriter, _ *ht
...
-	if err := ws.readinessChecker.IsServerReady(r.Context()); err != nil {
-		log.Debug("Readiness probe failing", "err", err.Error())
+	start := time.Now()
+	ctx := r.Context()
+	err := ws.readinessChecker.IsServerReady(ctx)
+	elapsed := time.Since(start)
+	deadline, hasDeadline := ctx.Deadline()
+
+	if err != nil {
+		switch {
```
jmealo, line 148: Is there anyway to eek more details out of these `context cancelled` errors? It didn't look like it... This is the best that I could come up with.

pkg/management/postgres/webserver/webserver.go
```diff
@@ -118,12 +118,26 @@ func (ws *Webserver) Start(ctx context.Context) error {
 	return nil
 }
 
+// sendTextResponse sends a generic text response
+//
+//nolint:unparam
```
jmealo, line 123: Linter complained that `statusCode` is always 200 and to remove the parameter. But it's not guaranteed to be so.

pkg/management/postgres/webserver/webserver.go
```diff
@@ -32,6 +32,10 @@ const (
 	DefaultReadTimeout = 20 * time.Second
 	// DefaultReadHeaderTimeout is the default value to be used by the webservers
 	DefaultReadHeaderTimeout = 3 * time.Second
+	// DefaultWriteTimeout is the default value to be used by the webservers
+	DefaultWriteTimeout = 60 * time.Second
```
jmealo, line 36: Need to check with team if we have any endpoints that aren't expected to return within 60 seconds during normal operation. @sxd ?

jmealo:
@sxd: This code works fine over 1.25.0: https://github.com/jmealo/cloudnative-pg/tree/fix-over-1.25.0-rc6 🤷

PR #6614, opened 2025-01-16
feat: support defining security contexts for pods and containers

### LINKED ISSUE #2821: add support for `securityContext` in postgres containers
Hello,

I've noticed that the `postgres` container of the pods created by the operator have a very basic `securityContext`:
```yaml
    securityContext:
      runAsUser: 1001
```
I saw there's a way to set the `seccompProfile` but I was hoping one can also set the full `securityContext` of the container, which doesn't seem to be the case. This is a problem when the restricted PSS is enforced.

This could become part of the `ClusterSpec` object - e.g. keeping the current value as defaults, but adding the option to override if required.

I don't know how much work this would represent, and I would be happy to help if I can - I would just need some guidance to get started.

Fred

x0ddf:
This commit introduces the ability to configure `securityContext` settings at both the pod and container level, enhancing control over runtime security policies and permissions.

Closes #2821


x0ddf:
@sxd could you please review and merge this one?

npgretz:
Thank you so much for this PR! I need this feature to be allowed to deploy CloudNative-PG at my org.

I also need the initDb jobs created as part of cluster.bootstrap to have a securityContext with runAsUser and runAsGroup. It looks like this PR does not enable that, correct?

cccsss01:
Very nice!  Can we get this added?

sxd:
@x0ddf hi!

I haven't had time to look into this with full time but the first things that comes to my eyes is, where is the documentation and explanation for the usage of this and also all the risks that can come from a bad configuration of a security context? Something like that should explain what kind of things we can expect if something is badly configured, for example, PostgreSQL not starting because the security context define a different user from the one on the image, or that was defined previously ergo you'll see permissions issue.

Also, you're touching the pgbouncer one, that should be handled using the PodTemplate not using the options from the cluster.

On the proposal we allow to change the Security Context even on running pods, should we allow that?

Just to finish a quick review, please, this is lacking some test and E2E tests.

Regards,

x0ddf:
@sxd  hi!
> 
> I haven't had time to look into this with full time but the first things that comes to my eyes is, where is the documentation and explanation for the usage of this and also all the risks that can come from a bad configuration of a security context? Something like that should explain what kind of things we can expect if something is badly configured, for example, PostgreSQL not starting because the security context define a different user from the one on the image, or that was defined previously ergo you'll see permissions issue.

 I do agree with the statement that misconfigured SecurityContext(SC) is a risk. Still, the lack of opportunity to configure it is a risk also and could be a blocker to adopting it in some environments.  For example, in the OpenShift/OKD environments, UID/GID could be randomized, and now you have to do tricks with AdmissionControllers/Kyverno and `Cluster` manifest simultaneously(just to set SC for the cnpg pods)! Also, at the moment, `spec.postgres[UID, GID]` is not propagated to the SC level, which is a big problem for me.

> 
> Also, you're touching the pgbouncer one, that should be handled using the PodTemplate not using the options from the cluster.
> 

If you look at my change around the pgbouncer, you'll see that I'm not doing anything new (the seccomp profile was consumed from the `Cluster` spec before my changes).
After a deeper code evaluation, you'll find these two helpers used to create the deployment for pooler/pgbouncer:
- [WithSecurityContext](https://github.com/cloudnative-pg/cloudnative-pg/blob/main/pkg/podspec/builder.go#L85)
- [WithContainerSecurityContext](https://github.com/cloudnative-pg/cloudnative-pg/blob/main/pkg/podspec/builder.go#L268)

It's already working as you want it to: if SC is defined in the `PodTemplate`, it will be utilized; otherwise, it will use the cnpg defaults.

sxd:
Hi @x0ddf 

> I do agree with the statement that misconfigured SecurityContext(SC) is a risk. Still, the lack of opportunity to configure it is a risk also and could be a blocker to adopting it in some environments. For example, in the OpenShift/OKD environments, [UID/GID could be randomized](https://www.redhat.com/en/blog/a-guide-to-openshift-and-uids), and now you have to do tricks with AdmissionControllers/Kyverno and `Cluster` manifest simultaneously(just to set SC for the cnpg pods)! Also, at the moment, `spec.postgres[UID, GID]` is not propagated to the SC level, which is a big problem for me.
> 
But the operator already work on OpenShift and OKD, by actually not adding anything of the (Pod)SecurityContext and let OpenShift define their own, on this case, the current behavior it's as it should. Can you elaborate more in the problem?

> > Also, you're touching the pgbouncer one, that should be handled using the PodTemplate not using the options from the cluster.
> 
> If you look at my change around the pgbouncer, you'll see that I'm not doing anything new (the seccomp profile was consumed from the `Cluster` spec before my changes). After a deeper code evaluation, you'll find these two helpers used to create the deployment for pooler/pgbouncer:
> 
> * [WithSecurityContext](https://github.com/cloudnative-pg/cloudnative-pg/blob/main/pkg/podspec/builder.go#L85)
> * [WithContainerSecurityContext](https://github.com/cloudnative-pg/cloudnative-pg/blob/main/pkg/podspec/builder.go#L268)
> 
> It's already working as you want it to: if SC is defined in the `PodTemplate`, it will be utilized; otherwise, it will use the cnpg defaults.

Yeah, shouldn't be the case I guess, specially because those two can be deployed in totally different environments.

@x0ddf in general, I'm in pro to apply this changes, but I think that we should change the API a little bit here, and that may break some changes, for example, having the PodSecurityContext and the SecurityContext elements in the Cluster definition will make the SeccompProfile a deprecated options, since we're already giving the possibility to modify the SecurityContext as is, so, why not allow the SeccompProfile too? it's a sub-object of the others two, so, you may want to have a different one per container and pod. With that being said, I think it's a possible change that we can move forward.

On the source code side, I can leave you some comments, since there's a few improvements.

@x0ddf are you going to continue working on this? Sorry for the late reply, but so many things require attention, that sometimes we need that some one pokes :)

Regards,

PS: There's still documentation missing =)

gbartolini:
Hi @x0ddf,

We are trying to determine if this pull request (PR) can be included in version 1.27.0. I agree with @sxd's comments regarding your proposed changes (which suggest a significant change to the API, but we can consider postponing it).

Could you please let us know if you can work on this feature in the coming weeks? If not, we will need to move it to version 1.28.

Thank you,  
Gabriele

gbartolini:
Waiting for @x0ddf to reply, I am moving this to 1.28 so that we can focus on 1.27.0-rc1.

x0ddf:
> Waiting for @x0ddf to reply, I am moving this to 1.28 so that we can focus on 1.27.0-rc1.

I'm good with moving to 1.28. I will work on this next week.

mnencia (approved):

armru (approved):

gbartolini (approved):

PR #6623, opened 2025-01-18
feat: startup and readiness probes for replicas

### LINKED ISSUE #6621: [Feature]: Startup and readiness probes for replicas

The current implementation of the startup probe uses the same endpoint as the liveness probe, which is `/healthz`. This endpoint relies on the [`pg_isready` command](https://www.postgresql.org/docs/current/app-pg-isready.html). The liveness probe reports success when `pg_isready` returns either 0 (indicating that the system is ready for connections) or 1 (indicating that the system is starting up, such as during crash recovery).

The readiness probe begins after this point. However, in some cases, such as with replicas, this implementation may pose a limitation.

We would like to differentiate the startup probe to define multiple waiting stages before the readiness probe kicks in:

- `pg_isready`: pg_isready returns 0 (connection is possible)
- `streaming`: in case of a streaming connected replica, the WAL receiver is up and `lag` is within a desired bound

The issue arose from a request by one of our customers at EDB to "tune" the instances participating in the synchronous replication quorum, with a focus on prioritising data durability. Following the shutdown of their only synchronous replica, the replica was automatically added to the `synchronous_standby_names` list as "ready" when it restarted. This caused write operations on the primary to halt until the replica became fully synchronous.

Our initial approach was to address this issue at the level of synchronous replication, pushing the problem to PostgreSQL's handling of replica readiness. However, after extensive discussions and scenario analysis between @leonardoce and me, we realised that modifying the readiness probe wasn't the optimal solution (better, wasn't the right place to solve it). Instead, the better approach would have been to focus on the startup probe, which operates exclusively during the pod's startup phase.

### Additional context

N/A

### Backport?

No

No

### Code of Conduct

leonardoce:
Extend the startup and readiness probes configured through the `.spec.probes.startup` and `.spec.probes.readiness` sections by adding two additional parameters:

- `type`: Defines the criteria for considering the probe successful. Accepted values include:
    - `pg_isready`: This setting marks the probe as successful when the `pg_isready` command exits with a status of `0`. This is the default for both primary instances and replicas.
    - `query`: This setting marks the probe as successful when a basic query is executed locally on the `postgres` database.
    - `streaming`: This setting marks the probe successful when the replica starts streaming from its source and meets the specified lag requirements (details below).
- `lag`: Specifies the maximum acceptable replication lag, measured in bytes (expressed using Kubernetes quantities). This parameter is applicable only when `type` is set to `streaming`. If the `lag` parameter is not specified, the replica is considered successfully started/ready as soon as it begins streaming.

Consequently, the liveness probe has been streamlined to verify solely that the instance manager is operational, without monitoring the underlying PostgreSQL instance.

Closes: #6621 

## Release Notes

**Improved Startup and Readiness Probes for Replicas**: Enhanced support for Kubernetes startup and readiness probes in PostgreSQL instances, providing greater control over replicas based on the streaming lag.


ardentperf:
IIUC, when replicas re-join they need to catch-up WAL - which has two effects:

1. clients can connect as soon as the readiness probe succeeds and during catch-up their queries may get results that lag more than usual.
2. the impetus for this PR: with `preferred` DataDurability (new feature in 1.25) all writes on the primary will completely hang during catch-up. the primary needs to write WAL synchronously and the latest WAL won't get acknowledged by the replica until it's caught up, leading to the hang on the primary.

+1 using a startup probe seems like a good idea

if someone is choosing `preferred` dataDurability then i don't think they would ever want the primary to hang. as a user, my expectation is that replicas (re)joining a cluster should be seamless. a primary database hanging for even a second can be catastrophic during peak on a large workload. this means a replica should never be considered ready as soon as it starts streaming, when dataDurability is `preferred`. furthermore, i suspect that users who choose `preferred` are specifically doing it because they want high availability.

my initial thought is that this should default to a very small number of bytes and might not need to be configurable at all, unless as a setting for debugging or troubleshooting.

this would result in a slightly longer delay for replicas becoming ready even with dataDurability=`required` however it also means we eliminate the period on startup of higher-than-usual lag, which doesn't seem like a bad idea to me.

*note: what i've written above is based on my understand but i have not had time to test or verify it, so there might be mistakes. i haven't looked yet, but it could also be interesting to check how patroni approaches this.*

gbartolini:
> IIUC, when replicas re-join they need to catch-up WAL - which has two effects:
> 
> 1. clients can connect as soon as the readiness probe succeeds and during catch-up their queries may get results that lag more than usual.
> 2. the impetus for this PR: with `preferred` DataDurability (new feature in 1.25) all writes on the primary will completely hang during catch-up. the primary needs to write WAL synchronously and the latest WAL won't get acknowledged by the replica until it's caught up, leading to the hang on the primary.

That's correct.

leonardoce:
e2e: https://github.com/EnterpriseDB/cloudnative-pg/actions/runs/12933848954

ardentperf:
I think Patroni addresses this with the `maximum_lag_on_failover` setting. sounds very similar to the `lag` setting in this PR. seeing that it's configurable in patroni, makes sense to have a similar configurable setting here.

should the name be changed to `maximum_lag` in this PR?

ardentperf:
Is this PR in patroni solving a similar problem?

https://github.com/patroni/patroni/pull/1786

ardentperf:
interesting - from this recent pgcon talk it sounds like actually patroni might even still have the pause

https://youtu.be/CWrFPPG5USA?feature=shared&t=1190

leonardoce:
> interesting - from this recent pgcon talk it sounds like actually patroni might even still have the pause [...]

Yes. I think there is no solution this problem but only compromises.
Having the cluster paused reduces the time needed for the replica to get in sync.
If the cluster is not paused the replica will need more time.
If may not even be able to get in sync again - depending on the workload of the primary.

ardentperf:
for the parameter, do you have thoughts on whether `maximum_lag` might be a better name than `lag`? or stick with the PR as it is?

leonardoce:
E2e: https://github.com/EnterpriseDB/cloudnative-pg/actions/runs/12945712758

Unfortunately, I need to run this E2e tests in the EDB Organization to spare some OSS workers.

leonardoce:
> for the parameter, do you have thoughts on whether `maximum_lag` might be a better name than `lag`? or stick with the PR as it is?

I like it. It is definitely clearer than just `lag`. What about `maximumLag`?
We're using camelCase unless when referring to some external tool/parameter, like `pg_isready`.

leonardoce:
E2e tests are green!!!

leonardoce:
E2e: https://github.com/EnterpriseDB/cloudnative-pg/actions/runs/12953656542

leonardoce:
I rebased it over the latest `main`. E2e tests are still good when run locally.

jmealo:
Would really love to have the `streaming` feature!

pkg/management/postgres/webserver/probes/streaming.go
```diff
@@ -0,0 +1,123 @@
...
+	ConfiguredLag uint64
+
+	// LatestEndTime is the time of last write-ahead log location reported to
+	// origin WAL sender
+	LatestEndTime string
+}
+
+func (e *ReplicaLaggingError) Error() string {
+	return fmt.Sprintf(
+		"streaming replica lagging; detectedLag=%v configuredLag=%v latestEndTime=%s",
```
jmealo, line 118: beautiful. when this gets cast to a string for printing it'll be helpful.

pkg/management/postgres/webserver/remote.go
```diff
@@ -193,32 +192,92 @@ func (ws *remoteWebserverEndpoints) cleanupStaleCollections(ctx context.Context)
...
+		)
+		return
+	}
+
+	// If `pg_rewind` is running, it means that the Pod is starting up.
 	// We need to report it healthy to avoid being killed by the kubelet.
-	// Same goes for instances with fencing on.
 	if ws.instance.PgRewindIsRunning || ws.instance.MightBeUnavailable() {
-		log.Trace("Liveness probe skipped")
+		log.Trace("startup probe skipped")
```
jmealo, line 217: **Quibble:** can we keep the trace log-line prefix the same case for folks grepping the logs? It's `Startup` on the following lines.

jmealo (approved):
:+1:  LGTM. If we could change `startup` to `Startup` on the one trace log line that'd be swell; but, not worth holding things up :)

gbartolini (approved):

armru (approved):

NiccoloFei (approved):

mnencia (approved):

PR #6659, opened 2025-01-24
fix(instance-manager): use pgdata content to discover PostgreSQL version

### LINKED ISSUE #6658: [Bug]: Configuration logic in the instance manager uses heuristic instead of on-disk version

### Contact Details

marco.nenciarini@enterprisedb.com

### Version

trunk (main)

1.32

YAML manifest

### What happened?

The configuration generator in the instance manager attempts to guess the PostgreSQL version using a heuristic based on the image name that is present in the status. This is inaccurate, as the instance manager has first-hand access to the data directory.

The instance manager should inspect the data directory on disk to determine the PostgreSQL major version, ensuring accurate configuration generation regardless of the cluster status's content.

### Code of Conduct

mnencia:
Use a precise check of the data directory to determine the PostgreSQL major version during PostgreSQL configuration generation.

Closes #6658

### DIFF
--- a/internal/cmd/manager/instance/run/lifecycle/run.go
+++ b/internal/cmd/manager/instance/run/lifecycle/run.go
@@ -150,7 +150,7 @@ func configureInstancePermissions(ctx context.Context, instance *postgres.Instan
 		return nil
 	}
 
-	majorVersion, err := postgresutils.GetPgdataVersion(instance.PgData)
+	pgVersion, err := postgresutils.GetPgdataVersion(instance.PgData)
 	if err != nil {
 		return fmt.Errorf("while getting major version: %w", err)
 	}
@@ -180,7 +180,7 @@ func configureInstancePermissions(ctx context.Context, instance *postgres.Instan
 		return err
 	}
 
-	err = configurePgRewindPrivileges(majorVersion, hasSuperuser, tx)
+	err = configurePgRewindPrivileges(pgVersion, hasSuperuser, tx)
 	if err != nil {
 		_ = tx.Rollback()
 		return err
--- a/internal/cmd/manager/instance/run/lifecycle/run.go
+++ b/internal/cmd/manager/instance/run/lifecycle/run.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"sync"
 
+	"github.com/blang/semver"
 	"github.com/cloudnative-pg/machinery/pkg/log"
 	"github.com/jackc/pgx/v5"
 
@@ -149,7 +150,7 @@ func configureInstancePermissions(ctx context.Context, instance *postgres.Instan
 		return nil
 	}
 
-	majorVersion, err := postgresutils.GetMajorVersion(instance.PgData)
+	majorVersion, err := postgresutils.GetPgdataVersion(instance.PgData)
 	if err != nil {
 		return fmt.Errorf("while getting major version: %w", err)
 	}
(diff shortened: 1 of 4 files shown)


internal/cmd/manager/instance/run/lifecycle/run.go
```diff
@@ -228,10 +228,10 @@ func configureStreamingReplicaUser(tx *sql.Tx) (bool, error) {
 }
 
 // configurePgRewindPrivileges ensures that the StreamingReplicationUser has enough rights to execute pg_rewind
-func configurePgRewindPrivileges(majorVersion *semver.Version, hasSuperuser bool, tx *sql.Tx) error {
+func configurePgRewindPrivileges(majorVersion semver.Version, hasSuperuser bool, tx *sql.Tx) error {
 	// We need the superuser bit for the streaming-replication user since pg_rewind in PostgreSQL <= 10
 	// will require it.
-	if majorVersion != nil && majorVersion.Major <= 10 {
```
armru: I made the semver as required, given that we cannot make a decision otherwise.

armru (approved):

pkg/management/postgres/utils/version.go
```diff
@@ -38,25 +38,30 @@ func GetPgVersion(db *sql.DB) (*semver.Version, error) {
...
+		Minor: (versionInt / 100) % 100,
+		Patch: versionInt % 100,
 	}, nil
 }
 
-// GetMajorVersion read the PG_VERSION file in the data directory
+// GetPgdataVersion read the PG_VERSION file in the data directory
 // returning the major version of the database
-func GetMajorVersion(pgData string) (int, error) {
+func GetPgdataVersion(pgData string) (semver.Version, error) {
```
jsilvela, line 55: I'm not sure I agree with the new signature.
The PG_VERSION contrains only the major number, nothing else.
I don't think this is the right place to return a semver.Version

internal/cmd/manager/instance/run/lifecycle/run.go
```diff
@@ -227,10 +228,10 @@ func configureStreamingReplicaUser(tx *sql.Tx) (bool, error) {
 }
 
 // configurePgRewindPrivileges ensures that the StreamingReplicationUser has enough rights to execute pg_rewind
-func configurePgRewindPrivileges(majorVersion int, hasSuperuser bool, tx *sql.Tx) error {
+func configurePgRewindPrivileges(majorVersion semver.Version, hasSuperuser bool, tx *sql.Tx) error {
```
jsilvela, line 231: and here we change the signature to consume a semver but only require the major. Not an improvement IMO, except perhaps we can use `unit64` better, as would be done in Semver

pkg/management/postgres/configuration.go
```diff
@@ -65,11 +67,12 @@ func (instance *Instance) RefreshConfigurationFilesFromCluster(
 	cluster *apiv1.Cluster,
 	preserveUserSettings bool,
 ) (bool, error) {
-	postgresConfiguration, sha256, err := createPostgresqlConfiguration(cluster, preserveUserSettings)
+	majorVersion, err := postgresutils.GetPgdataVersion(instance.PgData)
```
jsilvela, line 70: and here again you get the full semver but only consume the major

jsilvela (commented):
I'm not seeing the point of returning a Semver when we read PG_VERSION.
Isn't it better to just return and consume a `uint64`?

jsilvela:
Clarified my objections about the signature with Marco.

jsilvela (approved):

pkg/management/postgres/utils/version.go
mnencia: The goal is to eliminate confusion and assign a clear, specific type to the value. We've already made similar adjustments for the int/uint version in the past, with consensus that using `semver` helps prevent misinterpretation when the source of the values isn't immediately apparent.

Issue #6653, opened 2025-01-24
[Bug]: Hibernate not keeping replica PVCs


chris-gooch:
### Is there an existing issue already for this bug?

- [x] I have searched for an existing issue, and could not find anything. I believe this is a new bug.

### I have read the troubleshooting guide

- [x] I have read the troubleshooting guide and I think this is a new bug.

### I am running a supported version of CloudNativePG

- [x] I have read the troubleshooting guide and I think this is a new bug.

### Contact Details

chris.gooch@hotmail.co.uk

### Version

1.25 (latest patch)

### What version of Kubernetes are you using?

1.30

### What is your Kubernetes environment?

Cloud: Amazon EKS

### How did you install the operator?

YAML manifest

### What happened?

As per documentation https://cloudnative-pg.io/documentation/current/declarative_hibernation/#hibernation 
"Replica PVCs will be kept in addition to the primary's PVC"

I have a 3 instance cluster and ran
`kubectl cnpg hibernate <clustername>`

However, only the primary PVCs were retained and both the replica PVCs have been deleted. This is not expected.

### Cluster resource

```shell

```

### Relevant log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

chris-gooch:
Raised PR #6662

mnencia:
The declarative hibernation feature is distinct from the imperative hibernation process (described in detail at https://cloudnative-pg.io/documentation/current/kubectl-plugin/#cluster-hibernation). These two features serve different purposes and are not intended to be interchangeable.

With that clarification, I’ll proceed to close this issue.

chris-gooch:
Hi @mnencia

Why should imperative and declarative hibernate behaviour be different?
As a user, I want the same behaviour regardless how initiated. Preserving the replica PVCs makes it much faster to recover when invoked as well.

mnencia:
The key difference between the two types of hibernation lies in their handling of resources. The plugin-based hibernation does not preserve the Cluster resource and retains only the primary instance PVCs, whereas the declarative approach keeps the Cluster and all associated resources intact, removing only the pods.

While the design's validity is open for discussion, it is certainly not a bug.

PR #6663, opened 2025-01-25
Keep replica PVCs during hibernation

chris-gooch:
Fix for #6653 to keep replica PVCs during hibernation


mnencia:
I’m closing this PR because the original issue was based on a misunderstanding of the imperative hibernation feature and is, therefore, invalid.

chris-gooch:
Why should imperative and declarative hibernate behaviour be different? 
As a user, I want the same behaviour either way.

PR #6664, opened 2025-01-25
feat: declarative offline in-place major upgrades of PostgreSQL

### LINKED ISSUE #4682: Declarative support for PostgreSQL major in-place upgrades
Currently, Major upgrades of PostgreSQL can be done in an imperative way, by relying on the native logical replication capabilities and the `cnpg` plugin. See: https://www.gabrielebartolini.it/articles/2024/03/cloudnativepg-recipe-5-how-to-migrate-your-postgresql-database-in-kubernetes-with-~0-downtime-from-anywhere/

However, we should provide a way to support major upgrades just by changing the `major` field like in the following example:

```yaml
kind: Cluster
metadata:
  name: cluster-example
spec:
  instances: 3
  imageCatalogRef:
    apiGroup: postgresql.cnpg.io
    kind: ImageCatalog
    name: postgresql
    major: 16
  storage:
    size: 1Gi
```

To:

```yaml
kind: Cluster
metadata:
  name: cluster-example
spec:
  instances: 3
  imageCatalogRef:
    apiGroup: postgresql.cnpg.io
    kind: ImageCatalog
    name: postgresql
    major: 17
  storage:
    size: 1Gi
```

Alternative approaches are valid too, for example through CRDs that evoke imperative names (like upgrades) and that are useful too as they prepare the underlying PVCs for the new major.

mnencia:
This PR introduces the ability to upgrade PostgreSQL clusters to a higher major version, enabling a fully declarative approach for major version upgrades within CloudNativePG.

**IMPORTANT:** While the upgrade job area still requires refactoring, we’ve decided to merge this patch into the main branch and release version 1.26.0-rc1 to gather early community feedback at KubeCon Europe 2025 in London. The refactoring will likely require a second release candidate before the final 1.26.0 release.

With this update, users can specify a new image with a higher major version using any supported method. The operator tracks the previous image version in `.status.majorVersionUpgradeFromImage`, which also serves as an indicator that a major version upgrade is in progress.

The upgrade process starts by deleting all pods and jobs to ensure the data directory is inaccessible. A dedicated upgrade job is created for the primary instance, using an `initContainer` to copy the old PostgreSQL installation and running `pg_upgrade --link` for an efficient migration. Once the upgrade completes, all non-primary instances are removed, and new standbys are cloned to restore the desired number of instances.

Closes #4682 

# Release Notes

**Declarative Offline In-Place Major Upgrades of PostgreSQL**: Introduced support for offline in-place major upgrades when a new operand container image with a higher PostgreSQL major version is applied to a cluster. All cluster pods are shut down during the upgrade to ensure data consistency. A new job is created to validate upgrade conditions, run `pg_upgrade`, and set up new directories for `PGDATA`, WAL files, and tablespaces as needed. Once the upgrade is complete, replicas are re-created. Failed upgrades can be rolled back declaratively. (#6664)


pschichtel:
Just to be sure I understand this correctly: This is an offline migration as in: The cluster gets taken down completely, the primary upgraded in-place, started up again and new replicas are cloned from that, right?

mnencia:
> Just to be sure I understand this correctly: This is an offline migration as in: The cluster gets taken down completely, the primary upgraded in-place, started up again and new replicas are cloned from that, right?

Yes exactly. However, it will be speedy, as no data will be copied around.

internal/cmd/manager/instance/upgrade/prepare/cmd.go
```diff
@@ -0,0 +1,133 @@
...
+		}
+
+		contextLogger.Info("Creating the directory", "directory", destDir)
+		err = os.MkdirAll(destDir, 0o750)
+		if err != nil {
+			return fmt.Errorf("failed to create the directory: %w", err)
+		}
+
+		contextLogger.Info("Copying the files", "source", sourceDir, "destination", destDir)
+		err = exec.Command("cp", "-a", sourceDir+"/.", destDir).Run() //nolint:gosec
```
saolof, line 118: This could possibly be sped up with the `--reflink=auto` flag depending on the filesystem used in the pvc. Reflinking whenever possible to make copies instant is the default behaviour of cp in coreutils 9.x, but not on 8.x which is still widely used.

pschichtel:
just FYI, not in any way a request or anything: In my homelab I'm using [zalando's operator](https://github.com/zalando/postgres-operator), which is based around patroni and their spilo images. While there are a bunch of things I don't like about that operator, it's in-place major upgrades work quite well. I tested that on the weekend and it seems it keeps the replicas around while upgrading the master, so read queries could still be answered. Might be worth looking at what they are doing.

internal/cmd/manager/instance/upgrade/execute/cmd.go
```diff
@@ -0,0 +1,426 @@
...
+
+	clusterObjectKey := ctrl.ObjectKey{Name: instance.GetClusterName(), Namespace: instance.GetNamespaceName()}
+	if err = management.WaitForGetClusterWithClient(ctx, client, clusterObjectKey); err != nil {
+		return err
+	}
+
+	// Create a fake reconciler just to download the secrets and
+	// the cluster definition
+	metricExporter := metricserver.NewExporter(instance)
+	reconciler := controller.NewInstanceReconciler(instance, client, metricExporter)
```
leonardoce, line 130: In this code, we’re creating the instance manager reconciler just to call functions that download the secrets and make sure the WAL volume is set up correctly. The problem is, the first function doesn’t return any error, so we’re not really sure if the secrets were downloaded or not.

A similar issue pops up with the "initdb" call and the one that creates the metrics server. The "initdb" call should be encapsulated, and the metrics server creation shouldn't even be here in the first place.

Before merging this, we need to clean up the implementation to make it more maintainable and avoid repeating code, kind of like what we did with the subreconciler in the operator. This will take some time, and if we go down that path, we might not have enough time to hit the 1.26 release as planned.

I’d love to hear what the other maintainers think. What’s your opinion?

mnencia: I suggest proceeding incrementally to ensure both functionality and maintainability. We should first validate the feature mechanics in their current form to confirm they work as expected. At the same time, we should complete the implementation of the patch's E2E tests to demonstrate how the feature operates, providing a reliable safety net before making structural changes. Once this validation is complete and the E2E tests are in place, we can refactor the code to improve its structure and maintainability while preserving functionality.

For the 1.26.0 release, if we don’t have enough time to complete the full work before 1.26.0rc1, we should prioritize merging the patch with the E2E tests in place, leaving the refactor for later. This will allow users to try it out and provide feedback. Based on that input, we can complete the refactoring and issue a new RC release.

leonardoce: I agree, but only partially. I won't call that refactor - there's no tests in place, I call it clean up.
It is not the first time we postponed the clean up phase, and the duplicated code is still there needing a clean up.
One example is that `RefreshSecrets` call, another one is the `NewExporter` call.
In this patch we add new duplication for the `initdb` command.

If this patch reaches a release we give value and help people immediately - and that's our goal - but we won't be able to continue helping later.

gbartolini: > For the 1.26.0 release, if we don’t have enough time to complete the full work before 1.26.0rc1, we should prioritize merging the patch with the E2E tests in place, leaving the refactor for later. This will allow users to try it out and provide feedback. Based on that input, we can complete the refactoring and issue a new RC release.

This confirms that there will definitely be an RC2. Is this what you mean, @mnencia? Help me understand. Thanks.

mnencia: > > For the 1.26.0 release, if we don’t have enough time to complete the full work before 1.26.0rc1, we should prioritize merging the patch with the E2E tests in place, leaving the refactor for later. This will allow users to try it out and provide feedback. Based on that input, we can complete the refactoring and issue a new RC release.
> 
> This confirms that there will definitely be an RC2. Is this what you mean, @mnencia? Help me understand. Thanks.

Yes, in that scenario, an RC2 would be necessary. If we merge the patch with E2E tests in place but defer the refactor, we will need an additional RC release once the cleanup is complete to ensure a properly maintained and polished implementation.

internal/cmd/manager/instance/upgrade/prepare/cmd.go
```diff
@@ -0,0 +1,123 @@
...
+// copyPostgresInstallation replicates the PostgreSQL installation to the specified destination directory
+// for use by the pg_upgrade command as the old binary directory.
+//
+// Steps performed:
+// 1. Removes the existing destination directory if it exists.
+// 2. Retrieves the PostgreSQL binary, library, and shared directories using pg_config.
+// 3. Creates the corresponding directories in the destination path.
+// 4. Copies the contents of the PostgreSQL directories to the destination.
+// 5. Creates a bindir.txt file in the destination directory with the path to the binary directory.
+func copyPostgresInstallation(ctx context.Context, pgConfig string, dest string) error {
```
gbartolini, line 69: @mnencia isn't this identical to the `execute` implementation?

docs/src/postgres_upgrades.md
```diff
@@ -0,0 +1,189 @@
...
+
+```sh
+kubectl cnpg psql cluster-example -- -qAt -c 'SELECT version()'
+```
+
+This will return output similar to:
+
+```console
+PostgreSQL 16.x ...
+```
```
gbartolini, line 162: I need to fix this.

mnencia:
As maintainers, we’ve decided to merge this patch despite the need for further refactoring. The core functionality is in place and working with passing end-to-end tests, and we believe it’s important to make it available now so users can begin testing and provide feedback.

Speaking as the EDB team lead, I confirm our commitment to refactoring this code before the final release of version 1.26.0. Depending on how the testing of RC1 progresses, we are also open to producing a second release candidate (RC2) if needed.

armru:
> As maintainers, we’ve decided to merge this patch despite the need for further refactoring. The core functionality is in place and working with passing end-to-end tests, and we believe it’s important to make it available now so users can begin testing and provide feedback.
> 
> Speaking as the EDB team lead, I confirm our commitment to refactoring this code before the final release of version 1.26.0. Depending on how the testing of RC1 progresses, we are also open to producing a second release candidate (RC2) if needed.

In my opinion, given there is a clear commitment in the refactoring, this PR can be merged

armru (approved):

gbartolini:
I am fine with merging this commit, provided it is clear that refactoring is a blocker for version 1.26.0 and that allowing early access for testing offers greater benefits.

gbartolini (approved):

fcanovai (approved):

Issue #6069, opened 2024-11-12
[Bug]: initdb doesn't wait for secret


evilhamsterman:
### Is there an existing issue already for this bug?

- [x] I have searched for an existing issue, and could not find anything. I believe this is a new bug.

### I have read the troubleshooting guide

- [x] I have read the troubleshooting guide and I think this is a new bug.

### I am running a supported version of CloudNativePG

- [x] I have read the troubleshooting guide and I think this is a new bug.

### Contact Details

dmills@qumulo.com

### Version

1.24.0

### What version of Kubernetes are you using?

1.30

### What is your Kubernetes environment?

Self-managed: k3s

### How did you install the operator?

Helm

### What happened?

The initdb option to create a new DB doesn't appear to wait for the secret to exist before starting. Then later if the secret is created the owner is not created unless you separately use a role.

We use external-secrets to save secrets in a secure vault, but if we deploy the external-secret and the cluster at the same time they both start working even though external-secrets may have not created the Kubernetes secret. So you end up with a DB that doesn't have a owner can the user cannot connect. If you recreate the cluster once the secret exists everything proceeds normally

### Cluster resource

```shell
---
apiVersion: postgresql.cnpg.io/v1
kind: Cluster
metadata:
  name: db
  namespace: dv
spec:
  instances: 2
  storage:
    size: 10Gi
  bootstrap:
    initdb:
      database: db
      owner: db
      secret:
        name: db-secret

---
apiVersion: external-secrets.io/v1beta1
kind: ExternalSecret
metadata:
  name: db-secret
  namespace: db
spec:
  refreshInterval: 1h
  secretStoreRef:
    name: secretstore
    kind: ClusterSecretStore
  target:
    name: db-secret
    template:
      type: kubernetes.io/basic-auth
      data:
        username: "{{ .login }}"
        password: "{{ .password }}"
        host: "{{ .host }}"
  dataFrom:
    - extract:
        key: db-secret
```

### Relevant log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

armru:
It would be helpful to have the full logs of both the operator and the primary instance (while there is no 'owner')

firstdorsal:
[...]

PR #6668, opened 2025-01-26
fix(instance-controller): error on absence of user secrets

pkoutsovasilis:
Fix for https://github.com/cloudnative-pg/cloudnative-pg/issues/6069 that makes the existence of user-related secrets required to proceed


gbartolini:
We believe that this change is quite invasive and deserves more discussion. Moving this to the next patch release.

PR #6675, opened 2025-01-28
fix: reject if no size is used in shared_buffers parameter

### LINKED ISSUE #6657: [Bug]: inconsistent default calculation for shared_buffers GUCs

### Contact Details

neel.patel@enterprisedb.com

### Version

1.25 (latest patch)

1.32

YAML manifest

### What happened?

Webhook validation is failing when user provide `request.limits` along with shared_buffers GUCs ( provide value without unit )

Validation is  failing with `Memory request is lower than PostgreSQL `shared_buffers`

If user provide `shared_buffers` value as "256", it convert to 256 MB and which will fail against the `request.memory` validation.

**Expected result:**

If user do not provide any unit, it should not convert to MB as default.

### Cluster resource

```shell
kind: Cluster
metadata:
  name: test-shared-buffers
  namespace: default
spec:
  bootstrap:
    initdb:
      database: app
      encoding: UTF8
      localeCType: C
      localeCollate: C
      owner: app
  imageName: postgres:16.6
  instances: 1
  postgresql:
    parameters:
      shared_buffers: "540"
  resources:
    limits:
      cpu: 500m
      memory: 512Mi
    requests:
      cpu: 500m
      memory: 512Mi
  storage:
    resizeInUseVolumes: true
    size: 1Gi
```

### Relevant log output

```shell
❯ k apply -f test-shared_buffers.yaml
The Cluster "test-shared-buffers" is invalid: spec.resources.requests.memory: Invalid value: "512Mi": Memory request is lower than PostgreSQL `shared_buffers` value
```

### Code of Conduct

pchovelon:
Fix #6657 

It is a proposal to use 8kB as the default size if no unit is used in the shared_buffers parameters (in yaml).


internal/webhook/v1/cluster_webhook.go
```diff
@@ -1146,9 +1146,10 @@ func validateWalSizeConfiguration(
 // into kubernetes resource.Quantity values
 // Ref: Numeric with Unit @ https://www.postgresql.org/docs/current/config-setting.html#CONFIG-SETTING-NAMES-VALUES
 func parsePostgresQuantityValue(value string) (resource.Quantity, error) {
-	// If no suffix, default is MB
-	if _, err := strconv.Atoi(value); err == nil {
-		value += "MB"
+	// If no suffix, take it as 8kB (default PostgreSQL's value)
+	if size, err := strconv.Atoi(value); err == nil {
+		size = size * 8
+		value = strconv.Itoa(size) + "kB"
```
neel5481, line 1152: I don't think, we can assume 8 KB. If you see, same function is getting used for `min_wal_size` & `max_wal_size` where unit is different in `pg_settings` table.

e.g.

```
-[ RECORD 1 ]---+-------------------------------------------------------------
name            | shared_buffers
setting         | 16384
unit            | 8kB
category        | Resource Usage / Memory
short_desc      | Sets the number of shared memory buffers used by the server.
extra_desc      | 
context         | postmaster
vartype         | integer
source          | configuration file
min_val         | 16
max_val         | 1073741823
enumvals        | 
boot_val        | 16384
reset_val       | 16384
sourcefile      | /var/lib/postgresql/data/pgdata/postgresql.conf
sourceline      | 129
pending_restart | f


name            | max_wal_size
setting         | 1024
unit            | MB
category        | Write-Ahead Log / Checkpoints
short_desc      | Sets the WAL size that triggers a checkpoint.
extra_desc      | 
context         | sighup
vartype         | integer
source          | configuration file
min_val         | 2
max_val         | 2147483647
enumvals        | 
boot_val        | 1024
reset_val       | 1024
sourcefile      | /var/lib/postgresql/data/pgdata/postgresql.conf
sourceline      | 257
pending_restart | f


```

pchovelon: Nothing to say :) You are totally right ! Thanks

pchovelon: Other idea :
Add a new arg called `unit`, that must be in kB
`func parsePostgresQuantityValue(value string, unit int) (resource.Quantity, error)`

And when calling the function, we have to give the correct value of the unit depending on the PostgreSQL parameter : 

Examples : 
`sharedBuffers, err := parsePostgresQuantityValue(rawSharedBuffer, 8) ...
or 
`maxWalSizeValue, err := parsePostgresQuantityValue(maxWalSize, 1024) ...` 

The only think is that the unit will be hardcoded.

@neel5481 What do you think ?

GabriFedi97:
The `shared_buffer` unit is expressed in `block_size` which is typically and by default `8kB`. However, the `block_size` depends on the `BLCKSZ` value, which is a build time parameter, so it's not really a fixed value, but it is evaluated by [Postgres when the GUC is registered](https://github.com/postgres/postgres/blob/f6ff75f79689d2d0c5bb241435d7fc6a63273223/src/backend/utils/misc/guc_tables.c#L2322). 
I don't think we can retrieve the `block_size` value from the webhook though, which would be needed for performing the correct validation of a value without the unit.
The safest approach might be rejecting values without the unit from the configuration I think.
Otherwise we could assume the `block_size` to be always `8kB` and perform the validation with this assumption, but it wouldn't be totally correct IMO.

Ref:
- `shared_buffers`:  https://www.postgresql.org/docs/17/runtime-config-resource.html#RUNTIME-CONFIG-RESOURCE-MEMORY
- `block_size`: https://www.postgresql.org/docs/current/runtime-config-preset.html#RUNTIME-CONFIG-PRESET

pchovelon:
Here is a new try.

I also change the link to now send to the CloudNativePG's documentation instead of PostgreSQL one. :ok_hand:

sxd:
@pchovelon there's a linter error, can you please fix it so we can trigger a test round?

pchovelon:
Mistake ...

PR #6695, opened 2025-01-30
feat: add kubernetes `client-gen` tool support

### LINKED ISSUE #6585: [Feature]: Add support for `client-gen`

Due to a single missing field (`SchemeGroupVersion`), the `api/v1` API is not compatible with the k8s [`client-gen`](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-api-machinery/generating-clientset.md) tool for generating golang CRD clients. [The project already has this field](https://github.com/cloudnative-pg/cloudnative-pg/blob/main/api/v1/groupversion_info.go#L55), but it uses the wrong name.

Add a `SchemeGroupVersion = GroupVersion` field to the `groupversion` file for compatibility with `client-gen`.

Patch the files locally every time they're generated in every project:
`find $PATH_TO_GENERATED_CLIENT -type f -name '*.go' -exec sed 's/SchemeGroupVersion/GroupVersion/' {} \;`

### Additional context

_No response_

### Backport?

Yes

Yes

### Code of Conduct

solidDoWant:
This patch introduces support for the Kubernetes `client-gen` tool, enabling the automated generation of Go clients for all custom resources defined by cloudnative-pg CRDs.

Closes #6585 

## Release notes

Added support for Kubernetes `client-gen`, allowing the automated generation of Go clients for all custom resources defined by cloudnative-pg CRDs.

### DIFF
--- a/api/v1/groupversion_info.go
+++ b/api/v1/groupversion_info.go
@@ -14,9 +14,6 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-// Package v1 contains API Schema definitions for the postgresql v1 API group
-// +kubebuilder:object:generate=true
-// +groupName=postgresql.cnpg.io
 package v1
 
 import (
@@ -51,14 +48,11 @@ const (
 )
 
 var (
-	// GroupVersion is group version used to register these objects
-	GroupVersion = schema.GroupVersion{Group: "postgresql.cnpg.io", Version: "v1"}
-
-	// SchemeGroupVersion is the group version used to register these objects.
-	SchemeGroupVersion = GroupVersion
+	// SchemeGroupVersion is group version used to register these objects
+	SchemeGroupVersion = schema.GroupVersion{Group: "postgresql.cnpg.io", Version: "v1"}
 
 	// SchemeBuilder is used to add go types to the GroupVersionKind scheme
-	SchemeBuilder = &scheme.Builder{GroupVersion: GroupVersion}
+	SchemeBuilder = &scheme.Builder{GroupVersion: SchemeGroupVersion}
 
 	// AddToScheme adds the types in this group-version to the given scheme.
 	AddToScheme = SchemeBuilder.AddToScheme
--- a/api/v1/groupversion_info.go
+++ b/api/v1/groupversion_info.go
@@ -54,6 +54,9 @@ var (
 	// GroupVersion is group version used to register these objects
 	GroupVersion = schema.GroupVersion{Group: "postgresql.cnpg.io", Version: "v1"}
 
+	// SchemeGroupVersion is the group version used to register these objects.
+	SchemeGroupVersion = GroupVersion
+
 	// SchemeBuilder is used to add go types to the GroupVersionKind scheme
 	SchemeBuilder = &scheme.Builder{GroupVersion: GroupVersion}
 
--- a/api/v1/backup_funcs.go
+++ b/api/v1/backup_funcs.go
@@ -236,8 +236,8 @@ func (backup *Backup) GetVolumeSnapshotConfiguration(
 // By setting the GVK, we ensure that components such as the plugins have enough metadata to typecheck the object.
 func (backup *Backup) EnsureGVKIsPresent() {
 	backup.SetGroupVersionKind(schema.GroupVersionKind{
-		Group:   GroupVersion.Group,
-		Version: GroupVersion.Version,
+		Group:   SchemeGroupVersion.Group,
+		Version: SchemeGroupVersion.Version,
 		Kind:    BackupKind,
 	})
 }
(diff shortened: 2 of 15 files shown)


sxd:
Hi @solidDoWant 

Can you create some kind of explanation how to reproduce what you're talking about? also probably this may deserve at least some documentation showing how users can use this for their own purposes.

Regards!

solidDoWant:
Sure thing @sxd! Here's an how I am generating a [client-go](https://github.com/kubernetes/client-go) based client for interacting with CNPG via a Go tool I'm writing.

Note that the git references can be updated to point to a release tag, another commit, another repo, etc.

---

The [kubernetes/code-generator](https://github.com/kubernetes/code-generator/) project can be used to create a Go client for controllers who have Go structs defined for CRDs. This requires `+genclient` comments on the CRD structs, which [CNPG provides](https://github.com/cloudnative-pg/cloudnative-pg/blob/main/api/v1/cluster_types.go#L2313).

Here's an example of how to generate a CNPG Go client for your project:
```shell
#!/bin/bash

# 1. Download the codegen script. Note that using the script outside of a Go module directory requires a change that has been merged, but not released yet.
curl -fsSL -o /tmp/kube_codegen.sh https://raw.githubusercontent.com/kubernetes/code-generator/e7e5741e3a3ed0043276619b3f8026526dbed675/kube_codegen.sh


# 2. Clone the CNPG repository
git clone --quiet --branch v1.25.0 --single-branch https://github.com/cloudnative-pg/cloudnative-pg.git /tmp/cnpg-repo

# 3. Create a Go project for use with the client (optional, this can be replaced with an existing project)
mkdir /tmp/cnpg-client-example
pushd /tmp/cnpg-client-example
go mod init github.com/example/cnpg-client-example

# 4. Generate the client
# The generator may produce a `grep: /tmp/cnpg-client-example/generated/clientset: No such file or directory` error, which can be safely ignored.
pushd /tmp/cnpg-repo
KUBE_CODEGEN_TAG=latest . /tmp/kube_codegen.sh
kube::codegen::gen_client --output-dir /tmp/cnpg-client-example/generated --output-pkg github.com/example/cnpg-client-example/generated --boilerplate /dev/null .

# 4.1 FOR THIS PR ONLY, because the change hasn't merged yet, the generated files need to be patched to include the effects of this PR. This should not be included in any docs, as it is unnecessary with these changes.
find /tmp/cnpg-client-example/generated -type f -name '*.go' -exec sed -i 's/SchemeGroupVersion/GroupVersion/' {} \;

# 5. Use the client (optional)
# This example program will list CNPG clusters in the "default" namespace, using a kubeconfig file at `~/.kube/config`.
popd
go get "k8s.io/client-go" # Add deps
go mod tidy
cat << EOF > main.go
package main

import (
	"context"
	"log"
	"os"
	"path/filepath"

	"github.com/example/cnpg-client-example/generated/clientset/versioned"
	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
	"k8s.io/client-go/tools/clientcmd"
)

// Adjust these as needed
var kubeconfigPath = filepath.Join(os.Getenv("HOME"), ".kube", "config")
var namespace = "default"

func main() {
	// Simplified but fairly standard way to load a kubeconfig
	config, err := clientcmd.BuildConfigFromFlags("", kubeconfigPath)
	if err != nil {
		log.Fatalf("unable to load kubeconfig from %s: %v", kubeconfigPath, err)
	}

	cnpgClient := versioned.NewForConfigOrDie(config)
	clusters, err := cnpgClient.PostgresqlV1().Clusters(namespace).List(context.Background(), metav1.ListOptions{})
	if err != nil {
		log.Fatalf("unable to list clusters: %v", err)
	}

	for _, cluster := range clusters.Items {
		log.Printf("Cluster: %s", cluster.Name)
	}
}
EOF
```

Output:
```console
vscode ➜ /tmp/cnpg-client-example $ go run main.go
2025/01/31 20:52:21 Cluster: testing
```
---

It would be really convenient if the CNPG project managed a generated client, rather than having every consumer maintain (an essentially identical) one. This is what most other projects that I worked with do. Examples:
* cert-manager: https://github.com/cert-manager/cert-manager/tree/master/pkg/client/clientset/versioned
* external-snapshotter: https://github.com/kubernetes-csi/external-snapshotter/tree/master/client/clientset/versioned
* kubernetes: https://github.com/kubernetes/client-go/tree/master/kubernetes

I would be open to adding this feature myself in another PR if CNPG would accept this work. Depending on how CNPG would want this implemented, it could be as easy as [adding a single makefile target](https://github.com/cert-manager/cert-manager/blob/3d7ec7baad42c41ad34fc9cd9d5a2b2e833f0a7a/make/ci.mk#L48).

mnencia:
For reference, here is the issue in Kubebuilder where they are proposing the change to the default scaffolding: https://github.com/kubernetes-sigs/kubebuilder/issues/3692

api/v1/groupversion_info.go
```diff
@@ -14,9 +14,6 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-// Package v1 contains API Schema definitions for the postgresql v1 API group
-// +kubebuilder:object:generate=true
-// +groupName=postgresql.cnpg.io
```
mnencia, line 19: I removed this as they are already in the doc.go file

mnencia (approved):

armru (approved):

PR #6701, opened 2025-01-30
feat(plugin): add `--ttl` option to `pgbench` command

### LINKED ISSUE #4374: [Feature]: Give the possibility to automatically clean pgbench jobs - ttlSecondsAfterFinished option

pgbench jobs are not cleaned by default. It would be great to have to possibility to clean them automatically after a certain time.

The **ttlSecondsAfterFinished** option of a Kubernetes Job could be implemented. See [https://kubernetes.io/docs/concepts/workloads/controllers/job/#clean-up-finished-jobs-automatically]( https://kubernetes.io/docs/concepts/workloads/controllers/job/#clean-up-finished-jobs-automatically)

It could be configured in the YAML file, in a job or pgbench / section . For example for one hour TTL : 

```
kind: Cluster
metadata:
  name: perf-cluster
spec:
  imageName: ghcr.io/cloudnative-pg/postgresql:16.2
  instances: 1
  pgbench:
    ttlSecondsAfterFinished: '3600'
...
```

By default, we can keep it disabled.

none 

### Additional context

_No response_

### Backport?

No

Yes

### Code of Conduct

pchovelon:
This patch introduces the `--ttl` option to the `pgbench` plugin command,
allowing users to configure automatic cleanup of completed jobs after a
specified duration. By default, jobs do not expire, preserving existing
behavior.

Closes #4374

## Release notes

```
Add an optional `--ttl` flag to the `pgbench` plugin command, enabling
automatic deletion of completed jobs after a user-specified duration. By
default, jobs do not expire.
```

### DIFF
--- a/internal/cmd/plugin/pgbench/cmd.go
+++ b/internal/cmd/plugin/pgbench/cmd.go
@@ -67,8 +67,8 @@ func NewCmd() *cobra.Command {
 	pgBenchCmd.Flags().Int32Var(
 		&run.ttlSecondsAfterFinished,
 		"ttl",
-		86400,
-		"Time to live of the pgbench job. Defaults to: one day",
+		0,
+		"Time to live of the pgbench job. Defaults to no TTL.",
 	)
 
 	pgBenchCmd.Flags().BoolVar(
--- a/internal/cmd/plugin/pgbench/cmd.go
+++ b/internal/cmd/plugin/pgbench/cmd.go
@@ -64,6 +64,13 @@ func NewCmd() *cobra.Command {
 		"The name of the database that will be used by pgbench. Defaults to: app",
 	)
 
+	pgBenchCmd.Flags().Int32Var(
+		&run.ttlSecondsAfterFinished,
+		"ttl",
+		86400,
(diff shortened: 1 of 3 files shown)


jsilvela:
Why have a default? Seems a bit "aggressive" as the behavior out of the box.
I.e. it makes sense as a default for people who'll be doing a lot of pgbench-ing, but not for the general user.
Also, if you're set on having this as a default behavior, would be good to document in `benchmarking.md` and/or `kubectl-plugin.md`

leonardoce:
Hi @pchovelon! I just removed the default value to avoid changing the meaning of `kubectl cnpg pgbench` when you don't specify any option.
Do you agree with that?

leonardoce (approved):

pchovelon:
> Hi @pchovelon! I just removed the default value to avoid changing the meaning of `kubectl cnpg pgbench` when you don't specify any option. Do you agree with that?

Yes ! Thanks.

jsilvela (approved):

mnencia (approved):

mnencia:
/ok-to-merge E2E not needed for this change (the pgbench command is not covered in E2E)

PR #6749, opened 2025-02-03
docs: clarify the bootstrap section

jsilvela:


docs/src/bootstrap.md
```diff
@@ -718,6 +730,9 @@ and diverge from the source.
 For this reason, it is advised to stop all write operations to the source database
 before migrating to the target database.
 
+Note that this limitation applies only if the target cluster is not defined as
+a replica cluster.
+
```
gbartolini, line 735: What do you mean exactly here @jsilvela?

jsilvela: I think this section could lead to confusion. Seems to be saying that once the target cluster backup is completed, it diverges from the source.
This is not an inherent limitation of the `pg_basebackup` method, as we can make a replica cluster using it, which would not have that limiation.
So that was my goal here.

gbartolini (approved):

mnencia (approved):

PR #6794, opened 2025-02-07
feat: 5299 check for nodes cordoned by karpenter

MarcusDunn:
Superseded by #6928

### DIFF
--- a/internal/controller/cluster_predicates.go
+++ b/internal/controller/cluster_predicates.go
@@ -74,7 +74,13 @@ var (
 		UpdateFunc: func(e event.UpdateEvent) bool {
 			oldNode, oldOk := e.ObjectOld.(*corev1.Node)
 			newNode, newOk := e.ObjectNew.(*corev1.Node)
-			return oldOk && newOk && oldNode.Spec.Unschedulable != newNode.Spec.Unschedulable
+			if !oldOk || !newOk {
+				return false
+			}
+
+			oldDrained := isNodeBeingDrained(*oldNode)
+			newDrained := isNodeBeingDrained(*newNode)
+			return oldDrained != newDrained
 		},
 		CreateFunc: func(_ event.CreateEvent) bool {
 			return false
--- a/internal/controller/replicas.go
+++ b/internal/controller/replicas.go
@@ -166,6 +166,22 @@ func (r *ClusterReconciler) reconcileTargetPrimaryForNonReplicaCluster(
 	return mostAdvancedInstance.Pod.Name, r.setPrimaryInstance(ctx, cluster, mostAdvancedInstance.Pod.Name)
 }
 
+var drainTaints = map[string]struct{}{
+	"karpenter.sh/disruption":     {},
+	corev1.TaintNodeUnschedulable: {},
+}
+
+// isNodeBeingDrained checks if a node is currently being drained.
+// Copied from https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/7bacf2d36f397bd098b3388403e8759c480be7e5/cmd/hooks/prestop.go#L91
+func isNodeBeingDrained(node corev1.Node) bool {
+	for _, taint := range node.Spec.Taints {
+		if _, isDrainTaint := drainTaints[taint.Key]; isDrainTaint {
+			return true
+		}
+	}
+	return false
+}
+
 // isNodeUnschedulable checks whether a node is set to unschedulable
 func (r *ClusterReconciler) isNodeUnschedulable(ctx context.Context, nodeName string) (bool, error) {
 	var node corev1.Node
@@ -173,7 +189,9 @@ func (r *ClusterReconciler) isNodeUnschedulable(ctx context.Context, nodeName st
 	if err != nil {
 		return false, err
 	}
-	return node.Spec.Unschedulable, nil
+	isBeingDrained := isNodeBeingDrained(node)
+
+	return node.Spec.Unschedulable || isBeingDrained, nil
 }
 
 // Pick the next primary on a schedulable node, if the current is running on an unschedulable one,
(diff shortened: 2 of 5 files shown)


chris-gooch:
Can this be reviewed please @sxd @gbartolini ? Based on all the issue thumbs up it’s quite a popular issue

MarcusDunn:
I'm hesitantly marking this as ready for review as my test passes.

sxd:
hello @MarcusDunn 

Can you fix the DCO issue first? we can trigger a E2E tests but without that fixed will not work =)

Regards!

MarcusDunn:
@sxd done!

chris-gooch:
Hi @MarcusDunn 

Can I make a quick recommendation please? Annoyingly different versions of karpenter use different taints and so does K8 Cluster Autoscaler.
Could you make the code block like below to handle all the possible taints?

```
const clusterAutoscalerTaint = "ToBeDeletedByClusterAutoscaler"
const v1KarpenterTaint = "karpenter.sh/disrupted"
const v1beta1KarpenterTaint = "karpenter.sh/disruption"

// drainTaints includes taints used by K8s or autoscalers that signify node draining or pod eviction.
var drainTaints = map[string]struct{}{
	v1.TaintNodeUnschedulable: {}, // Kubernetes common eviction taint (kubectl drain)
	clusterAutoscalerTaint:    {},
	v1KarpenterTaint:          {},
	v1beta1KarpenterTaint:     {},
}
```

sxd:
Hello @MarcusDunn 

Looks like there could be more than one custom taint out there, why not make this a configuration option so you don't need a new version of the operator to support a new taint ?

Regards,

chris-gooch:
Hi @sxd 

I would personally advise against making it a configuration as there are too many variables and most users will not know what it should even be. Me included until investigating this issue. 

Best to copy what the AWS EBS CSI Driver and other controllers do to cover all known taint possibilities out of the box. 

Thanks
Chris

chris-gooch:
Hi @sxd 

I would personally advise against making it a configuration as there are too many variables and most users will not know what it should even be. Me included until investigating this issue. 

Best to copy what the AWS EBS CSI Driver and other controllers do to cover all known taint possibilities out of the box. 

Thanks
Chris

chris-gooch:
I am facing issue with the taints not being detected by the reconciliation loop. Any pointers/clues to what could be missing @sxd @gbartolini ?

MarcusDunn:
> Hello @MarcusDunn
> 
> Looks like there could be more than one custom taint out there, why not make this a configuration option so you don't need a new version of the operator to support a new taint ?
> 
> Regards,

Do you think a default of well-known taints would be acceptable?

gbartolini:
First, thanks for suggesting this. My opinion is the same as Jonathan's. We should avoid hard-coding them. But I think we can get to a compromise if:

- we use an operator configuration option that everyone can then customize (see https://cloudnative-pg.io/documentation/current/operator_conf/#available-options)
- we provide default values for that option that include well-known taints, including Karpenter

How does it sound?

chris-gooch:
Hi @gbartolini sorry I have to disagree. 
Other very popular kubernetes-sig repositories have implemented a fixed drain taints logic for example:

- https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/526b830b93b0e7f78c1808ced4482f75440abccc/pkg/backend/endpoint_utils.go#L17
- https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/dba6d7090f33e5ccd1452214691e0d1433d7881a/cmd/hooks/prestop.go#L47
- https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/c1ff2f8593ac2167ab3af8d299b1e44384cd97ad/pkg/azurediskplugin/hooks/prestop.go#L50

The majority of CloudNativePG users will already be using these.

I have also raised a new PR, which works locally #6920. Please can you review this?

Thanks,
Chris

gbartolini:
Thanks for the pointers, very useful. Do you happen to have more examples of projects?

chris-gooch:
Hi @gbartolini 

Sure, here are a few more example projects I can find using fixed drain taints:

https://github.com/Dynatrace/dynatrace-operator/blob/f31881e9a4782d4a6a5baa2f6b6004c863fe3e08/pkg/controllers/nodes/config.go#L15
https://github.com/oracle/oci-native-ingress-controller/blob/9987b182d5ab7d065b826febd80bfe464cd5740b/pkg/util/testutil.go#L29
https://github.com/openshift/aws-load-balancer-controller/blob/970ea56f31a3e660efbbed1deb50bf5ed9262c92/pkg/backend/endpoint_utils.go#L16
https://github.com/openshift/aws-ebs-csi-driver/blob/ff230b96ba8c2e1818b7da8edd7df96289228d00/cmd/hooks/prestop.go#L46

leonardoce:
Is there a single truth containing the list of these taints?
It sounds like every project has its own list, and sometimes they are not consistent.

chris-gooch:
Hi @leonardoce 

As far as I know, there are two main autoscalers used:

Native Autoscaler - https://github.com/kubernetes/autoscaler
AWS Karpenter (which has version specific taints)- https://aws.amazon.com/blogs/aws/introducing-karpenter-an-open-source-high-performance-kubernetes-cluster-autoscaler/

My PR #6920 covers all these.

chris-gooch:
I cannot contribute to @MarcusDunn code so I have created my own and tested in our environment using karpenter. 

Please ignore this PR as mine works with the reconcile loops, which this does not.

MarcusDunn:
I'll close this one then. Thanks @chris-gooch

MarcusDunn:
@chris-gooch 

Just for my own sake, I'm not seeing the difference between the two PRs in how they handle triggering reconciliation. What was this PR missing that was causing the loop not to be triggered?

chris-gooch:
Hi @MarcusDunn 

For the reconcile loop to trigger when nodes are tainted it needed to update this:

https://github.com/cloudnative-pg/cloudnative-pg/blob/4e2a2e79a4201ce2218e888eb76b41f6d99782b3/internal/controller/cluster_controller.go#L1367

PR #6796, opened 2025-02-07
fix: reject if no size is used in shared_buffers (2nd PR)

### LINKED ISSUE #6657: [Bug]: inconsistent default calculation for shared_buffers GUCs

pchovelon:
New PR after mistake. Fix #6657 
Following this closed one : https://github.com/cloudnative-pg/cloudnative-pg/pull/6675.

(So sorry, for this second PR ...)


pchovelon:
Trying to get all tests OK.
As we reject when no unit is used, I added `MB` in the tests where values haven't got unit.
Does it seems correct to you, the way I change those tests ?

armru:
This method would render valid cluster YAML suddenly invalid; I don't think we can pursue it.

We should take a more gentle approach by first introducing a warning and giving users time to migrate. Afterward, we can introduce the `error`.

pchovelon:
Seems a smarter approach !

PR #6807, opened 2025-02-10
 Fixes #6750: PVC naming is unconventional

### LINKED ISSUE #6750: [Bug]: PVC naming is unusual

### Contact Details

sebas@dbyond.nl

### Version

1.25 (latest patch)

other (unsupported)

Other

OLM

### What happened?

If we create a CNPG cluster, we get 2 PVC's for every pod.
The WAL pvc is named [podname]-wal, which is consistent with other common deployment types (such as StatefulSet's).
The data pvc however is named same as the pod, which is unusual.
This in combination with a vsphere-csi bug makes our vCenter been flooded with API calls bringing us to a DoS.

Although this is a vsphere-csi issue primarily, I think we can do better (more in line with common deployment patterns) by naming the pvc to [podname]-data.
I am offering to submit a PR. We could add enable/disable with a config flag, and could even use conversions to keep the old pattern if unset in cluster CR.

**note** I pre-discussed this in slack channel...

### Cluster resource

```shell

```

### Relevant log output

```shell

```

### Code of Conduct

sebasmannem:
With this PR we make the naming of PVC's configurable through the configmap.
With a Configmap as follows:

```
kind: ConfigMap
apiVersion: v1
metadata:
  name: cnpg-controller-manager-config
  namespace: cnpg-system
data:
  DATA_VOLUME_SUFFIX: '-data2'
  WAL_ARCHIVE_VOLUME_SUFFIX: ‘-wal2’
```

the PVC's will be named {pod-name}-data2 and {pod-name}-wal2 accordingly.
Without these values set in teh configmap the operator stays with the original naming:
(e.a. {pod-name} for data and {pod-name}-wal for WAL).
As I went trhough the code I noticed that I needed to change a lot of hardcoded -wal settings, which is fixed in the first commit.


sebasmannem:
If @gbartolini or someone else of the dev team could at least approve the direction, we can go proceed with what we currently have and resolve this customer issue internally ahead of a new release. We could start now.

sebasmannem:
By the way, I don;t think we need to backport this one.
I implemented it as a feature, and I would therefore expect this to be part of the next release, not being backported.

devotional-phoenix-97:
Nice...

benoitschipper:
Looking good @sebasmannem ! LGTM!

gbartolini:
@sebasmannem Thanks for taking the time to write a patch for CloudNativePG.

Here are a couple of points to consider:

- The naming convention is not unusual. We have followed a clear and consistent approach, applied to both WAL and tablespaces. Specifically, there is a storage section for the main volume and a `walStorage` section for WALs.
- I personally struggle to see the benefits of this patch beyond your specific and, so far, isolated issue, especially given the potential disruption it could cause to existing deployments (we need an E2E test for this). The PVC name should not be a concern when its identity is guaranteed within the namespace.
- This PR introduces additional complexity requiring ongoing maintenance without a clear justification.

Unless proven otherwise, the current behaviour of CloudNativePG’s custom controller is perfectly valid, as the PVC is correctly identified by the triplet: type, namespace, and name.

However, I would like to hear the opinion of other maintainers too.

sebasmannem:
@gbartolini ,

Thanks for your comment. I agree with a lot.

So, about the 'unusual' keyword:
I agree that unusual is not the exact correct word. Unconventional would be better. A short summery:

- Deployments have randomness in podnaming and not in PVC naming, so they are always different
- Statefulsets have the Pod name suffixed with the Volume name, so they are always different
- I checked most other PostgreSQL operators and they all use StatefulSets as their PostgreSQL deployment resource
- I checked some PostgreSQL helm charts and the one I could find use Statefulsets
- Minio tenants are based on StatefulSets
- Strimzi Clusters are based on StatefulSets
- I didn't check anything else, but I think CNPG probably covers about 100% of the cases where pods and PVC's are named the same.

Although I totally agree with you, that it should technically work, let's also agree that this is unconventional. To me it compares to compiling PostgreSQL with blocksize 16k.
You can, but it is unconventional, and if issues in this area arise, you will probably be the one facing them.
It is ill advised to run as such.

This is also where the benefits come into play.
- By allowing for pvc names to differ from pod names you can be more conventional in the way PVC naming is implemented. At least you allow the community options to dodge if they are hit by similar issues.
- Also, I think that more people run CNPG with vSphere CSI, and they probably all have this issue. Then again, they probably don't run our scale, and as such probably don't suffer as much as we do.
- Finally, there might be other situations (and future degradations) where systems beyond CNPG don't work well with this naming scheme.

Focussing on resolving the issue of our customer, I see the following options:
1: We could work together to improve the solution. Like I can add an e2e test if that helps. I can also work on an option to auto migrate to a conventional storage naming end solution.
2: Maybe we could alternatively focus on helping our customer facing this issue otherwise. 
Maybe EDB would be willing to support us with the altered CNPG operator until VMWare has fixed the issue on their end. This would require 2 storage migrations, which is cumbersome, but doable. I would be happy to discuss internally.

Looking forward to hearing the opinion of other maintainers too.

gbartolini:
> Unconventional would be better.

I still don't agree. :)

As you mentioned, most operators rely on StatefulSets. However, when this project began at 2ndQuadrant in 2019, we chose a different approach: building a custom controller that directly managed persistent volumes. We also adopted a clear and intuitive naming convention, based on the Kubernetes principle that any object can be uniquely identified by its type, namespace, and name.

To illustrate this, let's examine the [default `cluster-example` manifest](https://cloudnative-pg.io/documentation/current/samples/cluster-example.yaml). The cluster is named `cluster-example`, and the primary section for specifying the data volume is called `storage`. This results in the creation of three Persistent Volume Claims (PVCs): `cluster-example-1`, `cluster-example-2`, and `cluster-example-3`. These PVCs belong to the cluster, with a sequence number appended as part of our naming convention.  

When we introduced WAL volumes, we added a new section called `walStorage`. To differentiate these volumes, we appended `-wal` to the PVC names, such as `cluster-example-1-wal`.  

Later, when we introduced tablespaces, we defined a new stanza named `tablespaces`, accommodating the open-ended and arbitrary number of possible tablespaces. Our convention was to append `-tbs-%`, where `%` represents the tablespace name (e.g., `cluster-example-1-tbs-tmp1`).  

> * Strimzi Clusters are based on StatefulSets

Strimzi [dropped stateful sets in 0.35.0](https://github.com/strimzi/strimzi-kafka-operator/releases/tag/0.35.0). However, I am not familiar with Strimzi.

> * I didn't check anything else, but I think CNPG probably covers about 100% of the cases where pods and PVC's are named the same.

Thanks for your input. At the cost of sounding repetitive, I see no formal issue with this naming convention. CNPG covers 100% of active PostgreSQL operators that directly manage PVCs. This approach aligns perfectly with the fundamental Kubernetes principle of resource identity, making it a logical and consistent design choice.

> * Also, I think that more people run CNPG with vSphere CSI, and they probably all have this issue. Then again, they probably don't run our scale, and as such probably don't suffer as much as we do.

If this is the case, we should work to fix the issue upstream.

> * Finally, there might be other situations (and future degradations) where systems beyond CNPG don't work well with this naming scheme.

Why do you say that? What makes you think this might be the case?

> Focussing on resolving the issue of our customer, I see the following options: 1: We could work together to improve the solution. Like I can add an e2e test if that helps. I can also work on an option to auto migrate to a conventional storage naming end solution. 2: Maybe we could alternatively focus on helping our customer facing this issue otherwise. Maybe EDB would be willing to support us with the altered CNPG operator until VMWare has fixed the issue on their end. This would require 2 storage migrations, which is cumbersome, but doable. I would be happy to discuss internally.

Although I understand that EDB is the creator and main contributor, CloudNativePG is a vendor-neutral project. As maintainers, our primary responsibility is to ensure its long-term stability and vendor neutrality. Our decisions must prioritize the project's broader interests and maintain backward compatibility for users.

If EDB wishes to pursue this independently through their fork, I certainly won't stand in the way. However, my position as a maintainer for this open-source project remains unchanged: we should avoid introducing changes that are purely cosmetic, unnecessarily increase complexity, and overlook the impact on existing deployments.

That said, if we find clear evidence that CNPG’s naming convention violates Kubernetes principles or requirements, I will be the first to treat it as a bug and work on a solution that benefits all users, with proper handling for existing deployments.

Alternatively, if there is strong consensus on adopting a suffix like `-pgdata` for PVCs, this is something we could thoughtfully consider as part of a [future v2 API release](https://github.com/cloudnative-pg/cloudnative-pg/milestone/11).

Looking forward to other maintainers' opinion.

leonardoce:
Hi Sebastiaan! Thank you for contributing.

I believe problems need to be fixed in the correct layer of the stack.
If this is a problem with the vSphere CSI driver, it should be fixed there.
Is this confirmed? Is this happening on every vShere CSI driver installation, or need/can be configured?
Will Broadcom fix it?
Can the vSphere CSI driver be prevented from synchronizing CNPG Pods by annotating them?
That would be straightforward and won't require changing the PVC logic, which is already difficult to maintain.

About the patch, what would happen to the existing clusters after enabling this new parameter?
In my tests, I'm getting new blank PVCs with the new naming convention for existing clusters, too.
The clusters I had are now marked as "Waiting for the instances to become active".

sxd:
Hi,

I don't think this is the right path to go, as I already mentioned, I don't think it's a good, probably as was mentioned, we can think further on this matter for v2, but for now, this is a VMWare CSI issue, they should fix it.

Regards,

sebasmannem:
@gbartolini :
- I appreciate you sharing the history behind this. Thanks.
- Strimzi: I fired one up, and indeed they create pods and PVC's themselves too.
  I noticed [here](https://github.com/strimzi/strimzi-kafka-operator/blob/main/documentation/modules/configuring/ref-list-of-kafka-cluster-resources.adoc) that they prefix their PVC's, So yeah they don't use StatefulSets, but no they also don't use the same naming scheme for pods and PVC's
- I agree 'We' should work to fix this upstream'. We filed requests. But I still don't see that fix it anytime soon
- Future degradations: We ran into this once, so we will hit it again. That is not you, it probably is the community running CNPG. And then it will be the fault of something beyond CNPG codebase again. There is only one way to precursor it, and that is to create resources similar to how most other operators do, and that is not with anming pods and PVC's the same. It just isn't.
- I will check with EDB and our customer, and together we will find a way to get beyond this. Hopefully EDB is willing to help. If not, we will have to come up with something ourselves. It is what it is...

@leonardoce :
- I agree that the issue should be resolved with vSphere. We shall see when our customer will have a fix. I was hoping to deliver a resolution rather sooner than later. But it is what it is.
- Existing clusters start up and are reported healthy (at least in my case). The new PVC's are created, but dangling, and the operator mentions messages that they are not bound yet. I resolved this by migrating storage the way we did before (migrating from RDB to vsphere-csi): by removing PVC's and pods one by one. Operator creates a new set's up as a standby and they follow new convention. I could work towards maybe reusing PV's bound to newly named PVC's, but the more we automate, the messier it gets, I think. I could also add docs for storage migration.

@sxd :
- thanks for your comments.
- I was thinking about doing this with a Conversion WebHook, but I didn't know you might have plans to migrate to v2. If you want to we can talk about that approach too. I agree, it would be much cleaner.

sebasmannem:
We are working with EDB to see if we can get support on this.
Furthermore I will work on seemless migrations, so that we end up in a nicer state if we use this feature.
I will keep this PR open as draft, so I can communicate here.

Feel free to track my progress, discuss any opinions, ask, etc.
Let's use this PR for any open discussion on this front.

sebasmannem:
I have just (manually) succesfully changed PVC names without replacing the PV.
Looking into adding that to the code. That would mean that changing this parameter would be nearly seemless.
Except for the obvious: pvc names would be changed.

sebasmannem:
With these latest commit I think I have overcome all issues as mentioned in the PR comments.

I have added unittesting
I have added automatic PVC renaming
As on support from EDB, AFAIK, we are in the final stage of making this happen.
Would be awesome if this would still be merged, but if not, at least my customer can resolve storage issues they have run into sofar.

sebasmannem:
@gbartolini I have added my latest changes, and I have changed this PR from Draft to Open.

PR #6857, opened 2025-02-13
fix(backup): improve instance-manager backup mode resiliency

### LINKED ISSUE #6761: [Bug]: If 3 volume snapshots fail in a row, the instance manager stops functioning (connection pool exhausted)

### Contact Details

jeffm@gmail.com

### Version

1.25 (latest patch)

1.29

Cloud: Azure AKS

YAML manifest

### What happened?

When 3 scheduled volume snapshots fail in a row. The super user connection pool becomes exhausted in the instance manger. This causes all endpoints except `/healthz` to fail. This causes a near total failure of CNPG.

**Impact:**
- All scheduled backups start to fail (due to endpoints hanging)
- You run out of WAL space
- Metrics are no longer collected
- Hibernate fails
- `cnpg status` hangs
- All HTTP endpoints except /healthz hang
- Instance manager can no longer be updated in-place

Depending on when you catch it, the real reason for the volume snapshot failure will be buried by `context cancelled: deadline exceeded` (because only the first 2-3 attempts can start a backup, the other ones hang getting the cluster status).

This PR has the beginnings of a fix, but there's more to be done:
https://github.com/jmealo/cloudnative-pg/pull/1

# Big shout outs to the Gophers in the Go Slack:
### Cluster resource

```shell
Not relevant
```

### Relevant log output

```shell
postgres=# SELECT pid, age(clock_timestamp(), query_start), usename, query 
FROM pg_stat_activity 
WHERE query != '<IDLE>' AND query NOT ILIKE '%pg_stat_activity%' 
ORDER BY query_start desc;
 pid  |       age       | usename  |                      query                       
------+-----------------+----------+--------------------------------------------------
   26 |                 |          | 
   27 |                 |          | 
   28 |                 |          | 
  204 |                 |          | 
 3886 | 00:03:55.430776 | postgres | SELECT pg_backup_start(label => $1, fast => $2);
 2518 | 01:04:35.711006 | postgres | SELECT pg_backup_start(label => $1, fast => $2);
 1112 | 02:06:57.669102 | postgres | SELECT pg_backup_start(label => $1, fast => $2);
(7 rows)

postgres=# 
```

### Code of Conduct

armru:
This is a fix on the instance manager side to avoid connection leaks and sequentially connection saturation.

Partially Closes #6761

### DIFF
--- a/internal/controller/backup_controller.go
+++ b/internal/controller/backup_controller.go
@@ -555,7 +555,14 @@ func updateClusterWithSnapshotsBackupTimes(
 
 // isErrorRetryable detects is an error is retryable or not
 func isErrorRetryable(err error) bool {
-	return apierrs.IsServerTimeout(err) || apierrs.IsConflict(err) || apierrs.IsInternalError(err)
+	retryableApiErrs := apierrs.IsServerTimeout(err) || apierrs.IsConflict(err) || apierrs.IsInternalError(err)
+	if retryableApiErrs {
+		return true
+	}
+
+	// check if it contains a wrapped ErrRetryable error
+	var retryableErr *volumesnapshot.ErrRetryable
+	return errors.As(err, &retryableErr)
 }
 
 // getBackupTargetPod returns the pod that should run the backup according to the current
--- a/pkg/management/postgres/instance.go
+++ b/pkg/management/postgres/instance.go
@@ -752,19 +752,22 @@ func (instance *Instance) GetPgVersion() (semver.Version, error) {
 func (instance *Instance) ConnectionPool() *pool.ConnectionPool {
 	const applicationName = "cnpg-instance-manager"
 	if instance.pool == nil {
-		socketDir := GetSocketDir()
-		dsn := fmt.Sprintf(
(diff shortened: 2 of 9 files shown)


jmealo:
@armru If E2E tests pass here, I'll be happy to test this in my cluster. If the E2E tests fail, could we please test my branch:
https://github.com/cloudnative-pg/cloudnative-pg/pull/6860

armru:
Hello @jmealo,

regarding 

```
Improved Upgrade Process

Removed instance status check from upgrade endpoint
Enables successful in-place upgrades even during connection pool exhaustion
Enables support for triggering in-place restart via curl PUT with current instance manager binary even when the super user pool is exhausted
```

Those improvements, in my opinion, should be part of a different PR.


Regarding the backup, a more structural approach would be better. 
- The force flag, for example doesn't make a lot of sense
- Multiple POST and PUT can still happen concurrently (unintended scenarios)


But I'm happy to hear the other maintainers' opinions on that.

armru:
Hello @jmealo, what would be the best way to reproduce the issue in an E2E test environment, in your opinion?

armru:
refer to #6879

PR #6873, opened 2025-02-18
chore(github): mark files in `config/crd/bases` as generated

yardenshoham:
This lets GitHub know it should hide the diff of these files by default like it does for `api/v1/zz_generated.deepcopy.go`.

More info: https://docs.github.com/en/repositories/working-with-files/managing-files/customizing-how-changed-files-appear-on-github

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+config/crd/bases/** linguist-generated=true


gbartolini:
Hi @yardenshoham, thanks for your contribution.

gbartolini (approved):

leonardoce (dismissed):

sxd:
Hello,

My concern on `auto-hide` or `auto-collapse` is that if the changes in this file specifically is that, when the changes are way too big, are collapsed by default, but when is not, it actually show us useful information when updating the controller-gen and controller-tools, so for updates is really week to see that what we updated was just the version and nothing else, if we hide we will have to click file by file what we updated.

I think it's actually useful to be able to quickly check on the interface what changed when those files are touched cc @gbartolini @leonardoce 

Regards,

leonardoce:
On second thought, I agree with @sxd.

That folder contains our CRDs... a single line adding a "required" field could make us accidentally breaking the API compatibility. I know we could spot that my looking at the source code, too, but it's better to be safe than sorry.

PR #6879, opened 2025-02-19
fix: prevent leaking connection when snapshot backup fail

### LINKED ISSUE #6761: [Bug]: If 3 volume snapshots fail in a row, the instance manager stops functioning (connection pool exhausted)

leonardoce:
The instance manager allocates a PostgreSQL connection every time a
volume snapshot backup need to be taken and then runs
`pg_backup_start` on it.  That connection will stay up until the backup
is done.

If the snapshotting process fails, the backup will be marked as failed
but the connection won't be closed.

This patch ensures that a successive backup is able to proceed correctly
by closing the previous connection.

A successive commit will allow the operator to close that stale commit
even when there's no successive backup.

Fix: #6761 

# Release notes

```
Allow the instance manager to close a stale PostgreSQL connection left by a previous
failed volume snapshot backup.
```

### DIFF
--- a/pkg/management/postgres/webserver/backup_connection.go
+++ b/pkg/management/postgres/webserver/backup_connection.go
@@ -85,6 +85,13 @@ func (bc *backupConnection) closeConnection(backupName string) error {
 	return bc.conn.Close()
 }
 
+func (bc *backupConnection) forceCloseConnection() error {
+	bc.sync.Lock()
+	defer bc.sync.Unlock()
+
+	return bc.conn.Close()
+}
+
 func (bc *backupConnection) executeWithLock(backupName string, cb func() error) {
 	bc.sync.Lock()
 	defer bc.sync.Unlock()
--- a/pkg/management/postgres/webserver/remote.go
+++ b/pkg/management/postgres/webserver/remote.go
@@ -277,7 +277,7 @@ func (ws *remoteWebserverEndpoints) backup(w http.ResponseWriter, req *http.Requ
 				sendUnprocessableEntityJSONResponse(w, "PROCESS_ALREADY_RUNNING", "")
 				return
 			}
-			if err := ws.currentBackup.closeConnection(p.BackupName); err != nil {
+			if err := ws.currentBackup.forceCloseConnection(); err != nil {
 				if !errors.Is(err, sql.ErrConnDone) {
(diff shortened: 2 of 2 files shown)


leonardoce:
I ran locally the Volume Snapshot E2e test suite and they're fine. We still need a specific E2e for this to prevent it from happening again.

armru (approved):

mnencia (approved):

PR #6883, opened 2025-02-19
6599: Use runtime instead of instance status checks for arch during u…

jmealo:
Replaced with:
https://github.com/cloudnative-pg/cloudnative-pg/pull/6884

To attempt to remove the "Merging is blocked" error on this PR.


jmealo:
image
Cannot get this to go away ^

PR #6916, opened 2025-02-21
test(e2e,snapshot): disable wait for archive for primary

### LINKED ISSUE #6917: [Bug]: E2e tests of volume snapshot backup for primary randomly fail

### Contact Details

### Version

1.25 (latest patch)

1.32

YAML manifest

### What happened?

The E2e test of volume snapshot backup from the primary randomly fails because the timeout is set to 5 minutes, the same as `archive_timeout`.
We call `pg_backup_stop(waitForArchive => true)` but we don't wait for enough time to have a new WAL.

### Cluster resource

```shell

```

### Relevant log output

```shell

```

### Code of Conduct

leonardoce:
Disables the waitForArchive flag in the E2e tests of Volume Snapshot backups, where the
primary is explicitely chosen.

Since the target cluster have no workload, no WAL segment will be generated and this makes
the E2e fail after 5 minutes of timeout. Note: archive_timeout is set to 5 minutes too.

Fix: #6917 

# Release notes

```
NONE
```

### DIFF
--- a/tests/e2e/fixtures/volume_snapshot/declarative-backup-on-primary.yaml.template
+++ b/tests/e2e/fixtures/volume_snapshot/declarative-backup-on-primary.yaml.template
@@ -7,3 +7,4 @@ spec:
   target: primary
   cluster:
     name: cluster-declarative-backup
+  waitForArchive: false


leonardoce:
This fixes the bug. The E2e test that failed is fixed by https://github.com/cloudnative-pg/cloudnative-pg/pull/6890

armru (approved):

PR #6920, opened 2025-02-21
feat: 5299 check for nodes drain tainted by well known autoscalers

### LINKED ISSUE #5299: [Feature]: Check for nodes cordoned by karpenter

At the moment, when a primary runs on EKS nodes managed by karpenter, you will have to manually failover if a node is replaced or removed. This interrupts the automatic provisioning/deprovisioning of nodes.

Kubernetes uses the "node.kubernetes.io/unschedulable" taint and the "spec.unschedulable" field on a node to mark it as cordoned. The operator checks nodes for this information.
If a primary database is running on a node which was cordoned the operator will promote a secondary on another node.
This allows to evict the cordoned node without blocking the process due to the PDB of the primary.

Karpenter does not rely on "node.kubernetes.io/unschedulable" taint and/or the "spec.unschedulable" field.
Next to other things, the project basically needs to know that karpenter was the one that initiated the tainting.

Karpenter taints nodes with: "karpenter.sh/disruption:NoSchedule"
https://karpenter.sh/docs/concepts/disruption/#disruption-controller

It would be nice if the operator could also check for an additional/alternative taint like this.

You can set "enablePDB: false" - but this is officially not recommended for production environments.

### Additional context

_No response_

### Backport?

N/A

No

### Code of Conduct

chris-gooch:
Allows the controller to detect all well-known k8 autoscaler taints instead of only relying on `node.Spec.Unschedulable`

Fixes #5299 
Supercedes #6794

### DIFF
--- a/internal/controller/cluster_controller.go
+++ b/internal/controller/cluster_controller.go
@@ -67,7 +67,18 @@ const (
 	poolerClusterKey              = ".spec.cluster.name"
 	disableDefaultQueriesSpecPath = ".spec.monitoring.disableDefaultQueries"
 	imageCatalogKey               = ".spec.imageCatalog.name"
+	clusterAutoscalerTaint        = "ToBeDeletedByClusterAutoscaler"
+	v1KarpenterTaint              = "karpenter.sh/disrupted"
+	v1beta1KarpenterTaint         = "karpenter.sh/disruption"
 )
+
+// drainTaints includes taints used by K8s or autoscalers that signify node draining or pod eviction.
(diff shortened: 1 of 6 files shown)


MarcusDunn:
why has this been closed?

FF-Ninja:
@MarcusDunn 

please can you reopen your PR and make the relevant changes. I’ve had to close for personal reasons sorry

MarcusDunn:
@FF-Ninja no worries, thanks for the pointers on fixing my branch!

PR #6928, opened 2025-02-21
feat: allow customization of taints for node drain detection

### LINKED ISSUE #5299: [Feature]: Check for nodes cordoned by karpenter

MarcusDunn:
Introduce the `DRAIN_TAINTS` operator configuration option, enabling users to specify which taint keys should be recognized as indicators of node drain.

By default, this includes commonly used taints from `kubectl`, Cluster Autoscaler, and Karpenter:

- `node.kubernetes.io/unschedulable`
- `ToBeDeletedByClusterAutoscaler`
- `karpenter.sh/disrupted`
- `karpenter.sh/disruption`

This enhancement provides greater flexibility in handling node drain scenarios, allowing adaptation to different cluster configurations and scheduling strategies.

Fixes: #5299


internal/controller/cluster_controller.go
FF-Ninja, line 1: I think you need to keep this logic but change to isNodeBeingDrained. Without it, every node update will trigger a complete reconciliation loop, which will get expensive, especially on large clusters that have node churn.

MarcusDunn:
@sxd could you take a look at this one? apologies for the three separate PRs this seemed to turn into.

FF-Ninja:
Hi @MarcusDunn 

Hope you’re well. 

I think you may need to add back the drain taints check to the cluster_controller.go. Without that, I think it will trigger NodeToCluster reconciliation loop whenever a node is changed, added, removed from the cluster and it should only do that if a node is draining I believe. 

Thanks

MarcusDunn:
I would have liked to have a more clever check in `cluster_controller` but because the drain taints are configured per-cluster, I don't think there was a great way to do it.

For now (I think!) it only runs if any taints have changed.

Suggestions welcome.

MarcusDunn:
That seems pretty reasonable.

MarcusDunn:
@armru Is there anything I can do to move this along?

armru:
@mnencia we should put this in the queue so we can take a look and give feedback

leonardoce:
@MarcusDunn, thank you for this contribution. It was really good!
I just changed a few nits and added some unit tests. Do you agree with what I changed?

MarcusDunn:
> @MarcusDunn, thank you for this contribution. It was really good! I just changed a few nits and added some unit tests. Do you agree with what I changed?

Changes look great, thanks for taking a look! 

just for my own sake, when are struct fields PaselCase vs camelCase? I noticed you swapped `DrainTaints` to `drainTaints`

leonardoce:
> just for my own sake, when are struct fields PaselCase vs camelCase? I noticed you swapped DrainTaints to drainTaints. This is my first attempt at writing any go, so it's all pretty new.

`PaselCase` fields [are exported while](https://go.dev/ref/spec#Exported_identifiers) `camelCase` fields are not.
Yeah, it sounds weird if you are not used to it.
Your patch looked pretty good, and it is surprising that this is your first attempt. 
That means you've a strong background. Congrats!

leonardoce (approved):

armru (approved):

mnencia (approved):

gbartolini (approved):

PR #6945, opened 2025-02-24
docs: streamline documentation on failure modes

mnencia:
This page included specific failure scenarios. Since these largely follow standard Kubernetes behavior, we have streamlined the content to avoid duplication of information that belongs to the underlying Kubernetes stack and is not specific to CloudNativePG.

### DIFF
--- a/.wordlist-en-custom.txt
+++ b/.wordlist-en-custom.txt
@@ -1,4 +1,3 @@
-tolerationSeconds
 AES
 API's
 APIs
@@ -1329,6 +1328,7 @@ timeoutSeconds
 tls
 tmp
 tmpfs
+tolerationSeconds
 tolerations
 topologies
 topologyKey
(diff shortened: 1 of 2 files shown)


jsilvela (approved):

gbartolini:
After extensive contemplation and a thorough reassessment of the entire page, I realized that much of the content was simply a duplication of concepts that already exist in the Kubernetes ecosystem. We have already demonstrated our inability to keep this page updated, which has led to confusion among users. For instance, the failure of a worker node should not affect whether it is a CNPG workload or not.

gbartolini (approved):

sxd (approved):

ardentperf:
are there any specific pages from the kubernetes documentation that would be appropriate to link from this? for example https://kubernetes.io/docs/tasks/debug/debug-application/

PR #6990, opened 2025-02-26
test: simplify test depth code

jsilvela:
The struct that was used to hold the depth number initialized a
test environment structure unnecessarily


jsilvela:
E2E in shadow repo https://github.com/EnterpriseDB/cloudnative-pg/actions/runs/13542996730

PR #6991, opened 2025-02-26
fix: delay rollout until every Pods received the same configuration

### LINKED ISSUE #6956: [Bug]: double switchover on configuration upgrade

### Contact Details

_No response_

### Version

trunk (main)

1.32

YAML manifest

### What happened?

In a cluster with `PrimaryUpgradeMethod: true`, there is a nonzero probability that an update of a parameter that requires a restart triggers two switchovers, first restarting the primary, and later the newly promoted instance.

This can happen if an event happens after the primary has applied the new configuration while the other instances have not.

### Cluster resource

```shell

```

### Relevant log output

```shell

```

### Code of Conduct

leonardoce:
This patch delays the rollout of Pods to the point where every instance manager received the latest PostgreSQL configuration and applied it.

This prevents the operator from switching over two times to apply the same configuration when the primary received the configuration and, immediately after that, a reconciliation loop has been triggered on the cluster.

Fixes: #6956

# Release notes

```
Prevent a race condition causing the operator to perform two switchovers to update
the PostgreSQL configuration.
```


armru (approved):

leonardoce:
https://github.com/EnterpriseDB/cloudnative-pg/actions/runs/13563660081

internal/controller/cluster_controller.go
```diff
@@ -928,6 +929,23 @@ func (r *ClusterReconciler) reconcilePods(
 		return ctrl.Result{RequeueAfter: 1 * time.Second}, ErrNextLoop
 	}
 
+	// Stop acting here if the Pods don't have received
```
jsilvela, line 932:
```suggestion
	// Stop acting here if the Pods haven't received
```

internal/controller/cluster_controller.go
```diff
@@ -928,6 +929,23 @@ func (r *ClusterReconciler) reconcilePods(
...
+	// the latest PostgreSQL configuration.
```
jsilvela, line 933: Actually I think this whole comment is stale. The following code paragraph clarifies we request (stop acting) if there are pods in different configs

pkg/management/postgres/probes.go
```diff
@@ -86,6 +92,30 @@ func (instance *Instance) GetStatus() (result *postgres.PostgresqlStatus, err er
...
+			// and we still didn't load a generated configuration.
+			//
+			// This unfortunately will report an error
+			// into the PostgreSQL logs, but there's no
+			// current way to do it without raising an
+			// error at this moment.
+			//
+			// It should not represent a major issue since
+			// a correct configuration is going to be
+			// loaded immediately.
```
jsilvela, line 112: Here I wonder. So, we set the configuration hash to "", and then when we do `reconcilePods`, we see that not all pods have a hash, so we do a rolling upgrade, and this could repeat?

jsilvela:
https://github.com/EnterpriseDB/cloudnative-pg/actions/runs/13678905930

jsilvela (approved):

mnencia (approved):

PR #7055, opened 2025-03-03
chore(setup-cluster.sh): remove `prepare` command

### LINKED ISSUE #7050: [chore]: Remove `prepare` subcommand from `hack/setup-cluster.sh`
The `prepare` subcommand of `hack/setup-cluster.sh` was previously useful for preparing the environment for end-to-end (E2E) execution using that script. However, this process is now entirely managed by the GitHub workflow. Therefore, the workflow should be responsible for installing the appropriate versions of kind and kubectl.

armru:
Closes #7050

### DIFF
--- a/.github/workflows/continuous-delivery.yml
+++ b/.github/workflows/continuous-delivery.yml
@@ -659,7 +659,7 @@ jobs:
         with:
           install_only: true
           version: ${{ env.KIND_VERSION }}
-          kubectl_version: v${{ env.K8S_VERSION }}
+          kubectl_version: ${{ env.K8S_VERSION }}
       -
         name: Prepare patch for customization
         env:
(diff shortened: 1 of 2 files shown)


mnencia (commented):
This is half of the required work. The "prepare" command is still called in `.github/workflows/continuous-delivery.yml` and needs to be replaced with the appropriate github action to install kind and kubectl.

NiccoloFei:
E2E test run: https://github.com/cloudnative-pg/cloudnative-pg/actions/runs/13658655585

NiccoloFei (approved):

.github/workflows/continuous-delivery.yml
```diff
@@ -654,19 +654,12 @@ jobs:
...
+        uses: helm/kind-action@v1.12.0
         with:
-          timeout_seconds: 300
-          max_attempts: 3
-          on_retry_command: |
-            # Clear-ups before retries
-            sudo rm -rf /usr/local/bin/kind /usr/local/bin/kubectl
-          command: |
-            sudo apt-get update
-            sudo apt-get install -y gettext-base
```
mnencia, line 668: AFAIR, the `gettext-base` dependency is necessary to ensure that `envsubst` is available. Removing this dependency is risky.

mnencia (approved):

PR #7064, opened 2025-03-04
[draft] Automatically switch cluster to read-only on a high disk usage

usernamedt:
Hello, folks!

I propose to implement the periodic free disk space checks on behalf of cloudnative-pg instance manager. These checks will occur every `DiskSpaceCheckInterval` seconds.

If disk usage exceeds the `ReadOnlyThreshold`, instance will be automatically switched to read-only mode, i.e. all databases except `postgres` and template ones will be altered with `default_transaction_read_only` set to true.

If disk usage drops below the `ReadOnlyThreshold`, instance will be automatically switched back to read-write mode.

To ensure the propagation of the instance changes, all client connections will be dropped after the RW / RO mode switch.

Example:

```
postgres@postgresql-cluster-3:/$ fallocate -l9.4G /var/lib/postgresql/data/test

...

logs:

│ postgres {"level":"info","ts":"2025-03-03T15:23:32.381741445Z","logger":"DiskWatcher","msg":"default_transaction_readonly state changed, dropping existing user connections for changes to take effect","l │
│ ogger":"instance-manager","logging_pod":"postgresql-cluster-3","disk_usage":"99.23","threshold":"95.00"} 

...

postgres=# SELECT datname,
        COALESCE(setconfig @> '{default_transaction_read_only=true}'::text[], false) AS read_only
        FROM pg_database
        LEFT JOIN pg_db_role_setting
        ON (setdatabase=oid)
        WHERE datname != 'postgres' AND NOT datistemplate;
  datname   | read_only
------------+-----------
 usernamedt | t
(1 row)

postgres@postgresql-cluster-3:/$ rm /var/lib/postgresql/data/test

│ postgres {"level":"info","ts":"2025-03-03T15:29:32.381334211Z","logger":"DiskWatcher","msg":"default_transaction_readonly state changed, dropping existing user connections for changes to take effect","l │
│ ogger":"instance-manager","logging_pod":"postgresql-cluster-3","disk_usage":"1.91","threshold":"95.00"}                                                                                                    │

postgres=# SELECT datname,
        COALESCE(setconfig @> '{default_transaction_read_only=true}'::text[], false) AS read_only
        FROM pg_database
        LEFT JOIN pg_db_role_setting
        ON (setdatabase=oid)
        WHERE datname != 'postgres' AND NOT datistemplate;
  datname   | read_only
------------+-----------
 usernamedt | f
(1 row)


```

Existing connections are dropped during the RW / RO mode changes:
```
➜  ~ /opt/homebrew/Cellar/postgresql@16/16.4_1/bin/pgbench -i -s 100 "host=HOST\
   port=5432 \
   dbname=usernamedt \
   user=usernamedt"
Password:
dropping old tables...
NOTICE:  table "pgbench_accounts" does not exist, skipping
NOTICE:  table "pgbench_branches" does not exist, skipping
NOTICE:  table "pgbench_history" does not exist, skipping
NOTICE:  table "pgbench_tellers" does not exist, skipping
creating tables...
generating data (client-side)...
pgbench: error: query failed: FATAL:  terminating connection due to administrator command
FATAL:  server conn crashed?
server closed the connection unexpectedly
	This probably means the server terminated abnormally
	before or while processing the request.
pgbench: detail: Query was: insert into pgbench_tellers(tid,bid,tbalance) values (728,73,0)
```

if we try to write now there will be an expected error:

```
➜  ~ /opt/homebrew/Cellar/postgresql@16/16.4_1/bin/pgbench -i -s 100 "host=HOST\
   port=5432 \
   dbname=usernamedt \
   user=usernamedt"
Password:
dropping old tables...
pgbench: error: query failed: ERROR:  cannot execute DROP TABLE in a read-only transaction
pgbench: detail: Query was: drop table if exists pgbench_accounts, pgbench_branches, pgbench_history, pgbench_tellers
```

if default_transaction_read_only is altered manually, user can create / drop the tables, etc.

```
➜  ~ /opt/homebrew/Cellar/postgresql@16/16.4_1/bin/psql "host=HOST\
   port=5432 \
   dbname=usernamedt \
   user=usernamedt"
Password for user usernamedt:
psql (16.4 (Homebrew), server 16.6 (...))
Type "help" for help.

usernamedt=> set default_transaction_read_only = false;
SET

usernamedt=> drop table public.pgbench_accounts ;
DROP TABLE
```

### DIFF
--- a/api/v1/cluster_funcs.go
+++ b/api/v1/cluster_funcs.go
@@ -666,6 +666,20 @@ func (cluster *Cluster) GetMaxStopDelay() int32 {
 	return 1800
 }
 
+func (cluster *Cluster) GetReadOnlyDiskUsageThreshold() float64 {
+	if cluster.Spec.DiskWatcher != nil && cluster.Spec.DiskWatcher.ReadOnlyThreshold != nil {
+		return float64(*cluster.Spec.DiskWatcher.ReadOnlyThreshold)
+	}
+	return DefaultReadOnlyDiskUsageThreshold
+}
+
+func (cluster *Cluster) GetDiskWatcherCheckInterval() time.Duration {
+	if cluster.Spec.DiskWatcher != nil && cluster.Spec.DiskWatcher.DiskSpaceCheckInterval != nil {
+		return time.Duration(*cluster.Spec.DiskWatcher.DiskSpaceCheckInterval) * time.Second
+	}
+	return time.Duration(DefaultDiskWatcherCheckInterval) * time.Second
+}
+
 // GetSmartShutdownTimeout is used to ensure that smart shutdown timeout is a positive integer
 func (cluster *Cluster) GetSmartShutdownTimeout() int32 {
 	if cluster.Spec.SmartShutdownTimeout != nil {
--- a/api/v1/cluster_types.go
+++ b/api/v1/cluster_types.go
@@ -481,6 +481,10 @@ type ClusterSpec struct {
 	// in the PostgreSQL Pods.
 	// +optional
 	Probes *ProbesConfiguration `json:"probes,omitempty"`
+
+	// The configuration of the disk watcher
+	// +optional
+	DiskWatcher *DiskWatcherConfiguration `json:"diskWatcher,omitempty"`
 }
 
 // ProbesConfiguration represent the configuration for the probes
@@ -494,6 +498,20 @@ type ProbesConfiguration struct {
 
 	// The readiness probe configuration
 	Readiness *Probe `json:"readiness,omitempty"`
+}
+
+// DiskWatcherConfiguration defines parameters
+// for cluster free disk space checks. If the free disk space
+// falls below the specified threshold, the instance manager
+// will switch all databases in cluster to read-only mode.
+type DiskWatcherConfiguration struct {
+	// Disk usage threshold, in percents. After the disk usage exceeds this value,
+	// the instance manager will switch all databases in the cluster
+	// to read-only mode.
+	ReadOnlyThreshold *int32 `json:"readOnlyThreshold,omitempty"`
+
+	// Interval between free disk space checks, in seconds.
+	DiskSpaceCheckInterval *int32 `json:"diskSpaceCheckInterval,omitempty"`
 }
 
 // Probe describes a health check to be performed against a container to determine whether it is
@@ -1211,6 +1229,15 @@ const (
 	// FailureThreshold of startupProbe, the formula is `FailureThreshold = ceiling(startDelay / periodSeconds)`,
 	// the minimum value is 1
 	DefaultStartupDelay = 3600
+
+	// DefaultReadOnlyDiskUsageThreshold is the disk usage percentage threshold after which the instance will
+	// be put to ReadOnly mode. After the disk usage drops below the threshold, the instance will be put back to
+	// ReadWrite mode.
+	DefaultReadOnlyDiskUsageThreshold = 95
+
+	// DefaultDiskWatcherCheckInterval is the default interval in seconds
+	// for the disk watcher to check the disk usage
+	DefaultDiskWatcherCheckInterval = 30
 )
 
 // SynchronousReplicaConfigurationMethod configures whether to use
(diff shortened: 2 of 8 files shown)


usernamedt:
I am curious to get feedback regarding this feature, so I'll ignore the CI checks for now.

leonardoce:
Hi, thank you for contributing this.

I like the approach. It allows you to  connect to PostgreSQL and do emergency activities after having `SET TRANSACTION READ WRITE` to free space quickly.
A few questions:

1. Is `statfs` reliable? What would happen with NFS where the free disk space may be used by multiple volumes? NFS and PG may not be best friends, but there are people using it. A similar observation may apply to people using the file system of the underlying node (like https://github.com/rancher/local-path-provisioner)
2. What would happen with LVM thin provisioning, where you have may have free disk space declared but you won't be able to really use it?
3. PostgreSQL already goes so "read only mode" when the space for PGData is finished. The main problem the space for WALs, we may want to check just that
4. Instead of directly including it in the operator code base, I think this would be doable out-of-tree with a plugin. Why not creating a sidecar that periodically checks that and does the right thing?

jmealo:
**Overall, I think this feature could make sense, however, it will require careful consideration:**

- Do metrics/reporting and readiness status still makes sense?
- When a cluster changes from read to read-write --- would recovering from the low disk condition change it back automatically?
- I think it may need to be opt-in ... because making metrics/alerts react to the read-only versus read-write cluster could be challenging? (Assuming that the out of disk space condition isn't enough to alert on)
- If you have a primary that's in read-only status, it's not-ready? Then how would traffic get routed to it? 🤔 [ I assume making the cluster read-only, would mark this instance as READY]
- Where would the read-write poolers point to? Would it just queue waiting for connections?

usernamedt:
Hi @leonardoce @jmealo, thanks for your feedback!

> 1. Is `statfs` reliable? What would happen with NFS where the free disk space may be used by multiple volumes? NFS and PG may not be best friends, but there are people using it. A similar observation may apply to people using the file system of the underlying node (like https://github.com/rancher/local-path-provisioner)

Yes, for some storages we can't rely on it. My idea was to leave the decision to user. If he uses the NFS storage, he should not enable these checks.

> 2. What would happen with LVM thin provisioning, where you have may have free disk space declared but you won't be able to really use it?

Same logic as in (1), I guess I can write a list of requirements for storage to support this feature.

> 3. PostgreSQL already goes so "read only mode" when the space for PGData is finished. The main problem the space for WALs, we may want to check just that

Yes, but it turned out that the process of restoration from this read only mode is quite tedious, for example if storage does not support online disk resize. Also, in cloud providers where end users are decoupled from administrators we need some sort of "seat belt" to prevent users from making their cluster inoperable.

> 4. Instead of directly including it in the operator code base, I think this would be doable out-of-tree with a plugin. Why not creating a sidecar that periodically checks that and does the right thing?

Good idea, I'll check if it is easy to implement. Since we don't interfere with the CNPG logic here but just run it as a background job it should be easy to decouple this logic. However, I am not sure how the user should configure it in this case. As of now, it can be done easily by setting the appropriate config settings in the cluster's CRD.


> * Do metrics/reporting and readiness status still makes sense?
Yes, since we don't actually shut down the database. We only change the PostgreSQL session settings to prevent user from exhausting the disk space. So all the metrics and readiness checks should work just fine.

> * When a cluster changes from read to read-write --- would recovering from the low disk condition change it back automatically?

Yes, there is already logic implemented to bring cluster back to RW after the disk space usage falls below the threshold. 

> * I think it may need to be opt-in ... because making metrics/alerts react to the read-only versus read-write cluster could be challenging? (Assuming that the out of disk space condition isn't enough to alert on)

Sorry, I did not get this one. Can you clarify which metrics can be broken? 

Also, I thought about adding the metrics that will send the `default_transaction_read_only` signal for each database in cluster, but I thought that this might be too verbose.

> * If you have a primary that's in read-only status, it's not-ready? Then how would traffic get routed to it? 🤔 [ I assume making the cluster read-only, would mark this instance as READY]

I thought that this RW/RO switch should not affect the instance status. Technically, it is still in perfectly operable condition.  

> * Where would the read-write poolers point to? Would it just queue waiting for connections?

They still point to the ReadWrite (primary) instance. User can connect to cluster as usual. For example, user can manually set the `default_transaction_read_only` in his RW session to free up the space (drop tables / run vacuum / etc).

PR #7115, opened 2025-03-10
chore: delete the registry when cleaning up the test cluster

jsilvela:
When applying `hack/setup-cluster.sh destroy`, the registry is disconnected
and the Kind cluster destroyed, but the registry container remains, sometimes
leading to incompatibility when re-creating the dev cluster.

### DIFF
--- a/hack/setup-cluster.sh
+++ b/hack/setup-cluster.sh
@@ -213,6 +213,8 @@ destroy_kind() {
   docker network disconnect "kind" "${registry_name}" &>/dev/null || true
   kind delete cluster --name "${cluster_name}" || true
   docker network rm "kind" &>/dev/null || true
+  docker container stop "${registry_name}" &>/dev/null || true
+  docker rm "${registry_name}" &>/dev/null || true
 }
 
 ##


gabriele-wolfox (approved):

sxd:
I don't think it's a good idea to delete the registry, the whole reason for that registry to exists is to cache the images, otherwise, is better to not use it at all.

What's the issue with the registry? isn't something that we can fix?

jsilvela:
Just saw your comment @sxd 
Lately there have been many instances of outdated registry when we re-create the test cluster.
E.g.
```
ERROR: failed to solve: failed to push registry.dev:5000/cloud-native-postgresql-testing:latest: failed to do request: Head "https://registry.dev:5000/v2/cloud-native-postgresql-testing/blobs/sha256:093bfad5e1c6d4059cc78cf025bda6090d28b00ef9fb9de8fc29f61f38d42a97": dial tcp: lookup registry.dev on 127.0.0.11:53: no such host
```

jsilvela:
The thing to do there is delete and recreate.
As to the efficiency lost in doing this: since it only affects dev clusters, not much is lost.
You're destroying the cluster anyway.

PR #7126, opened 2025-03-11
fix: consider plugin changes while evaluating instance rollout

armru:
requires: https://github.com/cloudnative-pg/cnpg-i/pull/169

### DIFF
--- a/internal/cnpi/plugin/mapping.go
+++ b/internal/cnpi/plugin/mapping.go
@@ -22,15 +22,16 @@ import (
 	"github.com/cloudnative-pg/cnpg-i/pkg/lifecycle"
 )
 
-// The OperationVerb corresponds to the Kubernetes API method
+// The OperationVerb corresponds to the CNPG-I lifecycle operation verb
 type OperationVerb string
 
-// A Kubernetes operation verb
+// A lifecycle operation verb
 const (
-	OperationVerbPatch  OperationVerb = "PATCH"
-	OperationVerbUpdate OperationVerb = "UPDATE"
-	OperationVerbCreate OperationVerb = "CREATE"
-	OperationVerbDelete OperationVerb = "DELETE"
+	OperationVerbPatch    OperationVerb = "PATCH"
+	OperationVerbUpdate   OperationVerb = "UPDATE"
+	OperationVerbCreate   OperationVerb = "CREATE"
+	OperationVerbDelete   OperationVerb = "DELETE"
+	OperationVerbEvaluate OperationVerb = "EVALUATE"
 )
 
 // ToOperationType_Type converts an OperationVerb into a lifecycle.OperationType_Type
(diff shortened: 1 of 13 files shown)


jsilvela:
I've been playing with this a bit. It wasn't immediately clear how this worked.
Perhaps the name `Evaluate` threw me off.
So, seems the plugin enriches the generated pod spec ahead of the pod spec being applied?
Maybe PreCreate or PreProvision? Evaluate would give no clear message that the call modifies anything.
In any case, running it from a plugin and managing to inject an environment variable, so 👍

jsilvela:
Or, since we're talking of the `lifecycle` plugin, maybe `Gestate`? 😄

leonardoce (approved):

internal/cnpi/plugin/mapping.go
```diff
@@ -22,15 +22,16 @@ import (
...
 const (
-	OperationVerbPatch  OperationVerb = "PATCH"
-	OperationVerbUpdate OperationVerb = "UPDATE"
-	OperationVerbCreate OperationVerb = "CREATE"
-	OperationVerbDelete OperationVerb = "DELETE"
+	OperationVerbPatch    OperationVerb = "PATCH"
+	OperationVerbUpdate   OperationVerb = "UPDATE"
+	OperationVerbCreate   OperationVerb = "CREATE"
+	OperationVerbDelete   OperationVerb = "DELETE"
+	OperationVerbEvaluate OperationVerb = "EVALUATE"
```
jsilvela, line 34: In my opinion, Evaluate as a verb would give the impression of having no side-effects.
As this seems to be a mutation applied on the PodSpec, before actually creating the pod, I might suggest other verbs.
BTW made the same comments. Attaching here so they show up in reviews.

mnencia (approved):

internal/cnpi/plugin/mapping.go
leonardoce: I agree with you on the `Evaluate` concept. In the current implementation it has no side effects as the result is only used to evaluate whether the Pod definition is different than the one that would be generated.
The `Create` verb is the one actually used to create the new Pod.

Naming is difficult... at this moment I don't perceive `Evaluate` as misleading.

leonardoce:
> I've been playing with this a bit. It wasn't immediately clear how this worked. Perhaps the name `Evaluate` threw me off. So, seems the plugin enriches the generated pod spec ahead of the pod spec being applied? Maybe PreCreate or PreProvision? Evaluate would give no clear message that the call modifies anything. In any case, I am running it from a plugin and managing to inject an environment variable on instances, so 👍

Yes, it is complex for me too.

The plugin enriches the Pod spec the same way it could do it for every other resource managed by Kubernetes, by using the `CREATE`, `UPDATE` and the `PATCH` lifecycle method. See [the implementation in plugin-barman-cloud](https://github.com/cloudnative-pg/plugin-barman-cloud/blob/bcfe1fd8ddcc1cb868d329ab93c548d9095726e6/internal/cnpgi/operator/lifecycle.go#L41-L46).

In the case of CNPG Pods, the enrichment can be done using two methods:

- `EVALUATE` is used to create a Pod Spec that is used to compare with the current Pod to know whether to trigger a rollout or not
- `CREATE` to create the Pod spec that would be actually used.

Plugins can use the same implementation for both methods.

gbartolini (approved):

PR #7148, opened 2025-03-14
feat(database): allow `template1` as a database name

### LINKED ISSUE #7145: [Feature]: Enable `template1` in `Database` CRD

Right now, if you try to add a new extension to `template1`, you receive this error:

```
spec.name: Invalid value: "string": the name template1 is reserved
```

While earlier `template1` definition might now have made sense, with the new extension support we should enable users to add their extensions to `template1`. 

We should just keep `postgres`  and `template0` reserved for now.

N/A

### Additional context

_No response_

### Backport?

No

Yes

### Code of Conduct

gbartolini:
With the recent support for extensions in database management, users should be able to enable extensions in the `template1` database.

**Note:** Since the `app` user is created by default, we cannot guarantee that extensions will be added to `template1` beforehand. Therefore, if you use the `template1` approach via the `Database` resource, you should also apply the same configuration to the `app` database.

Closes #7145

### DIFF
--- a/api/v1/database_types.go
+++ b/api/v1/database_types.go
@@ -55,7 +55,6 @@ type DatabaseSpec struct {
 	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="name is immutable"
 	// +kubebuilder:validation:XValidation:rule="self != 'postgres'",message="the name postgres is reserved"
 	// +kubebuilder:validation:XValidation:rule="self != 'template0'",message="the name template0 is reserved"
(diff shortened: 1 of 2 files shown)


gabriele-wolfox:
@gbartolini 
In the scope of this PR, we should also update the docs here https://github.com/cloudnative-pg/cloudnative-pg/blob/main/docs/src/declarative_database_management.md#reserved-database-names

gbartolini:
I'm having second thoughts on this. Will comment later

gbartolini:
I'm having second thoughts on this. Will comment later

Issue #6564, opened 2025-01-10
[Bug]: final replica instance never finishes initializing


PeterGrace:
### Is there an existing issue already for this bug?

- [x] I have searched for an existing issue, and could not find anything. I believe this is a new bug.

### I have read the troubleshooting guide

- [x] I have read the troubleshooting guide and I think this is a new bug.

### I am running a supported version of CloudNativePG

- [x] I have read the troubleshooting guide and I think this is a new bug.

### Contact Details

pete.grace@gmail.com

### Version

1.25 (latest patch)

### What version of Kubernetes are you using?

1.30

### What is your Kubernetes environment?

Self-managed: k3s

### How did you install the operator?

Helm

### What happened?

I installed cloudnative-pg via helm chart, deploying chart version 0.23.0.  I created a basic cluster resource and waited for it to finish provisioning.  The final instance of the cluster never comes online, just continuously re-creates, terminates, then re-creates again.  The other two instances seem to have come up fine.  I removed the cluster and pvcs and tried a second time and the same result occurred.

### Cluster resource

```shell
apiVersion: postgresql.cnpg.io/v1
kind: Cluster
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"postgresql.cnpg.io/v1","kind":"Cluster","metadata":{"annotations":{},"name":"paperless-ngx","namespace":"paperless-ngx"},"spec":{"instances":3,"monitoring":{"enablePodMonitor":true},"storage":{"size":"100Gi"}}}
  creationTimestamp: "2025-01-10T17:31:24Z"
  generation: 1
  name: paperless-ngx
  namespace: paperless-ngx
  resourceVersion: "185987761"
  uid: 628a18ec-74e7-4834-b943-d8b2015f2cbd
spec:
  affinity:
    podAntiAffinityType: preferred
  bootstrap:
    initdb:
      database: app
      encoding: UTF8
      localeCType: C
      localeCollate: C
      owner: app
  enablePDB: true
  enableSuperuserAccess: false
  failoverDelay: 0
  imageName: ghcr.io/cloudnative-pg/postgresql:17.2
  instances: 3
  logLevel: info
  maxSyncReplicas: 0
  minSyncReplicas: 0
  monitoring:
    customQueriesConfigMap:
    - key: queries
      name: cnpg-default-monitoring
    disableDefaultQueries: false
    enablePodMonitor: true
  postgresGID: 26
[...]

PR #7167, opened 2025-03-20
fix: only compare image tag for cluster upgrade

PeterGrace:
This PR amends `cluster_upgrade.go` to only compare the tag value of the specified image, instead of the whole image name.

The rationale for this PR is that some organizations use MutatingWebHooks to force images to be pulled from pull-through caches or other security devices.  Comparing the full string value in this case will fail (as in #6564).

Personally, I think that comparing url string literals in the first place may be the wrong choice for this logic, but the alternatives (such as ensuring sha values match) might be more work than is necessary to solve the problem at hand.

### DIFF
--- a/internal/controller/cluster_upgrade.go
+++ b/internal/controller/cluster_upgrade.go
@@ -22,6 +22,7 @@ import (
 	"errors"
 	"fmt"
 	"reflect"
+	"strings"
 
 	"github.com/cloudnative-pg/machinery/pkg/log"
 	corev1 "k8s.io/api/core/v1"
@@ -498,7 +499,12 @@ func checkPodImageIsOutdated(pod *corev1.Pod, cluster *apiv1.Cluster) (rollout, 
 		return rollout{}, err
 	}
 
-	if pgCurrentImageName == targetImageName {
+	tagsMatch, err := compareImageTags(configuration.Current.PostgresImageName, pgCurrentImageName)
+	if err != nil {
+		return rollout{}, err
+	}
+
+	if tagsMatch {
 		return rollout{}, nil
 	}
 
@@ -520,7 +526,12 @@ func checkPodInitContainerIsOutdated(pod *corev1.Pod, _ *apiv1.Cluster) (rollout
 		return rollout{}, err
 	}
 
-	if opCurrentImageName == configuration.Current.OperatorImageName {
+	tagsMatch, err := compareImageTags(configuration.Current.OperatorImageName, opCurrentImageName)
+	if err != nil {
+		return rollout{}, err
+	}
+
+	if tagsMatch {
 		return rollout{}, nil
 	}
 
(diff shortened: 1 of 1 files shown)


armru:
Hello, thanks for the contribution, I think our current stance on this is that the correct image should be applied to the cluster so the mutation does not happen.

If any other maintainer has differing opinions feel free to correct me

PeterGrace:
> Hello, thanks for the contribution, I think our current stance on this is that the correct image should be applied to the cluster so the mutation does not happen.
> 
> If any other maintainer has differing opinions feel free to correct me

This stance would exempt organizations looking to run cloudnative-pg in air-gapped kubernetes clusters, since the image path would need to be rewritten to pull from a custom cluster registry such as harbor or artifactory.

armru:
could help me to understand further?

If you have the image pull URL that is `example.com/postgres:30` that gets mutated to `something.com/postgres:30` could you not apply the  `something.com/postgres:30` in the cluster manifest directly? Does the value change at each apply?

PeterGrace:
> I think, I'm likely, missing part of the problem
> 
> If you have the image pull URL that is `example.com/postgres:30` that gets mutated to `something.com/postgres:30` (assuming the local registry) could you not apply the `something.com/postgres:30` in the cluster manifest directly?

For the actual postgres image, yes, you can specify a custom path, but the logic in `cluster_upgrade.go` for the operator controller pod, by default, compares the image url to an image hosted at `ghcr.io` -- the only way to stop this is to force `OPERATOR_IMAGE_NAME` to be your new image path, but then in so doing, you wind up locking the operator image to the version you specify, requiring additional work for an upgrade that should not be required in a normal upgrade scenario.  This issue would not exist if we simply did the comparison on tag version instead of full repository URL.  

As I mentioned in the initial PR description, even doing this with a tag comparison is not a great choice as there's no guarantee that an airgapped image is uploaded with the same tag value.  A refactor of this logic that doesn't require string literal image url comparisons is likely the right play, but this PR would at least help those of us who use pull-through caching through harbor or artifactory.

gbartolini:
Hi @PeterGrace, thank you for your input. CloudNativePG is indeed used in air-gapped environments by many organisations.

Regarding this pull request, I believe we should not merge this patch as it does not resolve the underlying issue. The problem requires a broader solution than an image tag comparison and I fear that only addressing digests could provide a certain level of guarantee.

As far as I see it, [image catalog](https://cloudnative-pg.io/documentation/current/image_catalog/) resources have been specifically created for this purpose, so those operating in air-gapped environments should consider this approach.

If you choose not to use image catalogs, you can specify the correct image name that points to your local registry. However, the only challenge lies with the default image, which is addressed by the `OPERATOR_IMAGE_NAME` variable - which can be set in the operator's config map. Am I overlooking anything? Please help me better understand. Thanks.

Issue #7065, opened 2025-03-04
[Feature]: Deprecate `.spec.backup.retentionPolicy`


gbartolini:
### Is there an existing issue already for this feature request/idea?

- [x] I have searched for an existing issue, and could not find anything. I believe this is a new feature request to be evaluated.

### What problem is this feature going to solve? Why should it be added?

For historical reasons, retention policies are currently included in the `.spec.backup` section. This approach was necessary because object stores were the only supported method at the time, and we had not yet considered the use of plugins.

With the introduction of plugins, we need to establish a consistent way to define retention policies. This may involve allowing each plugin to manage its own retention policies.

### Describe the solution you'd like

- Implement the retention policies specification in the plugin
- Document that this feature will become obsolete and provide steps on how to move it in the plugin
- Log at cluster start that this feature is deprecated

### Describe alternatives you've considered

N/A

### Backport?

No

### Are you willing to actively contribute to this feature?

Yes

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

Issue #6876, opened 2025-02-19
[Feature]: Support Barman Cloud as a plugin


gbartolini:
### Is there an existing issue already for this feature request/idea?

- [x] I have searched for an existing issue, and could not find anything. I believe this is a new feature request to be evaluated.

### What problem is this feature going to solve? Why should it be added?

Since day one, **Barman Cloud** has been the primary backup solution for CloudNativePG. As its creators, this was a natural initial choice. However, over the years, we have recognised the need for **CloudNativePG to evolve into a Postgres backup/recovery-agnostic framework**, providing a **generic interface** that allows alternative backup solutions to emerge and thrive. This vision led to the development of **CNPG-I**.

Note: **Volume Snapshots**, as standard Kubernetes resources, will be natively supported - although we cannot exclude a transition to a plugin.

### Describe the solution you'd like

Starting with **CloudNativePG 1.26**, we plan to **deprecate native support for Barman Cloud** and transition it into a plugin over the course of two minor releases (1.28). 

To ensure a **smooth migration**, we will:

- Provide clear **documentation** on how to transition from native Barman Cloud to the Barman Cloud Plugin.
- Ensure a **transparent and well-documented migration path** for users.

### Describe alternatives you've considered

N/A

### Additional context

N/A

### Backport?

No

### Are you willing to actively contribute to this feature?

Yes

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

vishaltanawade:
Hi @gbartolini , @mnencia ,
I am using CNPG 1.26.0 . So when I use Barman Cloud I got  below warning: 
Warning: Native support for Barman Cloud backups and recovery is deprecated and will be completely removed in CloudNativePG 1.28.0. Found usage in: spec.backup.barmanObjectStore. Please migrate existing clusters to the new Barman Cloud Plugin to ensure a smooth transition.

so **Barman cloud plugin** is supported or not in version 1.26.0.
If not then in which version it will be supported.
Thanks and Regards,
vishal.tanawade@emerson.com

PR #7248, opened 2025-03-28
docs: Remove references to parallel WAL restore from docs

fhoekstra:
Seeing as according to bug https://github.com/cloudnative-pg/cloudnative-pg/issues/3459 this feature is not planned, I propose removing the references from the docs.

### DIFF
--- a/docs/src/recovery.md
+++ b/docs/src/recovery.md
@@ -1,5 +1,4 @@
 # Recovery
-<!-- SPDX-License-Identifier: CC-BY-4.0 -->
 
 In PostgreSQL terminology, recovery is the process of starting a PostgreSQL
 instance using an existing backup. The PostgreSQL recovery mechanism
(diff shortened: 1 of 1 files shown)


mnencia:
Thank you for your contribution! I appreciate the effort you've put into this PR. However, I would like to clarify a few points regarding the original issue:

* The retrieval of WAL files is not limited to recovery scenarios. It is also a critical process when a replica has been disconnected for some time and needs to synchronize with the primary.

* Parallel WAL retrieval during recovery operates effectively when using CNPG-I backup plugins, such as `plugin-barman-cloud`.

* As part of our ongoing improvements, we are deprecating the in-tree backup implementation in favor of CNPG-I backup plugins. The related documentation updates are being managed in #6876 and #7065.

In light of these considerations, I will be closing this PR. Thank you once again for your contribution, and we encourage you to continue contributing to the project!

PR #7265, opened 2025-03-31
refactor(instance,secrets): extract sub-reconciler for instance secrets management

armru:
This patch introduces a dedicated subreconciler to manage PostgreSQL cryptographic material.
The refactored logic is now reused across multiple components:

* the instance manager reconciliation loop
* the "join" bootstrap job
* the PostgreSQL major upgrade process

## Note for reviewers

- decide if we should backport
- depends on #7260

PR #7266, opened 2025-03-31
refactor(instance,storage): extract sub-reconciler for instance storage

armru:
This patch introduces a dedicated subreconciler to manage PostgreSQL separate WAL storage.
The refactored logic is now reused across multiple components:

* the PostgreSQL startup process
* the PostgreSQL major upgrade process

## Note for reviewers
- decide if we should backport
- depends on #7265


mnencia (approved):

leonardoce:
I'll backport this too

leonardoce (approved):

PR #7281, opened 2025-04-02
fix(plugin): expand `logs pretty` cmd buffer size

### LINKED ISSUE #7183: [Bug/Feature]: kubectl-cnpg logs pretty exit on large log lines

### Contact Details

_No response_

### Version

1.25 (latest patch)

1.30

Other

YAML manifest

### What happened?

When trying to push logs into `kubectl cnpg logs pretty`, the process silently exits when reading a line too large to fit in the Scanner.Buffer.

By default, Scanner.Scan uses an internal buffer with size 64 * 1024 bytes.

Unfortunately, postgresql log lines may include very large query, and reach that limit.
It would be helpful to have an option to increase that buffer size when running `kubectl cnpg logs pretty`.

### Cluster resource

```shell

```

### Relevant log output

```shell

```

### Code of Conduct

armru:
# Release Notes

`cnpg logs pretty` command now has a bigger default buffer size.

Closes #7183


mnencia:
In both the logpipe and the podlog writer, we utilize `scanner.Buffer(make([]byte, 0, 4096), 1024*1024)`. This code allocates a 4KB buffer that can expand dynamically, if necessary, up to a maximum size of 1MB. I believe we can apply the same approach here and eliminate the need for the additional parameter.

leonardoce (approved):

sxd (approved):

mnencia (approved):

PR #7284, opened 2025-04-03
feat(cnpgi,snapshot): support WAL recovery via plugins during snapshot restore

### LINKED ISSUE #7275: Segfault when using barman cloud plugin for restoration
There was an issue #6295 earlier, and we are having a similar issue when doing a restore by referencing WAL from S3 using the barman cloud plugin.  

> We are still having an issue.  I am not very confident if it is related to this. we are using CNPG 1.25 and barman cloud plugin v0.3.0 
> 
> Image
>  

 _Originally posted by @anjuls in [#6295](https://github.com/cloudnative-pg/cloudnative-pg/issues/6295#issuecomment-2770304233)_

armru:
This patch enables CNPG-i plugins to serve as a source of WALs during recovery from volume snapshots.

Previously, plugins were required to provide both WALs and PGDATA, which prevented their use in snapshot-based recovery scenarios. With this change, plugins can now be used solely for WAL retrieval during recovery.

Closes #7275

### DIFF
--- a/internal/cmd/manager/instance/pgbasebackup/cmd.go
+++ b/internal/cmd/manager/instance/pgbasebackup/cmd.go
@@ -180,5 +180,5 @@ func (env *CloneInfo) configureInstanceAsNewPrimary(ctx context.Context, cluster
 	// In the future, when we will support recovering WALs in the
 	// designated primary from an object store, we'll need to use
 	// the environment variables of the recovery object store.
-	return env.info.ConfigureInstanceAfterRestore(ctx, cluster)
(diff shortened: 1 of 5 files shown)


mnencia (approved):

fcanovai (changes requested):
This is not working. If I try to restore a cluster from a snapshot with wals from a plugin, the logic believes that it has to perform a restore via plugin, and then errors out with `no plugin supports the restore job hooks capability`. It doesn't restore the snapshot and use the WALs from the plugin.

fcanovai (approved):

PR #7285, opened 2025-04-03
fix(cnpgi,instance,backup): do not assume that the cluster plugins are available to the instance

armru:
The main branch code incorrectly assumes that all reported plugins in the cluster are available to the instance. 

The patch aims to correctly load only the expected backup plugin.


mnencia (approved):

mnencia:
Superseded by #7242

PR #7297, opened 2025-04-07
feat: remove Azure CSI driver resize workaround

### LINKED ISSUE #7296: [Feature]: Remove ENABLE_AZURE_PVC_UPDATES configuration parameter

The `ENABLE_AZURE_PVC_UPDATES` configuration was introduced as a workaround for a [missing feature](https://github.com/kubernetes-sigs/azuredisk-csi-driver/issues/273) in the AKS platform that was preventing CloudnativePG users from resizing volumes. 

This feature was implemented at the end of 2022, so it is now time to remove the `ENABLE_AZURE_PVC_UPDATES` configuration.

Remove the `ENABLE_AZURE_PVC_UPDATES` configuration in version 1.26.0

N/A

### Additional context

https://github.com/kubernetes-sigs/azuredisk-csi-driver/issues/273

### Backport?

No

Yes

### Code of Conduct

armru:
This commit removes `ENABLE_AZURE_PVC_UPDATES` configuration and the related
status field. 

The configuration is no longer needed for the operator to resize Azure volumes correctly. This change has been incorporated in the Azure CSI driver since version 1.11.0. 

Closes #7296 

## Release notes

```
Support for `ENABLE_AZURE_PVC_UPDATES` configuration has been removed, as it is no longer needed for the operator to resize Azure volumes correctly. The fix is included in the Azure CSI driver since version [1.11.0](https://github.com/kubernetes-sigs/azuredisk-csi-driver/releases/tag/v1.11.0).
```


gbartolini (approved):

jsilvela:
There's a whole section of the documentation that we need to rethink.
And, should we not add an item in the release notes with the deprecation notice?

```
### Expanding PVC volumes on AKS
```

nfaust-edb:
When triggering two operations (VM +disk update) at the same time, the operator asks the pod restart for VM upgrade to complete.  Azure preserves both PVC conditions at the same time, while AWS & GCP removes the `Resizing` one when it adds the `FileSystemResizePending`.  This will force the operator to wait and not recreate the pod on the new VM.

```
diff --git a/pkg/reconciler/persistentvolumeclaim/resources.go b/pkg/reconciler/persistentvolumeclaim/resources.go
index 1ca30256f..bcaaef34a 100644
--- a/pkg/reconciler/persistentvolumeclaim/resources.go
+++ b/pkg/reconciler/persistentvolumeclaim/resources.go
@@ -65,15 +65,21 @@ external:
        return true
 }

-// isResizing returns true if PersistentVolumeClaimResizing condition is present
+// isResizing returns true if PersistentVolumeClaimResizing condition is present and not PersistentVolumeClaimFileSystemResizePending
 func isResizing(pvc corev1.PersistentVolumeClaim) bool {
+       isResizing := false
+       isFileSystemResizePending := false
+
        for _, condition := range pvc.Status.Conditions {
                if condition.Type == corev1.PersistentVolumeClaimResizing {
-                       return true
+                       isResizing = true
+               }
+               if condition.Type == corev1.PersistentVolumeClaimFileSystemResizePending {
+                       isFileSystemResizePending = true
                }
        }

-       return false
+       return (isResizing && !isFileSystemResizePending)
 } 
```

valeriodelsarto:
I agree with @nfaust-edb , this PR is not only about removing support for a feature we don't need anymore due to Azure having improved (finally) the persistent volume expansions not requiring anymore a disk to be in a "detached" state from the node to progress, but also about fixing a corner case in which if both the disk and the node are changed at the same time, we end in a state in which the new pod is not getting created because of the PVC still having the `Resizing` condition set to `true`, and this only happens on Azure (not on AWS nor on GCP) of the 3 major CSPs.

mnencia (approved):

mnencia:
@nfaust-edb @valeriodelsarto I see your point about the proposed change, but I believe it deserves its own issue and PR for easier tracking. Could you open a dedicated issue?

valeriodelsarto:
Thanks @mnencia, @mbiagetti created one here https://github.com/cloudnative-pg/cloudnative-pg/issues/7324. I was hoping that this PR would have also fixed that issue, but needs to be confirmed by doing a test.

PR #7298, opened 2025-04-07
fix(instance_replica): include dbname in primary_conninfo for replication configuration

### LINKED ISSUE #5917: [Bug]: Cannot use PG 17 replication slot failover (dbname must be specified in primary_conninfo)

### Contact Details

_No response_

### Version

1.24.0

1.31

Self-managed: k3s

Helm

### What happened?

I'm trying to test the new failover slots into PG 17. I am able to turn on hot_standby_feedback and sync_replication_slots but the slots never synchronize. The replica is logging an error from the "slotsync worker" that says that dbname must be specified in the primary_conninfo in order to enable failover sync. [According to the PG docs](https://www.postgresql.org/docs/current/runtime-config-replication.html#GUC-PRIMARY-CONNINFO) the dbname parameter is used exclusively for slot synchronization, so the issue might be relatively straightforward to solve.

### Cluster resource

_No response_

### Relevant log output

```shell
{"level":"info","ts":"2024-10-21T23:57:29.986731099Z","logger":"postgres","msg":"record","logging_pod":"best-cluster-1","record":{"log_time":"2024-10-21 23:57:29.986 UTC","process_id":"4376","session_id":"6716ea69.1118","session_line_num":"2","session_start_time":"2024-10-21 23:57:29 UTC","virtual_transaction_id":"92/0","transaction_id":"0","error_severity":"ERROR","sql_state_code":"22023","message":"replication slot synchronization requires \"dbname\" to be specified in \"primary_conninfo\"","backend_type":"slotsync worker","query_id":"0"}}
```

### Code of Conduct

quajones:
closes #5917 include dbname within primary_conninfo for replica configuration to support replication slot synchronization and enable logical replication failover

> replication slot synchronization requires "dbname" to be specified in "primary_conninfo"

### DIFF
--- a/pkg/management/postgres/instance_replica.go
+++ b/pkg/management/postgres/instance_replica.go
@@ -64,7 +64,7 @@ func (instance *Instance) RefreshReplicaConfiguration(
 
 func (instance *Instance) writeReplicaConfigurationForReplica(cluster *apiv1.Cluster) (changed bool, err error) {
 	slotName := cluster.GetSlotNameFromInstanceName(instance.GetPodName())
-	primaryConnInfo := instance.GetPrimaryConnInfo() + fmt.Sprintf(" dbname=%v", cluster.GetApplicationDatabaseName())
+	primaryConnInfo := instance.GetPrimaryConnInfo() + " dbname=postgres"
 	return UpdateReplicaConfiguration(instance.PgData, primaryConnInfo, slotName)
 }
 
--- a/pkg/management/postgres/instance_replica.go
+++ b/pkg/management/postgres/instance_replica.go
@@ -64,10 +64,7 @@ func (instance *Instance) RefreshReplicaConfiguration(
 
 func (instance *Instance) writeReplicaConfigurationForReplica(cluster *apiv1.Cluster) (changed bool, err error) {
 	slotName := cluster.GetSlotNameFromInstanceName(instance.GetPodName())
-	primaryConnInfo := instance.GetPrimaryConnInfo()
-	if applicationDB := cluster.GetApplicationDatabaseName(); applicationDB != "" {
-		primaryConnInfo += fmt.Sprintf("dbname=%v", applicationDB)
-	}
+	primaryConnInfo := instance.GetPrimaryConnInfo() + fmt.Sprintf(" dbname=%v", cluster.GetApplicationDatabaseName())
 	return UpdateReplicaConfiguration(instance.PgData, primaryConnInfo, slotName)
 }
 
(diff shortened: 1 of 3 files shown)


pkg/management/postgres/instance_replica.go
```diff
@@ -64,7 +64,8 @@ func (instance *Instance) RefreshReplicaConfiguration(
 
 func (instance *Instance) writeReplicaConfigurationForReplica(cluster *apiv1.Cluster) (changed bool, err error) {
 	slotName := cluster.GetSlotNameFromInstanceName(instance.GetPodName())
-	return UpdateReplicaConfiguration(instance.PgData, instance.GetPrimaryConnInfo(), slotName)
+	primaryConnInfo := instance.GetPrimaryConnInfo() + fmt.Sprintf(" dbname=%v", cluster.GetApplicationDatabaseName())
```
mnencia, line 67: I would use the "postgres" database here because it always exists and does not require any additional logic.

quajones: thanks! this has been updated

kndoni:
would it be good to have this configurable, because there are use cases when dbname is not postgres but its another database

mnencia:
> would it be good to have this configurable, because there are use cases when dbname is not postgres but its another database

The instance manager always uses the `postgres` database when connecting to the local instance, so its existence is mandatory. The dbname parameter in the primary conninfo has no other uses than to specify a database that must exist, so I see no reason to make it configurable.

kndoni:
> > would it be good to have this configurable, because there are use cases when dbname is not postgres but its another database
> 
> The instance manager always uses the `postgres` database when connecting to the local instance, so its existence is mandatory. The dbname parameter in the primary conninfo has no other uses than to specify a database that must exist, so I see no reason to make it configurable.

But in my usecase I have enabled failover for logical replication in postgresql 17 instance. And when one of the standby replica-s is promoted as primary the logical replication slot won't switchover because of the missing dbname in primary_conninfo. But in my case I am having the logical replication in a different database not in the postgresql. So in this case I need to configure dbname parameter in primary_conninfo

mnencia:
> But in my usecase I have enabled failover for logical replication in postgresql 17 instance. And when one of the standby replica-s is promoted as primary the logical replication slot won't switchover because of the missing dbname in primary_conninfo. But in my case I am having the logical replication in a different database not in the postgresql. So in this case I need to configure dbname parameter in primary_conninfo

I think you misunderstood the usage of the `dbname` parameter here. It doesn't need to be the name of the database where the replication slot operates, but just a valid database where the slot replication code could run the functions to inspect the actual slot's position. So, setting it to `postgres` is perfectly safe, as the whole operator would stop working if you delete or rename that database. See the Postgres patch for further details https://github.com/postgres/postgres/commit/93db6cbda037f1be9544932bd9a785dabf3ff712

mnencia (approved):

sxd (approved):

PR #7303, opened 2025-04-08
test: allow configuring a custom image repository in major upgrade E2Es

### LINKED ISSUE #7302: [Test]: Major Upgrade E2E failures in postgres-trunk

### Contact Details

_No response_

### Version

1.26.0-rc1

1.32

YAML manifest

### What happened?

Major Upgrade E2Es have been failing on https://github.com/cloudnative-pg/postgres-trunk-containers/, see the latest [CD](https://github.com/cloudnative-pg/postgres-trunk-containers/actions/workflows/continuous-delivery.yml) runs.

We need to create proper target images that we can use in the E2Es to upgrade to v18, and also modify the test to allow using these new images. 

### Cluster resource

```shell

```

### Relevant log output

```shell

```

### Code of Conduct

NiccoloFei:
This patch goes together with https://github.com/cloudnative-pg/postgres-trunk-containers/pull/80.
The goal is to allow configuring a custom repository (in this case the `postgresql-trunk`) instead of the default ones (via an env variable), which will be used to fetch the target images required to perform the major upgrade scenarios.
This will allow us to test major upgrades to development versions of PG18.

Closes #7302

### DIFF
--- a/tests/e2e/cluster_major_upgrade_test.go
+++ b/tests/e2e/cluster_major_upgrade_test.go
@@ -54,11 +54,12 @@ import (
 
 var _ = Describe("Postgres Major Upgrade", Label(tests.LabelPostgresMajorUpgrade), func() {
 	const (
-		level                  = tests.Medium
-		namespacePrefix        = "cluster-major-upgrade"
-		postgisEntry           = "postgis"
-		postgresqlEntry        = "postgresql"
-		postgresqlMinimalEntry = "postgresql-minimal"
+		level                     = tests.Medium
+		namespacePrefix           = "cluster-major-upgrade"
+		postgisEntry              = "postgis"
+		postgresqlEntry           = "postgresql"
+		postgresqlMinimalEntry    = "postgresql-minimal"
+		customImageRegistryEnvVar = "MAJOR_UPGRADE_IMAGE_REGISTRY"
 	)
 
 	var namespace string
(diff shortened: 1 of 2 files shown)


jsilvela (approved):

NiccoloFei:
After 7dc39a3e1954c392a5b04d6b942369ae6fc4fbb8, major upgrade E2Es to v18 using minimal and standard images are able to succeed.

PostGIS images instead keep failing with the following error:
```
"msg":"/controller/old/usr/lib/postgresql/17/bin/postgres: error while loading shared libraries: libssl.so.1.1: 
cannot open shared object file: No such file or directory","pipe":"stderr"
```
This happens because the images we have on https://github.com/cloudnative-pg/postgis-containers are still based on bullseye, while the ones on postgres-trunk are on bookworm. 
There's a discrepancy in the version of OpenSSL being used between the 2 images, which makes them incompatible.
```
$ docker run --rm -it ghcr.io/cloudnative-pg/postgis:17 ls -la /usr/lib/x86_64-linux-gnu | grep libssl.so
-rw-r--r--  1 root root    597792 Nov  3 04:59 libssl.so.1.1

$ docker run --rm -it ghcr.io/cloudnative-pg/postgresql-trunk:18-postgis-bookworm ls -la /usr/lib/x86_64-linux-gnu | grep libssl.so
-rw-r--r--  1 root root    688160 Oct 27 14:16 libssl.so.3
```
Unfortunately we don't have PostGIS bookworm images yet, not until they either get added to https://github.com/postgis/docker-postgis or we start generating them ourselves..

mnencia (approved):

PR #7357, opened 2025-04-14
add hint on server-side encryption

JensErat:
When using cloudnative-pg with an object storage system that does not support or does not have enabled server-side encryption by default, uploading files will result with error messages not hinting towards the root cause at all. With a ceph-based object storage, I only realized I'm missing SSE when digging into object storage server logs -- something a regular user will not even have available.

### DIFF
--- a/docs/src/appendixes/object_stores.md
+++ b/docs/src/appendixes/object_stores.md
@@ -150,6 +150,32 @@ spec:
         [...]
 ```
 
+!!! Important
+    By default, backups will be encrypted. Encryption requires support for
+    server-side encryption in the object storage system. While AWS S3 and
+    many others enable server-side encryption by default, not all third
+    party systems do.
+
+    If the object storage does not support server side encryption and you
+    can accept unencrypted backups, disable encryption by setting
+    `encryption` to the empty string.
+
+```yaml
+apiVersion: postgresql.cnpg.io/v1
+kind: Cluster
+[...]
+spec:
+  backup:
+    barmanObjectStore:
+      [...]
+      wal:
+        encryption: ""
+      data:
+        encryption: ""
+      
+```
+
+
 ### Using Object Storage with a private CA
 
 Suppose you configure an Object Storage provider which uses a certificate


armru:
Hello, thanks for the contribution, adding it to the review queue for 1.27.0

jsilvela:
Hi @JensErat ,
looking over your patch. Thank you!
I'm surprised though: in the code, the `encryption` defaults to the empty string.
To be specific:
> // Allowed options are empty string (use the bucket policy, default),

In your case, are you perhaps using some base mechanism for deployment?
I worry that the behavior you see is specific about ceph, or your particular details.
If that were the case, the warning box would still be useful, but might need different wording.

JensErat:
Given I applied the change using gitops and then it worked, I'm pretty sure _something_ must have been going on. Some hint might be that while we have an empty string default, `wal` is a pointer. https://github.com/cloudnative-pg/barman-cloud/blob/main/pkg/api/config.go#L199

First idea was that maybe not setting wal introduces another default, but I cannot see any code path where this would have happened. The encryption flag is not set unless both wal/data and encryption are defined, so explicitly setting it to the emptry string should also result in not having an encryption flag set.

I guess I'll have to run the operator in the debugger to see what's going on.

jsilvela:
This is not fresh in my memory so may not be accurate, but in the past when investigating bug reports around encryption, I've had to explicitly set the encryption.
So, one thing is your particular setup which may have led to this "encryption-by-default".
The other thing is server-side encryption failing silently, or if not silently at least unspecifically.
Do you have saved log messages with the unspecific errors?
A warning box on this type of issue might be very useful.

gbartolini:
I am moving this to 1.27.1 as we are planning to release next week.

JensErat:
I was able to reproduce with the newest releases, and also understood where this is coming from. Actually pretty trivial: I'm installing using the helm chart, which defaults encryption to AES256 when templating the cluster resource.

Caused by values.yaml defaulting to AES256: https://github.com/cloudnative-pg/charts/blob/cluster-v0.3.1/charts/cluster/values.yaml#L424-L431

So this PR is not 100% accurate, given this is caused by the defaulting to AES256 which is applied by the helm chart but not default of the underlying resource. Not sure what's the best way to tackle this. What's your thought? Changing the helm chart values to the empty string would be the most expected way to have this. The way it fails right now is pretty hard to track down, as you have to run the required commands manually, the real issue is not showing up in the logs (the encryption flag is, but no trace to the encryption being root cause at all):

```
{"level":"info","ts":"2025-08-09T21:48:53.03815008Z","msg":"Failed archiving WAL: PostgreSQL will retry","logger":"instance-manager","logging_pod":"postgresql-test-cluster-1","controller":"instance-cluster","controllerGroup":"postgresql.cnpg.io","controllerKind":"Cluster","Cluster":{"name":"postgresql-test-cluster","namespace":"database"},"namespace":"database","name":"postgresql-test-cluster","reconcileID":"66b188b0-6dac-4cca-ab35-3103160fe7db","logging_pod":"postgresql-test-cluster-1","walName":"/var/lib/postgresql/data/pgdata/pg_wal/000000010000000000000002","startTime":"2025-08-09T21:48:52.426258347Z","endTime":"2025-08-09T21:48:53.038146886Z","elapsedWalTime":0.61188854,"error":"unexpected failure invoking barman-cloud-wal-archive: exit status 4"}
{"level":"info","ts":"2025-08-09T21:48:54.772862168Z","logger":"barman-cloud-wal-archive","msg":"2025-08-09 21:48:54,772 [177] ERROR: Barman cloud WAL archiver exception: An error occurred (InvalidRequest) when calling the PutObject operation: None","pipe":"stderr","logging_pod":"postgresql-test-cluster-1"}
{"level":"error","ts":"2025-08-09T21:48:54.815460111Z","msg":"Error invoking barman-cloud-wal-archive","logger":"instance-manager","logging_pod":"postgresql-test-cluster-1","controller":"instance-cluster","controllerGroup":"postgresql.cnpg.io","controllerKind":"Cluster","Cluster":{"name":"postgresql-test-cluster","namespace":"database"},"namespace":"database","name":"postgresql-test-cluster","reconcileID":"66b188b0-6dac-4cca-ab35-3103160fe7db","logging_pod":"postgresql-test-cluster-1","walName":"/var/lib/postgresql/data/pgdata/pg_wal/000000010000000000000001","options":["--gzip","-e","AES256","--endpoint-url","[redacted]","--cloud-provider","aws-s3","s3://postgres-test-backup/","postgresql-test-cluster","/var/lib/postgresql/data/pgdata/pg_wal/000000010000000000000001"],"exitCode":-1,"error":"exit status 4","stacktrace":"github.com/cloudnative-pg/machinery/pkg/log.(*logger).Error\n\tpkg/mod/github.com/cloudnative-pg/machinery@v0.1.0/pkg/log/log.go:125\ngithub.com/cloudnative-pg/barman-cloud/pkg/walarchive.(*BarmanArchiver).Archive\n\tpkg/mod/github.com/cloudnative-pg/barman-cloud@v0.1.0/pkg/walarchive/cmd.go:83\ngithub.com/cloudnative-pg/barman-cloud/pkg/walarchive.(*BarmanArchiver).ArchiveList.func1\n\tpkg/mod/github.com/cloudnative-pg/barman-cloud@v0.1.0/pkg/walarchive/cmd.go:115"}

```

Maybe kind of related to https://github.com/cloudnative-pg/charts/pull/614 -- but I don't see this would change the defaulting.

jsilvela:
@JensErat sorry forgot about this. Your finding about the helm chart makes total sense.
I think the right place to add this as documentation, troubleshooting, gotchas etc. would be the helm charts repo.
And it may make sense to take away that default in the chart, since it is so dependent on implementation.
IMO either the changing of the default, or the documentation of edge cases, should be discussed with @itay-grudev and potentially put in the charts repo.

itay-grudev:
@JensErat The chart aims to have a set of sane secure defaults that make it very easy to get started with CloudNativePG. Because most users use S3 or GCS, that works fine.

You can always override those values by setting them here:

https://github.com/cloudnative-pg/charts/blob/0dcefbe60a6031b7d8508b6b9efa0faf2fa8e0ec/charts/cluster/values.yaml#L422

As such, I'm not that inclined to change it unless it really presents a problem for the majority of our users. AFAIK, that's not the case.

jsilvela:
I think this could be documented and formalized a bit. If the cluster chart has different defaults than plain CloudNativePG, that's totally legitimate, but worth a warning box at least. And the charts repo is the right place for this, in my opinion.

By the way I was looking at the documentation in the `cloudnative-pg/cloudnative-pg` main repo, and I don't think it says anything about the cluster chart. Which makes sense, but then perhaps, @itay-grudev , you might add a blog post in the website explaining about the cluster chart, its approach and caveats?
I do see quite a few users of it :)

gbartolini:
I am 100% certain that this information should not be in the operator's document. I am closing this ticket here. @JensErat, thanks for proposing it. It might be worth checking with @itay-grudev and @jsilvela on how to add it to the chart docs.

PR #7359, opened 2025-04-15
feat: allow to specify a scrapeClass on PodMonitors

steinwelberg:
This PR adds the option to configure a scrapeClass on the PodMonitors created by the operator.

### DIFF
--- a/.wordlist-en-custom.txt
+++ b/.wordlist-en-custom.txt
@@ -1099,6 +1099,7 @@ podCount
 podMetricsEndpoints
 podMonitorMetricRelabelings
 podMonitorRelabelings
+podMonitorScrapeClass
 podName
 podStatuses
 podmonitor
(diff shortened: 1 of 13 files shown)


gbartolini:
Thank you for the patch, @steinwelberg.

As you've probably noticed, we're actively working to reduce the scope of CloudNativePG with initiatives like CNPG-I.

The existing `PodMonitor` support in CloudNativePG was designed to provide basic functionality, and rather than incrementally adding pieces from the upstream API, we believe it is better to delegate entirely to it for advanced and customised configurations.

In line with this, we're considering starting the deprecation process for our built-in `PodMonitor` support, while ensuring we provide clear documentation on how to integrate CloudNativePG with Prometheus using native resources.

Would you be interested in helping us with the documentation aspect of this transition as we close this PR?

Thanks again for taking the time to contribute and help improve the project.

gbartolini:
Closing this, as any work will proceed with #8075

PR #7366, opened 2025-04-15
Correct PGMAJOR for Image Catalog Example in image_catalog.md

theadamwright:
###

The two image catalog examples list PGMAJOR as 16, but the actual image is version 17. This change simply updates PGMAJOR to 17, ensuring it matches the PGMAJOR in the YAML matches the PGMAJOR of the container image. 

https://cloudnative-pg.io/documentation/1.25/image_catalog/

Before: 
    - major: 16 image: ghcr.io/cloudnative-pg/postgresql:17.4

After: 
    - major: 16 image: ghcr.io/cloudnative-pg/postgresql:17.4


jsilvela (approved):
Ah, good find. This seems to have happened because of `github-actions[bot]` updating the repo with the latest PG version.
So I wonder if it would be better to keep the major at 16 and fix the image to be 16.8, to avoid having to correct again in the future.
That said, the fix works.

sxd (approved):

sxd:
@theadamwright can you please fix the DCO issue?

gbartolini:
Duplicate, already merged.

PR #7394, opened 2025-04-18
chore: always disable wal_sender_timeout in join-related pg_basebackup

fcanovai:
We used to do it only for PG version 12+. Since supported versions start at 13, we can drop the conditional.


fcanovai:
Superseded by https://github.com/cloudnative-pg/cloudnative-pg/pull/7396

PR #7409, opened 2025-04-23
Docs: Improve operator logging documentation

jmealo:


docs/src/logging.md
```diff
@@ -52,6 +52,53 @@ The log level for the operator can be configured by editing the `Deployment`
 definition of the operator and setting the `--log-level` command line argument
 to the desired value.
 
+!!! Important
+    Changing the log-level of the operator will not affect the log level of the
+    [instance manager](/instance_manager/) or other components.
+
+### Adding a custom log-level
```
jsilvela, line 59: The first section, in my opinion, should handle modifying the existing value,
which is added by default in plain deployments.

docs/src/logging.md
```diff
@@ -42,7 +42,7 @@ specification using the `logLevel` option. Available log levels are: `error`,
     Changes to the log level in the cluster specification after the cluster has
     started will only apply to new pods, not existing ones.
 
-## Operator Logs
+## Operator Logs {#operator-logs}
```
jsilvela, line 45: I had no idea about this! I guess Mkdocs magic? Trying to find a ref for it...

PR #7443, opened 2025-04-28
docs: qualified resource type names

### LINKED ISSUE #6308: [Docs]: Prefix the cnpg resource type in kubectl examples

lbruun:
Add best practice: qualify the resource type name. 

Fixes #6308.

The potential conflicts on resource names already exists as noted by others. So not just theoretical scenario.

Note, that the problem is amplified by the fact that you cannot look into the future. Who knows what kind of operators/CRDs may be installed on your K8s cluster in the future?

Because Kubernetes project will not change how `kubectl`  works ([ref](https://github.com/kubernetes/kubectl/issues/1704)) the best that can be done from CloudNativePG is to promote use a fully qualified names.

### DIFF
--- a/docs/src/cloudnative-pg.v1.md
+++ b/docs/src/cloudnative-pg.v1.md
@@ -17,6 +17,25 @@
 - [ScheduledBackup](#postgresql-cnpg-io-v1-ScheduledBackup)
 - [Subscription](#postgresql-cnpg-io-v1-Subscription)
 
+
+!!! Important
+    Because the Kubernetes resource type names used by CloudNativePG are fairly generic, for example, "backup", they 
+    may in their short form conflict with resource types from other operators unrelated to CloudNativePG. When used 
+    with tools such as `kubectl` it is therefore advisable to use the _fully qualified form_ of the resource type name 
+    which includes the namespace.
+
+    For example:
+
+    `kubectl delete cluster.postgresql.cnpg.io mycluster`
+
+    is future proof and unambiguous.
+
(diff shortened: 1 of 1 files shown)


jsilvela (changes requested):
I think this is the wrong place, as the cloudnative-pg.v1.md file is autogenerated.
IMO the effect would be better by putting it in one of the template files, e.g. docs/markdown/pkg.tpl
Additionally, not convinced that this warning does any good in the API doc, which nobody will read in order.

PR #7470, opened 2025-05-02
docs:  add OpenBao as OSS alternative to Vault

gusfcarvalho:
Hashicorp Vault, though commonly used, does have a restrictive, BUSL license that most organizations cannot use.

Adding mention to https://openbao.org , which is MPL, thus really open source for users :smile:


gbartolini (approved):

sxd (approved):

leonardoce:
Thanks @gusfcarvalho !!

PR #7476, opened 2025-05-02
fix(cnpg-plugin): Primary start time

### LINKED ISSUE #7428: [Bug]: kubectl-cnpg status  - Primary start time

### Contact Details

_No response_

### Version

1.26.0-rc1

1.29

Other

YAML manifest

### What happened?

this issue is a follow up on the [ slack discussion](https://cloud-native.slack.com/archives/C08MAUJ7NPM/p1745496187763419?thread_ts=1745337190.130179&cid=C08MAUJ7NPM) 

The `Primary start time`: 
is the time since that instance has been chosen as (designated) primary.  It is actually updated only on the initial bootstrap, or a primary change.

The expected  `Primary start time` should be the time where the primary server was started up

### Cluster resource

```shell

```

### Relevant log output

```shell

```

### Code of Conduct

smiyc:
fixes #7428

Change the status output from "Primary start time" to "Primary promotion time"

Signed-off-by: Daniel Chambre <smiyc@pm.me>

### DIFF
--- a/internal/cmd/plugin/status/status.go
+++ b/internal/cmd/plugin/status/status.go
@@ -280,7 +280,7 @@ func (fullStatus *PostgresqlStatus) printBasicInfo(ctx context.Context, k8sClien
 		// Avoid printing the start time when hibernated or fenced
 		primaryStartTime := getPrimaryStartTime(cluster)
 		if len(primaryStartTime) > 0 {
-			summary.AddLine("Primary start time:", primaryStartTime)
+			summary.AddLine("Primary promotion time:", primaryStartTime)
 		}
 		summary.AddLine("Status:", fullStatus.getStatus(cluster))
 	}


smiyc:
@armru do you mind reviewing this

armru (approved):

internal/cmd/plugin/status/status.go
```diff
@@ -280,7 +280,7 @@ func (fullStatus *PostgresqlStatus) printBasicInfo(ctx context.Context, k8sClien
 		// Avoid printing the start time when hibernated or fenced
 		primaryStartTime := getPrimaryStartTime(cluster)
 		if len(primaryStartTime) > 0 {
-			summary.AddLine("Primary start time:", primaryStartTime)
+			summary.AddLine("Primary promotion time:", primaryStartTime)
```
sxd, line 283: Why not add start/promotion time string? because it's both right? I haven't check yet the full code

armru: The primary could technically restart and the time wouldn't change AFAIK (that's why it isn't reliable as start time)

PR #7572, opened 2025-05-14
feat: introduce experimental support for failover quorum

### LINKED ISSUE #7481: [Bug]: reconcileTargetPrimaryForNonReplicaCluster does not verify quorum

### Contact Details

schneider@ardentperf.com

### Version

trunk (main)

1.32

YAML manifest

### What happened?

i'm reviewing the code in `reconcileTargetPrimaryForNonReplicaCluster()` and it might be missing some logic?

i'm thinking automatic promotion as part of self-healing is only safe (ie. no data loss) in conjunction with sync replication. i'm not sure we should fully trust any methodologies based on trying to flush WAL.

* Quorum-Based: only consider promoting the first N entries in instancesStatus[], where N is `synchronous.number - unreachable_replicas` *(if synchronous.number == 0 then we should not auto promote anyone. possibly we should default synchronous.number to `1` on all cnpg clusters with at least one replica.)*
* Priority-Based: only consider promoting entries in instancesStatus[] that are known sync replicas

this is fairly complex to reason through, so would be good to discuss further

### Cluster resource

```shell

```

### Relevant log output

```shell

```

### Code of Conduct

leonardoce:
This commit adds opt-in, experimental support for "failover quorum", also known as "quorum-based failover", in CloudNativePG.

Failover quorum is a mechanism designed to improve data durability and safety during failover events by ensuring that the promoted replica contains all synchronously committed data.

With synchronous replication, a transaction is acknowledged only after all required synchronous standbys have received the WAL data. However, this alone doesn't guarantee that the operator can always promote the most advanced replica during a failure.

The failover quorum mechanism addresses this risk by requiring the operator to verify that a quorum of replicas agrees on the promotion. This ensures that the selected replica has all required committed data. If this quorum is not met, failover will not proceed, reducing the risk of data loss.

To enable the feature, set the annotation: `cnpg.io/failoverQuorum="true"` in the `Cluster` resource.

Once stabilised, a dedicated configuration field will replace this annotation.

For further information, refer to the included documentation and the related issue.

Closes #7481


armru (approved):

mnencia (approved):

fcanovai:
We should add a check to make sure that the value of `synchronous.number` is higher than the amount of names in `standbyNamesPre + standbyNamesPost`, otherwise people that are using synchronous replica cluster with the synchronous quorum could lock themselves out of failover. We could send them a warning. This is the case of third example in the original issue.

gbartolini (approved):

fcanovai (approved):

gbartolini:
@ardentperf, I apologise for not mentioning your contributions in the commit itself. I will make sure we add you in #8170.

PR #7612, opened 2025-05-21
Fix: Default synchronous_commit method to 'any'

GellertBabel:
When creating a Cluster via OLM with only default settings, the `spec.postgresql.synchronous.method` field was being submitted as an empty string, leading to a validation error: "Unsupported value: "": supported values: "any", "first"".

This change modifies the defaulting logic in `api/v1/cluster_defaults.go` to set `spec.postgresql.synchronous.method` to "any" if the `synchronous` block is present but the `method` is not specified.

Unit tests have been added to `api/v1/cluster_defaults_test.go` to verify:
- The method defaults to "any" when the synchronous block is present and the method is empty.
- The synchronous block remains nil if not initially defined.
- An existing method value is not overwritten by the default.

### DIFF
--- a/api/v1/cluster_defaults.go
+++ b/api/v1/cluster_defaults.go
@@ -102,6 +102,13 @@ func (r *Cluster) setDefaults(preserveUserSettings bool) {
 		r.Spec.PostgresConfiguration.Parameters = sanitizedParameters
 	}
 
+	// Default synchronous_commit method if not specified.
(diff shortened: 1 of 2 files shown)


gbartolini:
Isn't this related to #7599 ?

mnencia:
Yes, this is related to #7599. The solution here is not to add a default value to the `method` field, but rather to remove the default for the `dataDurability` field, leaving that entire section empty.

PR #7662, opened 2025-05-28
feat(dev-env): allow customizing the image loading timeout

hanshal101:
### Title
Increase context timeout in dev setup script to avoid deadline error

### Description

This PR updates the development setup script to increase the context timeout from 300 seconds to 600 seconds.
The change addresses intermittent context deadline exceeded errors observed during environment initialization.

### Changes

- Increased timeout value to ensure setup completes reliably on slower systems or with unstable network conditions.

- No other modifications made.

This is a minor but necessary adjustment to improve developer experience and reduce setup failures.


leonardoce (changes requested):

leonardoce (changes requested):

hack/setup-cluster.sh
```diff
@@ -562,10 +570,10 @@ pyroscope() {
 main() {
   if ! getopt -T > /dev/null; then
     # GNU enhanced getopt is available
-    parsed_opts=$(getopt -o e:k:n:r -l "engine:,k8s-version:,nodes:,registry" -- "$@") || usage
+    parsed_opts=$(getopt -o e:k:n:r -l "engine:,k8s-version:,nodes:,registry,fluentd-timeout:,csi-timeout:" -- "$@") || usage
```
leonardoce, line 573: we still miss `f:c:` in the list of short options

hack/setup-cluster.sh
```diff
@@ -598,10 +606,30 @@ main() {
...
+        echo "ERROR: CSI timeout must be a positive integer!" >&2
+        echo >&2
+        usage
+        fi
+        ;;
     --)
       shift
       break
-      ;;
+        ;;
```
leonardoce, line 632: I think the indentation we had was fine

hack/setup-cluster.sh
```diff
@@ -562,10 +570,10 @@ pyroscope() {
 main() {
   if ! getopt -T > /dev/null; then
     # GNU enhanced getopt is available
-    parsed_opts=$(getopt -o e:k:n:r -l "engine:,k8s-version:,nodes:,registry" -- "$@") || usage
+    parsed_opts=$(getopt -o e:k:n:r -l "engine:,k8s-version:,nodes:,registry,fluentd-timeout:,csi-timeout:" -- "$@") || usage
   else
     # Original getopt is available
-    parsed_opts=$(getopt e:k:n:r "$@") || usage
+    parsed_opts=$(getopt e:k:n:r:f:c "$@") || usage
```
leonardoce, line 576: By doing that, we mark `r` as requiring an argument, meanwhile `c`is not requiring it.
I think we should use `e:k:n:f:c:r ` instead of `e:k:n:r:f:c`, and the same applies to the previous getopt call.

leonardoce (changes requested):

PR #7725, opened 2025-06-04
feat(auth): support user maps for predefined users to enable self-managed certificates

### LINKED ISSUE #7697: [Feature]: support user maps for fixed users or allow changing those users

CNPG uses certificates for "system" user connections. Both `streaming_replica` and `cnpg_pooler_pgbouncer` are forced to use cert auth with no option of overriding it as `pg_hba` entries are checked in order and matching stops on the first match.

That means #3534 actually didn't fix the issue described in #3362 - you cannot alias any name to `streaming_replica` because there is no way to add the `pg_ident` mapping to those entries.

Provide a way to specify `pg_ident` entries - easiest way would be to just add fixed ident maps to each entry, like it's done for the `local` authentication. With custom `pg_ident` entries already supported that should be a simple change.

Allow using different users for local replication. This could reuse the  replication's`self` entry from the external clusters section. In my opinion `pg_ident` solution is better and simpler in both implementation and usage.

### Additional context

In environments where those specific CNs are not allowed (e.g. in our env all certificates are required to use full domains) it's not possible to scale cluster instances beyond 1. You can create cluster, you can even connect an external one (using a custom user) but cannot join a local replica.

As for the pooler, I believe it might not even be possible to provide a custom certificate, I don't see any option to use custom certs at all, only the fully manual mode (which seems to use passwords exclusively).

### Backport?

No

Yes

### Code of Conduct

Agalin:
Add support for default user maps for predefined users, allowing administrators to remap certificate identities and enable the `streaming_replica` user to authenticate with certificates using different Common Names.

This improves compatibility and flexibility when using self-managed PKI infrastructures.

Closes #7697

### DIFF
--- a/docs/src/certificates.md
+++ b/docs/src/certificates.md
@@ -265,18 +265,18 @@ the following parameters:
 
 #### Customizing the `streaming_replica` client certificate
 
-In some environments it might not be possible to generate a certificate with the common
-name set to `streaming_replica`  due to a company policy or other security concerns,
-like CA shared across multiple clusters. In such cases the user mapping feature can be
-used to allow authentication as the `streaming_replica` user with certificates
-containing different common names.
+In some environments, it may not be possible to generate a certificate with the
+common name `streaming_replica` due to company policies or other security
+concerns, such as a CA shared across multiple clusters. In such cases, the user
+mapping feature can be used to allow authentication as the `streaming_replica`
+user with certificates containing different common names.
 
-The only thing necessary to configure such setup is to add a `pg_ident.conf` entry
-for a predefined map named `cnpg_streaming_replica`.
+To configure this setup, add a `pg_ident.conf` entry for the predefined map
+named `cnpg_streaming_replica`.
 
-For example, to enable `streaming_replica` authentication using a certificate with
-a common name `streaming-replica.cnpg.svc.cluster.local` this should be added to the
-cluster definition:
+For example, to enable `streaming_replica` authentication using a certificate
+with the common name `streaming-replica.cnpg.svc.cluster.local`, add the
+following to your cluster definition:
 
 ```yaml
 apiVersion: postgresql.cnpg.io/v1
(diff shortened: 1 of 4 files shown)


armru:
Hello, thanks for the contribution, adding to the review queue for 1.27.0

mnencia:
@Agalin This patch is very interesting, but we cannot rebase it because maintainers don't have permission to write in the branch. Could you please rebase the patch and consider granting maintainers write permissions on the branch?

Agalin:
I've rebased the branch. Is it even possible to give access to a single branch in Github?

Agalin:
@mnencia considering the frequency of merges, I've invited all the maintainers to the fork project with write access.

mnencia:
> I've rebased the branch. Is it even possible to give access to a single branch in Github?

When you open a pull request, a flag gives the maintainer access to the branch. When I open my PRs, this flag is enabled by default.

mnencia:
> @mnencia considering the frequency of merges, I've invited all the maintainers to the fork project with write access.

I've received nothing and the push doesn't work:

```
➤  git push --force-with-lease
remote: Permission to operasoftware/cloudnative-pg.git denied to mnencia.
fatal: impossibile accedere a 'https://github.com/operasoftware/cloudnative-pg.git/': The requested URL returned error: 403
```

Agalin:
> > I've rebased the branch. Is it even possible to give access to a single branch in Github?
> 
> When you open a pull request, a flag gives the maintainer access to the branch. When I open my PRs, this flag is enabled by default.

Maybe it's related to org's feature plan? I definitely don't see such an option when creating a PR. No flags at all.

Ah. Looking into [docs](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork), it seems that it only works for personal forks. Will discuss internally if we can skip forking to org.

> People with push access to the upstream repository of a fork in a personal account can commit to the forked branches.

> I've received nothing and the push doesn't work

You're definitely on the invited list @mnencia. 
Zrzut ekranu 2025-07-21 o 18 25 08

mnencia:
Thanks, it worked. I've been able to rebase the branch.

mnencia (approved):

gbartolini (approved):

PR #7792, opened 2025-06-09
feat: extend service account usage

QuantumEnigmaa:
This PR is a response to [this discussion](https://github.com/cloudnative-pg/cloudnative-pg/discussions/4219) mentioning the need to be able to make the cluster resource use an existing service account.

This introduced change replaces the `serviceAccountTemplate` section in the cluster resource's spec which only allow the user to specify the metadata for the service account that will be created by a `ServiceAccount` section allowing the user to choose between using an existing service account or creating one. 

This is a first draft and I'm pretty sure there are a lot of things to improve here but that's still a beginning.


QuentinBisson:
Hey @sxd is there any way to push this PR forward?

Hazmi35:
We’re very interested in this feature. Our setup uses Azure federated credentials, which require exact ServiceAccount names (no wildcards or patterns allowed).

Right now, because CNPG creates a new ServiceAccount per cluster, we’d have to register each one in Azure. That makes workflows like PITR testing and especially spinning up PITR “copy” clusters (temporary databases for testing, validation, or other use cases) impractical, since every new cluster requires additional Azure config.

lieberlois:
100% agree. I can implement this if somebody from the maintainers gives me some pointers where to look, it's my first time contributing here. 

Also need it for Azure Workload Identity.

QuentinBisson:
Hey @armru as you're testing PRs, do you know if we can expect this to be merged?

armru:
This PR is not being tested by me, and more in general I don't do the testing in the project, if there is some mentions of that it should be corrected because it stale and/or wrong

QuentinBisson:
Oh my bad, i've seen you on so much PRs that I assumed you were a maintainer 🤦🏻

PR #7794, opened 2025-06-10
Fixes Issue 7793: Stuck reconciliation fixes

### LINKED ISSUE #7793: [Bug]: Cluster reconciliation loop gets stuck when scaling replicas due to failed snapshot recovery jobs

### Contact Details

jeffm@gisual.com

### Version

1.26 (latest patch)

1.31

Cloud: Azure AKS

YAML manifest

### What happened?

### To Reproduce

### Expected behavior

### Actual behavior

### Workaround

   ```bash
   ```

   ```bash
   ]' --subresource=status
   ```

### Supporting data

  ```yaml
  status:
    phase: Creating a new replica
  ```

- **Job Pod Events:**
  ```
  Events:
  ```

### System information

### Additional context

### Cluster resource

```shell

```

### Relevant log output

```shell

```

### Code of Conduct

jmealo:
# Fix stuck reconciliation loop for failed snapshot recovery jobs #7793

## Problem
CloudNativePG clusters get stuck in "Creating a new replica" phase when snapshot recovery jobs fail, even when the cluster already has the correct number of replicas. Currently requires manual intervention to recover.

## Solution
Adds automatic detection and recovery:
- Detects stuck/failed jobs (10min timeout)
- Validates if cluster already has correct replica count
- Clears stuck scaling phase and deletes failed jobs
- Returns cluster to normal operation

### New Files
- `pkg/utils/jobs.go` - Job state detection utilities
- `pkg/utils/jobs_test.go` - Unit tests

### Modified
- `internal/controller/cluster_controller.go` - Added stuck detection/recovery logic
- `api/v1/cluster_types.go` - Added scaling phase constants

### Example Scenario
```
1. Cluster scaling 9→10 replicas
2. Snapshot recovery job fails (missing PVC)
3. Instance created via streaming replication (now has 10 replicas)
4. Cluster stuck in "Creating a new replica" despite having 10 instances

Fix automatically:
✅ Detects stuck state
✅ Verifies has 10 instances (correct count)
✅ Clears phase → Healthy
✅ Deletes failed job
```

### Before Merging
- [ ] Test normal scaling still works
- [ ] Verify stuck job detection (10min timeout)
- [ ] Confirm phase clears when replica count correct
- [ ] Test recovery with missing PVCs
- [ ] Check no impact on healthy clusters

### Test Instructions
1. Create cluster, scale up
2. Delete PVC to cause job failure
3. Verify automatic recovery

## Configuration
- Stuck job timeout: **10 minutes** (hardcoded)
- Equilibrium timeout: **15 minutes** (hardcoded)
- Future: Make configurable

## Risks
- 10min timeout might be aggressive for large restores
- Need to ensure no race conditions with legitimate scaling

## Future Work
- Configurable timeouts
- Prometheus metrics
- Root cause fixes for job failures

---

**Questions for Reviewers:**
1. Are the timeout values (10/15min) appropriate?
2. Should timeouts be configurable now vs later?
3. Any additional test scenarios needed?

## ⚠️

### DIFF
--- a/api/v1/cluster_types.go
+++ b/api/v1/cluster_types.go
@@ -691,6 +691,12 @@ const (
 
 	// PhaseCannotCreateClusterObjects is set by the operator when is unable to create cluster resources
 	PhaseCannotCreateClusterObjects = "Unable to create required cluster objects"
+
+	// PhaseScalingUp when scaling up the cluster
+	PhaseScalingUp = "Scaling up"
+
+	// PhaseScalingDown when scaling down the cluster
+	PhaseScalingDown = "Scaling down"
 )
 
 // EphemeralVolumesSizeLimitConfiguration contains the configuration of the ephemeral
--- a/internal/controller/cluster_controller.go
+++ b/internal/controller/cluster_controller.go
@@ -72,6 +72,20 @@ const (
 	poolerClusterKey              = ".spec.cluster.name"
 	disableDefaultQueriesSpecPath = ".spec.monitoring.disableDefaultQueries"
 	imageCatalogKey               = ".spec.imageCatalog.name"
+
+	// Timeout constants for stuck reconciliation handling
+	// defaultStuckJobTimeout is the duration after which a job with no progress is considered stuck
+	// This should be long enough to allow for normal PostgreSQL operations like backup restoration
+	// but short enough to prevent indefinite blocking of cluster operations
+	defaultStuckJobTimeout = 10 * time.Minute
+
+	// equilibriumStateTimeout is the duration after which we consider the cluster to be in
+	// an equilibrium state (stuck without progress). This should be longer than stuckJobTimeout
+	// to allow for job cleanup and retry cycles
+	equilibriumStateTimeout = 15 * time.Minute
+
+	// runningJobRecheckInterval is how often we recheck running jobs for failures or stuck states
+	runningJobRecheckInterval = 5 * time.Second
 )
 
 var apiSGVString = apiv1.SchemeGroupVersion.String()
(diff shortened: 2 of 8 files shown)


mnencia:
@jmealo What is missing to move this in ready for review?

jmealo:
@mnencia I haven't been able to test it on a cluster yet. I've only run the automated tests.

jmealo:
@mnencia I still haven't had a chance to test this in a cluster. I likely won't have time for another month.

jmealo:
@armru @mnencia I'm marking this as ready-for-review, but I could use a hand testing it. The changes made make sense, their implementation may need some work.

internal/controller/cluster_controller.go
```diff
@@ -723,6 +723,103 @@ func (r *ClusterReconciler) setDefaults(ctx context.Context, cluster *apiv1.Clus
...
+// but already has the correct number of instances, and clears the phase if needed
+func (r *ClusterReconciler) checkAndClearStuckScalingPhase(
+	ctx context.Context,
+	cluster *apiv1.Cluster,
+	resources *managedResources,
+) error {
+	contextLogger := log.FromContext(ctx)
+
+	// Check if we're in a scaling-related phase
+	scalingPhases := []string{
```
jmealo: Do these need to come from a const/be-extracted to a helper function `isInScalingPhase`? 🤔

internal/controller/cluster_controller.go
```diff
@@ -723,6 +723,103 @@ func (r *ClusterReconciler) setDefaults(ctx context.Context, cluster *apiv1.Clus
...
+}
+
+// clearStuckScalingPhaseAfterJobDeletion clears scaling phases after deleting failed jobs
+func (r *ClusterReconciler) clearStuckScalingPhaseAfterJobDeletion(
+	ctx context.Context,
+	cluster *apiv1.Cluster,
+) error {
+	contextLogger := log.FromContext(ctx)
+
+	// Check if we're in a scaling-related phase that might be stuck due to failed jobs
```
jmealo, line 811: **TODO:** Extract this to `isInScalingPhase` method.

internal/controller/cluster_controller.go
```diff
@@ -731,9 +828,60 @@ func (r *ClusterReconciler) reconcileResources(
...
+	if err := r.checkAndClearStuckScalingPhase(ctx, cluster, resources); err != nil {
+		contextLogger.Error(err, "Error while checking stuck scaling phase")
+		return ctrl.Result{}, err
+	}
+
 	// Act on Pods and PVCs only if there is nothing that is currently being created or deleted
 
 	if len(runningJobs) > 0 {
+		// Let's check for failed or stuck jobs and handle them
+		stuckJobTimeout := 10 * time.Minute // Jobs stuck for more than 10 minutes are considered failed
```
jmealo: This feels arbitrary but I'm not sure what the cut-off should be?

internal/controller/cluster_controller.go
```diff
@@ -1561,3 +1709,96 @@ func (r *ClusterReconciler) markPVCReadyForCompletedJobs(
...
+// where jobs are running but making no progress, potentially due to missing PVCs or other issues
+func (r *ClusterReconciler) checkForEquilibriumState(
+	ctx context.Context,
+	cluster *apiv1.Cluster,
+	resources *managedResources,
+) error {
+	contextLogger := log.FromContext(ctx)
+
+	// Check if we have long-running jobs that might be stuck
+	equilibriumTimeout := 15 * time.Minute // Consider equilibrium after 15 minutes
```
jmealo: This is somewhat arbitrary, should this be configurable? What would be a sane value.

internal/controller/cluster_controller.go
```diff
@@ -1561,3 +1709,96 @@ func (r *ClusterReconciler) markPVCReadyForCompletedJobs(
...
+	for _, job := range resources.jobs.Items {
+		// Skip completed or failed jobs
+		if utils.IsJobComplete(job) || utils.IsJobFailed(job) {
+			continue
+		}
+
+		// Check if job has been running for too long without progress
+		if job.CreationTimestamp.Add(equilibriumTimeout).Before(time.Now()) {
+			// Check if job has no active pods (stuck in pending)
+			if job.Status.Active == 0 && job.Status.Succeeded == 0 && job.Status.Failed == 0 {
```
jmealo, line 1748: If the job is failing to schedule for a long time, I'd be more inclined to call it stuck and consider it a failure than if it gets scheduled.

internal/controller/cluster_controller.go
```diff
@@ -1561,3 +1709,96 @@ func (r *ClusterReconciler) markPVCReadyForCompletedJobs(
...
+				contextLogger.Warning("Detected job in equilibrium state - no pods created",
+					"jobName", job.Name,
+					"creationTime", job.CreationTimestamp.Time,
+					"age", time.Since(job.CreationTimestamp.Time))
+
+				// Check if there are missing PVCs that might be causing the issue
+				if err := r.checkForMissingPVCs(ctx, cluster, &job); err != nil {
+					return fmt.Errorf("equilibrium state detected: job %s stuck due to missing PVCs: %w",
```
jmealo, line 1756: It might be useful to surface the different error types in the cluster status somehow? 🤔

jmealo:
**Issues I see:**
- Magic values for timeouts / stuck job criteria
- We should extract the `isInScalingPhase` to its own function, and instead of hard coding the strings, get them from a constant if possible

jmealo:
I'm working on fixing the issues I flagged here. Let me know if there's anything else that needs to be cleaned up.

I'm trying to find the best way to surface this condition to  folks since cleaning up stuck pods might make it difficult to troubleshoot the issue if it persists.

For example, `kubectl cnpg status` should provide helpful information.

internal/controller/cluster_controller.go
```diff
@@ -723,6 +723,103 @@ func (r *ClusterReconciler) setDefaults(ctx context.Context, cluster *apiv1.Clus
...
+
+	// Check if we're in a scaling-related phase that might be stuck due to failed jobs
+	scalingPhases := []string{
+		"Creating a new replica",
+		"Scaling up",
+		"Scaling down",
+	}
+
+	isInScalingPhase := false
+	for _, phase := range scalingPhases {
```
jmealo: Remove loop

internal/controller/cluster_controller.go
```diff
@@ -72,6 +72,20 @@ const (
 	poolerClusterKey              = ".spec.cluster.name"
 	disableDefaultQueriesSpecPath = ".spec.monitoring.disableDefaultQueries"
 	imageCatalogKey               = ".spec.imageCatalog.name"
+
+	// Timeout constants for stuck reconciliation handling
+	// defaultStuckJobTimeout is the duration after which a job with no progress is considered stuck
+	// This should be long enough to allow for normal PostgreSQL operations like backup restoration
+	// but short enough to prevent indefinite blocking of cluster operations
+	defaultStuckJobTimeout = 10 * time.Minute
```
jmealo, line 80: I have no idea what's a reasonable value here. Thoughts?

api/v1/cluster_types.go
```diff
@@ -673,6 +673,12 @@ const (
 
 	// PhaseCannotCreateClusterObjects is set by the operator when is unable to create cluster resources
 	PhaseCannotCreateClusterObjects = "Unable to create required cluster objects"
+
+	// PhaseScalingUp when scaling up the cluster
+	PhaseScalingUp = "Scaling up"
```
jmealo, line 696: We didn't have these in a const, so I added them here, not sure if that's the way to go?

jmealo:
@armru @mnencia I'm closing this PR because it's unwieldy and tries to do too much. Now that I've had some time to think about it, a more iterative approach would be best.

This PR is superseded by:
#10004
#10005
  
I've implemented the approach below:

### Proposed Approach

**PR 1: Handle failed/deleted jobs (core fix)**
- When a job enters a `Failed` state, ensure it doesn't block reconciliation (`runningJobNames()` excluding failed jobs)
- Clean up failed jobs in `cleanupCompletedJobs`
- Add a safety net: when the cluster is in a scaling phase but `currentInstances == desiredInstances` with zero running jobs, clear the phase to Healthy. This handles the "manually delete the job" workaround and any other edge cases where the phase gets stuck despite the cluster being fine.

All signals are deterministic (no timeouts).

**PR 2: Fallback from snapshot to pg_basebackup**
- When a snapshot-recovery job for a replica fails, fall back to `pg_basebackup` instead of retrying the broken snapshot
- Self-healing rather than just unblocking

**Future: Stuck job detection**
Timeout-based detection of jobs stuck in Pending is useful but tricky to get right. Probably best to start narrow (job age > N minutes with `Active == 0 && Succeeded == 0 && Failed == 0`) and make timeouts configurable. Separate PR to keep scope manageable. I think this is the most difficult to "get right", so let's punt on it and get the concrete fixes out sooner rather than later.

PR #7801, opened 2025-06-10
fix(import): do not create default user and database in monolith mode

### LINKED ISSUE #7447: [Bug]: wrong db owner when using monolith import and db name is app

### Contact Details

_No response_

### Version

1.26.0-rc2

1.29

Other

YAML manifest

### What happened?

when using .spec.bootstrap.import.monolith and the db name to be imported is "app" the initial db owner is replaced by the "app" user.

source cluster:
```
postgres=# \l
                                                List of databases
   Name    |  Owner   | Encoding | Locale Provider | Collate | Ctype | Locale | ICU Rules |   Access privileges   
-----------+----------+----------+-----------------+---------+-------+--------+-----------+-----------------------
 app       | xxx      | UTF8     | libc            | C       | C     |        |           | 

```

import monolith cluster:
```
postgres=# \l
                                                     List of databases
   Name    |  Owner   | Encoding | Locale Provider |   Collate   |    Ctype    | Locale | ICU Rules |   Access privileges   
-----------+----------+----------+-----------------+-------------+-------------+--------+-----------+-----------------------
 app       | app      | UTF8     | libc            | en_US.UTF-8 | en_US.UTF-8 |        |           | 
```

when the to be imported db is <> app, the correct owner is used

### Cluster resource

```shell
source cluster:

kind: Cluster
metadata:
  name: cluster-example
spec:
  instances: 1

  postgresql:
    pg_hba:
      - host all all all trust

  storage:
    size: 1Gi

  bootstrap:
    initdb:
      database: app
      owner: xxx



import.monolith cluster

kind: Cluster
metadata:
  name: cluster-example-monolith
spec:
  instances: 1
  
  storage:
    size: 1Gi

  bootstrap:
    initdb:
      encoding: UTF8
      dataChecksums: true
      localeCType: en_US.UTF-8
      localeCollate: en_US.UTF-8
      import:
        type: monolith
        databases:
          - app
        roles:
          - xxx
        source:
          externalCluster: cluster-example

  externalClusters:
    - name: cluster-example
      connectionParameters:
        host: cluster-example-rw
        dbname: postgres
```

### Relevant log output

```shell

```

### Code of Conduct

### LINKED ISSUE #7731: [Bug]: monolith import may create unintended default database

### Contact Details

jaime.silvela@mailfence.com

### Version

1.26 (latest patch)

1.33

YAML manifest

### What happened?

Investigating #7447 , the issue opened by @smiyc , I agree with his assessment that it is a bug,
and moreover I find the mis-behavior bigger than just that bug.

In this description I discuss the issue, compare with microservice approach, and propose the fix.
Section headings to hopefully aid navigation.

## monolith import may take in N source databases and create N + 1, or alter the "app" database owner

To recap from #7447 , when using monolith import, a `(database, owner)` of `("app", "xxx")` results in an imported database of `("app", "xxx")`, whereas a `(database, owner)` of `("foo", "xxx")` is imported, as expected, as `("foo", "xxx")`.

Now, in the second case, the monolith import doesn't create only `("foo", "xxx")`, but adds the default `("app", "app")` too.

Doing a `kubectl get -o yaml` of the running cluster reveals:

``` yaml
spec:
  <snipped>
  bootstrap:
    initdb:
      database: app
      import:
        databases:
        - foo
        roles:
        - xxx
        source:
          externalCluster: cluster-other
        type: monolith
      owner: app
```

In the documentation for the monolith import (https://cloudnative-pg.io/documentation/current/database_import/#the-monolith-type), the expectation is that N `.initdb.import.databases` will result in N databases in the created cluster.
This example shows this is not the case.

The cause is the defaulting webhook, which will create the default (database, user) if there are no `initdb.database` and `initdb.owner` specified.

## comparison with microservice

The defaulting webhook behavior makes sense for the microservice approach.
There, a single DB from the source cluster is chosen to be imported, and in the target cluster will result in `("app", "app")` by default, unless the user specifies a `(database, owner)`.

For monolith this does not make sense. Firstly, monolith is more useful for N databases to import rather than just one.
And in addition, the monolith approach seems tailor made for copying from the source cluster as faithfully as possible.

## proposal

The right fix, in my view, is that the defaulting webhook should not pre-populate the database/owner, if there is a monolith import specified.

Alternatives considered:

Force the user of monolith imports to specify explicitly which of the imported databases is "default" via the `initdb.database` and `initdb.owner`.
 
Also, the documentation needs some updating. And ... we don't have a sample file for monolith import, which shows that it's been taken less seriously than microservice.

### Cluster resource

```shell
kind: Cluster
metadata:
  name: cluster-example-new-monolith
spec:
  instances: 1
  
  storage:
    size: 1Gi

  bootstrap:
    initdb:
      encoding: UTF8
      database: app
      owner: xxx
      dataChecksums: true
      localeCType: en_US.UTF-8
      localeCollate: en_US.UTF-8
      import:
        type: monolith
        databases:
          - app
        roles:
          - xxx
        source:
          externalCluster: cluster-example

  externalClusters:
    - name: cluster-example
      connectionParameters:
        host: cluster-example-rw
        dbname: postgres
```

### Relevant log output

```shell

```

### Code of Conduct

jsilvela:
When importing in monolith mode, avoid automatically creating a default user and database. This approach ensures that the import process accurately mirrors the source environment, preventing the creation of unnecessary or unintended objects. Additionally, the documentation has been updated to clarify the import behavior and usage in monolith scenarios.

Closes #7731
Closes #7447

### DIFF
--- a/docs/src/database_import.md
+++ b/docs/src/database_import.md
@@ -252,6 +252,105 @@ There are a few things you need to be aware of when using the `monolith` type:
   database.
 - `postImportApplicationSQL` field is not supported
 
+## Comparing microservice vs monolith imports
+
+There is nothing to stop you from using the `monolith` approach to import a
+single database. It is interesting to see how the results of doing so would
+differ from using the `microservice` approach.
+
+Given a source cluster, for example the following, with a database named
+`mydb` owned by role `me`:
+
+``` yaml
+apiVersion: postgresql.cnpg.io/v1
+kind: Cluster
+metadata:
+  name: cluster-example
+spec:
+  instances: 1
+
+  postgresql:
+    pg_hba:
+      - host all all all trust
+
+  storage:
+    size: 1Gi
+
+  bootstrap:
+    initdb:
+      database: mydb
+      owner: me
+```
+
+We can import it via `microservice`:
+
+``` yaml
+apiVersion: postgresql.cnpg.io/v1
+kind: Cluster
+metadata:
+  name: cluster-example-microservice
+spec:
+  instances: 1
+  
+  storage:
+    size: 1Gi
+
+  bootstrap:
+    initdb:
+      import:
+        type: microservice
+        databases:
+          - mydb
+        source:
+          externalCluster: cluster-example
+
+  externalClusters:
+    - name: cluster-example
+      connectionParameters:
+        host: cluster-example-rw
+        dbname: postgres
+```
+
+as well as via monolith:
+
+``` yaml
+apiVersion: postgresql.cnpg.io/v1
+kind: Cluster
+metadata:
+  name: cluster-example-monolith
+spec:
+  instances: 1
+  
+  storage:
+    size: 1Gi
+
+  bootstrap:
+    initdb:
+      import:
+        type: monolith
+        databases:
+          - mydb
+        roles:
+          - me
+        source:
+          externalCluster: cluster-example
+
+  externalClusters:
+    - name: cluster-example
+      connectionParameters:
+        host: cluster-example-rw
+        dbname: postgres
+```
+
+In both cases, the database's contents will be imported, but:
+
+- In the microservice case, the database and owner both become `app`. You can
+  configure the database and onwer via the `bootstrap.initdb` stanza's
+  `database` and `owner` fields, as always
+- In the monolith case, the database and owner as kept as in the source cluster,
+  i.e. `mydb` and `me` respectivelly. No `app` database nor user will be
+  created.
+
 ## Import optimizations
 
 During the logical import of a database, CloudNativePG optimizes the
--- a/docs/src/database_import.md
+++ b/docs/src/database_import.md
@@ -345,12 +345,15 @@ spec:
 
 In both cases, the database's contents will be imported, but:
 
-- In the microservice case, the database and owner both become `app`. You can
-  configure a different database and owner via the `bootstrap.initdb` stanza's
-  `database` and `owner` fields.
+- In the microservice case, the imported database's name and owner both become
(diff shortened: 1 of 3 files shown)


jsilvela:
On your commit d2f631, @mnencia , I thought about that but,
if we have a monolith import and the cluster doesn't declare (database,owner) in the bootstrap,
why should we still apply the defaults for encoding?

mnencia:
> On your commit d2f631, @mnencia , I thought about that but,
> if we have a monolith import and the cluster doesn't declare (database,owner) in the bootstrap,
> why should we still apply the defaults for encoding?
> 

Because those are global and affect the catalog encoding

docs/src/database_import.md
fcanovai, line 1: The added paragraph seems mostly redundant. How to do microservices and monolith imports has been explained in the previous paragraphs, and the renaming of the database for the microservice import is in the hint  block.

I think we could just add the final item to the monolith import paragraph.

jsilvela: Not sure which added paragraph you mean specifically. Perhaps you mean the whole `## Comparing microservice vs monolith imports` subsection?
It can probably be cut down in size. (Ber in mind the hint block about the renaming came after this PR).
That said, I don't think it's obvious from the existing documentation what the difference between the methods is. Why have the microservice method at all? Could you not do a monolith import for a single DB?

gbartolini: I have added a 'hint' section in the monolith block and renamed the new section "A practical example."

mnencia (approved):

gbartolini (approved):

Issue #3242, opened 2023-10-30
[Feature]: Cluster without default user


baurmatt:
### Is there an existing issue already for this feature request/idea?

- [X] I have searched for an existing issue, and could not find anything. I believe this is a new feature request to be evaluated.

### What problem is this feature going to solve? Why should it be added?

We use the new declarative role management to create an admin user:

```
...
  managed:
    roles:
    - connectionLimit: -1
      createdb: true
      createrole: true
      ensure: present
      inherit: true
      login: true
      name: admin
      passwordSecret:
        name: cnpg-admin-secret
```

The user of the DB is supposed to handle user/db creation them self hence the default app user/db created by bootstrap.initdb isn't needed.

### Describe the solution you'd like

We like to see an option where none of the bootstrap options (initdb/pg_basebackup/recovery) is used e.g. by configuring `bootstrap: false`.

### Describe alternatives you've considered

I've checked, but couldn't find any option to disable the `app` db/user creation.

### Backport?

No

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

gbartolini:
> We like to see an option where none of the bootstrap options (initdb/pg_basebackup/recovery)

What does that even mean?

baurmatt:
That's correct, yes. Sorry for being inaccurate about this!

cterence:
We are having the same problem with clusters bootstrapped from external non-CNPG databases. The operator keeps trying to alter the app role after booting the base and fails with `ERROR: role \"app\" does not exist`.
We don't use the app database and role so we would like them to be togglable.

T1mK:
I also get this error after using initdb.import from a non-cnpg deployment of cnpg that doesn't have an "app" anything. I'm importing one database and one role, which works flawlessly. The database, owner, and password all appear correct and work. However, the <cluster-name>-app secret gets created nevertheless with the username as "app" and the Reconciler Error regularly appearing stating that <"error":"while updating database owner password: ERROR: role \"app\" does not exist (SQLSTATE 42704)"> . 

Is there a known documented solution to this that I'm missing, or some magic workaround other than manually creating the "app" role to make cnpg happy once again?

Thanks a lot for you help!

baurmatt:
Dear @github-actions, this issues might not have been touch in 60 days but is still a requested feature from 3 user which certainly still like to see it implemented.

baurmatt:
[...]

PR #7809, opened 2025-06-12
feat(bootstrap): stop defaulting database field

mnencia:
This pull request changes the default behavior of the `database` fields in all bootstrap methods (initdb, pg_basebackup, and recovery). If neither the `database` nor the `owner` fields are explicitly set, no application-specific database or role will be created during the bootstrap process. However, if only the `database` value is provided, the existing behavior will remain unchanged.

Partially implements #3242


dosmanak:
Hi. I'd like to see this feature merged because when I need to have a cluster with more than one database and I want to use declarative Database management I want to declare ALL my database as Database resources.
The initdb `app` db is only confusing then.

PR #7811, opened 2025-06-12
chore(initdb) remove app-db-name and app-user defaults

### LINKED ISSUE #7810: [Chore]: remove app-db-name and app-user defaults for initdb subcommand

In the following [code](https://github.com/EnterpriseDB/cloud-native-postgres/blob/5d2a1a3e994947d016775fcbff7fe330585333ef/internal/cmd/manager/instance/initdb/cmd.go#L137-L140) of the initdb subcommand 

```go
	cmd.Flags().StringVar(&appDBName, "app-db-name", "app",
		"The name of the application containing the database")
	cmd.Flags().StringVar(&appUser, "app-user", "app",
		"The name of the application user")
```

We require both the `appDBName` and `appUser` parameters to be filled. Currently, the defaults for the Cluster and those defined here are consistent, but this alignment could change in the future (see #3242). Therefore, we should remove the defaults from this section.

Additionally, it’s important to modify the initdb code to handle the scenario where these two parameters might be empty.

N/A

N/A

### Additional context

N/A

### Backport?

Yes

Yes

### Code of Conduct

mnencia:
Closes #7810

### DIFF
--- a/pkg/management/postgres/initdb.go
+++ b/pkg/management/postgres/initdb.go
@@ -356,11 +356,9 @@ func (info InitInfo) ConfigureNewInstance(instance *Instance) error {
 	}
 
 	var existsDB bool
-	dbRow := dbSuperUser.QueryRow(
-		"SELECT COUNT(*) > 0 FROM pg_catalog.pg_database WHERE datname = $1",
+	dbRow := dbSuperUser.QueryRow("SELECT COUNT(*) > 0 FROM pg_catalog.pg_database WHERE datname = $1",
 		info.ApplicationDatabase)
-	err = dbRow.Scan(&existsDB)
-	if err != nil {
+	if err = dbRow.Scan(&existsDB); err != nil {
 		return err
 	}
 
--- a/pkg/management/postgres/initdb.go
+++ b/pkg/management/postgres/initdb.go
@@ -339,16 +339,14 @@ func (info InitInfo) ConfigureNewInstance(instance *Instance) error {
 	var existsRole bool
 	userRow := dbSuperUser.QueryRow("SELECT COUNT(*) > 0 FROM pg_catalog.pg_roles WHERE rolname = $1",
 		info.ApplicationUser)
-	err = userRow.Scan(&existsRole)
-	if err != nil {
+	if err = userRow.Scan(&existsRole); err != nil {
 		return err
 	}
 
 	if !existsRole {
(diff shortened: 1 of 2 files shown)


jsilvela:
As you're on course to making #7801 irrelevant, I'd appreciate if you took a look and see if anything may be used @mnencia

pkg/management/postgres/initdb.go
```diff
@@ -341,6 +324,33 @@ func (info InitInfo) ConfigureNewInstance(instance *Instance) error {
...
+	filePath := filepath.Join(info.PgData, constants.CheckEmptyWalArchiveFile)
+	// We create the check empty wal archive file to tell that we should check if the
+	// destination path it is empty
+	if err := fileutils.CreateEmptyFile(filepath.Clean(filePath)); err != nil {
+		return fmt.Errorf("could not create %v file: %w", filePath, err)
+	}
+
+	if info.ApplicationUser == "" {
+		return nil
+	}
```
jsilvela, line 337: Does it make sense that if the ApplicationDatabase is empty we create the ApplicationRole but not viceversa?
Logic would be easier to require both/neither imo.

mnencia:
> As you're on course to making #7801 irrelevant, I'd appreciate if you took a look and see if anything may be used @mnencia

Thank you, Jaime. I will review PR #7801

pkg/management/postgres/initdb.go
mnencia: Machanically, you need the owner to create the database, but you don't need anything else to create the role. I've tried to make this code as linear as possible, deferring all the choices about defaults and dependencies to the main operator code.

armru (approved):

NiccoloFei (approved):

pkg/management/postgres/initdb.go
```diff
@@ -341,16 +324,41 @@ func (info InitInfo) ConfigureNewInstance(instance *Instance) error {
...
+		return err
+	}
+
+	if !existsRole {
+		if _, err = dbSuperUser.Exec(fmt.Sprintf(
+			"CREATE ROLE %v LOGIN",
+			pgx.Identifier{info.ApplicationUser}.Sanitize())); err != nil {
+			return err
+		}
+	}
```
mnencia, line 352: This change could potentially disrupt any initialization scripts that assume the application user is already created.

mnencia: I've created #7870 to address this.

PR #7826, opened 2025-06-14
chore: add usestdlibvars linter

### LINKED ISSUE #8484: [Bug]: add linter usestdlibvars

### Contact Details

_No response_

### Version

1.27 (latest patch)

1.33

YAML manifest

### What happened?

The linter `usestdlibvars` should be added to the list of linters, now we don't have it.

### Cluster resource

```shell

```

### Relevant log output

```shell

```

### Code of Conduct

PascalBourdier:
Add `usestdlibvars` linter to the list of golangci linters
Fix issues throw by the new linter

Closes #8484

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -38,6 +38,7 @@ linters:
     - unconvert
     - unparam
     - unused
+    - usestdlibvars
     - wastedassign
     - whitespace
   settings:
--- a/internal/cmd/manager/backup/cmd.go
+++ b/internal/cmd/manager/backup/cmd.go
@@ -64,7 +64,7 @@ func NewCmd() *cobra.Command {
 				return err
 			}
 
-			if resp.StatusCode != 200 {
+			if resp.StatusCode != http.StatusOK {
 				contextLogger.Info(
 					"Error while requesting backup",
 					"backupURL", backupURL,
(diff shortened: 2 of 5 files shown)


sxd:
Hello @PascalBourdier 

What are we adding here? from my point of view, another dependency to worry about that doesn't add too much value and also is going to add a lot of entropy because a macro could always come with a dangerous code inside, can you explain the benefit ?

All these questions are because there's no issue linked to explain why is needed 

Regards,

PascalBourdier:
> Hello @PascalBourdier
> 
> What are we adding here? from my point of view, another dependency to worry about that doesn't add too much value and also is going to add a lot of entropy because a macro could always come with a dangerous code inside, can you explain the benefit ?
> 
> All these questions are because there's no issue linked to explain why is needed
> 
> Regards,

It could be nice to be consistent with code already in this repo like https://github.com/cloudnative-pg/cloudnative-pg/blob/9a519b6287e525b7346d5af92a78f2c6b6374d67/pkg/management/postgres/webserver/client/local/cache.go#L105 or https://github.com/cloudnative-pg/cloudnative-pg/blob/9a519b6287e525b7346d5af92a78f2c6b6374d67/pkg/management/postgres/webserver/client/remote/instance.go#L264
 
And the `http` lib is already used in every files where I update code.

PascalBourdier:
@sxd if you prefer, I can remove the http lib usage for constant where it is already used elsewhere ?

sxd (approved):

mnencia (approved):

Issue #7713, opened 2025-06-02
[Feature]: No warning or status when Cluster resource is in namespace outside operator scope (clusterWide=false)


tekicat:
### Is there an existing issue already for this bug?

- [x] I have searched for an existing issue, and could not find anything. I believe this is a new bug.

### I have read the troubleshooting guide

- [x] I have read the troubleshooting guide and I think this is a new bug.

### I am running a supported version of CloudNativePG

- [x] I have read the troubleshooting guide and I think this is a new bug.

### Contact Details

tekicat@amibee.com

### Version

1.26 (latest patch)

### What version of Kubernetes are you using?

1.33

### What is your Kubernetes environment?

Cloud: Other

### How did you install the operator?

YAML manifest

### What happened?

**Title:** No warning or status when Cluster resource is in namespace outside operator scope (clusterWide=false)

**Environment:**

* Config: `--set config.clusterWide=false`

**Issue Summary:**
When deploying the CloudNativePG operator with `clusterWide=false`, creating a `Cluster` resource in any namespace other than the operator's own namespace results in silent failure:

* The resource is accepted and passes validation
* No pod, PVC, StatefulSet, or any related resource is created
* There are **no logs, events, or status conditions** indicating that reconciliation is skipped

This leads to a frustrating debugging experience. The user is left with a valid-looking `Cluster` object that is never acted on, with no indication as to why.

**Expected Behavior:**
There should be clear feedback when a `Cluster` resource is created outside of the operator's watch scope. Options could include:

1. Emit an event on the resource (if possible)
2. Add a `.status.conditions` entry like:

   ```yaml
   - type: Reconciled
     status: False
     reason: OutsideWatchScope
     message: This Cluster resource is in a namespace not watched by the operator
   ```
3. Log a warning at the operator level if the webhook admission occurs for an out-of-scope resource (even if reconciliation is not possible)

**Steps to Reproduce:**

1. Install CNPG with `clusterWide=false` into namespace `cloudnative-pg`
2. Create a valid `Cluster` resource in namespace `default`
3. Observe that the resource is validated but silently ignored

**Why this matters:**
This behavior breaks principle of least astonishment. It's hard to debug and confusing to operators and developers using CNPG, especially in multi-namespace setups.

**Request:**
Please add feedback (event/log/status) to signal to users that their cluster resource will not be reconciled due to namespace scope. This will significantly improve the usability and debuggability of the operator.

[...]

PR #7835, opened 2025-06-17
feat: Adding status when Cluster resource is in namespace outside operator scope. 

anishbista60:
- Adding status when Cluster resource is in namespace outside operator scope (clusterWide=false) #7713

### DIFF
--- a/internal/controller/cluster_controller.go
+++ b/internal/controller/cluster_controller.go
@@ -32,6 +32,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	policyv1 "k8s.io/api/policy/v1"
 	apierrs "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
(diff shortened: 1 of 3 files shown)


internal/controller/cluster_controller.go
```diff
@@ -178,6 +179,9 @@ func (r *ClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 		}
 		return ctrl.Result{}, err
 	}
+	if configuration.Current.WatchNamespace == "cnpg-system" && cluster.Namespace != configuration.Current.OperatorNamespace {
```
armru, line 182: a couple of question on top of my head:
- what happens if the operator is installed in another namespace?
- what happens if I have multiple operators installed each one watching a specific namespace?

anishbista60: That's reason I stopped working in this PR. I think I need more discussion .

PR #7853, opened 2025-06-19
fix: Removed the latest tag

anishbista60:
- Removed the latest tag 
- Used of a sha for a specific tag/release #7720


anishbista60:
@sxd Please take a look .

sxd:
Hello @anishbista60 

There's a full section on top of the Makefile with all the versions that we use in the Makefile, so the variables should be there. Also, these sha needs to be updated using the renovate syntax we already have in place in all that section, that's the main idea =)

Regards,

anishbista60:
@sxd Thank you. Would you mind to  have a look another look ?

Issue #7723, opened 2025-06-04
[Bug]: Scorecard should be using the sha256 for the images


sxd:
### Is there an existing issue already for this bug?

- [x] I have searched for an existing issue, and could not find anything. I believe this is a new bug.

### I have read the troubleshooting guide

- [x] I have read the troubleshooting guide and I think this is a new bug.

### I am running a supported version of CloudNativePG

- [x] I have read the troubleshooting guide and I think this is a new bug.

### Version

1.26 (latest patch)

### What version of Kubernetes are you using?

1.33

### What is your Kubernetes environment?

Self-managed: kind (evaluation)

### How did you install the operator?

YAML manifest

### What happened?

The files under `config/olm-scorecard/patches/` should have the images with a sha256 and not just the tag

### Cluster resource

```shell

```

### Relevant log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

anishbista60:
@sxd  I will take a look . Thank you

sxd:
Not an issue anymore or at least not detected by the scorecard review
Closing for now

PR #7856, opened 2025-06-19
fix: Added the images with a sha256 instead tag

anishbista60:
- Added the images with a sha256 instead just tag #7723

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -285,7 +285,7 @@
     {
       groupName: 'operator framework',
       separateMajorMinor: false,
-      pinDigests: false,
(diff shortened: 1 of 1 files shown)


anishbista60:
@sxd FYL.

sxd:
Hello @anishbista60 

I don't think this is the way, probably is better to leave the work to renovate and start using the pinDigest set to true so by default will update, I think that we're skipping the images here or we're not setting the pinDigest properly

Regards,

anishbista60:
> I think that we're skipping the images here or we're not setting the pinDigest properly

ok i have configured `pinDigests: true`

PR #7858, opened 2025-06-19
fix: Pin the go releaser version in workflow

anishbista60:
- Pin the go releaser version in workflow #7749


anishbista60:
@sxd here as well. Thank you for review so far. I don't see any make file and script where go releaser needs to be pinned.
please correct me if i am wrong

sxd:
Hello @anishbista60 

This is not exactly what we're looking for, right now but it's good, how is this going to be updated? renovate will update the values?

Regards,

PR #7876, opened 2025-06-23
chore(cluster): add support for pprof via `alpha.cnpg.io/enableInstancePprof` annotation

### LINKED ISSUE #4056: [Feature]: Add the pprof service to the instance manager

Currently we don't know the amount of resources consumed by the instance manager inside every PostgreSQL pod, which let us without knowledge if it's behaving properly or not 

Add the possibility to start the pprof server inside every instance to report the proper information to the pyroscope/grafana to gather the data and generate the proper graphs.

This needs to be disabled or enabled independently to the pprof server running on the operator itself since it may not be desired to have both running at the same time.

The proposal is to add this as another parameter passed to the pods at starting point, for every process including: bootstrap, run, join, etc. the meaning it's to facilitate the gather of data on every process.

Doing this profiling could help to improve the memory and cpu usage as well as the resources used by the goroutines.

There's no alternative solution to do a proper profiling that can be consistently used in the future

### Additional context

_No response_

### Backport?

Yes

Yes

### Code of Conduct

sxd:
This change introduces the ability to enable the Go pprof HTTP server in instance pods by setting the `alpha.cnpg.io/enableInstancePprof` annotation to "true" on a Cluster resource.

When enabled, each instance pod exposes a pprof server on port 6060/TCP, making it possible to collect and analyse profiling data for debugging and performance tuning. The operator automatically manages the rolling update needed to apply or remove the annotation.

Closes #4056


sxd:
Closing since it's not the desired implementation

mnencia:
Reopened as we want to have pprof usable in 1.27.0

NiccoloFei:
@sxd Could you review if this service that's being created is actually needed? Given that pyroscope will scrape the operator pod directly via the `profiles.grafana.com` annotations, I don't think we need to create this service at all.
Please clarify if it's needed for other purposes or if instead we can just avoid creating it.

```
  service_file="${TEMP_DIR}/pyroscope_service.yaml"

  cat >"${service_file}" <<-EOF
apiVersion: v1
kind: Service
metadata:
  name: cnpg-pprof
  name: cnpg-controller-manager-pprof
spec:
  ports:
  - targetPort: 6060
    port: 6060
  selector:
    app: cnpg-pprof
  type: ClusterIP
  selector:
    app.kubernetes.io/name: cloudnative-pg
EOF
  kubectl -n cnpg-system apply -f "${service_file}"
```

NiccoloFei (approved):

mnencia (approved):

gbartolini (approved):

PR #7885, opened 2025-06-24
fix(pvc): support `VolumeAttributesClass` reconciliation

### LINKED ISSUE #7800: [Feature]: Support Volume Attributes Classes

Currently, if we want to update storage capabilities other than just the size (such as throughput or IOPS), we're forced to create new replicas with the desired settings (with a different StorageClass for example).

This process is both annoying and slow.

[Volumes Attributes Classes](https://kubernetes.io/docs/concepts/storage/volume-attributes-classes/), a feature currently in beta in Kubernetes, allows for the dynamic modification of certain storage parameters. It would be highly beneficial if CloudNative-PG could leverage this feature to enable users to update storage parameters.

Currently, this functionality is only exposed within the [pvcTemplate](https://cloudnative-pg.io/documentation/1.26/cloudnative-pg.v1/#postgresql-cnpg-io-v1-StorageConfiguration) field, which is exclusively used during the creation of PVCs. I believe this parameter could be more effectively placed alongside the existing storage size parameter, making it accessible for updates beyond initial provisioning.

I'm not aware of many alternatives. To my knowledge, the only other way to dynamically change these parameters typically involves using cloud-provider-specific annotations or labels directly on PVCs.

### Additional context

_No response_

### Backport?

No

Yes

### Code of Conduct

armru:
Add missing support for reconciling `VolumeAttributesClass` in `pvcTemplate`, leveraging the feature promoted to beta in Kubernetes 1.31 (disabled by default).

A `VolumeAttributesClass` allows administrators (typically cloud providers) to define mutable "classes" of storage, mapping to different quality-of-service levels, while Kubernetes remains unopinionated about their semantics.
See: https://kubernetes.io/docs/concepts/storage/volume-attributes-classes/

This patch also refactors the PVC reconciliation logic to simplify handling additional parameters.

Closes #7800 

## Release notes

The operator now ensures that the `VolumeAttributesClassName` in PVCs matches the configuration defined in the `pvcTemplate`.

### DIFF
--- a/docs/src/storage.md
+++ b/docs/src/storage.md
@@ -144,8 +144,8 @@ spec:
 
 ## Configuration via a PVC template
 
-To further customize the generated PVCs, you can provide a PVC template inside
-the custom resource, as shown below:
+To further customize the generated PVCs, you can provide a PVC template inside the custom resource,
+like in the following example:
 
 ```yaml
 apiVersion: postgresql.cnpg.io/v1
@@ -165,14 +165,6 @@ spec:
       storageClassName: standard
       volumeMode: Filesystem
 ```
-
-!!! Info
-    Depending on your Kubernetes cluster configuration, you can also specify a
-    `VolumeAttributesClass` in the `pvcTemplate`.
-    [Volume Attributes Classes](https://kubernetes.io/docs/concepts/storage/volume-attributes-classes/),
-    promoted to beta in Kubernetes 1.31 (disabled by default), allow administrators
-    (typically cloud providers) to define mutable "classes" of storage that map to
-    different quality-of-service levels.
 
 ## Volume for WAL
 
--- a/docs/src/storage.md
+++ b/docs/src/storage.md
@@ -144,8 +144,8 @@ spec:
 
 ## Configuration via a PVC template
 
-To further customize the generated PVCs, you can provide a PVC template inside the custom resource,
-like in the following example:
+To further customize the generated PVCs, you can provide a PVC template inside
+the custom resource, as shown below:
 
 ```yaml
 apiVersion: postgresql.cnpg.io/v1
@@ -165,6 +165,14 @@ spec:
       storageClassName: standard
       volumeMode: Filesystem
 ```
+
+!!! Info
+    Depending on your Kubernetes cluster configuration, you can also specify a
+    `VolumeAttributesClass` in the `pvcTemplate`.
+    [Volume Attributes Classes](https://kubernetes.io/docs/concepts/storage/volume-attributes-classes/),
+    promoted to beta in Kubernetes 1.31 (disabled by default), allow administrators
+    (typically cloud providers) to define mutable "classes" of storage that map to
+    different quality-of-service levels.
 
 ## Volume for WAL
 
--- a/internal/controller/cluster_controller.go
+++ b/internal/controller/cluster_controller.go
@@ -735,7 +735,7 @@ func (r *ClusterReconciler) reconcileResources(
 		cluster,
 		resources.instances.Items,
 		resources.pvcs.Items,
-	); !res.IsZero() || err != nil {
+	); err != nil || !res.IsZero() {
 		return res, err
 	}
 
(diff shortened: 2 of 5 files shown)


leonardoce:
A Kubernetes environment with the VolumeAttributeClass feature enabled is required to test this patch.
To enable it in the local development environment, you need to apply the following patch against `setup-cluster.sh`:

```diff
diff --git a/hack/setup-cluster.sh b/hack/setup-cluster.sh
index 6d8d008c3..9b69a4143 100755
--- a/hack/setup-cluster.sh
+++ b/hack/setup-cluster.sh
@@ -120,6 +120,15 @@ apiVersion: kind.x-k8s.io/v1alpha4
 networking:
   apiServerAddress: "0.0.0.0"

+featureGates:
+  # any feature gate can be enabled here with "Name": true
+  # or disabled here with "Name": false
+  # not all feature gates are tested, however
+  "VolumeAttributesClass": true
+
+runtimeConfig:
+  storage.k8s.io/v1beta1: "true"
+
 # add to the apiServer certSANs the name of the docker (dind) service in order to be able to reach the cluster through it
 kubeadmConfigPatchesJSON6902:
   - group: kubeadm.k8s.io
```

And then, the VolumeAttributeClass feature gate need to be enabled in the external-provisioner sidecar deployed by the hostpath CSI driver.
You can use the `--feature-gates` command line option in the `csi-provisioner` sidecar of the `csi-hostpathplugin` statefulset to do that:

```yaml
      - args:
        - -v=5
        - --csi-address=/csi/csi.sock
        - --feature-gates=Topology=true,VolumeAttributesClass=true # <===
        image: registry.k8s.io/sig-storage/csi-provisioner:v5.2.0
        imagePullPolicy: IfNotPresent
        name: csi-provisioner
        resources: {}
        securityContext:
          privileged: true
        terminationMessagePath: /dev/termination-log
        terminationMessagePolicy: File
        volumeMounts:
        - mountPath: /csi
          name: socket-dir
```

After that, the `ModifyVolume` GRPC call must be enabled in the CSI driver with the following change:

```yaml
      - args:
        - --drivername=hostpath.csi.k8s.io
        - --v=5
        - --endpoint=$(CSI_ENDPOINT)
        - --nodeid=$(KUBE_NODE_NAME)
        - --enable-controller-modify-volume=true  # <=====
        env:
        - name: CSI_ENDPOINT
          value: unix:///csi/csi.sock
        - name: KUBE_NODE_NAME
          valueFrom:
            fieldRef:
              apiVersion: v1
              fieldPath: spec.nodeName
        image: registry.k8s.io/sig-storage/hostpathplugin:v1.16.1
        imagePullPolicy: IfNotPresent
        livenessProbe:
          failureThreshold: 5
          httpGet:
            path: /healthz
            port: healthz
            scheme: HTTP
          initialDelaySeconds: 10
          periodSeconds: 2
          successThreshold: 1
          timeoutSeconds: 3
        name: hostpath
```

Having done that, VolumeAttributeClasses will be supported by the Kind cluster and the CSI hostpath driver.

This is a basic example of a VolumeAttributeClass:

```yaml
apiVersion: storage.k8s.io/v1beta1
kind: VolumeAttributesClass
metadata:
  name: silver
driverName: hostpath.csi.k8s.io
parameters:
  provisioned-iops: "3000"
  provisioned-throughput: "50"
```

And can be referenced from the cluster with:

```yaml
kind: Cluster
metadata:
  name: cluster-example
spec:
  instances: 3
  imageName: ghcr.io/cloudnative-pg/postgresql:13

  storage:
    size: 1Gi
    storageClass: csi-hostpath-sc
    pvcTemplate:
      volumeAttributesClassName: silver
```

You can create multiple volume attribute classes, and when you change the PVC template in the cluster the volume attribute class name will be reconciled in the existing PVCs.

leonardoce (approved):

gbartolini (approved):

gbartolini:
After chatting with @armru, we believe that this is more a bug than a feature and should be backported. He made me reflect on the piece of documentation that I had added. When we talk about a PVC template, users expect us to reconcile what Kubernetes already provides. From this perspective, this is an operator "deficiency" and thus a bug. @leonardoce thoughts, given that you reviewed the patch?

leonardoce:
It is fine with me! Let's backport it.

PR #7891, opened 2025-06-25
fix(postgres,instance): set postgres workers `oom_score_adj`

### LINKED ISSUE #7132: [Bug]: Instance-manager running as PID1 makes the pod exit during high memory usage

### Version

1.25 (latest patch)

1.30

Cloud: Google GKE

Helm

### What happened?

Hello,

We are happily running CloudNativePG in production (thanks for this amazing project!)

We run spiky workloads on those databases, with very high memory usage.

To avoid getting our backend processes killed by the OOM-killer, which in cascade would kill postmaster and take down the whole instance, we disabled memory overcommit on the host (`sysctl -w vm.overcommit_memory=2`), as advised by the PostgreSQL documentation: https://www.postgresql.org/docs/15/kernel-resources.html#LINUX-MEMORY-OVERCOMMIT

With this parameter, once the host no longer has free virtual memory available ([CommitLimit](https://www.kernel.org/doc/html/v4.19/vm/overcommit-accounting.html)) `malloc()` will fail in the PG backend, which will terminate the query gracefully with error code 53200 (Out of memory), without killing postmaster.

However, as this is a host setting, when we hit this CommitLimit, if the instance-manager tries to allocate memory at the same time, it will crash. Since the instance-manager is the PID1 of the pod, the pod itself will exit even though PG itself is running.

We are wondering if there would be ways to make a CNPG cluster more resilient to these types of crashes.

For example by not running the instance-manager as PID1, and instead use a nanny process, not/less prone to crash when memory allocations fail, that will launch the instance-manager (and restart it when it crashes).
If done wrong I suppose this could break the inplace update of the instance manager.

We also thought of other solutions, none that were very Kubernetes idiomatic (trying to play around with cgroups for example).

We are also trying to fine tune the database parameters and the type of queries we send to the database to prevent these memory spikes from happening, but due to the nature of our application, it is not possible to completely prevent this issue from happening.

We are interested in your opinion on this issue, and if there is a solution for this problem and you deem it worth adding to CNPG, we are willing to take care of the development.

## Steps to reproduce the issue
It is very manual but allows you to reproduce this memory pressure issue on a node to observe the behavior of the instance-manager.
Do not run it in production as it may crash other pods of the node

- Start a CNPG cluster on a Kubernetes cluster
- Disable memory overcommit on the node hosting the CNPG cluster (you need to be root):

```bash
sysctl -w vm.overcommit_memory=2
```

- Allocate memory so that all the memory of the host is committed
    - To see the committed memory (`Committed_AS`) and the memory limit (`CommitLimit` ):
    
    ```bash
    cat /proc/meminfo | grep Commit
    ```
    
    - To allocate memory:
    
    ```bash
    head -c <quantity of memory to allocate> /dev/zero | tail | sleep 60
    ```
    
    For example:
    
    ```bash
    head -c 5000m /dev/zero | tail | sleep 60 # to allocate ~5GB
    ```
    
- Due to the lack of free memory, you should find the following Kernel error logs:
    
    > __vm_enough_memory: pid: <Process PID>, comm: <process name>, no enough memory for the allocation

    
    And the following Postgres logs
    
    > out of memory

    
    With attributes such as:
    
    > Failed on request of size 64 in memory context "MessageContext".

    
    Eventually, if the instance-manager allocates memory during this time, the following kernel error log will spawn:
    
    > __vm_enough_memory: pid: <Process PID>, comm: manager, no enough memory for the allocation

    
    And the pod of the CNPG cluster will exit.

### Cluster resource

```shell

```

### Relevant log output

```shell

```

### Code of Conduct

armru:
When the quality of service is set to "Guaranteed", CloudNativePG sets the `PG_OOM_ADJUST_VALUE` for the `postmaster` process to `0`.

This allows the `postmaster` to retain its low Out-Of-Memory (OOM) score of `-997`, while its child processes run with an OOM score adjustment of `0`. As a result, if the OOM killer is triggered, it will terminate the child processes before the `postmaster`.

This behaviour helps keep the PostgreSQL instance alive for as long as possible and enables a clean shutdown procedure in the event of an eviction.

Closes #7132

### DIFF
--- a/docs/src/resource_management.md
+++ b/docs/src/resource_management.md
@@ -44,19 +44,21 @@ For a PostgreSQL workload it is recommended to set a "Guaranteed" QoS.
     When the quality of service is set to "Guaranteed", CloudNativePG sets the
     `PG_OOM_ADJUST_VALUE` for the `postmaster` process to `0`, in line with the
     [PostgreSQL documentation](https://www.postgresql.org/docs/current/kernel-resources.html#LINUX-MEMORY-OVERCOMMIT).
-    This allows the `postmaster` to retain its low OOM score (`-997`), while its
-    child processes run with an OOM score adjustment of `0`. As a result, if the
-    OOM killer is triggered, it will terminate the child processes before the
-    `postmaster`. This behaviour helps keep the PostgreSQL instance alive and
-    enables a clean shutdown procedure in the event of an eviction.
+    This allows the `postmaster` to retain its low Out-Of-Memory (OOM) score of
+    `-997`, while its child processes run with an OOM score adjustment of `0`. As a
+    result, if the OOM killer is triggered, it will terminate the child processes
+    before the `postmaster`. This behaviour helps keep the PostgreSQL instance
+    alive for as long as possible and enables a clean shutdown procedure in the
+    event of an eviction.
 
 To avoid resources related issues in Kubernetes, we can refer to the best practices for "out of resource" handling
 while creating a cluster:
 
 -  Specify your required values for memory and CPU in the resources section of the manifest file.
-   This way, you can avoid the `OOM Killed` (where "OOM" stands for Out Of Memory) and `CPU throttle` or any other
+   This way, you can avoid the `OOM Killed` and `CPU throttle` or any other
    resource-related issues on running instances.
--  For your cluster's pods to get assigned to the "Guaranteed" QoS class, you must set limits and requests
+-  For your cluster's pods to get assigned to the "Guaranteed" QoS class, you
+   must set limits and requests
    for both memory and CPU to the same value.
 -  Specify your required PostgreSQL memory parameters consistently with the pod resources (as you would do
    in a VM or physical machine scenario - see below).
--- a/docs/src/resource_management.md
+++ b/docs/src/resource_management.md
@@ -40,6 +40,16 @@ section in the Kubernetes documentation.
 
 For a PostgreSQL workload it is recommended to set a "Guaranteed" QoS.
 
+!!! Info
+    When the quality of service is set to "Guaranteed", CloudNativePG sets the
+    `PG_OOM_ADJUST_VALUE` for the `postmaster` process to `0`, in line with the
+    [PostgreSQL documentation](https://www.postgresql.org/docs/current/kernel-resources.html#LINUX-MEMORY-OVERCOMMIT).
+    This allows the `postmaster` to retain its low OOM score (`-997`), while its
+    child processes run with an OOM score adjustment of `0`. As a result, if the
+    OOM killer is triggered, it will terminate the child processes before the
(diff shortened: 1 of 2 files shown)


BaudouinH:
@armru Thanks for this PR.
As mentioned in https://github.com/cloudnative-pg/cloudnative-pg/issues/7132, my organization is willing to help test this patch if needed.

How would you like to proceed?
We can build the CNPG operator with this PR's changes and do tests using it, would that work for you?

BaudouinH:
I ran the patched version on a cluster.
I may have done something wrong, but I don't see the adjusted scores, despite having a controller with the correct commit hash:
```bash
postgres@<pod name>:/$ ps -aux
USER         PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
postgres       1  0.3  0.4 1287844 65968 ?       Ssl  08:55   0:50 /controller/manager instance run --status-port-tls --log-level=info
postgres      22  0.1  2.0 4613776 337024 ?      S    08:55   0:14 postgres -D /var/lib/postgresql/data/pgdata
postgres@<pod name>:/$ cat /proc/22/oom_score_adj 
232
```

And indeed, even with the user `postgres`, I was unable to change it manually, which leads me to believe we have a permission issue:
```bash
postgres@<pod name>:/$ echo 230 > /proc/22/oom_score_adj
bash: echo: write error: Permission denied
```

[The Postgres documentation on `PG_OOM_ADJUST_VALUE` ](https://www.postgresql.org/docs/current/kernel-resources.html#LINUX-MEMORY-OVERCOMMIT) does mention it:
> in the PostgreSQL startup script just before invoking postgres. **Note that this action must be done as root**, or it will have no effect; so a root-owned startup script is the easiest place to do it. If you do this, you should also set these environment variables in the startup script before invoking postgres:
```bash
export PG_OOM_ADJUST_FILE=/proc/self/oom_score_adj
export PG_OOM_ADJUST_VALUE=0
```

Also, when I check the environment variables of the postmaster process with `/proc/<PID>/environ`, the environment variables have been set correctly.

I will do some tests to see if adding the capability `CAP_SYS_RESOURCE` to the pod would be enough

mnencia:
@BaudouinH Thanks for the update! Just a clarification regarding setting oom_score_adj:

Lowering the oom_score_adj value (making the process more protected from the OOM killer) requires the CAP_SYS_RESOURCE capability in Linux. However, a process can increase its own oom_score_adj value (making itself more likely to be killed) without this capability.

Since Kubernetes manages oom_score_adj for system containers using privileged processes, this usually works as expected for guaranteed QoS pods. We already recommend setting the QoS of the PostgreSQL instances to guaranteed (by setting equal requirements for requests and limits) for all production instances, so this change would be automatically beneficial for anyone who follows our recommendations.

If you think it would help, we could add a note about this to the documentation.

mnencia (approved):

gbartolini (approved):

PR #7893, opened 2025-06-25
docs: explain `wal_level` in Postgres configuration page

### LINKED ISSUE #7890: [Docs]: Document `wal_level` in the Postgres configuration section

By default, we set `wal_level` to `logical`. There have been some discussions in the `-users` Slack channel about this decision, which surprises veteran Postgres DBAs who expect `replica`.

Explain the decision and the options for users.

https://cloudnative-pg.io/documentation/current/postgresql_conf/

None

### Additional context

_No response_

### Backport?

Yes

### Code of Conduct

gbartolini:
Closes #7890

### DIFF
--- a/docs/src/postgresql_conf.md
+++ b/docs/src/postgresql_conf.md
@@ -133,24 +133,26 @@ archiving and replication.
 ### Write-Ahead Log Level
 
 The [`wal_level`](https://www.postgresql.org/docs/current/runtime-config-wal.html)
-parameter in PostgreSQL determines the amount of information written to the
+parameter in PostgreSQL determines the amount of data written to the
 Write-Ahead Log (WAL). It accepts the following values:
 
-- `minimal`: Writes only the information required for crash recovery
-  (CloudNativePG uses this setting only for single-instance clusters with WAL
-  archiving disabled).
+- `minimal`: Writes only the information required for crash recovery.
 - `replica`: Adds sufficient information to support WAL archiving and streaming
   replication, including the ability to run read-only queries on standby
   instances.
 - `logical`: Includes all information from `replica`, plus additional data
   required for logical decoding and replication.
 
-Unlike upstream PostgreSQL, CloudNativePG sets `wal_level` to `logical` by
-default. This is to enable logical replication out of the box, making it easier
-to support use cases such as migrations from external PostgreSQL servers.
+By default, upstream PostgreSQL sets `wal_level` to `replica`. CloudNativePG,
+instead, sets `wal_level` to `logical` by default to enable logical replication
+out of the box. This makes it easier to support use cases such as migrations
+from external PostgreSQL servers.
 
 If your cluster does not require logical replication, it is recommended to set
 `wal_level` to `replica` to reduce WAL volume and overhead.
+
+Finally, CloudNativePG allows `wal_level` to be set to `minimal` only for
+single-instance clusters with WAL archiving disabled.
 
 ### Replication Settings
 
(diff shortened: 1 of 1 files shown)


docs/src/postgresql_conf.md
```diff
@@ -130,6 +130,28 @@ Since the fixed parameters are added at the end, they can't be overridden by the
...
 
+### Write-Ahead Log Level
+
+The [`wal_level`](https://www.postgresql.org/docs/current/runtime-config-wal.html)
+parameter in PostgreSQL determines the amount of information written to the
+Write-Ahead Log (WAL). It accepts the following values:
+
+- `minimal`: Writes only the information required for crash recovery
+  (CloudNativePG uses this setting only for single-instance clusters with WAL
+  archiving disabled).
```
jsilvela: I think this bit would make more sense in the paragraph below that begins with "Unlike upstream ...". That way you get the full explanation more compactly, and reduce the scope of this enumeration.

docs/src/postgresql_conf.md
```diff
@@ -130,6 +130,28 @@ Since the fixed parameters are added at the end, they can't be overridden by the
...
+- `replica`: Adds sufficient information to support WAL archiving and streaming
+  replication, including the ability to run read-only queries on standby
+  instances.
+- `logical`: Includes all information from `replica`, plus additional data
+  required for logical decoding and replication.
+
+Unlike upstream PostgreSQL, CloudNativePG sets `wal_level` to `logical` by
```
jsilvela: Either in this paragraph or in the next one, I think it would be helpful to state that the Postgres default is `replica`. Otherwise you mention the disparity but don't explain it.

docs/src/postgresql_conf.md
```diff
@@ -130,6 +130,28 @@ Since the fixed parameters are added at the end, they can't be overridden by the
...
+parameter in PostgreSQL determines the amount of information written to the
+Write-Ahead Log (WAL). It accepts the following values:
+
+- `minimal`: Writes only the information required for crash recovery
+  (CloudNativePG uses this setting only for single-instance clusters with WAL
+  archiving disabled).
+- `replica`: Adds sufficient information to support WAL archiving and streaming
+  replication, including the ability to run read-only queries on standby
+  instances.
+- `logical`: Includes all information from `replica`, plus additional data
```
jsilvela: "all information, plus data".
In this post there are mentions to "data" and "information" interchangeably, but that can be a distraction IMO.
And in the first paragraph you talk of "amount of information", but IMO "information" is a qualitative word, where "data" is more commonly measured.

jsilvela (approved):
optional suggestions

gbartolini:
@jsilvela can you please check now? Thanks

jsilvela (approved):
👍

docs/src/postgresql_conf.md
jbattiato: @jsilvela  @gbartolini 

Based on the definition of both "data" and "information" on the Internet, information refers to a processed, organized, and structured raw data: "information is meaningful data".
Usually, humans rely on information to make decisions, after processing all the data. While data is a collection of raw values with no meaning or purpose.

The main problem here lies on the philosophical question: from which point of view should we interpret these two words?

1. In PostgreSQL the WAL file contains raw data from a human point of view,
2. but WAL files contain information for a PostgreSQL point of view.

I'd say "information" is a good use in this case.

jbattiato: Can we use "plus additional info" instead?
Also, use "amount of info" at the beginning of the paragraph.

mnencia (approved):

PR #7903, opened 2025-06-26
docs: clarify the defaults for microservice imports

jsilvela:
Make it explicit that if the `import.database` is not set, databases
imported via microservice may be renamed to `app`.

### DIFF
--- a/docs/src/database_import.md
+++ b/docs/src/database_import.md
@@ -161,9 +161,8 @@ There are a few things you need to be aware of when using the `microservice` typ
 !!! Hint
     The microservice approach follows CloudNativePG conventions and defaults for
     the destination cluster.
-    Notably, if you don't explicitly set `initdb.database` or `initdb.owner` for
-    the destination cluster, the default of `app` will be used for both, and
-    your chosen source database may be renamed to `app`.
+    Notably, if you don’t set `initdb.database` or `initdb.owner` for the
+    destination cluster, both will default to `app`.
 
 ## The `monolith` type
 
--- a/docs/src/database_import.md
+++ b/docs/src/database_import.md
@@ -83,7 +83,7 @@ performed in 4 steps:
 - execution of `ANALYZE VERBOSE` on the imported database
 
 In the below figure, a single PostgreSQL cluster with N databases is imported
-into N separate CloudNativePG clusters, each using a `microservice` import
+into separate CloudNativePG clusters, each using a microservice import
 for one of the N source databases.
 
 ![Example of microservice import type](./images/microservice-import.png)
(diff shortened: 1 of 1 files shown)


docs/src/database_import.md
```diff
@@ -82,6 +82,10 @@ performed in 4 steps:
   database via the `postImportApplicationSQL` parameter
 - execution of `ANALYZE VERBOSE` on the imported database
 
+In the below figure, a single PostgreSQL cluster with N databases is imported
+into separate CloudNativePG clusters, each using a microservice import
+for one of the N source databases.
+
```
jsilvela: I've added this paragraph because in the image the source cluster has the legend "MONOLITH", so there is potential for confusion.
Can the image be corrected?

docs/src/database_import.md
```diff
@@ -154,6 +158,13 @@ There are a few things you need to be aware of when using the `microservice` typ
 - Only one database can be specified inside the `initdb.import.databases` array
 - Roles are not imported - and as such they cannot be specified inside `initdb.import.roles`
 
+!!! Hint
+    The microservice approach follows CloudNativePG conventions and defaults for
+    the destination cluster.
+    Notably, if you don't explicitly set `initdb.database` or `initdb.owner` for
+    the destination cluster, the default of `app` will be used for both, and
+    your chosen source database may be renamed to `app`.
```
anishbista60:
```suggestion
    Notably, if you don’t set `initdb.database` or `initdb.owner` for the
    destination cluster, both will default to `app`.
```

jsilvela: Please note the formatting in my PR. Notably:
- four spaces at the beginning, since it's in a `!!!` block
- respect 80 columns
Also: `will default to app`  ... without any quote around `app`, this is confusing.

anishbista60, line 166: ok sorry i missed that formatting part .  I think this part is actual behaviour not a possibility

anishbista60: ok sorry for that i have not noticed formatting. My only doubt is I think this part `your chosen source database may be renamed to app.` is not a possibility, it's acutal behaviour

jsilvela: Unless your chosen database is `app`

anishbista60: ok cool that's correct as well. Rest LGTM. Thanks for this PR.

jsilvela: Further comment on your suggested changes before I can accept them.
On line 165, there should be a space between the period and `This`.
And since you put the phrasing in a more active, affirmative voice, the second sentence becomes almost redundant.
"... both will default to `app`, and will override the original source name."

anishbista60: updated. Thanks . if this suggested changes still have some issue then you can move further in your own by including the changes from my suggestion.

gbartolini (approved):

mnencia (approved):

mnencia:
/ok-to-merge docs only change, no need for E2E

PR #7942, opened 2025-06-27
feat(databases): add declarative FDW management to `Database` CRD

### LINKED ISSUE #4683: Foreign data wrappers for Postgres
Introduce declarative support for foreign data wrappers, starting with Postgres. In some cases, this might improve horizontal scalability by distributing data and queries on multiple instances (please refer to foreign tables in the Postgres docs).

EdwinaZhu:
Add support for managing PostgreSQL Foreign Data Wrappers (FDWs) declaratively via the `spec.fdws` field in the `Database` custom resource.

This feature enables Kubernetes-native, privilege-aware lifecycle management of FDWs without requiring direct SQL execution or superuser escalation by users.

It supports configuring `name`, `ensure`, `handler`, `validator`, `owner`, `usage` grants/revokes, and `options`, including the PostgreSQL `"-"` convention for removing handlers or validators.

Ownership changes are restricted to superusers, following PostgreSQL rules.

Only FDWs explicitly listed in `spec.fdws` are reconciled, leaving all others untouched.

The operator manages FDWs using PostgreSQL native `CREATE FOREIGN DATA WRAPPER`, `ALTER FOREIGN DATA WRAPPER`, and `DROP FOREIGN DATA WRAPPER` commands.

Closes #4683 

Implemented as part of the LFX Mentorship Program 2025 Term 2.


internal/management/controller/database_controller_sql.go
```diff
@@ -43,6 +43,24 @@ type schemaInfo struct {
...
+type fdwInfo struct {
+	Name      string `json:"name"`
+	Handler   string `json:"handler"`
+	Validator string `json:"validator"`
+	Owner     string `json:"owner"`
+	// Options   []apiv1.OptSpec `json:"options"`
+}
+
+// sqlObjectBuilder builds a SQL query to create a database object
+type sqlObjectBuilder func(b *strings.Builder, obj interface{}) error
```
leonardoce:
```
type sqlObjectBuilder func(b *strings.Builder, obj interface{}) error
```

I feel like `interface {}` is too broad for what we're trying to achieve here.
Could you please try to devise an interface for it, or to use generic types?

internal/management/controller/database_controller_sql.go
```diff
@@ -231,6 +249,34 @@ func dropDatabase(
 	return nil
 }
 
+func createDatabaseObject(
+	ctx context.Context,
+	db *sql.DB,
+	config sqlObjectConfig,
+	obj interface{},
```
leonardoce:
```
	obj interface{},
```

Same as the previous comment

internal/management/controller/database_controller_sql.go
```diff
@@ -404,3 +448,141 @@ func dropDatabaseSchema(ctx context.Context, db *sql.DB, schema apiv1.SchemaSpec
...
+			return nil, nil
+		}
+		return nil, fmt.Errorf("while scanning if FDW %q exists: %w", fdw.Name, err)
+	}
+
+	return &result, nil
+}
+
+func buildCreateFDWSQL(b *strings.Builder, obj interface{}) error {
+	fdw := obj.(apiv1.FDWSpec)
```
leonardoce:
```
	fdw := obj.(apiv1.FDWSpec)
```

If we should cast this type, I think we should also check if the cast is doable, like in:

```
	fdw, ok := obj.(apiv1.FDWSpec)
        if !ok {
                // oops, something bad happened!
        }
```

internal/management/controller/database_controller_sql.go
```diff
@@ -257,28 +303,28 @@ func getDatabaseExtensionInfo(ctx context.Context, db *sql.DB, ext apiv1.Extensi
 	return &result, nil
 }
 
-func createDatabaseExtension(ctx context.Context, db *sql.DB, ext apiv1.ExtensionSpec) error {
```
gbartolini: I would prefer if we leave the specialised functions that manage a specific object, like this one for "CREATE EXTENSION". It makes the code easier to maintain over time, as we can easily understand what a specific function is using. While I understand the generalisation process for a single "object create function", I see it primarily as a didactic/learning exercise. We cannot rely on every object being created identically in PostgreSQL - we might want to do something specific for a given SQL command. I also had problems looking at where FOREIGN DATA WRAPPERS were created. I suggest reverting this and adding specialised functions for the FDW objects.

internal/webhook/v1/database_webhook.go
```diff
@@ -212,3 +213,27 @@ func (v *DatabaseCustomValidator) validateSchemas(d *apiv1.Database) field.Error
 
 	return result
 }
+
+// validateFDWs validates the database Foreign Data Wrappers
+// FDWs must be unique in .spec.fdws
+func (v *DatabaseCustomValidator) validateFDWs(d *apiv1.Database) field.ErrorList {
+	var result field.ErrorList
+
+	FDWNames := stringset.New()
```
leonardoce:
```suggestion
	fdwNames := stringset.New()
```

internal/management/controller/database_controller_sql.go
```diff
@@ -43,6 +43,24 @@ type schemaInfo struct {
...
+}
+
+// sqlObjectBuilder builds a SQL query to create a database object
+type sqlObjectBuilder func(b *strings.Builder, obj interface{}) error
+
+type sqlObjectConfig struct {
+	objectType    string
+	name          string
+	builder       sqlObjectBuilder
+	logObjectName string
```
leonardoce:
```suggestion
```

Let's remove this and just use objectType

internal/management/controller/database_controller_sql.go
```diff
@@ -231,6 +249,34 @@ func dropDatabase(
...
+	}
+
+	_, err := db.ExecContext(ctx, sqlBuilder.String())
+	if err != nil {
+		contextLogger.Error(err, "while creating "+config.logObjectName,
+			"query", sqlBuilder.String())
+		return err
+	}
+
+	contextLogger.Info("created "+config.logObjectName, "name", config.name)
```
leonardoce:
```suggestion
	contextLogger.Info("Created database object")
```

internal/management/controller/database_controller_sql.go
```diff
@@ -231,6 +249,34 @@ func dropDatabase(
...
+		config.objectType,
+		pgx.Identifier{config.name}.Sanitize()))
+
+	if err := config.builder(&sqlBuilder, obj); err != nil {
+		return err
+	}
+
+	_, err := db.ExecContext(ctx, sqlBuilder.String())
+	if err != nil {
+		contextLogger.Error(err, "while creating "+config.logObjectName,
```
leonardoce:
```suggestion
		contextLogger.Error(err, "while creating database object",
```

internal/management/controller/database_controller_sql.go
```diff
@@ -231,6 +249,34 @@ func dropDatabase(
 	return nil
 }
 
+func createDatabaseObject(
+	ctx context.Context,
+	db *sql.DB,
+	config sqlObjectConfig,
+	obj interface{},
+) error {
+	contextLogger := log.FromContext(ctx)
```
leonardoce:
```suggestion
	contextLogger := log.FromContext(ctx).WithValues("name", config.name)
```

internal/management/controller/database_controller_sql.go
```diff
@@ -404,3 +415,228 @@ func dropDatabaseSchema(ctx context.Context, db *sql.DB, schema apiv1.SchemaSpec
...
+				continue
+			}
+			if !first {
+				sqlCreateFDW.WriteString(", ")
+			}
+			first = false
+			sqlCreateFDW.WriteString(fmt.Sprintf("%s '%s'", name, optionSpec.Value))
+		}
+		sqlCreateFDW.WriteString(")")
+	}
```
armru: I think we should try to refactor the string builder in a more declarative approach

internal/management/controller/database_controller_sql.go
```diff
@@ -404,3 +415,228 @@ func dropDatabaseSchema(ctx context.Context, db *sql.DB, schema apiv1.SchemaSpec
...
+				Value: parts[1],
+			}
+		}
+	}
+	result.Options = opts
+
+	return &result, nil
+}
+
+//nolint:dupl
```
armru: Minor note: We should try to remove all the nolint rules before merging the patch

api/v1/database_types.go
```diff
@@ -220,6 +224,39 @@ type ExtensionSpec struct {
...
+	// Name of the handler function (e.g., "postgres_fdw_handler")
+	Handler string `json:"handler,omitempty"`
+
+	// Name of the validator function (e.g., "postgres_fdw_validator")
+	Validator string `json:"validator,omitempty"`
+
+	// Owner specifies the database user who will own the Foreign Data Wrapper.
+	// By default, the owner of a new FDW is the current session user.
+	// Even if an owner is explicitly specified during creation, it will be ignored.
+	Owner string `json:"owner,omitempty"`
```
mnencia:
```suggestion
	// Owner specifies the database user who will own the Foreign Data Wrapper.
	// By default, the owner of a new FDW is the current session user.
	// Even if an owner is explicitly specified during creation, it will be ignored.
	// +optional
	Owner string `json:"owner,omitempty"`
```

Any field that has the JSON attribute `omitempty` needs to be marked as `+optional`, otherwise when we generate the documentation it is marked as required.

mnencia: Additionally, given that the owner can only be a superuser and I don't really see why anyone should change it from the default, I wonder if we should just drop this field.

api/v1/database_types.go
```diff
@@ -220,6 +224,39 @@ type ExtensionSpec struct {
...
 
+// FDWSpec configures an Foreign Data Wrapper in a database
+type FDWSpec struct {
+	DatabaseObjectSpec `json:",inline"`
+
+	// Name of the handler function (e.g., "postgres_fdw_handler")
+	Handler string `json:"handler,omitempty"`
+
+	// Name of the validator function (e.g., "postgres_fdw_validator")
+	Validator string `json:"validator,omitempty"`
```
mnencia: I've tried the patch and tested several FDW implementations, and commonly, it is not explained what you should put in these fields. Usually, the FDW is created from the extension initialization; we should probably make these fields really optional by making them string pointers and not modifying the current value if they are set to nil.

gbartolini:
Great job @EdwinaZhu!

I have been able to make the patch work as follows. Once deployed locally, I have created these resources:

```yaml
kind: Cluster
metadata:
  name: pg-a
spec:
  instances: 1
  storage:
    size: 1Gi

  bootstrap:
    initdb:
      postInitApplicationSQL:
        - CREATE TABLE n (i SERIAL PRIMARY KEY, m INTEGER)
        - INSERT INTO n (m) (SELECT generate_series(1, 10000))
        - ALTER TABLE n OWNER TO app
---
kind: Cluster
metadata:
  name: pg-b
spec:
  instances: 1
  storage:
    size: 1Gi
---
kind: Database
metadata:
  name: app
spec:
  name: app
  owner: app
  cluster:
    name: pg-b
  extensions:
  - name: postgres_fdw
  fdws:
  - name: postgres_fdw
    #ensure: present
    # We should make 'handler' and 'validator' optional
    handler: postgres_fdw_handler
    validator: postgres_fdw_validator
    # We need to inform that only superusers can own an FDW
    #owner: app
```

Then imperatively, I have been able to create a foreign server, a user mapping and import a foreign table.

```sh
PGAPPPWD=`kubectl get secret pg-a-app -o jsonpath="{.data.password}" | base64 -d`
kubectl cnpg psql pg-b -- app -c "CREATE SERVER pga FOREIGN DATA WRAPPER postgres_fdw OPTIONS (host 'pg-a-rw')"
kubectl cnpg psql pg-b -- app -c "GRANT USAGE ON FOREIGN SERVER pga TO app"
kubectl cnpg psql pg-b -- `kubectl get secret pg-b-app   -o jsonpath="{.data.uri}" | base64 -d` -c "CREATE USER MAPPING FOR app SERVER pga OPTIONS (user 'app', password '$PGAPPPWD')"
kubectl cnpg psql pg-b -- `kubectl get secret pg-b-app   -o jsonpath="{.data.uri}" | base64 -d` -c "CREATE FOREIGN TABLE n (i INTEGER, m INTEGER) SERVER pga"
```

At this point, the second cluster (`pg-b`) will have in the `app` database a foreign table called `n` with the same schema as the one in `pg-a`. You can try and get all the entries with:

```sh
kubectl cnpg psql pg-b -- `kubectl get secret pg-b-app   -o jsonpath="{.data.uri}" | base64 -d` -c "SELECT * FROM n ORDER BY 1"
```

You can also insert a new record in the `n` foreign table and take advantage of `postgres_fdw` write capability:

```sh
kubectl cnpg psql pg-b -- `kubectl get secret pg-b-app   -o jsonpath="{.data.uri}" | base64 -d` -c "INSERT INTO n(i, m) VALUES (-1, -1)"
```

And verify that the record was inserted also in the `pg-a` cluster:

```sh
kubectl cnpg psql pg-a -- `kubectl get secret pg-a-app   -o jsonpath="{.data.uri}" | base64 -d` -c "SELECT * FROM n ORDER BY 1"
```

I think you can use these instructions to guide you further.

api/v1/database_types.go
```diff
@@ -220,6 +224,47 @@ type ExtensionSpec struct {
...
+	Handler *string `json:"handler,omitempty"`
+
+	// The pointer to the name of the validator function (e.g., "postgres_fdw_validator")
+	// It would be nil if the validator is not specified, as commonly the default validator is
+	// registrated with handler together when creating the fdw extension.
+	// +optional
+	Validator *string `json:"validator,omitempty"`
+
+	// Owner specifies the database user who will own the Foreign Data Wrapper.
+	// By default, the owner of a new FDW is the current session user.
```
gbartolini: I would remove this comment as there is no "current session user" concept in CNPG. We should also specify that the owner must be a role with superuser privileges.

gbartolini:
@EdwinaZhu it would be good if you could add support to the `usage` stanza as follows:

```yaml
kind: Database
metadata:
  name: app
spec:
  name: app
  owner: app
  cluster:
    name: pg-b
  extensions:
  - name: postgres_fdw
  fdws:
  - name: postgres_fdw
    usage:
    - name: app
      type: grant
```

When `type` is `grant`, you need to verify that the `app` user doesn't yet have the `usage` privilege on the `postgres_fdw` FDW, and then run:

```sql
GRANT USAGE ON FOREIGN DATA WRAPPER postgres_fdw TO app
```

In case of `revoke`, verify that the permission exists and run:

```sql
REVOKE USAGE ON FOREIGN DATA WRAPPER postgres_fdw FROM app
```

See:

- https://www.postgresql.org/docs/current/sql-grant.html
- https://www.postgresql.org/docs/current/sql-revoke.html

gbartolini:
After today's meeting we've decided to make the handler and validator fields strings. The `-` string represents a special meaning based on how Postgres shows the foreign data wrappers `NO HANDLER` and `NO VALIDATOR` cases.

```yaml
    handler: '-'
    validator: '-'
```

When `handler` and `validator` are empty strings, we skip them in the reconciliation cycle. If they are `-`, they are treated as `NO HANDLER` or `NO VALIDATOR`, otherwise they are set according to that value.

api/v1/database_types.go
```diff
@@ -220,6 +224,61 @@ type ExtensionSpec struct {
...
+	Options map[string]OptionSpecValue `json:"options,omitempty"`
+
+	// Usages specifies usages for the FDW
+	// +optional
+	Usages []UsageSpec `json:"usages,omitempty"`
+}
+
+// OptionSpecValue holds both the value and the ensure field for an option
+type OptionSpecValue struct {
+	Value string `json:"value"`
```
leonardoce, line 271: Let's add a comment before the declaration of the `Value` field.

api/v1/database_types.go
```diff
@@ -220,6 +224,61 @@ type ExtensionSpec struct {
...
+	// the database. If set to `present`, the option will be
+	// created if it does not exist. If set to `absent`, the
+	// option will be removed if it exists.
+	// +kubebuilder:default:="present"
+	// +kubebuilder:validation:Enum=present;absent
+	// +optional
+	Ensure EnsureOption `json:"ensure,omitempty"`
+}
+
+type UsageSpec struct {
```
leonardoce, line 284: Let's add a comment before this structure.

docs/src/samples/database-example.yaml
```diff
@@ -6,4 +6,4 @@ spec:
   name: one
   owner: app
   cluster:
-    name: cluster-example
+    name: cluster-example
```
leonardoce: Can we remove this change from the PR?

docs/src/declarative_database_management.md
```diff
@@ -256,6 +256,68 @@ Each schema entry supports the following properties:
...
+Each FDW entry supports the following properties:
+
+- `name` *(mandatory)*: The name of the foreign data wrapper.
+- `ensure`: Indicates whether the FDW should be `present` or `absent` in the database (default is `present`).
+- `handler`: The name of the handler function used by the FDW. If not specified, the default handler defined by the FDW extension (if any) will be used.
+- `validator`: The name of the validator function used by the FDW. If not specified, the default validator defined by the FDW extension (if any) will be used.
+- `owner`: The owner of the FDW.
+- `options`: A map of FDW options to manage. Each option supports:
+  - `value`: The string value of the option.
+  - `ensure`: Indicates whether the option should be `present` or `absent`.
```
leonardoce: I think we need to document the `usage` stanza.

tests/.DS_Store
leonardoce, line 1: We need to remove this file from the PR

leonardoce (changes requested):

leonardoce:
@EdwinaZhu I added a few comments. Can you address them? There is nothing essential to change.

leonardoce (approved):

api/v1/database_types.go
```diff
@@ -220,6 +224,64 @@ type ExtensionSpec struct {
...
+	// +optional
+	Owner string `json:"owner,omitempty"`
+
+	// Options specifies options for the FDW(key is option name, value is option value)
+	// +optional
+	Options map[string]OptionSpecValue `json:"options,omitempty"`
+
+	// Usages specifies usages for the FDW
+	// +optional
+	Usages []UsageSpec `json:"usages,omitempty"`
```
gbartolini:
```suggestion
	Usages []UsageSpec `json:"usage,omitempty"`
```

I suggest we use `usage` instead of `usages`.

gbartolini:
I love the `usage` grant functionality (please rename `usages` into `usage`).

I tried this:

```yaml
kind: Cluster
metadata:
  name: pg-a
spec:
  instances: 1
  storage:
    size: 1Gi

  bootstrap:
    initdb:
      postInitApplicationSQL:
        - CREATE TABLE n (i SERIAL PRIMARY KEY, m INTEGER)
        - INSERT INTO n (m) (SELECT generate_series(1, 10000))
        - ALTER TABLE n OWNER TO app
---
kind: Cluster
metadata:
  name: pg-b
spec:
  instances: 1
  storage:
    size: 1Gi
---
kind: Database
metadata:
  name: app
spec:
  name: app
  owner: app
  cluster:
    name: pg-b
  extensions:
  - name: postgres_fdw
  fdws:
  - name: postgres_fdw
    usages:
    - name: app
      type: grant
```

And then ran:

```sh
PGAPPPWD=`kubectl get secret pg-a-app -o jsonpath="{.data.password}" | base64 -d`
kubectl cnpg psql pg-b -- `kubectl get secret pg-b-app   -o jsonpath="{.data.uri}" | base64 -d` -c "CREATE SERVER pga FOREIGN DATA WRAPPER postgres_fdw OPTIONS (host 'pg-a-rw')"
kubectl cnpg psql pg-b -- `kubectl get secret pg-b-app   -o jsonpath="{.data.uri}" | base64 -d` -c "CREATE USER MAPPING FOR app SERVER pga OPTIONS (user 'app', password '$PGAPPPWD')"
kubectl cnpg psql pg-b -- `kubectl get secret pg-b-app   -o jsonpath="{.data.uri}" | base64 -d` -c "CREATE FOREIGN TABLE n (i INTEGER, m INTEGER) SERVER pga"
```

As you can see, thanks to the `USAGE` grant to the `app` user, I can now create the foreign server as `app` user.

gbartolini:
As agreed in today's meeting, this PR is ready to be merged, provided the above changes are implemented. However, given that we are in a feature freeze situation since we have released 1.27.0-rc1, we have decided to wait to merge this on trunk immediately after 1.27.0 is released.

gbartolini:
Moreover, @EdwinaZhu has agreed to extend her mentorship contribution to include also management of foreign servers in #8063. She will provide another PR based on this to implement that additional feature.

armru (approved):

gbartolini (approved):

mnencia (approved):

Issue #7503, opened 2025-05-07
[Bug]: Backup with method: volumeSnapshot won't reconcile


bo0tzz:
### Is there an existing issue already for this bug?

- [x] I have searched for an existing issue, and could not find anything. I believe this is a new bug.

### I have read the troubleshooting guide

- [x] I have read the troubleshooting guide and I think this is a new bug.

### I am running a supported version of CloudNativePG

- [x] I have read the troubleshooting guide and I think this is a new bug.

### Contact Details

bo0tzz@immich.app

### Version

1.25 (latest patch)

### What version of Kubernetes are you using?

1.32

### What is your Kubernetes environment?

Self-managed: k3s

### How did you install the operator?

Helm

### What happened?

When trying to create a Backup of a Cluster with `method: volumeSnapshot`, the backup object seems to not reconcile at all. The operator logs acknowledge its existence, but It doesn't get a `status` subresource and no VolumeSnapshot is created. I've tried every permutation I can think of between declarative and imperative, all the variations of `onlineConfiguration`, etc.

I've previously taken successful snapshots of databases on this kube cluster, and I also know snapshotting is still working as creating VolumeSnapshots for other PVCs went flawlessly. I think the operator isn't even getting that far.

### Cluster resource

```shell
apiVersion: postgresql.cnpg.io/v1
kind: Cluster
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"postgresql.cnpg.io/v1","kind":"Cluster","metadata":{"annotations":{},"name":"test-cluster","namespace":"testing"},"spec":{"backup":{"target":"prefer-standby","volumeSnapshot":{"className":"zfs"}},"bootstrap":{"initdb":{}},"imageName":"ghcr.io/tensorchord/cloudnative-pgvecto.rs:14.12-v0.3.0","instances":1,"storage":{"size":"10Gi","storageClass":"zfs"}}}
  creationTimestamp: "2025-05-07T08:30:31Z"
  generation: 1
  name: test-cluster
  namespace: testing
  resourceVersion: "140557050"
  uid: f8acfb13-153d-49ba-862c-0fbeb4eb4262
spec:
  affinity:
    podAntiAffinityType: preferred
  backup:
    target: prefer-standby
    volumeSnapshot:
      className: zfs
      online: true
      onlineConfiguration:
        immediateCheckpoint: false
        waitForArchive: true
      snapshotOwnerReference: none
[...]

PR #7944, opened 2025-07-01
fix(backup,online): fail backup if targetPod becomes unhealthy

### LINKED ISSUE #7905: [Bug]: Backup remains running after target pod is killed

### Contact Details

agelwarg@galileosuite.com

### Version

1.26 (latest patch)

1.31

Helm

### What happened?

When a backup is running and the target pod is terminated/killed, the backup status appears to remain `running` indefinitely.  In our case, it looks like the pod was terminated because it exhausted the ephemeral-storage limit, but I don't think that part is particularly relevant.   I would expect the same if you simply delete the pod while the backup is in progress.

### Cluster resource

```shell
kind: Cluster
metadata:
  name: test-database
spec:
  instances: 1
  imageCatalogRef:
    apiGroup: postgresql.cnpg.io
    kind: ImageCatalog
    name: test
    major: 17
  primaryUpdateMethod: switchover
  bootstrap:
    initdb:
      database: 'test'
      owner: 'test'
  backup:
    retentionPolicy: 3d
    barmanObjectStore:
      destinationPath: "s3://some-bucket/database"
      s3Credentials:
        inheritFromIAMRole: true
      data:
        compression: "snappy"
  storage:
    size: 50Gi
    storageClass: gp3-encrypted
```

### Relevant log output

```shell

```

### Code of Conduct

armru:
This patch ensures that the backup process is immediately aborted if the targetPod becomes unhealthy during an online backup operation. Previously, the backup would continue even if the target instance was no longer in a healthy state, potentially resulting in incomplete or inconsistent backups. By monitoring the pod's health status, the process now fails fast, providing early feedback and increasing the accuracy of the Backup object status.

Related #7503
Closes #7905

### Release notes

```
Online backup operation now fails immediately if the targetPod becomes
unhealthy, increasing the accuracy of the Backup object status.
```

### DIFF
--- a/internal/controller/backup_controller.go
+++ b/internal/controller/backup_controller.go
@@ -210,7 +210,7 @@ func (r *BackupReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 	}
 
 	if isRunning && backup.GetOnlineOrDefault(&cluster) {
-		if err := r.ensureTargetPodHealthy(ctx, &backup, &cluster); err != nil {
+		if err := r.ensureTargetPodHealthy(ctx, r.Client, &backup, &cluster); err != nil {
 			contextLogger.Error(err, "while ensuring target pod is healthy")
 			_ = resourcestatus.FlagBackupAsFailed(ctx, r.Client, &backup, nil,
 				fmt.Errorf("while ensuring target pod is healthy: %w", err))
@@ -729,6 +729,7 @@ func (r *BackupReconciler) SetupWithManager(ctx context.Context, mgr ctrl.Manage
 
 func (r *BackupReconciler) ensureTargetPodHealthy(
 	ctx context.Context,
+	cli client.Client,
 	backup *apiv1.Backup,
 	cluster *apiv1.Cluster,
 ) error {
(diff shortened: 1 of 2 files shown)


internal/controller/backup_controller.go
```diff
@@ -65,6 +66,11 @@ const backupPhase = ".status.phase"
 // where the name of the cluster is written
 const clusterName = ".spec.cluster.name"
 
+// isRunningResult ensures that we periodically check running backups. It could happen that the targetPod is destroyed
+// or stops responding.
+// This should be used almost always when a backup is running
+var isRunningResult = ctrl.Result{RequeueAfter: 10 * time.Minute}
```
uhthomas: It would probably be better to have the requeue duration as a const than this global variable.

```suggestion
const isRunningRequeueDuration = 10 * time.Minute
```

```go
return ctrl.Result{RequeueAfter: isRunningRequeueDuration}
```

armru: What if we add a function that creates the result? Something on the line: getBackupRunningResult() 
This would:
- remove the need of a global variable
- returns always the same object (the global variable could be conceptually modified)
- The object would be ready to be used!

leonardoce: Yep, that's what I did

leonardoce (approved):

NiccoloFei (approved):

mnencia (approved):

Issue #4407, opened 2024-04-30
[Bug]: Recovery from WAL fails when tablespace was added


wvengen:
### Is there an existing issue already for this bug?

- [X] I have searched for an existing issue, and could not find anything. I believe this is a new bug.

### I have read the troubleshooting guide

- [X] I have read the troubleshooting guide and I think this is a new bug.

### I am running a supported version of CloudNativePG

- [X] I have read the troubleshooting guide and I think this is a new bug.

### Contact Details

willem+cloudnativepg@thequestionmark.org

### Version

1.23.0

### What version of Kubernetes are you using?

1.26

### What is your Kubernetes environment?

Cloud: Other

### How did you install the operator?

Helm

### What happened?

1. Start with a database with backup setup (base + WAL) and make sure a base backup is present.
2. Add a (temporary) tablespace by updating the resource, and make sure WALs are backed up. I also deleted the cluster.
3. Create a new cluster by bootstrapping from recovery, from the backup.
4. The recovery process has a fatal error on the redo of the `Tablespace/CREATE` record.

I was able to recover to just before affected WAL record without data loss, but I can imagine that this could bring unexpected downtime during recovery in certain situations, hence this report.

### Cluster resource

```shell
apiVersion: postgresql.cnpg.io/v1
kind: Cluster
metadata:
  name: db2
  labels:
    cnpg.io/reload: 'true'
spec:
  instances: 1
  imageName: ghcr.io/cloudnative-pg/postgresql:16.2

  storage:
    size: 80Gi

  # tried restoring both with and without this tablespace
  # in both cases, restore failed because the tablespace was not setup
  # ideally, I would be able to restore with a different temporary tablespace setup
  tablespaces:
    - name: temporary
      storage:
        size: 40Gi
      temporary: true

  env:
    - name: PSQL_HISTORY
      value: /run/psql_history

  bootstrap:
    recovery:
      source: clusterBackup
      database: db
      owner: db
      secret:
        name: db-secret

  resources:
    limits:
      memory: 6000Mi
    requests:
      memory: 3000Mi

  postgresql:
    parameters:
      checkpoint_completion_target: "0.9"
      default_statistics_target: "100"
      effective_cache_size: 4608MB
      effective_io_concurrency: "300"
      maintenance_work_mem: 384MB
      max_connections: "20"
      max_standby_streaming_delay: "300000"
      max_wal_size: 4GB
      min_wal_size: 1GB
[...]

PR #7972, opened 2025-07-03
Add note on WAL recovery and tablespace changes

wvengen:
as discussed in #4407

### DIFF
--- a/docs/src/recovery.md
+++ b/docs/src/recovery.md
@@ -375,11 +375,10 @@ recovery target.
     the volume snapshot is before the recovery target timestamp.
 
 !!! Warning
-    If you recently added or removed a [tablespace](tablespaces.md) in your
-    cluster, replaying the WAL will fail. You need a base backup between the
-    time of the tablespace change and the recovery target timestamp. When you
-    plan to add or remove a tablespace and rely on PITR recovery, testing is
-    recommended.
+    If you added or removed a [tablespace](tablespaces.md) in your cluster
+    since the last base backup, replaying the WAL will fail. You need a base
+    backup between the time of the tablespace change and the recovery target
+    timestamp.
 
 ### Recovery targets
 
--- a/docs/src/recovery.md
+++ b/docs/src/recovery.md
@@ -374,6 +374,13 @@ recovery target.
     It's your responsibility to ensure that the end time of the base backup in
     the volume snapshot is before the recovery target timestamp.
 
+!!! Warning
+    If you recently added or removed a [tablespace](tablespaces.md) in your
+    cluster, replaying the WAL will fail. You need a base backup between the
+    time of the tablespace change and the recovery target timestamp. When you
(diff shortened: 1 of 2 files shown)


jsilvela (commented):
Good addition, thanks!
I'd consider:
- putting this in a `!!! Warning` or `!!! Important` box.
- perhaps adding a reference in the restore page

docs/src/tablespaces.md
```diff
@@ -356,3 +356,6 @@ Instances status
 
 Currently, you can't remove tablespaces from an existing CloudNativePG
 cluster.
+
+When you add or remove a tablespace in an existing cluster, recovery
+from WAL will fail, until you take a new base backup. Testing recommended.
```
jsilvela: IMO the comma makes the "fail" seem a bit too permanent.
```suggestion
from WAL will fail until you take a new base backup. Testing recommended.
```

wvengen:
Thank you for your review! I've made the suggested changes. I found it hard to find a suitable place for the warning on the recovery page, ideally in the PITR section but after the rest (to avoid burding a first-time reader with too many details like this). If you have a better suggestion, looking forward to hearing it.

docs/src/recovery.md
```diff
@@ -374,6 +374,13 @@ recovery target.
     It's your responsibility to ensure that the end time of the base backup in
     the volume snapshot is before the recovery target timestamp.
 
+!!! Warning
+    If you recently added or removed a [tablespace](tablespaces.md) in your
+    cluster, replaying the WAL will fail. You need a base backup between the
```
jsilvela: "recently" does not send the right message. I think you mean, since the last base backup.

docs/src/tablespaces.md
```diff
@@ -254,6 +254,10 @@ tablespace map) both on object stores and volume snapshots.
     backup. The lag will be resolved in a maximum of 5 minutes, with the next
     reconciliation.
 
+!!! Warning
+    When you add or remove a tablespace in an existing cluster, recovery
+    from WAL will fail until you take a new base backup. Testing recommended.
```
jsilvela: In the previous paragraph, and here too "testing recommended" is not specific enough to be useful.
What are you saying? That after making a change you need to test? Test what specifically?
IMO you can either remove it or make it more specific.

wvengen: Yeah, I agree that it could be more specific. I added it, because I have not specifically tested various scenario's of recovery after a tablespace change, just that it should work after a base backup according to knowledge of the workings of PostgreSQL. But that would be too much explanation, in my view, so I'll leave it without the testing: having these sections is better than not having them :)

wvengen:
Thank you for your reviews, helpful!

gbartolini (approved):

jsilvela (approved):

gbartolini:
(apologies, I made a mistake and merged without ensuring conventional commit message)

PR #7979, opened 2025-07-04
feat(pooler): enable `stats_users` to be configured

### LINKED ISSUE #7973: [Feature]: allow stats_users to be configured in Pooler

We would like to setup pgbouncer monitoring via Datadog. Unfortunately, since the pgbouncer console is locked down to local access only, we are unable to setup a connection to it from Datadog.

The least work for the developers would be to allow `stats_users` to be configurable. With this, we can add our Datadog user and add that user to the existing DB to piggyback off of the auth_query.

I tried customizing the pg_hba to allow a remote connection to the `pgbouncer` user but that's not allowed. I also tried setting `stats_users` but that is rejected by the webhook.

I also tried doing something like

```
max_prepared_statements: |
  0
  stats_users = "datadog"
```

but unfortunately, [cleanupPgBouncerValue()](https://github.com/cloudnative-pg/cloudnative-pg/blob/main/pkg/management/pgbouncer/config/strings.go#L78) catches this case correctly.

### Additional context

_No response_

### Backport?

Yes

Yes

### Code of Conduct

gbartolini:
Enable PgBouncer's `stats_users` parameter in the webhook.

Closes #7973

### DIFF
--- a/docs/src/connection_pooling.md
+++ b/docs/src/connection_pooling.md
@@ -374,6 +374,7 @@ are the ones directly set by PgBouncer.
 - [`server_tls_ciphers`](https://www.pgbouncer.org/config.html#server_tls_ciphers)
 - [`server_tls_protocols`](https://www.pgbouncer.org/config.html#server_tls_protocols)
 - [`stats_period`](https://www.pgbouncer.org/config.html#stats_period)
+- [`stats_users`](https://www.pgbouncer.org/config.html#stats_users)
 - [`suspend_timeout`](https://www.pgbouncer.org/config.html#suspend_timeout)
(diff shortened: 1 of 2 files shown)


mnencia (approved):

gbartolini:
I tried the patch on the plane and - as I was afraid - it doesn't work. I am not actually convinced we can access the administration database remotely - but I have not been used PgBouncer in a while and I don't have much time to dig now.

oddur:
I would love to be able to use the datadog native pgBouncer integration, which requires this.

Denton-L:
@gbartolini Sorry for the super delayed response. I tested this out and it works. I just had to also set the corresponding `auth_dbname` variable but after that, auth seemed to work well.

PR #7991, opened 2025-07-07
feat: dynamic loading of PostgreSQL extensions

### LINKED ISSUE #7188: [Feature]: Dynamic loading of PostgreSQL extensions

Currently, extensions must be part of the base image of the operand, requiring pre-determination of the entire content of the Postgres image and preventing dynamic loading of extensions.

The reason is detailed in this blog article that I wrote: https://www.gabrielebartolini.it/articles/2025/03/the-immutable-future-of-postgresql-extensions-in-kubernetes-with-cloudnativepg/

Following the exploration work done in #6546 and the PostgreSQL 18 feature for `extension_control_path` we contributed to (see https://github.com/postgres/postgres/commit/4f7f7b0375854e2f89876473405a8f21c95012af), we should enable dynamic loading of extensions via the new `ImageVolume` feature introduced in Kubernetes 1.33. This guarantees immutability of container images.

Rollout is required by image volumes.

This feature will only be available with Postgres 18 and Kubernetes 1.33+. Also, extensions must be distributed as self-contained images - this is part of another initiative, outside CloudNativePG.

Image catalogs must be considered as well.

We have considered mutable solutions (use a mutable volume to install needed extensions), but we preferred to contribute to Postgres improvements and exploit innovation in Kubernetes with Image Volumes.

### Additional context

See the blog article cited above.

### Backport?

No

Yes

### Code of Conduct

NiccoloFei:
This feature allows you to mount a PostgreSQL extension, packaged as an OCI-compliant container image, as a read-only, immutable volume inside an instance pod.

The new `.spec.postgresql.extensions` stanza of a `Cluster` resource accepts an ordered list of extension images to be added to the PostgreSQL cluster. Each item in the `extensions` stanza is identified by `name` and provides the following options:

- `extension_control_path`: A list of relative paths within the container image to be appended to PostgreSQL’s `extension_control_path`, allowing it to locate extension control files (default: `share`).
- `dynamic_library_path`: A list of relative paths within the container image to be appended to PostgreSQL’s `dynamic_library_path`, enabling it to locate shared library files for extensions (default: `lib`).
- `ld_library_path`: A list of relative paths within the container image to be appended to the `LD_LIBRARY_PATH` environment variable of the instance manager process, allowing PostgreSQL to locate required system libraries at runtime (default: empty).

Each image volume is mounted at `/extensions/<EXTENSION_NAME>`. By default, CloudNativePG automatically manages the relevant PostgreSQL GUCs, setting:

- `extension_control_path` to `/extensions/<EXTENSION_NAME>/share`, allowing PostgreSQL to locate any extension control file within `/extensions/<EXTENSION_NAME>/share/extension`
- `dynamic_library_path` to `/extensions/<EXTENSION_NAME>/lib`

These values are appended in the order in which the extensions are defined in the `extensions` list, ensuring deterministic path resolution within PostgreSQL.

Requirements:
- PostgreSQL 18 or higher (for the `extension_control_path` option)
- Kubernetes 1.33+ with the `ImageVolume` feature enabled

Closes #7188


NiccoloFei:
# Testing

1. Deploy a kind cluster from this branch, running:  `hack/setup-cluster.sh create load deploy`
    (The imageVolume feature gate will be enabled by default)

2. Apply a PG 18beta1 Cluster with PostGIS and pgVector extensions
```yaml
kind: Cluster
metadata:
  name: postgresql-with-extensions
spec:
  instances: 3
  imageName: ghcr.io/cloudnative-pg/postgresql:18beta1-minimal-bookworm

  bootstrap:
    initdb:
      database: app
      owner: app

  postgresql:
    extensions:
      - name: pgvector
        image:
          reference: ghcr.io/niccolofei/pgvector:18beta1-master-bookworm
      - name: postgis
        ld_library_path:
          - system
        image:
          reference: ghcr.io/niccolofei/postgis:18beta1-master-bookworm

  storage:
    size: 1Gi
---
kind: Database
metadata:
  name: app
spec:
  name: app
  owner: app
  cluster:
    name: postgresql-with-extensions
  extensions:
  - name: vector
    ensure: present
  - name: postgis
    ensure: present
  - name: postgis_raster
    ensure: present
  - name: postgis_sfcgal
    ensure: present
  - name: fuzzystrmatch
    ensure: present
  - name: address_standardizer
    ensure: present
  - name: address_standardizer_data_us
    ensure: present
  - name: postgis_tiger_geocoder
    ensure: present
  - name: postgis_topology
    ensure: present
```

# Image Extensions

https://github.com/NiccoloFei/pg-extensions-containers is the repo that contains the image extensions.
It currently builds the following extensions: 
* pgvector: https://github.com/NiccoloFei/pg-extensions-containers/pkgs/container/pgvector
* postgis: https://github.com/NiccoloFei/pg-extensions-containers/pkgs/container/postgis

for Debian bookworm and bullseye. All extensions are built from source since there are no packages yet for PG18beta1.

gbartolini:
This looks great Niccolò. I just tried it and it seems to work, but I will continue in the next days to review it.

Volumes are mount:

```sh
kubectl exec -ti postgresql-with-extensions-1 -c postgres -- mount
```

Returns:

```console
...
overlay on /extensions/pgvector type overlay (ro,relatime,lowerdir=/var/lib/containerd/io.containerd.snapshotter.v1.overlayfs/snapshots/70/fs:/var/lib/containerd/io.containerd.snapshotter.v1.overlayfs/snapshots/69/fs,upperdir=/var/lib/containerd/io.containerd.snapshotter.v1.overlayfs/snapshots/141/fs,workdir=/var/lib/containerd/io.containerd.snapshotter.v1.overlayfs/snapshots/141/work,uuid=on,volatile)
overlay on /extensions/postgis type overlay (ro,relatime,lowerdir=/var/lib/containerd/io.containerd.snapshotter.v1.overlayfs/snapshots/74/fs:/var/lib/containerd/io.containerd.snapshotter.v1.overlayfs/snapshots/73/fs:/var/lib/containerd/io.containerd.snapshotter.v1.overlayfs/snapshots/72/fs:/var/lib/containerd/io.containerd.snapshotter.v1.overlayfs/snapshots/71/fs,upperdir=/var/lib/containerd/io.containerd.snapshotter.v1.overlayfs/snapshots/142/fs,workdir=/var/lib/containerd/io.containerd.snapshotter.v1.overlayfs/snapshots/142/work,uuid=on,volatile)
...
```

I also ran this:

```sh
kubectl exec -ti postgresql-with-extensions-1 -c postgres -- find /extensions -type f | pbcopy
```
And got this:

```console
/extensions/postgis/bin/raster2pgsql
/extensions/postgis/bin/shp2pgsql
/extensions/postgis/bin/postgis_restore
/extensions/postgis/bin/postgis
/extensions/postgis/bin/pgtopo_import
/extensions/postgis/bin/pgtopo_export
/extensions/postgis/bin/pgsql2shp
/extensions/postgis/system/libSFCGAL.so.1
/extensions/postgis/system/libcfitsio.so.10
/extensions/postgis/system/libgeotiff.so.5
/extensions/postgis/system/libnetcdf.so.19
/extensions/postgis/system/libprotobuf-c.so.1
/extensions/postgis/system/libarpack.so.2
/extensions/postgis/system/libnspr4.so
/extensions/postgis/system/libdfalt.so.0
/extensions/postgis/system/libhdf5_serial_hl.so.100
/extensions/postgis/system/libsmime3.so
/extensions/postgis/system/libsuperlu.so.5
/extensions/postgis/system/libnghttp2.so.14
/extensions/postgis/system/libdeflate.so.0
/extensions/postgis/system/libtiff.so.6
/extensions/postgis/system/libfygm.so.0
/extensions/postgis/system/libjpeg.so.62
/extensions/postgis/system/libjbig.so.0
/extensions/postgis/system/libmpfr.so.6
/extensions/postgis/system/libnss3.so
/extensions/postgis/system/libexpat.so.1
/extensions/postgis/system/libhdf5_serial.so.103
/extensions/postgis/system/libwebp.so.7
/extensions/postgis/system/libogdi.so.4.1
/extensions/postgis/system/libcurl.so.4
/extensions/postgis/system/libLerc.so.4
/extensions/postgis/system/libgdal.so.32
/extensions/postgis/system/libspatialite.so.7
/extensions/postgis/system/librtmp.so.1
/extensions/postgis/system/liblapack.so.3
/extensions/postgis/system/libdav1d.so.6
/extensions/postgis/system/libnssutil3.so
/extensions/postgis/system/libbrotlicommon.so.1
/extensions/postgis/system/libgfortran.so.5
/extensions/postgis/system/libgif.so.7
/extensions/postgis/system/libfreetype.so.6
/extensions/postgis/system/libx265.so.199
/extensions/postgis/system/libkmlengine.so.1
/extensions/postgis/system/libmfhdfalt.so.0
/extensions/postgis/system/libfyba.so.0
/extensions/postgis/system/libjson-c.so.5
/extensions/postgis/system/libplds4.so
/extensions/postgis/system/libfontconfig.so.1
/extensions/postgis/system/libboost_serialization.so.1.74.0
/extensions/postgis/system/libsz.so.2
/extensions/postgis/system/libgeos_c.so.1
/extensions/postgis/system/liblcms2.so.2
/extensions/postgis/system/libblas.so.3
/extensions/postgis/system/libpsl.so.5
/extensions/postgis/system/libopenjp2.so.7
/extensions/postgis/system/libcurl-gnutls.so.4
/extensions/postgis/system/libplc4.so
/extensions/postgis/system/libsnappy.so.1
/extensions/postgis/system/librttopo.so.1
/extensions/postgis/system/libkmldom.so.1
/extensions/postgis/system/libpoppler.so.126
/extensions/postgis/system/libodbcinst.so.2
/extensions/postgis/system/libfyut.so.0
/extensions/postgis/system/liburiparser.so.1
/extensions/postgis/system/libltdl.so.7
/extensions/postgis/system/libtirpc.so.3
/extensions/postgis/system/libxerces-c-3.2.so
/extensions/postgis/system/libkmlbase.so.1
/extensions/postgis/system/libarmadillo.so.11
/extensions/postgis/system/libpng16.so.16
/extensions/postgis/system/libproj.so.25
/extensions/postgis/system/libfreexl.so.1
/extensions/postgis/system/libssh2.so.1
/extensions/postgis/system/libaec.so.0
/extensions/postgis/system/libqhull_r.so.8.0
/extensions/postgis/system/libodbc.so.2
/extensions/postgis/system/libaom.so.3
/extensions/postgis/system/libblosc.so.1
/extensions/postgis/system/libgmpxx.so.4
/extensions/postgis/system/libgeos.so.3.11.1
/extensions/postgis/system/libnuma.so.1
/extensions/postgis/system/libminizip.so.1
/extensions/postgis/system/libbrotlidec.so.1
/extensions/postgis/system/libmariadb.so.3
/extensions/postgis/system/libde265.so.0
/extensions/postgis/system/libheif.so.1
/extensions/postgis/share/man/man1/pgsql2shp.1
/extensions/postgis/share/man/man1/pgtopo_import.1
/extensions/postgis/share/man/man1/postgis_restore.1
/extensions/postgis/share/man/man1/postgis.1
/extensions/postgis/share/man/man1/pgtopo_export.1
/extensions/postgis/share/man/man1/shp2pgsql.1
/extensions/postgis/share/extension/postgis_topology--unpackaged--3.6.0dev.sql
/extensions/postgis/share/extension/postgis--unpackaged.sql
/extensions/postgis/share/extension/postgis.control
/extensions/postgis/share/extension/address_standardizer_data_us--3.6.0dev--3.6.0devnext.sql
/extensions/postgis/share/extension/postgis_raster.control
/extensions/postgis/share/extension/postgis_tiger_geocoder--3.6.0dev.sql
/extensions/postgis/share/extension/address_standardizer--ANY--3.6.0dev.sql
/extensions/postgis/share/extension/postgis_topology--3.6.0dev.sql
/extensions/postgis/share/extension/postgis_topology--ANY--3.6.0dev.sql
/extensions/postgis/share/extension/postgis_sfcgal.control
/extensions/postgis/share/extension/postgis_sfcgal--ANY--3.6.0dev.sql
/extensions/postgis/share/extension/postgis_raster--ANY--3.6.0dev.sql
/extensions/postgis/share/extension/address_standardizer.control
/extensions/postgis/share/extension/postgis_tiger_geocoder--ANY--3.6.0dev.sql
/extensions/postgis/share/extension/postgis--TEMPLATED--TO--ANY.sql
/extensions/postgis/share/extension/postgis_topology--unpackaged.sql
/extensions/postgis/share/extension/postgis_raster--TEMPLATED--TO--ANY.sql
/extensions/postgis/share/extension/postgis_topology.control
/extensions/postgis/share/extension/address_standardizer_data_us--3.6.0devnext--3.6.0dev.sql
/extensions/postgis/share/extension/postgis--3.6.0dev.sql
/extensions/postgis/share/extension/postgis_sfcgal--unpackaged.sql
/extensions/postgis/share/extension/postgis--ANY--3.6.0dev.sql
/extensions/postgis/share/extension/postgis_topology--TEMPLATED--TO--ANY.sql
/extensions/postgis/share/extension/address_standardizer_data_us.control
/extensions/postgis/share/extension/postgis_raster--unpackaged--3.6.0dev.sql
/extensions/postgis/share/extension/postgis_sfcgal--unpackaged--3.6.0dev.sql
/extensions/postgis/share/extension/address_standardizer_data_us--ANY--3.6.0dev.sql
/extensions/postgis/share/extension/postgis_tiger_geocoder--unpackaged--3.6.0dev.sql
/extensions/postgis/share/extension/address_standardizer--TEMPLATED--TO--ANY.sql
/extensions/postgis/share/extension/postgis_tiger_geocoder.control
/extensions/postgis/share/extension/address_standardizer--3.6.0dev.sql
/extensions/postgis/share/extension/address_standardizer_data_us--3.6.0dev.sql
/extensions/postgis/share/extension/address_standardizer--1.0--3.6.0dev.sql
/extensions/postgis/share/extension/postgis--unpackaged--3.6.0dev.sql
/extensions/postgis/share/extension/postgis_sfcgal--TEMPLATED--TO--ANY.sql
/extensions/postgis/share/extension/postgis_raster--unpackaged.sql
/extensions/postgis/share/extension/postgis_sfcgal--3.6.0dev.sql
/extensions/postgis/share/extension/postgis_raster--3.6.0dev.sql
/extensions/postgis/share/extension/address_standardizer_data_us--TEMPLATED--TO--ANY.sql
/extensions/postgis/share/extension/postgis_tiger_geocoder--TEMPLATED--TO--ANY.sql
/extensions/postgis/share/contrib/postgis-3.6/legacy_minimal.sql
/extensions/postgis/share/contrib/postgis-3.6/legacy_gist.sql
/extensions/postgis/share/contrib/postgis-3.6/uninstall_postgis.sql
/extensions/postgis/share/contrib/postgis-3.6/topology_comments.sql
/extensions/postgis/share/contrib/postgis-3.6/spatial_ref_sys.sql
/extensions/postgis/share/contrib/postgis-3.6/rtpostgis.sql
/extensions/postgis/share/contrib/postgis-3.6/uninstall_topology.sql
/extensions/postgis/share/contrib/postgis-3.6/sfcgal.sql
/extensions/postgis/share/contrib/postgis-3.6/postgis.sql
/extensions/postgis/share/contrib/postgis-3.6/uninstall_legacy.sql
/extensions/postgis/share/contrib/postgis-3.6/uninstall_sfcgal.sql
/extensions/postgis/share/contrib/postgis-3.6/sfcgal_upgrade.sql
/extensions/postgis/share/contrib/postgis-3.6/rtpostgis_upgrade.sql
/extensions/postgis/share/contrib/postgis-3.6/legacy.sql
/extensions/postgis/share/contrib/postgis-3.6/postgis_upgrade.sql
/extensions/postgis/share/contrib/postgis-3.6/uninstall_rtpostgis.sql
/extensions/postgis/share/contrib/postgis-3.6/raster_comments.sql
/extensions/postgis/share/contrib/postgis-3.6/postgis_comments.sql
/extensions/postgis/share/contrib/postgis-3.6/topology_upgrade.sql
/extensions/postgis/share/contrib/postgis-3.6/topology.sql
/extensions/postgis/share/contrib/postgis-3.6/sfcgal_comments.sql
/extensions/postgis/share/contrib/postgis-3.6/rtpostgis_legacy.sql
/extensions/postgis/lib/postgis_raster-3.so
/extensions/postgis/lib/address_standardizer-3.so
/extensions/postgis/lib/bitcode/address_standardizer-3.index.bc
/extensions/postgis/lib/bitcode/postgis-3.index.bc
/extensions/postgis/lib/bitcode/postgis_raster-3/rtpg_raster_properties.bc
/extensions/postgis/lib/bitcode/postgis_raster-3/rtpg_legacy.bc
/extensions/postgis/lib/bitcode/postgis_raster-3/rtpg_utility.bc
/extensions/postgis/lib/bitcode/postgis_raster-3/rtpg_gdal.bc
/extensions/postgis/lib/bitcode/postgis_raster-3/rtpostgis.bc
/extensions/postgis/lib/bitcode/postgis_raster-3/rtpg_wkb.bc
/extensions/postgis/lib/bitcode/postgis_raster-3/rtpg_geometry.bc
/extensions/postgis/lib/bitcode/postgis_raster-3/rtpg_mapalgebra.bc
/extensions/postgis/lib/bitcode/postgis_raster-3/rtpg_spatial_relationship.bc
/extensions/postgis/lib/bitcode/postgis_raster-3/rtpg_band_properties.bc
/extensions/postgis/lib/bitcode/postgis_raster-3/rtpg_internal.bc
/extensions/postgis/lib/bitcode/postgis_raster-3/rtpg_inout.bc
/extensions/postgis/lib/bitcode/postgis_raster-3/rtpg_create.bc
/extensions/postgis/lib/bitcode/postgis_raster-3/rtpg_pixel.bc
/extensions/postgis/lib/bitcode/postgis_raster-3/rtpg_statistics.bc
/extensions/postgis/lib/bitcode/address_standardizer-3/analyze.bc
/extensions/postgis/lib/bitcode/address_standardizer-3/lexicon.bc
/extensions/postgis/lib/bitcode/address_standardizer-3/std_pg_hash.bc
/extensions/postgis/lib/bitcode/address_standardizer-3/hash.bc
/extensions/postgis/lib/bitcode/address_standardizer-3/export.bc
/extensions/postgis/lib/bitcode/address_standardizer-3/address_parser.bc
/extensions/postgis/lib/bitcode/address_standardizer-3/address_standardizer.bc
/extensions/postgis/lib/bitcode/address_standardizer-3/gamma.bc
/extensions/postgis/lib/bitcode/address_standardizer-3/standard.bc
/extensions/postgis/lib/bitcode/address_standardizer-3/err_param.bc
/extensions/postgis/lib/bitcode/address_standardizer-3/parseaddress-api.bc
/extensions/postgis/lib/bitcode/address_standardizer-3/pagc_tools.bc
/extensions/postgis/lib/bitcode/address_standardizer-3/tokenize.bc
/extensions/postgis/lib/bitcode/postgis_topology-3.index.bc
/extensions/postgis/lib/bitcode/postgis_topology-3/postgis_topology.bc
/extensions/postgis/lib/bitcode/postgis_sfcgal-3.index.bc
/extensions/postgis/lib/bitcode/postgis_raster-3.index.bc
/extensions/postgis/lib/bitcode/postgis-3/lwgeom_btree.bc
/extensions/postgis/lib/bitcode/postgis-3/lwgeom_out_geojson.bc
/extensions/postgis/lib/bitcode/postgis-3/lwgeom_functions_temporal.bc
/extensions/postgis/lib/bitcode/postgis-3/lwgeom_window.bc
/extensions/postgis/lib/bitcode/postgis-3/lwgeom_in_gml.bc
/extensions/postgis/lib/bitcode/postgis-3/gserialized_estimate.bc
/extensions/postgis/lib/bitcode/postgis-3/lwgeom_sqlmm.bc
/extensions/postgis/lib/bitcode/postgis-3/geobuf.pb-c.bc
/extensions/postgis/lib/bitcode/postgis-3/lwgeom_inout.bc
/extensions/postgis/lib/bitcode/postgis-3/vector_tile.pb-c.bc
/extensions/postgis/lib/bitcode/postgis-3/gserialized_typmod.bc
/extensions/postgis/lib/bitcode/postgis-3/lwgeom_out_mvt.bc
/extensions/postgis/lib/bitcode/postgis-3/lwgeom_itree.bc
/extensions/postgis/lib/bitcode/postgis-3/mvt.bc
/extensions/postgis/lib/bitcode/postgis-3/lwgeom_geos_clean.bc
/extensions/postgis/lib/bitcode/postgis-3/lwgeom_functions_basic.bc
/extensions/postgis/lib/bitcode/postgis-3/lwgeom_generate_grid.bc
/extensions/postgis/lib/bitcode/postgis-3/gserialized_spgist_nd.bc
/extensions/postgis/lib/bitcode/postgis-3/lwgeom_functions_analytic.bc
/extensions/postgis/lib/bitcode/postgis-3/geometry_inout.bc
/extensions/postgis/lib/bitcode/postgis-3/lwgeom_geos.bc
/extensions/postgis/lib/bitcode/postgis-3/gserialized_spgist_2d.bc
/extensions/postgis/lib/bitcode/postgis-3/lwgeom_in_geojson.bc
/extensions/postgis/lib/bitcode/postgis-3/lwgeom_geos_prepared.bc
/extensions/postgis/lib/bitcode/postgis-3/lwgeom_box.bc
/extensions/postgis/lib/bitcode/postgis-3/lwgeom_out_flatgeobuf.bc
/extensions/postgis/lib/bitcode/postgis-3/postgis_legacy.bc
/extensions/postgis/lib/bitcode/postgis-3/geography_measurement.bc
/extensions/postgis/lib/bitcode/postgis-3/lwgeom_in_marc21.bc
/extensions/postgis/lib/bitcode/postgis-3/geography_centroid.bc
/extensions/postgis/lib/bitcode/postgis-3/gserialized_supportfn.bc
/extensions/postgis/lib/bitcode/postgis-3/lwgeom_spheroid.bc
/extensions/postgis/lib/bitcode/postgis-3/brin_2d.bc
/extensions/postgis/lib/bitcode/postgis-3/postgis_module.bc
/extensions/postgis/lib/bitcode/postgis-3/geography_btree.bc
/extensions/postgis/lib/bitcode/postgis-3/lwgeom_box3d.bc
/extensions/postgis/lib/bitcode/postgis-3/lwgeom_in_geohash.bc
/extensions/postgis/lib/bitcode/postgis-3/lwgeom_remove_small_parts.bc
/extensions/postgis/lib/bitcode/postgis-3/lwgeom_rectree.bc
/extensions/postgis/lib/bitcode/postgis-3/lwgeom_out_marc21.bc
/extensions/postgis/lib/bitcode/postgis-3/lwgeom_remove_irrelevant_points_for_view.bc
/extensions/postgis/lib/bitcode/postgis-3/lwgeom_union.bc
/extensions/postgis/lib/bitcode/postgis-3/lwgeom_out_geobuf.bc
/extensions/postgis/lib/bitcode/postgis-3/geography_inout.bc
/extensions/postgis/lib/bitcode/postgis-3/lwgeom_export.bc
/extensions/postgis/lib/bitcode/postgis-3/lwgeom_triggers.bc
/extensions/postgis/lib/bitcode/postgis-3/brin_nd.bc
/extensions/postgis/lib/bitcode/postgis-3/brin_common.bc
/extensions/postgis/lib/bitcode/postgis-3/geobuf.bc
/extensions/postgis/lib/bitcode/postgis-3/lwgeom_accum.bc
/extensions/postgis/lib/bitcode/postgis-3/lwgeom_in_flatgeobuf.bc
/extensions/postgis/lib/bitcode/postgis-3/flatgeobuf.bc
/extensions/postgis/lib/bitcode/postgis-3/geography_measurement_trees.bc
/extensions/postgis/lib/bitcode/postgis-3/postgis_libprotobuf.bc
/extensions/postgis/lib/bitcode/postgis-3/lwgeom_in_kml.bc
/extensions/postgis/lib/bitcode/postgis-3/lwgeom_geos_relatematch.bc
/extensions/postgis/lib/bitcode/postgis-3/lwgeom_transform.bc
/extensions/postgis/lib/bitcode/postgis-3/gserialized_gist_2d.bc
/extensions/postgis/lib/bitcode/postgis-3/lwgeom_functions_lrs.bc
/extensions/postgis/lib/bitcode/postgis-3/lwgeom_geos_predicates.bc
/extensions/postgis/lib/bitcode/postgis-3/lwgeom_dumppoints.bc
/extensions/postgis/lib/bitcode/postgis-3/lwgeom_in_encoded_polyline.bc
/extensions/postgis/lib/bitcode/postgis-3/gserialized_spgist_3d.bc
/extensions/postgis/lib/bitcode/postgis-3/lwgeom_dump.bc
/extensions/postgis/lib/bitcode/postgis-3/gserialized_gist_nd.bc
/extensions/postgis/lib/bitcode/postgis-3/lwgeom_ogc.bc
/extensions/postgis/lib/bitcode/postgis_sfcgal-3/postgis_sfcgal_legacy.bc
/extensions/postgis/lib/bitcode/postgis_sfcgal-3/lwgeom_sfcgal.bc
/extensions/postgis/lib/postgis-3.so
/extensions/postgis/lib/postgis_topology-3.so
/extensions/postgis/lib/postgis_sfcgal-3.so
/extensions/pgvector/lib/vector.so
/extensions/pgvector/lib/bitcode/vector.index.bc
/extensions/pgvector/lib/bitcode/vector/src/ivfvacuum.bc
/extensions/pgvector/lib/bitcode/vector/src/hnswvacuum.bc
/extensions/pgvector/lib/bitcode/vector/src/hnswinsert.bc
/extensions/pgvector/lib/bitcode/vector/src/halfutils.bc
/extensions/pgvector/lib/bitcode/vector/src/hnswbuild.bc
/extensions/pgvector/lib/bitcode/vector/src/halfvec.bc
/extensions/pgvector/lib/bitcode/vector/src/ivfbuild.bc
/extensions/pgvector/lib/bitcode/vector/src/hnsw.bc
/extensions/pgvector/lib/bitcode/vector/src/hnswutils.bc
/extensions/pgvector/lib/bitcode/vector/src/bitvec.bc
/extensions/pgvector/lib/bitcode/vector/src/ivfkmeans.bc
/extensions/pgvector/lib/bitcode/vector/src/ivfinsert.bc
/extensions/pgvector/lib/bitcode/vector/src/ivfutils.bc
/extensions/pgvector/lib/bitcode/vector/src/vector.bc
/extensions/pgvector/lib/bitcode/vector/src/sparsevec.bc
/extensions/pgvector/lib/bitcode/vector/src/ivfflat.bc
/extensions/pgvector/lib/bitcode/vector/src/bitutils.bc
/extensions/pgvector/lib/bitcode/vector/src/hnswscan.bc
/extensions/pgvector/lib/bitcode/vector/src/ivfscan.bc
/extensions/pgvector/share/extension/vector--0.2.3--0.2.4.sql
/extensions/pgvector/share/extension/vector--0.2.2--0.2.3.sql
/extensions/pgvector/share/extension/vector.control
/extensions/pgvector/share/extension/vector--0.2.4--0.2.5.sql
/extensions/pgvector/share/extension/vector--0.4.0--0.4.1.sql
/extensions/pgvector/share/extension/vector--0.1.0--0.1.1.sql
/extensions/pgvector/share/extension/vector--0.7.2--0.7.3.sql
/extensions/pgvector/share/extension/vector--0.4.4--0.5.0.sql
/extensions/pgvector/share/extension/vector--0.3.0--0.3.1.sql
/extensions/pgvector/share/extension/vector--0.7.3--0.7.4.sql
/extensions/pgvector/share/extension/vector--0.1.3--0.1.4.sql
/extensions/pgvector/share/extension/vector--0.4.1--0.4.2.sql
/extensions/pgvector/share/extension/vector--0.7.0--0.7.1.sql
/extensions/pgvector/share/extension/vector--0.6.1--0.6.2.sql
/extensions/pgvector/share/extension/vector--0.1.6--0.1.7.sql
/extensions/pgvector/share/extension/vector--0.1.4--0.1.5.sql
/extensions/pgvector/share/extension/vector--0.2.7--0.3.0.sql
/extensions/pgvector/share/extension/vector--0.8.0.sql
/extensions/pgvector/share/extension/vector--0.3.1--0.3.2.sql
/extensions/pgvector/share/extension/vector--0.6.2--0.7.0.sql
/extensions/pgvector/share/extension/vector--0.2.6--0.2.7.sql
/extensions/pgvector/share/extension/vector--0.2.0--0.2.1.sql
/extensions/pgvector/share/extension/vector--0.7.1--0.7.2.sql
/extensions/pgvector/share/extension/vector--0.3.2--0.4.0.sql
/extensions/pgvector/share/extension/vector--0.1.1--0.1.3.sql
/extensions/pgvector/share/extension/vector--0.1.8--0.2.0.sql
/extensions/pgvector/share/extension/vector--0.5.1--0.6.0.sql
/extensions/pgvector/share/extension/vector--0.5.0--0.5.1.sql
/extensions/pgvector/share/extension/vector--0.4.2--0.4.3.sql
/extensions/pgvector/share/extension/vector--0.6.0--0.6.1.sql
/extensions/pgvector/share/extension/vector--0.2.1--0.2.2.sql
/extensions/pgvector/share/extension/vector--0.2.5--0.2.6.sql
/extensions/pgvector/share/extension/vector--0.1.5--0.1.6.sql
/extensions/pgvector/share/extension/vector--0.1.7--0.1.8.sql
/extensions/pgvector/share/extension/vector--0.7.4--0.8.0.sql
/extensions/pgvector/share/extension/vector--0.4.3--0.4.4.sql
```

And this:

```sh
kubectl cnpg psql postgresql-with-extensions -- app -qA -c '\dx'
```

Returning:

```console
List of installed extensions
Name|Version|Default version|Schema|Description
address_standardizer|3.6.0dev|3.6.0dev|public|Used to parse an address into constituent elements. Generally used to support geocoding address normalization step.
address_standardizer_data_us|3.6.0dev|3.6.0dev|public|Address Standardizer US dataset example
fuzzystrmatch|1.2|1.2|public|determine similarities and distance between strings
plpgsql|1.0|1.0|pg_catalog|PL/pgSQL procedural language
postgis|3.6.0dev|3.6.0dev|public|PostGIS geometry and geography spatial types and functions
postgis_raster|3.6.0dev|3.6.0dev|public|PostGIS raster types and functions
postgis_sfcgal|3.6.0dev|3.6.0dev|public|PostGIS SFCGAL functions
postgis_tiger_geocoder|3.6.0dev|3.6.0dev|tiger|PostGIS tiger geocoder and reverse geocoder
postgis_topology|3.6.0dev|3.6.0dev|topology|PostGIS topology spatial types and functions
vector|0.8.0|0.8.0|public|vector data type and ivfflat and hnsw access methods
(10 rows)
```

gbartolini:
I also verified this:

```sh
rm -f /extensions/pgvector/lib/vector.so
rm: cannot remove '/extensions/pgvector/lib/vector.so': Read-only file system
```

api/v1/cluster_types.go
```diff
@@ -1416,6 +1416,37 @@ type PostgresConfiguration struct {
...
+	ExtensionControlPath []string `json:"extension_control_path,omitempty"`
+
+	// The list of directories inside the image which should be added to dynamic_library_path.
+	// If not defined, defaults to "/lib".
+	// +optional
+	DynamicLibraryPath []string `json:"dynamic_library_path,omitempty"`
+
+	// The list of directories inside the image which should be added to ld_library_path.
+	// +optional
+	LdLibraryPath []string `json:"ld_library_path,omitempty"`
```
theory, line 1462: Long term we'll need to find a way to do away with `LD_LIBRARY_PATH`, as more and more OSes drop support for it. Will likely have to do something with `rpath` in the Postgres binaries and extensions compiled for it. But I think the core will need to take a position on it first.

config/crd/bases/postgresql.cnpg.io_clusters.yaml
```diff
@@ -4011,6 +4011,67 @@ spec:
                       This should only be used for debugging and troubleshooting.
                       Defaults to false.
                     type: boolean
+                  extensions:
+                    description: The configuration of the extensions to be added
+                    items:
+                      description: |-
+                        ExtensionConfiguration is the configuration used to add
+                        a PostgreSQL extensions to the Cluster
```
theory:
```suggestion
                        PostgreSQL extensions to the Cluster.
```

config/crd/bases/postgresql.cnpg.io_clusters.yaml
```diff
@@ -4011,6 +4011,67 @@ spec:
...
+                                This field is optional to allow higher level config management to default or override
+                                container images in workload controllers like Deployments and StatefulSets.
+                              type: string
+                          type: object
+                          x-kubernetes-validations:
+                          - message: An image reference is required
+                            rule: has(self.reference)
+                        ld_library_path:
+                          description: The list of directories inside the image which
+                            should be added to ld_library_path.
```
theory, line 4061:
```suggestion
                          description: |-
                            The list of directories inside the image which
                            should be added to ld_library_path.
```

docs/src/cloudnative-pg.v1.md
```diff
@@ -2820,6 +2820,61 @@ storage</p>
...
+## ExtensionConfiguration     {#postgresql-cnpg-io-v1-ExtensionConfiguration}
+
+
+**Appears in:**
+
+- [PostgresConfiguration](#postgresql-cnpg-io-v1-PostgresConfiguration)
+
+
+<p>ExtensionConfiguration is the configuration used to add
+a PostgreSQL extensions to the Cluster</p>
```
theory:
```suggestion
PostgreSQL extensions to the Cluster</p>
```

docs/src/imagevolume_extensions.md
```diff
@@ -0,0 +1,332 @@
+# Image Volume Extensions
+<!-- SPDX-License-Identifier: CC-BY-4.0 -->
+
+CloudNativePG supports the **dynamic loading of PostgreSQL extensions** into a
+running `Cluster` using the [Kubernetes `ImageVolume` feature](https://kubernetes.io/docs/tasks/configure-pod-container/image-volumes/)
```
theory:
```suggestion
`Cluster` at startup using the [Kubernetes `ImageVolume` feature](https://kubernetes.io/docs/tasks/configure-pod-container/image-volumes/)
```

api/v1/cluster_types.go
```diff
@@ -1416,6 +1416,37 @@ type PostgresConfiguration struct {
...
 	EnableAlterSystem bool `json:"enableAlterSystem,omitempty"`
+
+	// The configuration of the extensions to be added
+	// +optional
+	Extensions []ExtensionConfiguration `json:"extensions,omitempty"`
+}
+
+// ExtensionConfiguration is the configuration used to add
+// a PostgreSQL extensions to the Cluster
+type ExtensionConfiguration struct {
```
theory, line 1440: Doesn't it also need a list of modules to add to `shared_preload_libraries`? Or perhaps an option to specify how a module needs to be pre-loaded, as in [this PGXN Meta v2 property](https://github.com/pgxn/rfcs/pull/3/files#diff-3a6d7e2784cf9bc361c4255353765fc71bc523285341b9f77b6787223165a19cR526-R530):

> 
> ```markdown
> *   **preload**: A [String](#string) that indicates that the extension's
>     libraries **MAY** be loaded in advance. Its two possible values are:
>     `server` if the module requires loading on server start, and `session`
>     if the module can be loaded in a session. Omit this field if the
>     module does not require preloading.
> ```

docs/src/imagevolume_extensions.md
```diff
@@ -0,0 +1,332 @@
...
+
+Extension images are defined in the `.spec.postgresql.extensions` stanza of a
+`Cluster` resource, which accepts an ordered list of extensions to be added to
+the PostgreSQL cluster.
+
+!!! Info
+    For field-level details, see the
+    [API reference for `ExtensionConfiguration`](cloudnative-pg.v1.md#postgresql-cnpg-io-v1-ExtensionConfiguration).
+
+Each image volume is mounted at `/extensions/<EXTENSION_NAME>`.
```
theory, line 61: PostgreSQL will not find extensions at that path, because it always wants some directory in the path to be named "postgresql" or "pgsql". We may want to consider, using `/extensions/postgresql/<EXTENSION_NAME>` or some such.

docs/src/imagevolume_extensions.md
```diff
@@ -0,0 +1,332 @@
...
+
+!!! Info
+    Depending on how your extension container images are built and their layout,
+    you may need to adjust the default `extension_control_path` and
+    `dynamic_library_path` values to match the image structure.
+
+!!! Important
+    The extension container image must match the PostgreSQL container used by
+    your cluster in PostgreSQL major version, Operating system distribution, and
+    CPU architecture to ensure compatibility and prevent runtime issues.
```
theory: This is not an issue for many non-native extensions. Probably worth it to build separate images for each Postgres version for such extensions, but not OS or architecture, I suspect.

docs/src/imagevolume_extensions.md
```diff
@@ -0,0 +1,332 @@
...
+
+## How to add a new extension
+
+Adding an extension to a database in CloudNativePG involves two steps:
+
+1. Attach the extension image to the `Cluster` resource so that PostgreSQL can
```
theory: Below you say "adding", not "attaching". "Attaching" sounds to me like it needs to be mounted in a volume or something.

docs/src/imagevolume_extensions.md
```diff
@@ -0,0 +1,332 @@
...
+    automatically trigger a [rolling update](rolling_update.md) to attach the new
+    image volume to each pod. Before adding a new extension in production,
+    ensure you have thoroughly tested it in a staging environment to prevent
+    configuration issues that could leave your PostgreSQL cluster in an unhealthy
+    state.
+
+Once mounted, CloudNativePG will automatically configure PostgreSQL by appending:
+
+- `/extensions/foo/share` to `extension_control_path`
+- `/extensions/foo/lib` to `dynamic_library_path`
```
theory, line 139: Postgres will rewrite those as

```
/extension/foo/share/postgreql/share
/extension/foo/share/postgreql/lib
```

unless the default path is updated to contain the string "pgsql" or "postgresql".

docs/src/imagevolume_extensions.md
```diff
@@ -0,0 +1,332 @@
...
+
+This ensures that the PostgreSQL container is ready to serve the `foo`
+extension when requested by a database, as described in the next section. The
+`CREATE EXTENSION foo` command, triggered automatically during the
+[reconciliation of the `Database` resource](declarative_database_management.md/#managing-extensions-in-a-database),
+will work without additional configuration, as PostgreSQL will locate:
+
+- the extension control file at `/extensions/foo/share/extension/foo.control`
+- the shared library at `/extensions/foo/lib/foo.so`
```
theory, line 148: Not all extensions have shared libraries, and those that do may not use the same name for the library file.

docs/src/imagevolume_extensions.md
```diff
@@ -0,0 +1,332 @@
...
+    name: foo-18
+  extensions:
+    - name: foo
+      version: 1.0
+```
+
+CloudNativePG will automatically reconcile this resource, executing the
+`CREATE EXTENSION foo` command inside the `app` database if it is not
+already installed, ensuring your desired state is maintained without manual
+intervention.
```
theory, line 178: How does it work to upgrade extensions?

docs/src/imagevolume_extensions.md
```diff
@@ -0,0 +1,332 @@
...
+  appended to the `LD_LIBRARY_PATH` environment variable of the instance
+  manager process, allowing PostgreSQL to locate required system libraries at
+  runtime.
+
+This flexibility enables you to support complex or non-standard extension
+images while maintaining clarity and predictability.
+
+### Setting Custom Paths
+
+If your extension image does not use the default `lib` and `share` directories
```
theory, line 209: Is it worth using `lib` as the default for `ld_library_path`, as well? Maybe use `modules` for the shared modules?

docs/src/imagevolume_extensions.md
```diff
@@ -0,0 +1,332 @@
...
+This allows PostgreSQL to discover your extension’s control files and shared
+libraries correctly, even with a non-standard layout.
+
+### Multi-extension Images
+
+You may need to include multiple extensions within the same container image,
+adopting a structure where each extension’s files reside in their own
+subdirectory.
+
+For example, to package PostGIS and pgRouting together in a single image:
```
theory: Why not put all the `share` files from both extension in one `share` directory? Same for `lib`.

docs/src/imagevolume_extensions.md
```diff
@@ -0,0 +1,332 @@
...
+
+CloudNativePG will set the `LD_LIBRARY_PATH` environment variable to include
+`/extensions/postgis/system`, allowing PostgreSQL to locate and load these
+system libraries at runtime.
+
+!!! Important
+    Since `ld_library_path` must be set when the PostgreSQL process starts,
+    changing this value requires a **cluster restart** for the new value to take effect.
+    CloudNativePG does not currently trigger this restart automatically; you will need to
+    manually restart the cluster (e.g., using `cnpg restart`) after modifying `ld_library_path`.
```
theory, line 301: I'm confused. ISTM that extensions can _only_ be added via a restart, because

1. `ImageVolume` requires an init; volumes cannot be dynamically mounted in running Pods
2. `extension_control_path` and `dynamic_library_path` both require Postgres to restart in order for the directories to be present in all backends

docs/src/imagevolume_extensions.md
```diff
@@ -0,0 +1,332 @@
...
+
+## Image Specifications
+
+A standard extension container image for CloudNativePG includes two
+required directories at its root:
+
+- `share`: contains the extension control file (e.g., `<EXTENSION>.control`).
```
theory:
```suggestion
- `share`: contains the extension control file (e.g., `<EXTENSION>.control`) and any SQL files.
```

docs/src/imagevolume_extensions.md
```diff
@@ -0,0 +1,332 @@
...
+- `lib`: contains the extension's shared library (e.g., `<EXTENSION>.so`) and
+  any additional required libraries.
+
+Following this structure ensures that the extension will be automatically
+discoverable and usable by PostgreSQL within CloudNativePG without requiring
+manual configuration.
+
+!!! Important
+    We encourage PostgreSQL extension developers to publish OCI-compliant extension
+    images following this layout as part of their artifact distribution, making
```
theory, line 319: See also the [PGXN packaging layout RFC](https://github.com/pgxn/rfcs/pull/2/files#diff-7c377996b7e36b02007931f9cd3d72c0e5f81f77047a4778c4d5cee148810abcR184-R188); I plan to remove the `pgsql` parent directory, but the equivalent to `lib` in this PR is `pkglib` in the RFC.

theory: Long term I think CNPG will need to provide a lot of extension images that people can just use.

docs/src/imagevolume_extensions.md
theory: What happens if the version isn't available in the mounted image?

hack/setup-cluster.sh
```diff
@@ -174,6 +174,15 @@ EOF
     done
   fi
 
+  # Enable ImageVolume support from kindest/node v1.33.1
+  if [[ "$(printf '%s\n' "1.33.1" "${k8s_version#v}" | sort -V | head -n1)" == "1.33.1" ]]; then
```
theory, line 178: Will this work when 1.33.2 or 1.34.0 is released?

leonardoce: The ImageVolume feature was introduced in 1.31 as alpha. It was promoted to beta in 1.33, so we'll need to wait a bit to have it enabled by default.
That depends on the Kubernetes distribution as well.

docs/src/imagevolume_extensions.md
leonardoce: Yes, that's true when adding extensions.

That documentation paragraph is about changing `ld_library_path` on an extension that was already declared.
Unfortunately, this doesn't trigger an automatic Pod rollout for now.
Changing `extension_control_path` and `dynamic_library_path` triggers a Pod restart because PostgreSQL requires that.

docs/src/imagevolume_extensions.md
leonardoce: Yes, this could certainly be useful, but it would also be an important effort on our part as maintainers.
We'll need to seek help from the community.

config/crd/bases/postgresql.cnpg.io_clusters.yaml
leonardoce: This is a generated file.

tests/e2e/fixtures/imagevolume_extensions/cluster-with-extensions.yaml.template
```diff
@@ -0,0 +1,36 @@
...
+      log_statement: 'ddl'
+      log_temp_files: '1024'
+      log_autovacuum_min_duration: '1s'
+      log_replication_commands: 'on'
+    extensions:
+      - name: postgis
+        ld_library_path:
+          - system
+        image:
+          reference: ghcr.io/niccolofei/postgis:18beta1-master-bullseye # wokeignore:rule=master
```
leonardoce, line 23: We'll need to remove this hardcoded reference from here.  These tests are run with different PG major versions.

leonardoce:
E2e tests are fine. I'm approving it even if they don't test the feature because we still don't test PG 18.

leonardoce (approved):

api/v1/cluster_types.go
gbartolini: I comment only on the `ld_library_path` option, leaving the broader case open for future discussions. We would appreciate your thoughts on its naming. Initially, we considered calling it `system_library_path` to avoid creating a direct link to the current implementation of `LD_LIBRARY_PATH`. We then decided on `ld_library_path` so that the target is clearer. Thoughts?

api/v1/cluster_types.go
gbartolini: We were thinking of leaving this to the actual `Cluster` configuration through the `.spec.postgresql.shared_preload_libraries` special option [(docs)](https://cloudnative-pg.io/documentation/current/postgresql_conf/#shared-preload-libraries).

Similarly, only when you instantiate an extension in the `Database` resource, we trigger the `CREATE/ALTER/DROP EXTENSION`-based management process.

What do you think?

docs/src/imagevolume_extensions.md
gbartolini: I believed the purpose of `extension_control_path` is to allow the use of any path. The only requirement is the `extension` folder. Indeed, the patch is functioning correctly unless I misunderstood your point.

docs/src/imagevolume_extensions.md
gbartolini: How would you improve that box? Any ideas? Thanks!

docs/src/imagevolume_extensions.md
gbartolini: Changed, thanks!

> 1. Define the extension image in the `Cluster` resource so that PostgreSQL ...

docs/src/imagevolume_extensions.md
gbartolini: I now understand what you mean. PostgreSQL now requires the `extension` folder for the control file, and will expect it to be under `/extensions/pgvector/share/extension`.

docs/src/imagevolume_extensions.md
gbartolini: Yep. Therefore, we provide the option to make `dynamic_library_path` empty and customizable; that's merely a default setting, as we believe that most extensions will be like this. If you believe that should not be the case for the majority of extensions, we're happy to review this.

docs/src/imagevolume_extensions.md
gbartolini: We have been thinking about this and should explore it further when we also support image catalogs (the second part of this patch, probably for 1.28). We would like to complete the entire feature incrementally, allowing us to experiment for a few months with a few projects (hopefully, we will get traction among extension developers). The general idea was that the image catalog also lists extensions, providing the versions contained in the image.

gbartolini: Right now, you upgrade the image and make the new version of the extension available in the system. Then, you update the version in the `Database` resource to request an extension upgrade once the rollout has completed.

docs/src/imagevolume_extensions.md
gbartolini: That shouldn't be a problem. The important thing is to provide ways to configure the paths. That use case is just an example to prove that the multi-path support is possible and available.

docs/src/imagevolume_extensions.md
gbartolini: I am happy to revisit this choice. Although `make install` uses `lib`for the module's shared libraries, doesn't it?

docs/src/imagevolume_extensions.md
gbartolini: Yep, thanks Leonardo.

docs/src/imagevolume_extensions.md
gbartolini: > See also the [PGXN packaging layout RFC](https://github.com/pgxn/rfcs/pull/2/files#diff-7c377996b7e36b02007931f9cd3d72c0e5f81f77047a4778c4d5cee148810abcR184-R188); I plan to remove the `pgsql` parent directory, but the equivalent to `lib` in this PR is `pkglib` in the RFC.

So, if I understand correctly, we could rename `lib` into `pkglib` and `system` into `lib`, based on your RFC @theory?

gbartolini: > Long term I think CNPG will need to provide a lot of extension images that people can just use.

This could be a viable plan, but as Leonardo mentioned, it presents an opportunity for a new set of contributors to assist us in this effort. We are confident that, since we are already supporting PostGIS container images and providing support in the `standard` image for `pgaudit`, `pgvector`, and `pg_failoverslots`, we can start a project focused on just these images. We could call it `extension-containers`.

However, we will need help from more than just those of us working at EDB to ensure its long-term maintenance. Niccolò can certainly play a crucial role, but it would be excellent if other volunteers also stepped up to contribute.

hack/setup-cluster.sh
gbartolini: My understanding is that this will be enabled by default in 1.34 as beta (end of August), and there's a plan to move it to GA already in 1.35. The good news is also that OpenShift 4.20 will have this feature by default.

api/v1/cluster_types.go
theory: I like the clarity, and it ties it closely to the future of support for the environment variable itself. +1.

hack/setup-cluster.sh
theory: My question is: does the above code allow it to work with 1.34 or 1.35? Or is it locked to _only_ 1.33.1?

docs/src/imagevolume_extensions.md
theory: Why not have `ld_library_path` trigger a restart, as well?

docs/src/imagevolume_extensions.md
theory: > So, if I understand correctly, we could rename `lib` into `pkglib` and `system` into `lib`, based on your RFC @theory?

Yes, I think that's right. The point was to match the terms used by `pg_config`. But it's an RFC, subject to change.

config/crd/bases/postgresql.cnpg.io_clusters.yaml
theory: I didn't realize it was valid YAML to have a line break like that.

api/v1/cluster_types.go
theory: I think that the specification of what shared libraries need to be loaded for an extension ought to be closely associated with the configuration of that extension. But I can also see waiting to see how things evolve, to see how it feels in practice, since there's an existing feature that enables loading DSOs.

It should probably be mentioned here, though.

docs/src/imagevolume_extensions.md
theory: Ah, perhaps it does, but the PGXS support will append `postgresql` as a subdirectory to the path it it doesn't exist. So if you do:

```
make install prefix=/usr/local/extras
```

PGXS will put the files into

```
/usr/local/extras/postgresql/share
/usr/local/extras/postgresql/lib
```

That's what I was thinking of. I think you're right that the GUCs will work with any path. The question is: do we want the behavior of the extension mount to vary from that pattern. Probably not a big deal if it doesn't, since the extensions aren't installed by PGXS, but just put into directories in the image.

It does not allow any path. The way that postgres/postgres@4f7f7b0 was implemented was to operate in the same way as `dyanamic_library_path` works

docs/src/imagevolume_extensions.md
theory: I think we need to develop guidance for the creation of extension images. You'll note that PGDG Yum (and I presume Apt) uses "common" in place of the OS/arch combination when an extension package has no native dependencies. For the purpose of this box, perhaps just qualifying the native aspect is sufficient:

```
!!! Important
    If the extension image includes shared libraries, they must be compiled with
    the same major PostgreSQL version, OS distribution and CPU architecture to
    ensure compatibility and prevent runtime issues.
```

docs/src/imagevolume_extensions.md
theory: That too!

docs/src/imagevolume_extensions.md
theory: Oh, I was just reacting to the contrast between the name of the control file, which is specific to the extension name, and the library file, which can vary. But it usually doesn't, and since this is an example, I think it's fine.

docs/src/imagevolume_extensions.md
theory: Might be worth mentioning, then, something like

```
For example, to package PostGIS and pgRouting together in a single image, each in its own subdirectory:
```

docs/src/imagevolume_extensions.md
theory: It uses `pg_config --pkglib`.

docs/src/imagevolume_extensions.md
gbartolini: I think we are fine as long as we find a consistent and sensible way to manage extensions' content based on their relative path in the container image. Users should not worry where in the container they end up (and we can always change it).

armru (approved):

api/v1/cluster_types.go
gbartolini: Let me understand. Do you prefer `ld_library_path` or `system_library_path`?

gbartolini: Oh ... I get it. `ld_library_path`.

api/v1/cluster_types.go
gbartolini: Makes sense.

gbartolini: Resolved here: https://github.com/cloudnative-pg/cloudnative-pg/pull/7991/commits/8a5b3a98b06dbd96286d7b092b3d0cf66aa4867f

docs/src/imagevolume_extensions.md
gbartolini: See https://github.com/cloudnative-pg/cloudnative-pg/pull/7991/commits/fe426d41acd3171d3c32cb228393bc1c5cf39ae3

docs/src/imagevolume_extensions.md
gbartolini: This could be partially related to #2530

hack/setup-cluster.sh
mnencia: The line indicates that the flag should be enabled if the version is `1.33.1` or higher.

To clarify: it creates a list containing two values, `1.33.1` and the actual version. This list is then sorted according to semantic versioning principles, and it checks if the first value (the lower one) is `1.33.1`.

tests/e2e/imagevolume_extensions_test.go
```diff
@@ -0,0 +1,234 @@
...
+
+			Eventually(func(g Gomega) {
+				// Updating the Cluster
+				cluster, err := clusterutils.Get(env.Ctx, env.Client, namespace, clusterName)
+				g.Expect(err).NotTo(HaveOccurred())
+				cluster.Spec.PostgresConfiguration.Extensions = append(cluster.Spec.PostgresConfiguration.Extensions,
+					apiv1.ExtensionConfiguration{
+						Name: "pgvector",
+						ImageVolumeSource: corev1.ImageVolumeSource{
+							Reference: "ghcr.io/niccolofei/pgvector:18beta1-master-bullseye", // wokeignore:rule=master
```
mnencia, line 201: Another place where we should parametrize the extension image name.

docs/src/imagevolume_extensions.md
gbartolini: We have discussed internally and decided to rename `system` in `syslib`, as the entire content refers implicitly to the extension and the system libraries deviate from that.

theory: Good call.

mnencia (approved):

gbartolini (approved):

PR #8003, opened 2025-07-09
feat: decouple metrics scraping from metrics computation

### LINKED ISSUE #4962: [Feature]: decouple metrics scraping from metrics recomputation

Several metrics, including user defined metrics, are run each time the prometheus metrics endpoint 9187, is scraped. Depending on the configuration of Prometheus, and even perhaps just having some other process doing a GET on `:9187/metrics`, those SQL queries might run often.
If a user included some long running queries for monitoring, this could be a CPU hog and even pose a threat to the normal functioning of their cluster.

Generally, to respond to a GET on the metrics port, we should simply return the current values of the various prometheus counters, histograms etc.
A few select metrics might warrant a refresh on each scrape.

The recomputation of metrics should happen independently of scrapes.
The refresh interval should be configurable, and if not set default to something reasonable, e.g. 30 seconds, which seems the current default scraping interval for the Prometheus operator

``` go
type MonitoringConfiguration struct {
	<... snipped ...>
	RefreshInterval *metav1.Duration `json:"refreshInterval,omitempty"`
```

For the purpose of decoupling scraping from updating of metrics, it is sufficient to have a goroutine rerunning the metrics queries and creating Prometheus metrics, with the above refresh interval.

A future PR/Issue could tackle having query-dependent refresh intervals, so that very expensive queries could be monitored, while keeping a normal frequency with the smaller queries.

The prometheus HTTP library has a parameter `MaxRequestsInFlight` which, if set will respond with an HTTP 503 if exceeded.
That could be a very low effort low resolution safety mechanism.

### Additional context

_No response_

### Backport?

Yes

Yes

### Code of Conduct

jsilvela:
Closes #4962


pkg/management/postgres/metrics/collector.go
```diff
@@ -88,7 +103,7 @@ func (q *QueriesCollector) collectUserQueries(ch chan<- prometheus.Metric) error
 
 	for name, userQuery := range q.userQueries {
 		queryLogger := log.WithValues("query", name)
-		collector := QueryCollector{
+		queryRunner := QueryRunner{
```
jsilvela, line 148: OK, way too many things in the monitoring code are called a collector.
Computing the next value of a metric is not "collecting".
If I get the OK from maintainers, would like to do a lot of renaming. Perhaps a separate PR would be better though.

jsilvela:
Gah, metrics E2E failing. Manually taking out the OK-to-merge label (from smoke tests)

pkg/management/postgres/metrics/collector.go
jmealo: 100% agree on the changes; the metrics code was hard to follow. I'd support it on this PR or a separate one... My gut is that a separate PR might be easier to review [rename only, no logic changes].

pkg/management/postgres/metrics/collector.go
```diff
@@ -50,33 +50,48 @@ type QueriesCollector struct {
 
 	errorUserQueries      *prometheus.CounterVec
 	errorUserQueriesGauge prometheus.Gauge
+
+	computedMetrics []prometheus.Metric
```
jmealo, line 64: Does this field need synchronization? 🤔

pkg/management/postgres/metrics/collector.go
```diff
@@ -50,33 +50,48 @@ type QueriesCollector struct {
...
 	q.errorUserQueriesGauge.Collect(ch)
 	q.errorUserQueries.Collect(ch)
-
-	return nil
 }
 
-func (q *QueriesCollector) collectUserQueries(ch chan<- prometheus.Metric) error {
+func (q *QueriesCollector) createMetricsFromUserQueries() error {
+	// refresh
+	q.computedMetrics = []prometheus.Metric{}
```
jmealo: Do we need to lock before we write? 🤔

pkg/management/postgres/metrics/collector.go
```diff
@@ -123,24 +138,31 @@ func (q *QueriesCollector) collectUserQueries(ch chan<- prometheus.Metric) error
...
 
-			err = collector.collect(conn, ch)
+			computedMetrics, err := queryRunner.computeMetrics(db)
 			if err != nil {
 				queryLogger.Error(err, "Error collecting user query",
 					"targetDatabase", targetDatabase)
 				// Increment metrics counters.
 				q.reportUserQueryErrorMetric(name + " on db " + targetDatabase + ": " + err.Error())
 			}
+			q.computedMetrics = append(q.computedMetrics, computedMetrics...)
```
jmealo: Same question... do we need to lock before we append?

pkg/management/postgres/metrics/collector.go
```diff
@@ -123,24 +138,31 @@ func (q *QueriesCollector) collectUserQueries(ch chan<- prometheus.Metric) error
...
 		}
 	}
 	return nil
 }
 
+func (q *QueriesCollector) collectComputedMetrics(ch chan<- prometheus.Metric) {
+	for _, m := range q.computedMetrics {
```
jmealo: Do we need to lock before we read?

jmealo (commented):
I meant to leave this as a PR review not a comment. Ooops.

pkg/management/postgres/metrics/collector.go
```diff
@@ -50,33 +50,48 @@ type QueriesCollector struct {
...
 
-	err := q.collectUserQueries(ch)
+	err := q.createMetricsFromUserQueries()
 	if err != nil {
 		return err
 	}
+	return nil
+}
+
+// Collect sends the updated values to the output channel
```
jmealo:
```suggestion
// Collect sends the pre-computed metrics to the output channel.
// These metrics were cached during the last Update() call and are not
// fetched from the database during collection.
```

jmealo (commented):
**Looks good overall!**
When I was troubleshooting some issues with CNPG, I received help from some Gophers in the Go slack who hinted that the project may have some race conditions in patterns used throughout. This was months ago and I don't remember the specifics.  All of th Go I've done is just trying to fix bugs in CNPG; so I could be way off-base here.

## Possible Race Condition
I think there may be a race condition between `Update()` and `Collect()`. The `Update()` method writes to `computedMetrics` while `Collect()` reads
   from it. It looks like `Update()` is called from a ticker goroutine (every 30s) and `Collect()` is called by Prometheus HTTP handler (meaning they can execute concurrently without synchronization)? 🤔 

**I _think_ something like this could work?**   
```go
  type QueriesCollector struct {
      // ... existing fields ...
      computedMetrics []prometheus.Metric
      mu              sync.RWMutex  // Add this
  }
```
Then use `q.mu.Lock()` in `Update()` and `q.mu.RLock()` in `Collect()`.

**Alternatively:**
Instead of a mutex, we could possibly use an atomic pointer to swap the entire metrics slice atomically? (I would lean towards this)
```go
  type QueriesCollector struct {
      // ... existing fields ...
      computedMetrics atomic.Pointer[[]prometheus.Metric]
  }
  ```
  This would eliminate lock contention between `Update` and `Collect` operations.
 
## Error handling
Since the user queries are user supplied, would it make sense to add more context to errors in `createMetricsFromUserQueries()`? For example: when `instance.IsPrimary()` fails, it might be helpful to wrap the error with more context about what operation was being attempted.

File: pkg/management/postgres/metrics/collector.go (hunk 2 above)
Comment (jsilvela) [reply pkg/management/postgres/metrics/collector.go:64]: not at the moment, it's only updated in one function call. Maybe defensively though

File: pkg/management/postgres/metrics/collector.go (hunk 1 above)
Comment (jsilvela) [reply pkg/management/postgres/metrics/collector.go:148]: > My gut is that a separate PR might be easier to review [rename only, no logic changes].

Yep, my thoughts too

File: pkg/management/postgres/metrics/collector.go (hunk 3 above)
Comment (jsilvela) [reply pkg/management/postgres/metrics/collector.go]: well, only this function updates that, not sure if it's necessary, but might be a good thing

**Comment by jsilvela** [comment]
@jmealo you're right about the race condition. While it wouldn't lead to corruption/destruction, we could call Collect at a point where Update was in progress, and we could lose metrics during collection.
The mutex is right, I think

**Comment by jsilvela** [comment]
@jmealo thinking about the lock more, I think we might not need it. Instead of clearing the `computedMetrics` field at the beginning of Update, and progressively appending to it, we can build a slice with everything, and then do `q.computedMetrics = generatedMetrics` in one fell swoop.

**Comment by jmealo** [comment]
> @jmealo thinking about the lock more, I think we might not need it. Instead of clearing the `computedMetrics` field at the beginning of Update, and progressively appending to it, we can build a slice with everything, and then do `q.computedMetrics = generatedMetrics` in one fell swoop.

This looks right to me. :+1:

File: pkg/management/postgres/metrics/collector.go [hunk 7]
```diff
@@ -69,13 +69,15 @@ var isPathPattern = regexp.MustCompile(`[][*?]`)
 
 // Update recomputes the metrics from the user queries
 func (q *QueriesCollector) Update() error {
+	isPrimary, err := q.instance.IsPrimary()
+	if err != nil {
+		return fmt.Errorf("while updating query metrics, could not check for primary: %w", err)
```
jmealo, line 97: **Nit:** Should we explicitly state that the query is user provided/custom? (if we're not including a portion of the query to help identify it?) ... I'm thinking what if a user has multiple custom queries that are working, then they add one with a syntax error.... will it be obvious to know what's going wrong? [other than the absence of that query having results in the metrics?]

jsilvela: I'm not sure I get your meaning here. At this level of the code, we still need to go into the `createMetricsFromUserQueries` function, which will iterate over the queries, for all relevant databases.
The distinction between default/custom is not easy to see here, as the instance manager may be reloading a set of new metric queries from a secret or configmap.

jsilvela: Now, reading between the lines, I think you're asking, what happens if there is an error when we're running the metric query `foo`.
And farther down in the code, in the `createMetricsFromUserQueries`, we have two ways we're recording this: into a metric collecting errors, and via a log
See lines 145-150
``` go
if err != nil {
	queryLogger.Error(err, "Error collecting user query",
		"targetDatabase", targetDatabase)
	// Increment metrics counters.
	q.reportUserQueryErrorMetric(name + " on db " + targetDatabase + ": " + err.Error())
}
```

jsilvela: At the very bottom of this chain, we create a const metric or a const histogram.
And the error returned there generates a log, but the return is nil.
If that was your concern, then I do agree with it. Would probably be better to bubble the errors up.
That said, the existing behavior was this one (on error it logged but did not perform the "collection"), and I wanted to keep this PR from growing too much...

jsilvela: Yikes, sorry @jmealo , long answer but was not sure specifically what you were pointing out.
If you think worth doing the bubbling up of errors, I'd like the maintainers to take a look first, before the PR gets bigger. WDYT?

jmealo: Sorry, should have read further. That seems reasonable to me. I don't think any changes are necessary that should block the PR. 

A follow up issue/PR to improve error logging might make sense.

armru (approved):

leonardoce:
@jsilvela, I had a look at the implementation and while it definitely solves the frequent querying problem, it has a couple of issues that prevent me from merging it.

1. the ticker executes the queries every 30 seconds (or the configured amount of time) even when the metrics endpoint is not used - and this wastes resources when no one is scraping the metrics endpoint
2. it forces caching for in-memory metrics - i.e. the fencing status

I think a TTL-based approach would have all the benefits while avoiding all the overhead when the metrics endpoint is not used.
We could allow our users to:

* define a TTL for each user-defined query 
* and a global TTL to be used when the specific one is not configured

The system should only execute a query when the metrics endpoint is accessed and the last result has expired.
Defaulting the TTL to zero will preserve the existing behavior.
Would you be willing to invest your time in implementing this approach?
Unfortunately this would need a significant change in this patch.

I apologize that this design feedback didn't come earlier. I know you invested a lot of time into this. 
I'm involving the other maintainers to get more feedback.

jsilvela:
That's fine, thanks for the feedback @leonardoce, much appreciated.

So, I did think about something along the lines you comment with a TTL.
But I don't quite get your point 2.
Let me make sure I get it.
If we define that for a particular query, the TTL is 1 minute, the computed value needs to be held for 1 minute.
How do you hold the value without caching in some shape or form?
The solution in production right now is that the ad-hoc metric is computed on the fly, then scraped and discarded.
How would we ensure that two scrapes falling before TTL expiry produce the same value.
I can't see a way around storing the ad-hoc metric for the duration.
And, yes, as you point out, if the TTL is 0 then we can discard immediately and avoid caching.

Aside from the discussion, yes of course I'm willing to rework and reimplement.

jsilvela:
On your other objection, by the way, that this PR currently would rerun the queries even in the absence of scrapes:

- That's the Prometheus way, no? I mean, prometheus counters, gauges, counterVecs, are updated independently of scrapes.
- I find that the production behavior is troubling itself. I.e. if we run the queries on every scrape, we could have an inadvertent DOS. In my opinion, recomputing the queries every 30 seconds is better and more predictable. Just me of course.

gbartolini:
> That's fine, thanks for the feedback @leonardoce, much appreciated.

Thanks @leonardoce for the review and thanks @jsilvela for being willing to contribute to this patch.

> So, I did think about something along the lines you comment with a TTL. But I don't quite get your point 2. Let me make sure I get it. If we define that for a particular query, the TTL is 1 minute, the computed value needs to be held for 1 minute. How do you hold the value without caching in some shape or form? The solution in production right now is that the ad-hoc metric is computed on the fly, then scraped and discarded. How would we ensure that two scrapes falling before TTL expiry produce the same value. I can't see a way around storing the ad-hoc metric for the duration. And, yes, as you point out, if the TTL is 0 then we can discard immediately and avoid caching.

The point here is to define the scope and domain of responsibility of CloudNativePG. CloudNativePG should not perform any action unless explicitly requested and should only represent the same value if less than a specified amount of time has passed since the last request (acting as a cache).

I agree with Leonardo that we should not prematurely and preventively request metrics if there's no consumer: we don't provide a scraper, we only give an exporter to be queried. In this case, the cache could alleviate the load in some use cases.

We also need to start thinking that the exporter should run as a separate plugin, so that we can remove it from the operator.

jsilvela:
> The point here is to define the scope and domain of responsibility of CloudNativePG. CloudNativePG should not perform any action unless explicitly requested and should only represent the same value if less than a specified amount of time has passed since the last request (acting as a cache).

I still see this a different way. With the current behavior, the metrics queries are run on each scrape. If a trigger-happy metrics aficionado decides to refresh metrics every 5s, then that becomes what the cluster will do.
And if a second team decides to retrieve metrics from our cluster with a new PodMonitor, then the cluster now needs to serve them too, and recompute the queries on a second set of scrape requests. A cluster might start screeching to a halt because of this.

For the counters, gauges etc that the operator computes without need for queries, those are always updated by the processes in the operator, and do not depend on the scraping.
That is the Prometheus approach.

> CloudNativePG should not perform any action unless explicitly requested

This is a good point, but here, I think, this points to having a boolean to enable or disable the publishing of metrics via the Prometheus endpoint. Let the user of CNPG explicitly state they want metrics computed.

> We also need to start thinking that the exporter should run as a separate plugin

Interesting idea.

jsilvela:
>  requested and should only represent the same value if less than a specified amount of time has passed since the last request (acting as a cache).

Going back to the same thing bit. Leonardo's feedback point 2 is concerned with caching.
And the proposal is a TTL that defaults to 0, i.e. no caching.
That behavior troubles me for the reasons I've stated above.
Having a default TTL of, say, 30s, would be much safer IMO, but the this does imply caching, which is one of the objections.
I'd be fine with such a non-zero TTL, and I think it would not be difficult to refactor this code to achieve that.

jsilvela:
@leonardoce I've re-read your feedback, and issue 2

> it forces caching for in-memory metrics - i.e. the fencing status

Wait, no, the PR does not force caching for the counters, gauges etc computed without running SQL queries.
Is that what you meant?
But also, these in-memory metrics *are* cached, that's how Prometheus works.
Not sure I'm understanding your point properly...

leonardoce:
> So, I did think about something along the lines you comment with a TTL.
> But I don't quite get your point 2.
> Let me make sure I get it.
> If we define that for a particular query, the TTL is 1 minute, the computed value needs to be held for 1 minute.
> How do you hold the value without caching in some shape or form?
> The solution in production right now is that the ad-hoc metric is computed on the fly, then scraped and discarded.
> How would we ensure that two scrapes falling before TTL expiry produce the same value.
> I can't see a way around storing the ad-hoc metric for the duration.
> And, yes, as you point out, if the TTL is 0 then we can discard immediately and avoid caching.
> 
> Aside from the discussion, yes of course I'm willing to rework and reimplement.

Thank you!

My explanation was not clear. Yes, the latest value need to be cached.
For "in-memory metrics" I means those metrics that don't require a SQL query to be collected, such as the fencing status.
For those metrics caching is not needed and we can just return the updated value.

So, yes, we need to cache the latest value. And the system should only execute a query when the metrics endpoint is accessed and the last result has expired.

> That's the Prometheus way, no? I mean, prometheus counters, gauges, counterVecs, are updated independently of scrapes.

Yes, I agree with you, but unfortunately I argue that this poses unneeded stress on databases.
This way of working is perfectly suitable when updating the value is not expensive

> I find that the production behavior is troubling itself. I.e. if we run the queries on every scrape, we could have an inadvertent DOS.

Yes.

> In my opinion, recomputing the queries every 30 seconds is better and more predictable. Just me of course.

Yes, it is more predictable. But I argue that recomputing the queries when it is needed is even better.

jsilvela:
OK, I think I start to understand better, and I think we're not far.

>For "in-memory metrics" I means those metrics that don't require a SQL query to be collected, such as the fencing status.
For those metrics caching is not needed and we can just return the updated value.

My PR did not affect those in-memory metrics, so I was confused by the mention, and also I interpreted your comment as being against caching in general.

So, if we change the approach to a TTL of 30 seconds (or some other non-zero default), with caching of the computed ad-hoc metrics until TTL expiry, then this would be OK for you?

I'd be fine with that (with a warning not to make the TTL 0, explaining the DOS risk).
And if we did that, the PR is mostly there, since 90% of what this PR does is to separate the code into a function to compute the ad-hoc prometheus metrics and cache them, and another function to scrape the ad-hoc metrics previously cached.

So, are we in agreement there?

Now, back to your initial feedback:

> We could allow our users to:
>
>define a TTL for each user-defined query
and a global TTL to be used when the specific one is not configured

Actually in the Issue I initially proposed per-query refresh interval, but after looking at the work, I thought that that was more of a refinement or a second step.
Because i get complaints when my PRs are big, and this one was going to be even bigger with per-query refresh.
And because IMO the most urgent change is avoiding that potential DOS, and a blanket refresh interval is enough.

What do you think about a single TTL, just like the PR has a single RefreshInterval.
And leave for a future PR the per-query setting?
Or do you think we should do the per-query TTL now?

jmealo:
Single TTL seems fine for a first pass. Per-query would be nice but could be a follow-up task?

leonardoce:
> So, if we change the approach to a TTL of 30 seconds (or some other non-zero default), with caching of the computed ad-hoc metrics until TTL expiry, then this would be OK for you?
> 
> I'd be fine with that (with a warning not to make the TTL 0, explaining the DOS risk). And if we did that, the PR is mostly there, since 90% of what this PR does is to separate the code into a function to compute the ad-hoc prometheus metrics and cache them, and another function to scrape the ad-hoc metrics previously cached.
> 
> So, are we in agreement there?

Yes! Let's change the approach to a TTL.
I agree on the warning for the TTL=0, too.

> Now, back to your initial feedback:
> 
> > We could allow our users to:
> > define a TTL for each user-defined query
> > and a global TTL to be used when the specific one is not configured
> 
> Actually in the Issue I initially proposed per-query refresh interval, but after looking at the work, I thought that that was more of a refinement or a second step. Because i get complaints when my PRs are big, and this one was going to be even bigger with per-query refresh. And because IMO the most urgent change is avoiding that potential DOS, and a blanket refresh interval is enough.
> 
> What do you think about a single TTL, just like the PR has a single RefreshInterval. And leave for a future PR the per-query setting? Or do you think we should do the per-query TTL now?

Yes, a single TTL is a good first step.
Let's start with that!

jsilvela:
@leonardoce my apologies. Your initial feedback said

> 2. it forces caching for in-memory metrics - i.e. the fencing status

which I thought was inaccurate, but now that I'm modifying the code, you were right of course.

Will set the fencing status and others to be refreshed per scrape as they currently are.

pkg/management/postgres/webserver/metricserver/pg_collector.go
```diff
@@ -316,10 +317,46 @@ func (e *Exporter) Describe(ch chan<- *prometheus.Desc) {
 }
 
 // Collect implements prometheus.Collector, collecting the Metrics values to
-// export.
+// export. Note that those metrics are updated in the updatePgMetrics function
```
jsilvela:
```suggestion
// export.
```

jsilvela:
The E2E tests that failed in my triggered run yesterday, are failing in `main` and have nothing to do with metrics.

jsilvela:
OK, approach changed from thread running queries every 30 seconds, to a TTL during which metrics computed from queries are considered current and returned from cache.
Ready for review again, @leonardoce . And, if you don't mind @armru , please have a look at the changes.
Thank you!

armru (approved):

leonardoce (approved):

Issue #7632, opened 2025-05-23
[Bug]: Pinned dependencies missed


sxd:
### Is there an existing issue already for this bug?

- [x] I have searched for an existing issue, and could not find anything. I believe this is a new bug.

### I have read the troubleshooting guide

- [x] I have read the troubleshooting guide and I think this is a new bug.

### I am running a supported version of CloudNativePG

- [x] I have read the troubleshooting guide and I think this is a new bug.

### Version

1.26.0-rc3

### What version of Kubernetes are you using?

1.32

### What is your Kubernetes environment?

Self-managed: kind (evaluation)

### How did you install the operator?

YAML manifest

### What happened?

We have a couple of pinned dependencies missed that we should fix, probably using renovate:

```
Warn: third-party GitHubAction not pinned by hash: .github/workflows/continuous-delivery.yml:633: update your workflow using https://app.stepsecurity.io/secureworkflow/cloudnative-pg/cloudnative-pg/continuous-delivery.yml/main?enable=pin
Warn: third-party GitHubAction not pinned by hash: .github/workflows/continuous-integration.yml:605: update your workflow using https://app.stepsecurity.io/secureworkflow/cloudnative-pg/cloudnative-pg/continuous-integration.yml/main?enable=pin
Warn: third-party GitHubAction not pinned by hash: .github/workflows/require-labels.yml:21: update your workflow using https://app.stepsecurity.io/secureworkflow/cloudnative-pg/cloudnative-pg/require-labels.yml/main?enable=pin
Warn: containerImage not pinned by hash: Dockerfile:5
Warn: containerImage not pinned by hash: Dockerfile:10
Warn: goCommand not pinned by hash: hack/e2e/run-e2e-local.sh:70
Warn: goCommand not pinned by hash: hack/e2e/run-e2e.sh:74
Warn: pipCommand not pinned by hash: .github/workflows/continuous-delivery.yml:1533
Warn: downloadThenRun not pinned by hash: .github/workflows/continuous-integration.yml:855
Warn: pipCommand not pinned by hash: .github/workflows/latest-postgres-version-check.yml:31
Info: 92 out of 92 GitHub-owned GitHubAction dependencies pinned
Info: 110 out of 113 third-party GitHubAction dependencies pinned
Info: 0 out of 2 containerImage dependencies pinned
Info: 0 out of 2 goCommand dependencies pinned
Info: 0 out of 2 pipCommand dependencies pinned
[...]

PR #8023, opened 2025-07-13
Pin GitHub Action & OCI dependencies

datosh:
#7632 highlights that some GitHub Actions and OCI image references are missing the hash to properly pin the version

I have pinned the `free-disk-space` action to the latest commit on main, and moved the reference from `main` to `v1.3.1` as they reference the same commit.

The synk action repository is a collection of actions, and there are no releases for each action (or any recent releases for that matter). Therefore I pinned to the latest commit and left the reference as `master`. Renovate will still track and recommend updates when there are new commits on `master` branch.

The third reference in #7632 to be pinned (`.github/workflows/require-labels.yml:21`) was already fixed by https://github.com/cloudnative-pg/cloudnative-pg/commit/61a0537c45a394c94fa3d9e31312f95ab90afe2a

For the two OCI image references I pulled the latest version of the tag and added the SHA accordingly.

### DIFF
--- a/.github/workflows/continuous-delivery.yml
+++ b/.github/workflows/continuous-delivery.yml
@@ -633,7 +633,7 @@ jobs:
     steps:
       -
         name: Cleanup Disk
-        uses: jlumbroso/free-disk-space@main
+        uses: jlumbroso/free-disk-space@54081f138730dfa15788a46383842cd2f914a1be # v1.3.1
         with:
           android: true
           dotnet: true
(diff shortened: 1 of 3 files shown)


sxd:
I think this should be enough, there's more places but I think the future we can handle those.

@datosh this will require a commit message, will work on it after the CI pass
@mnencia this is the kind of solution we were talking about renovate right?

sxd (approved):

leonardoce (approved):

Issue #7812, opened 2025-06-12
[Bug]: Missing prometheus metrics cnpg_collector_last_available_backup_timestamp and cnpg_collector_first_recoverability_point


sbv-csis:
### Is there an existing issue already for this bug?

- [x] I have searched for an existing issue, and could not find anything. I believe this is a new bug.

### I have read the troubleshooting guide

- [x] I have read the troubleshooting guide and I think this is a new bug.

### I am running a supported version of CloudNativePG

- [x] I have read the troubleshooting guide and I think this is a new bug.

### Contact Details

sbv@csis.com

### Version

1.26 (latest patch)

### What version of Kubernetes are you using?

1.32

### What is your Kubernetes environment?

Cloud: Google GKE

### How did you install the operator?

YAML manifest

### What happened?

Looking at the metrics reported it seems like  cnpg_collector_last_available_backup_timestamp and cnpg_collector_first_recoverability_point never gets another value than 0 even when I have confirmed WAL archiving and backup status.

```
NAME                    AGE        METHOD   PHASE       ERROR
backup-20250611153500   21h        plugin   completed   
backup-20250612020000   5h4m       plugin   completed   
backup-20250612083230   4h21m      plugin   completed   
backup-20250612090510   3h49m      plugin   completed   
backup-20250612124800   6m28s      plugin   completed   
```
and both prometheus and the metrics endpoints still only show 0:
``` 
$ curl --silent -q $POD:9187/metrics | grep available_backup | grep -v "^#"     # across all replicas
cnpg_collector_last_available_backup_timestamp 0
``` 

It also seems to be observed in https://github.com/cloudnative-pg/cloudnative-pg/issues/7658 where I left a comment

Maybe it's related to the plugin version over the deprecated version of backup

### Cluster resource

```shell
apiVersion: postgresql.cnpg.io/v1
kind: Cluster
metadata:
  name: dmt-telos
  namespace: dmt-telos
spec:
  affinity:
    topologyKey: topology.kubernetes.io/zone
  enablePDB: true
  enableSuperuserAccess: true
  instances: 3
  serviceAccountTemplate:
    metadata:
      annotations:
        iam.gke.io/gcp-service-account: REPLACEME
  backup: {}
  plugins:
    - name: barman-cloud.cloudnative-pg.io
      enabled: true
      isWALArchiver: true
      parameters:
        barmanObjectName: google-store
  postgresql:
    parameters:
[...]

PR #8033, opened 2025-07-14
feat(instance): add cnpgi `metrics` capabilities to the instance webserver

armru:
# Proposed solution
This commit allows the plugins to export metrics to the instance metrics web server.
    
This work will also enable future patches to integrate the metrics that we previously had with in-tree barman, but that are now absent from the plugin.

Requires https://github.com/cloudnative-pg/cnpg-i/pull/219
Relates #7812 

# Alternate solution

Do not implement the metrics interface and require that the users scrape the sidecar containers directly


mnencia (approved):

leonardoce:
I tested it manually and verified that a plugin is able to expose custom metrics.

leonardoce (approved):

PR #8036, opened 2025-07-15
test: support beta versions in major upgrade E2Es

### LINKED ISSUE #8035: [Feature]: Support beta versions in major upgrade E2E tests

Major upgrade E2E tests currently do not support running with beta or rc versions, which is required given https://github.com/cloudnative-pg/cloudnative-pg/issues/7992.
Improve the test logic to recognize and allow upgrading to a BETA version.

This would also standardize and remove the necessity for a workaround that's currently being done in https://github.com/cloudnative-pg/postgres-trunk-containers/blob/main/.github/workflows/reusable-e2e.yml#L151-L152 to override the latest `DefaultImageName` of the codebase.

Beta and RC images do not (yet) have a major only tag (e.g `18` or `18-minimal-bookworm`) given that they are not stable images. We could improve the logic to detect BETA versions and then build the current image and the target image accordingly.

.

### Additional context

_No response_

### Backport?

Yes

Yes

### Code of Conduct

NiccoloFei:
Closes #8035


sxd (approved):

.github/pg_versions.json
```diff
@@ -1,4 +1,8 @@
 {
+  "18": [
+    "18beta1",
+    "18beta1-7"
```
gbartolini, line 4: Shouldn't these be beta2 @sxd?

mnencia: The update will occur automatically later, allowing also us to verify the automation for the new version.

fcanovai (approved):

PR #8072, opened 2025-07-21
fix(logpipe): add one second  before closing the file read

armru:
This will fix the rare cases where the logs from PostgreSQL are incomplete.

This is a mitigation and does not entirely fix the issue.

### DIFF
--- a/pkg/management/postgres/logpipe/logpipe.go
+++ b/pkg/management/postgres/logpipe/logpipe.go
@@ -156,7 +156,7 @@ func (p *LogPipe) collectLogsFromFile(ctx context.Context) error {
 	select {
 	case <-ctx.Done():
 		filenameLog.Info("Terminating log reading process")
-		err := f.SetDeadline(time.Now())
+		err := f.SetDeadline(time.Now().Add(1 * time.Second))
 		if err != nil {
 			filenameLog.Error(err,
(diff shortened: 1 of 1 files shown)


pkg/management/postgres/logpipe/logpipe.go
```diff
@@ -156,7 +156,7 @@ func (p *LogPipe) collectLogsFromFile(ctx context.Context) error {
 	select {
 	case <-ctx.Done():
 		filenameLog.Info("Terminating log reading process")
-		err := f.SetDeadline(time.Now())
+		err := f.SetDeadline(time.Now().Add(1 * time.Second))
```
mnencia, line 159: The file we are monitoring is not a standard file; it is a FIFO (First In, First Out) file. There are two possible scenarios to consider:

1. We are reading from a running PostgreSQL instance, which means we are performing an online upgrade of the instance manager. In this case, it does not matter when we stop reading, as the new instance manager will resume from the point where the previous one left off.

2. We are shutting down the pod. In this situation, PostgreSQL has already been stopped; otherwise, it would have been stuck when it attempted to write a log line.

mnencia:
I'm closing this because it should not cause any visible problems. After the release, we will evaluate whether to open another PR to improve the process's robustness.

PR #8087, opened 2025-07-21
feat(labels): adopt Kubernetes recommended labels for all CNPG resources

### LINKED ISSUE #8066: [Feature]: Add standard label to all CNPG-managed pods for policy targeting

In environments where PostgreSQL clusters are designed to be self-service, labels become essential to apply cluster-wide policies such as:

* Kubernetes `NetworkPolicy`
* Istio `PeerAuthentication`
* Kyverno or OPA-based policy enforcement
* Admission control via validating/mutating webhooks

Today there are no shared labels across all provisioned cnpg manged pods, meaning that a matchLabels statement cannot target all cnpg managed pods at once. this makes it quite difficult to apply these policies consistently across all provisioned objects.

I'd like to request that all pods managed by CloudNativePG be labeled with a consistent, well-known label such as:

```yaml
app.kubernetes.io/managed-by: cloudnative-pg
```

Adding a well-known label (such as `app.kubernetes.io/managed-by: cloudnative-pg`) to all pods created by CNPG-managed workloads (e.g. Postgres instances, init pods, wal-archive pods, etc) aligns with Kubernetes label best practices and makes it easier for platform operators to apply consistent policies.

you can find a list of well known recommended labels here: https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#labels

Today, it's possible to work around this by using a `matchExpressions` clause and checking for the existence of the `cnpg.io/cluster` label, for example:

```yaml
matchExpressions:
  - key: cnpg.io/cluster
    operator: Exists
```

However, some systems (such as Istio's `PeerAuthentication`) do not support `matchExpressions` and require explicit `matchLabels`. Without a consistent label, it becomes difficult to safely target only CNPG-managed pods with such policies. Today I am forced to rely on a consumer to add the label, or resort to creating a mutating policy to add the label, while these work, it certainly adds complexity to the process.

### Additional context

_No response_

### Backport?

Yes

Yes

### Code of Conduct

JefeDavis:
Apply Kubernetes recommended labels
(https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#labels)
to every object managed by CloudNativePG in a PostgreSQL cluster:

- `app.kubernetes.io/name`: postgresql
- `app.kubernetes.io/instance`: <cluster name>
- `app.kubernetes.io/version`: <PostgreSQL version of running pod>
- `app.kubernetes.io/component`: database
- `app.kubernetes.io/managed-by`: cloudnative-pg

This ensures consistent labelling and improves integration with
Kubernetes-native tooling and ecosystem practices.

Closes #8066


JefeDavis:
@jsilvela could you review when you get a moment? Let me know if you want me to change anything 😊

jsilvela (commented):
Very good stuff.
I think it would be good to add simple tests in the various test files for these.
E.g. in services_test.go, the unit tests check for labels like `"cnpg.io/instanceRole"`.
Will not add insight right now, but I'd consider it insurance against inadvertent removal.

JefeDavis:
@jsilvela I've added the requested tests! let me know if there's anything you want added or changed!

pkg/specs/jobs_test.go
```diff
@@ -195,4 +196,41 @@ var _ = Describe("Job created via InitDB", func() {
 		Expect(initdbFlags).ShouldNot(ContainSubstring("--locale="))
 		Expect(initdbFlags).Should(ContainSubstring("'--icu-rules=&A < z <<< Z'"))
 	})
+
+	It("contains correct labels", func() {
+		cluster := apiv1.Cluster{
```
jsilvela, line 201:
```suggestion
		cluster := apiv1.Cluster{
			ObjectMeta: metav1.ObjectMeta{
				Name: "cluster",
			},
```

pkg/specs/jobs_test.go
```diff
@@ -195,4 +196,41 @@ var _ = Describe("Job created via InitDB", func() {
...
+					},
+				},
+			},
+		}
+		job := CreatePrimaryJobViaInitdb(cluster, 0)
+		Expect(job.Labels).To(BeEquivalentTo(map[string]string{
+			utils.ManagedByLabelName:    utils.ManagerName,
+			utils.ClusterLabelName:      cluster.Name,
+			utils.JobRoleLabelName:      "initdb",
+			utils.InstanceNameLabelName: "-0",
```
jsilvela:
```suggestion
			utils.InstanceNameLabelName: "cluster-0",
```

pkg/specs/jobs_test.go
```diff
@@ -195,4 +196,41 @@ var _ = Describe("Job created via InitDB", func() {
...
+		}))
+		Expect(job.Spec.Template.Labels).To(BeEquivalentTo(map[string]string{
+			utils.ManagedByLabelName:    utils.ManagerName,
+			utils.ClusterLabelName:      cluster.Name,
+			utils.JobRoleLabelName:      "initdb",
+			utils.InstanceNameLabelName: "-0",
```
jsilvela:
```suggestion
			utils.InstanceNameLabelName: "cluster-0",
```

jsilvela (commented):
Good stuff.
The one bit is the tests where the instance name is "-0". Added a suggestion to fix that.
Worth fixing that way or some other way you may think.
Aside from that, looking good.

gbartolini:
In the maintainers' chat, @sxd rightly pointed out that we should add the other labels that @JefeDavis pointed out initially (see again https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#labels).

Ideally:

```yaml
    app.kubernetes.io/name: postgresql
    app.kubernetes.io/instance: <CLUSTER NAME>
    app.kubernetes.io/version: <POSTGRESQL VERSION OF THE PRIMARY OR THE RUNNING POD (for replicas)>
    app.kubernetes.io/component: database
    app.kubernetes.io/managed-by: cloudnative-pg
```

Thoughts?

leonardoce:
> Thoughts?

Yes, that is a good idea.

JefeDavis:
Would you like me to add those to this PR? Or I can do a follow up?

Would this affect the selector labels? Or should those be left alone?

leonardoce:
> Would you like me to add those to this PR? Or I can do a follow up?

@JefeDavis If you could add them to this PR, it would be great. I won't change the selectors.
Thanks!

JefeDavis:
@leonardoce do you want all those labels added to all objects or just the pods, jobs, and deployments?

JefeDavis:
would component always be database? i.e. pgBouncer?

JefeDavis:
I've added labels where i think it makes sense, let me know what you think

JefeDavis:
@leonardoce ready for review

JefeDavis:
@jsilvela ready for review

pkg/specs/jobs.go
```diff
@@ -318,6 +318,7 @@ func (role jobRole) getJobName(instanceName string) string {
 func CreatePrimaryJob(cluster apiv1.Cluster, nodeSerial int, role jobRole, initCommand []string) *batchv1.Job {
 	instanceName := GetInstanceName(cluster.Name, nodeSerial)
 	jobName := role.getJobName(instanceName)
+	version, _ := cluster.GetPostgresqlMajorVersion()
```
gbartolini, line 323: I would add at least the minor version. What do you think @JefeDavis ?

JefeDavis: Is the minor version available anywhere? I couldn't, from what I saw only the major version is recorded and available for the cluster.

JefeDavis:
@gbartolini @leonardoce I'm kind of a mind to split out these changes into a separate MR, I'm not sure if the scoping is right on what objects should receive what labels. My Original goal here was to add a common label so I can target all clusters for network policies and istio peerAuthentications, as this is a current pain point I'm experiencing. Any help here would be appreciated

pkg/specs/poddisruptionbudget_test.go
```diff
@@ -44,6 +45,14 @@ var _ = Describe("POD Disruption Budget specifications", func() {
 	It("have the same name as the PostgreSQL cluster", func() {
 		result := BuildReplicasPodDisruptionBudget(cluster)
 		Expect(result.Name).To(Equal(cluster.Name))
+		Expect(result.Labels).To(BeEquivalentTo(map[string]string{
+			utils.ClusterLabelName:   cluster.Name,
+			utils.AppLabelName:       utils.AppName,
+			utils.InstanceLabelName:  cluster.Name,
+			utils.VersionLabelName:   "17",
+			utils.ComponentLabelName: "database",
```
jsilvela: Why not use utils.DatabaseComponentName?

pkg/utils/labels_annotations.go
```diff
@@ -112,6 +121,21 @@ const (
 
 	// LivenessPingerAnnotationName is the name of the pinger configuration
 	LivenessPingerAnnotationName = AlphaMetadataNamespace + "/livenessPinger"
+
+	// ManagedByLabelName is the name of the label applied to all managed objects
+	ManagedByLabelName = "app.kubernetes.io/managed-by"
+
+	// NameLabelName is the name of the label used to indicate the name of the application
+	AppLabelName = "app.kubernetes.io/name"
```
jsilvela: AppNameLabel, no?
Also, the beginning of the comment is currently mismatching

pkg/specs/podmonitor.go
```diff
@@ -42,9 +44,18 @@ func (c ClusterPodMonitorManager) IsPodMonitorEnabled() bool {
...
+
 	meta := metav1.ObjectMeta{
 		Namespace: c.cluster.Namespace,
 		Name:      c.cluster.Name,
+		Labels: map[string]string{
+			utils.ManagedByLabelName: utils.ManagerName,
+			utils.AppLabelName:       utils.AppName,
+			utils.InstanceLabelName:  c.cluster.Name,
+			utils.VersionLabelName:   fmt.Sprint(version),
+			utils.ComponentLabelName: "database",
```
jsilvela: here also, why not use the constant you defined in utils?

pkg/specs/podmonitor.go
```diff
@@ -42,9 +44,18 @@ func (c ClusterPodMonitorManager) IsPodMonitorEnabled() bool {
...
 func (c ClusterPodMonitorManager) BuildPodMonitor() *monitoringv1.PodMonitor {
+	version, _ := c.cluster.GetPostgresqlMajorVersion()
+
 	meta := metav1.ObjectMeta{
 		Namespace: c.cluster.Namespace,
 		Name:      c.cluster.Name,
+		Labels: map[string]string{
+			utils.ManagedByLabelName: utils.ManagerName,
+			utils.AppLabelName:       utils.AppName,
+			utils.InstanceLabelName:  c.cluster.Name,
```
jsilvela: I find this a bit discontinuous.
I mean, in the K8s page on recommended labels, you see: application -> mysql, instance -> mysql-abcxyz
Instead here, we'd have application -> postgres,  instance -> acme-corp-cluster
Maybe something like "postgres-CLUSTER-NAME"?

jsilvela (commented):
This is turning out to be quite the first pull request!
Sorry, have some new comments. Basically about wording.
Good work!

jsilvela (commented):
This is turning out to be quite the first commit!
Sorry, have some new comments. Basically about wording.
Good work!

jsilvela (commented):
This is turning out to be quite the first commit!
Sorry, have some new comments. Basically about wording.
Good work!

jsilvela (commented):
This is turning out to be quite the first commit!
Sorry, I have some new comments. Basically about wording.
Good work!

jsilvela:
Apologies if you got multiple messages from me. Was trying to submit my review in the middle of a Github outage, it seems 🤷

pkg/utils/labels_annotations.go
jsilvela: Actually, no, disregard.

jsilvela: You have a discrepancy between the name and the comment. Perhaps half-decided?
And, maybe AppNameLabelName? Even if it stutters a bit, it's more self-descriptive.

pkg/specs/podmonitor.go
JefeDavis: the application and instance are necessarily linked, the instance is just the unique name of that entity grouping.

i.e. in the context a helm app.kubernetes.io/name is usually the name of the chart, while app.kuberentes.io/instance is usually the name of the release

i can deploy cloudnative-pg chart twice and the first one i use the release name database-operator, while the second one i name cnpg, the name will be cloudnative-pg for both but the instances will be database-operator and cnpg resepctively

JefeDavis:
@jsilvela I think that is everything, let me know if anything else is needed

pkg/specs/jobs.go
```diff
@@ -338,6 +344,11 @@ func CreatePrimaryJob(cluster apiv1.Cluster, nodeSerial int, role jobRole, initC
 						utils.InstanceNameLabelName: instanceName,
 						utils.ClusterLabelName:      cluster.Name,
 						utils.JobRoleLabelName:      string(role),
+						utils.AppLabelName:          utils.AppName,
+						utils.InstanceLabelName:     cluster.Name,
+						utils.VersionLabelName:      fmt.Sprint(version),
+						utils.ComponentLabelName:    "database",
```
jsilvela: "database" -> utils.DatabaseComponentName,

jsilvela (approved):
The magic constant "database" shouldn't need to be there since you have a const definition in utils.
And similarly for "pooler".
That said, those are nice-to-have so I'm giving this a thumbs up

pkg/specs/jobs.go
jsilvela: here and everywhere where you have it

JefeDavis:
@jsilvela ah I apologize I thought I got them all, I will update and push here in a bit

JefeDavis:
@jsilvela i have converted component labels to use the constant

jsilvela:
Want to go back to a previous point in the messages around this PR.

> Ideally:
>
>    app.kubernetes.io/name: postgresql
>    app.kubernetes.io/instance: <CLUSTER NAME>
>    app.kubernetes.io/version: <POSTGRESQL VERSION OF THE PRIMARY OR THE RUNNING POD (for replicas)>
>    app.kubernetes.io/component: database
>    app.kubernetes.io/managed-by: cloudnative-pg
>
> Thoughts?

I was looking at the existing codebase to see if there were places where we already used some of the `app.kubernetes.io/` labels.
And, `app.kubernetes.io/name` is used a lot, with the value `cloudnative-pg`, NOT `postgresql`.

E.g. in `config/manager/manager.yaml`:

```
metadata:
  name: controller-manager
  namespace: system
  labels:
    app.kubernetes.io/name: cloudnative-pg
```

so, do we keep the value for `app.kubernetes.io/name` as it was, or change it?

JefeDavis:
@jsilvela Not sure if that question is for me, but, To me cloudnative-pg is the controller, a postgres is the pods spawned from it.

I would consider it to be managed-by cloudnative-pg not cloudnative-pg itself

If these pods also used the same label values as the controller it could inadvertently be selected by the controller's replicaSet or service. It would be ill-advised to make them the same values

jsilvela:
@JefeDavis you're right about the specific example I used.
The app name label is used in many places.
Another place is monitoring. In `config/prometheus/monitor.yaml` we have

``` yaml
spec:
  selector:
    matchLabels:
      app.kubernetes.io/name: cloudnative-pg
```

If we're going to have the label set to `cloudnative-pg` in the controller pod / deployment (which may make sense as you suggest) but set to `postgres` in instance pods, we need to pay more attention to this kind of thing.

By the way, you're right, my comment was more for the maintainers than for you.

JefeDavis:
@jsilvela would it be better to split this merge and just do the managed by labels in this one?

 Seems like there might be more design discussion, my original issue was around the need to target all managed pods for network policies and istio peerAuthentications. I would love to see this merged soon as it is currently a pain point for me, however I feel this may have fallen victim to some scope creep

jsilvela:
@JefeDavis I don't know what to advise. On one hand your PR has come a long way and I think it's close.
On the other hand the maintainers need to weigh in: @leonardoce , @gbartolini

armru:
Hello, linter was failing pushed a chore to handle that

armru:
I think the PR should be split, but waiting for other maintainers opinion too

JefeDavis:
@gbartolini @leonardoce should this PR be split?

jsilvela:
OK, I'll add another reminder about this PR. @gbartolini @leonardoce

JefeDavis:
@gbartolini @leonardoce can I get a way forward on this?

gbartolini:
I take responsibility for the change in scope of this patch, but I believe we should maintain it as a single commit, despite being quite invasive (small changes in many files). The aim is to provide support for [Kubernetes' recommended labels](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#labels).
We can take your work from here on @JefeDavis. Thanks for your contribution!

gbartolini:
Moreover, we should extend this to plugins too. For example, Barman Cloud plugin should honour the same set of labels, and we should at least provide guidance for generic plugins to use those labels.

gbartolini:
Another point. Should we also add them to the `config/manager/default-monitoring.yaml` config map? Any similar resources to consider from the main manifest?

NiccoloFei:
The `app.kubernetes.io/managed-by: cloudnative-pg` has been added to every object, however for certain objects we currently only reconcile the content, and not the objectMeta, thus any change to labels/annotations in such objects will not be propagated. From what I've noticed, this is the list of those objects:
```
role, rolebinding, serviceaccount, TLS and CA secrets, `cnpg-default-monitoring` configmap
```
As such, the `app.kubernetes.io/managed-by` label in these object will only be applied for newly created Clusters, and will instead not be propagated in existing Clusters that upgrade to a new version of the operator.

If we want to start reconciling the object metadata for these resources , that should be done as part of a new separate issue.

NiccoloFei (approved):

gbartolini:
I think it is fine for now. This is just a nice to have recommendation after all. Looking forward, things will get adjusted over time. Thanks!

JefeDavis:
@NiccoloFei @gbartolini @leonardoce @jsilvela can this be merged? It seems like it was approved but never merged, it is a feature that will greatly improve the ability to target CNPG pods for network policies and istio auth policies.

mnencia:
After reviewing with the other maintainers, we agreed that the `app.kubernetes.io/version` label should contain the version of the CloudNativePG operator, rather than the version of Postgres. Additionally, I changed the application name to `postgresql.cnpg.io` to make it clear that this is not just a bare PostgreSQL deployment.

mnencia (approved):

mnencia:
After some more digging, I'm going to revert my last commit (2b9fae5c4383). I've had second thoughts and realized my changes weren't aligned with community best practices.

Long story short, the labels should describe the application (PostgreSQL), not the operator (CNPG).

The evidence that convinced me:

1.  **`app.kubernetes.io/name` should describe the workload, not the manager.**
    *   The [official Kubernetes documentation](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/) uses simple names like `mysql`. The goal is for the user to be able to ask, "Show me all my MySQL instances," not "Show me all the MySQL instances managed by a specific tool." Using `postgresql` aligns with this principle.

2.  **`app.kubernetes.io/version` should describe the application's version, not the manager's version.**
    *   The same Kubernetes docs specify that this label is for "The current version of the application."
    *   More importantly, the [Helm project's best practices](https://helm.sh/docs/chart_best_practices/labels/) provide a strong precedent. They use `app.kubernetes.io/version` for the application version (the `appVersion` in a Chart.yaml) and use a separate `helm.sh/chart` label for the Helm chart's own version. This clear separation of concerns, where the workload is described separately from the managing tool, is the standard we should follow.
    *   Using the PostgreSQL version provides direct value to users for version tracking and applying policies (e.g., "apply to all PG14 clusters"). Using the operator version removes this critical capability.

My commit was incorrect from a user's perspective. I'll push the revert shortly.

mnencia (approved):

PR #8088, opened 2025-07-21
test(logpipe): improve extra logs test reliability

armru:
should handle the flaky unit tests in main (hopefully)


mnencia (approved):

mnencia:
I don't know if this is the fix that fixes the test (on my machine, it never fails even with the previous code), but the logic is sound. I propose to merge it ASAP.

sxd (approved):

PR #8091, opened 2025-07-21
docs(cnpg-i): deprecate backup metrics and fields in `Cluster` status

### LINKED ISSUE #8051: [Docs]: Deprecate Cluster status fields related to in-tree backup implementation

The Cluster status includes five fields that display the newest and oldest available backups, as well as the name of the last failed backup. It's important to note that these fields only apply to in-tree backup implementations and cannot be used with CNPG-I backup plugins. Backup plugins manage their own retention policies and may store this information within their configuration objects; for example, the plugin-barman-cloud does this in the ObjectStore status.

I propose to deprecate the following fields in the Cluster status:

- `firstRecoverabilityPoint`
- `firstRecoverabilityPointByMethod`
- `lastSuccessfulBackup`
- `lastSuccessfulBackupByMethod`
- `lastFailedBackup`

_No response_

_No response_

### Additional context

I recommend initiating the deprecation process with the upcoming release of version 1.26.1. This move will clearly communicate that these fields do not function with backup plugins, which have been the recommended method for backups since version 1.26.0. Once the in-tree backup implementation is removed, these fields will be forcibly left blank.

### Backport?

Yes

### Code of Conduct

gbartolini:
Due to the transition to a plugin-based, backup-agnostic architecture in CloudNativePG, we are deprecating:

- `firstRecoverabilityPoint`, `firstRecoverabilityPointByMethod`, `lastSuccessfulBackup`, `lastSuccessfulBackupByMethod`, and `lastFailedBackup` in `Cluster.status`
- `cnpg_collector_first_recoverability_point`, `cnpg_collector_last_failed_backup_timestamp`, `cnpg_collector_last_available_backup_timestamp` metrics

If you have migrated to a plugin-based backup and recovery solution such as Barman Cloud, these fields/metrics are no longer synchronized and will not be updated. Users still relying on the in-core support for Barman Cloud and volume snapshots can continue to use these fields for the time being.

Closes #8051

### DIFF
--- a/api/v1/cluster_types.go
+++ b/api/v1/cluster_types.go
@@ -907,32 +907,32 @@ type ClusterStatus struct {
 	// The first recoverability point, stored as a date in RFC3339 format.
 	// This field is calculated from the content of FirstRecoverabilityPointByMethod.
 	//
-	// Deprecated: the field is not set for backup plugins. In-tree backup will be removed in 1.28.
+	// Deprecated: the field is not set for backup plugins.
 	// +optional
 	FirstRecoverabilityPoint string `json:"firstRecoverabilityPoint,omitempty"`
 
 	// The first recoverability point, stored as a date in RFC3339 format, per backup method type.
 	//
-	// Deprecated: the field is not set for backup plugins. In-tree backup will be removed in 1.28.
+	// Deprecated: the field is not set for backup plugins.
 	// +optional
 	FirstRecoverabilityPointByMethod map[BackupMethod]metav1.Time `json:"firstRecoverabilityPointByMethod,omitempty"`
 
 	// Last successful backup, stored as a date in RFC3339 format.
 	// This field is calculated from the content of LastSuccessfulBackupByMethod.
 	//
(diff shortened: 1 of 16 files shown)


mnencia (approved):

api/v1/cluster_types.go
```diff
@@ -905,24 +905,34 @@ type ClusterStatus struct {
 	Certificates CertificatesStatus `json:"certificates,omitempty"`
 
 	// The first recoverability point, stored as a date in RFC3339 format.
-	// This field is calculated from the content of FirstRecoverabilityPointByMethod
+	// This field is calculated from the content of FirstRecoverabilityPointByMethod.
+	//
+	// Deprecated: the field is not set for backup plugins. In-tree backup will be removed in 1.28.
```
gbartolini: This is not true. In 1.28 we will remove only native Barman Cloud support. Volume snapshots will continue to be there.

fcanovai: I've removed that

fcanovai (approved):

PR #8095, opened 2025-07-22
Adding troubleshooting section to help restoring clusters that were accidently deleted

lucasvmiguel:
I believe a section to explain how to recover a cluster that was accidentally deleted could be beneficial as I see many people having issues like that.

Reference: https://github.com/cloudnative-pg/cloudnative-pg/discussions/5253#discussioncomment-12651012

### DIFF
--- a/docs/src/troubleshooting.md
+++ b/docs/src/troubleshooting.md
@@ -830,8 +830,9 @@ If you accidentally delete a cluster and you don't have an up-to-date backup, yo
 Here are the steps to restore the cluster:
 
 1. Find the `PersistentVolume` (PV) that was associated to the PostgreSQL cluster.
-2. Remove the `claimRef` field from the PV so it can be attached to a new `PersistentVolumeClaim` (PVC).
-3. Create a new PVC with `volumeName` equal to the name of the PV found in step 1.
-4. Once the PVC is correctly attached to the PV, you can create `VolumeSnapshot` K8S resource from the PVC by setting `persistentVolumeClaimName` equal to the name of the PVC.
-5. Delete the recently created PVC so CNPG Operator can recreate a new PVC from the `VolumeSnapshot`.
-6. Finally, with the `VolumeSnapshot` resource created, you can follow the instructions of [recovery from VolumeSnapshot objects](recovery.md#recovery-from-volumesnapshot-objects).
+2. Save the value from `spec.claimRef.name` field - the value will be later used in step 4.
+3. Remove the `claimRef` field from the PV so it can be attached to a new `PersistentVolumeClaim` (PVC). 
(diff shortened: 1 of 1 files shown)


adrianmarcu18:
@lucasvmiguel I think the documentation should be a little bit more clear in terms of naming convention. 
I have tested several times and looks like if the volumesnapshot's source PVC has a different name than what the CNPG expects, the PVC does not get created, and the recovery will fail.

Let's say I want to recover cluster "myapp-postgres". I than create a PVC with the name "myapp-postgres" (without the index number), I do a volume snapshot, than I use that volume snapshot to recover the cluster, CNPG will ignore it and does not  create a properly named PVC.

Additionally, from my tests I see that this step is not needed.

> 5. Delete the recently created PVC so CNPG Operator can recreate a new PVC from the `VolumeSnapshot`.

Following the exact steps from your docs, but skipping this one, the CNPG cluster will be able to use it directly, without having to create it from the VolumeSnapshot. 

Why this is important? Because, if someone wants to automate this process in any way (e.g. build the logic within the CNPG Helm Chart), the VolumeSnapshot could be automatically created for recovery process, without having to delete in advance the existing PVC.

lucasvmiguel:
@adrianmarcu18

I addressed your comment about being specific with the pvc name that needs to be created. However, I still think that step 5 (now step 6) is a nice thing to do in order to prove that the operator is working and recovering the database as expected. In any case, I left as optional so people can choose if they want to do that or not.

docs/src/troubleshooting.md
```diff
@@ -819,3 +819,20 @@ You can work around this issue by setting `STANDBY_TCP_USER_TIMEOUT` in the
...
+!!! Important
+    Make sure you configure your Kubernetes PVs with `persistentVolumeReclaimPolicy` set to `Retain` so that the PV is not deleted when the cluster is deleted.
+
+Here are the steps to restore the cluster:
+
+1. Find the `PersistentVolume` (PV) that was associated to the PostgreSQL cluster.
+2. Save the value from `spec.claimRef.name` field - the value will be later used in step 4.
+3. Remove the `claimRef` field from the PV so it can be attached to a new `PersistentVolumeClaim` (PVC). 
+4. Create a new PVC with name from `spec.claimRef.name` (step 2) and with `volumeName` equal to the name of the PV found in step 1.
+5. Once the PVC is correctly attached to the PV, you can create `VolumeSnapshot` K8S resource from the PVC by setting `persistentVolumeClaimName` equal to the name of the PVC.
```
WentingWu666666, line 836: @lucasvmiguel CNPG currently support restore from PVC. I don't think we need step5 to create VolumeSnapshot from PVC for restore.

Reference: https://github.com/cloudnative-pg/cloudnative-pg/blob/3d8826820f75c49fcadfa71e9acf4203c84a4460/docs/src/cloudnative-pg.v1.md#bootstraprecovery
image

PR #8127, opened 2025-07-24
feat(webhook): add warnings for storage configuration

armru:
This commit introduces new admission warnings for the Cluster resource to inform users about potentially confusing storage configurations.

A warning is now issued when:
- Both `spec.storage.size` and `spec.storage.pvcTemplate.resources.requests.storage` are set, clarifying that `size` will be used.
- Both `spec.storage.storageClass` and  `spec.storage.pvcTemplate.storageClassName` are set, clarifying that`storageClass` will be used.

These checks are applied to both `spec.storage` and `spec.walStorage`.

### DIFF
--- a/internal/webhook/v1/cluster_webhook.go
+++ b/internal/webhook/v1/cluster_webhook.go
@@ -2421,7 +2421,58 @@ func (v *ClusterCustomValidator) getAdmissionWarnings(r *apiv1.Cluster) admissio
 	list := getMaintenanceWindowsAdmissionWarnings(r)
 	list = append(list, getInTreeBarmanWarnings(r)...)
 	list = append(list, getRetentionPolicyWarnings(r)...)
+	list = append(list, getStorageWarnings(r)...)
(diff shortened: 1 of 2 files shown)


internal/webhook/v1/cluster_webhook.go
```diff
@@ -2447,9 +2447,60 @@ func (v *ClusterCustomValidator) getAdmissionWarnings(r *apiv1.Cluster) admissio
...
+				),
+			)
+		}
+		requestsSpecified := !configuration.PersistentVolumeClaimTemplate.Resources.Requests.Storage().IsZero()
+		if configuration.Size != "" && requestsSpecified {
+			size := path.Child("size").String()
+			result = append(
+				result,
+				fmt.Sprintf(
+					"%s and %s  are both specified, %s value will be used.",
```
mnencia:
```suggestion
					"%s and %s are both specified, %s value will be used.",
```

jbattiato (approved):

leonardoce (approved):

PR #8140, opened 2025-07-25
fix: pointer slice loop

mnencia:
Fix the loop when converting item slices to pointer slices. The previous version unnecessarily allocates memory, and with an older compiler, it does not work as expected.

### DIFF
--- a/api/v1/generic_funcs.go
+++ b/api/v1/generic_funcs.go
@@ -59,9 +59,9 @@ func ensureManagedResourceExclusivity[T managedResourceComparer](t1 T, list []T)
 
 // toSliceWithPointers converts a slice of items to a slice of pointers to the items
 func toSliceWithPointers[T any](items []T) []*T {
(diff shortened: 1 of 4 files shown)


armru (approved):

sxd (approved):

jsilvela:
Is this really a "fix"? Since go 1.22 the scoping in for loops is right, so this is just an optimization to have pointers to the original array elements, rather than pointers to equivalent values, right?
Curious how this was triggering a bug?

mnencia:
> Is this really a "fix"? Since go 1.22 the scoping in for loops is right, so this is just an optimization to have pointers to the original array elements, rather than pointers to equivalent values, right? Curious how this was triggering a bug?

It unnecessarily allocates memory, and with an older compiler, it does not work as expected. Therefore, it is better to correct the code to avoid any surprises.

Issue #7566, opened 2025-05-14
[Feature]: Add support for ephemeral volumes/data loss on postgres instances


connorbradshaw10:
### Is there an existing issue already for this feature request/idea?

- [x] I have searched for an existing issue, and could not find anything. I believe this is a new feature request to be evaluated.

### What problem is this feature going to solve? Why should it be added?

I am trying to integrate Azure Container Storage to use CNPG. Azure Container Storage has a storage option that uses local NVMe disks to bring super high performance to DB workloads. However, the downside of local NVMe disks in Azure is that they are ephemeral. If a node is lost, we lose the data associated with any of the PVCs that were created on that node. In the case of CNPG, when we scale down or upgrade a node, the CNPG instance on that node is migrated to a different node. The underlying volume will be recreated fresh but the PVC/PV will not change. Whenever the CNPG instance starts back up, it expects /var/lib/postgresql/data/pgdata to exist. Since this directory does not exist anymore, the pod is stuck in a CrashLoopBackOff state.

### Describe the solution you'd like

I am looking for a solution where CNPG can recognize complete data loss on an instance and rebuild it as if it were a new instance automatically or support backing the postgres data directory with a generic ephemeral volume.

### Describe alternatives you've considered

There is a mitigation to run kubectl cnpg destroy on the instance that is stuck, but it would be nice to be able to automate this process.

### Additional context

@eh8

### Backport?

Yes

### Are you willing to actively contribute to this feature?

Yes

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

armru:
I think the issue you're facing with Azure Container Storage and ephemeral NVMe disks is part of a broader problem of "tainted" PVCs that are no longer safe to use after node migrations or corruptions.

To solve this issue, I propose we add a new condition to apply to the PVCs that, when present, will triggers a new logic in the reconciliation that will destroy the PVC/PVCGroup tainted by the new condition. This will result in the operator to recreate the instance/s with fresh storage.

This would eliminate the need for manual intervention facing race conditions with the operator reconciliation loop.

This solution requires an external component that adds the new condition to the bad PVCs when needed.

Would this approach address your use case? Are you willing to contribute to the feature? In case you want to contribute to this feature, we (mantainers) are willing to provide guidance.

NedAnd1:
The issue isn't that the PVCs/PVs are no longer safe to use,
[...]

PR #8178, opened 2025-07-29
feat: "unrecoverable" annotation

leonardoce:
This patch introduces the annotation `alpha.cnpg.io/unrecoverable=true`, which can be added to any replica instance Pod.
When this annotation is present, the operator will permanently delete the associated instance by removing both the PVCs and the Pod itself. Following this, the operator will recreate the missing instance from the primary instance.

Please note that this annotation is available only for replica instances.


leonardoce:
This patch is marked as a draft because we miss E2e tests and we should improve the documentation.

leonardoce:
I added a basic E2e test.

armru (approved):

GabriFedi97:
This annotation, and the resulting behavior, could also be useful for destroying an instance that isn't necessarily "unrecoverable." I'd suggest changing its name to something that more generally describes its function, rather than tying it to a specific use case.
Perhaps calling it `alpha.cnpg.io/destroy` might be more appropriate.

mnencia:
I find it a bit counterintuitive to put the annotation on the Pod to trigger deletion of the Pod and all its PVCs. The intent here is clearly storage/instance-scoped, while a Pod is ephemeral. If PVC deletion fails and the Pod disappears for any reason, the request is gone. I’d really prefer to see the trigger live on something more persistent (like the PVC or the Cluster spec/status) so the action is durable and auditable.

NedAnd1:
Should the relevant issue which this PR was intended for be linked to this PR?
It has an outstanding discussion: https://github.com/cloudnative-pg/cloudnative-pg/issues/7566

mnencia:
I spoke with @leonardoce, and we agreed that the patch is safe, provided we change the order of the delete operations. Specifically, I changed the patch to issue the Pod delete after the PVCs one. I will merge once the end-to-end (E2E) tests have passed.

mnencia (approved):

NedAnd1:
@mnencia @leonardoce
When pods are drained from one node & re-created on another node (via kubectl drain),
volumes for those pods backed by local storage are reprovisioned on that new node
with new, empty underlying storage.
Our CSI driver, which provides these local volumes,
reprovisions them if it notices that a volume has moved to a new node, based on volume annotations,
& it doesn't directly interact with application pods.

We've successfully tested an alternative approach that fixes, within CNPG, the CNPG crash loop observed,
without the need for reasoning about the complex question of how an external component,
especially a general-purpose one, should apply a CNPG-specific annotation to CNPG-specific pods,
when that component doesn't directly interact with external pods in the first place.

jmealo:
What's the difference between `kubectl cnpg destroy` and this annotation?

Issue #8182, opened 2025-07-30
[Bug]: fix changing image and PostgreSQL settings simultaneously with extensions


gbartolini:
### Is there an existing issue already for this bug?

- [x] I have searched for an existing issue, and could not find anything. I believe this is a new bug.

### I have read the troubleshooting guide

- [x] I have read the troubleshooting guide and I think this is a new bug.

### I am running a supported version of CloudNativePG

- [x] I have read the troubleshooting guide and I think this is a new bug.

### Version

1.26 (latest patch)

### What version of Kubernetes are you using?

1.33

### What is your Kubernetes environment?

Self-managed: kind (evaluation)

### How did you install the operator?

YAML manifest

### What happened?

The new dynamic loading of extensions feature is not compatible with PR #8115 that introduced support for changing image name and configuration at the same time. Please fix this for 1.27.0.

### Cluster resource

```shell

```

### Relevant log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

NishantSinghhhhh:
@gbartolini hey can I work on this issue, ? I am highly motivated to get this solved

gbartolini:
@NishantSinghhhhh this is blocking 1.27.0 release. Feel free to review what @armru has proposed.

gbartolini:
Implemented by #8241

PR #8184, opened 2025-07-30
feat: add `extensions.conf`

armru:
Relates #8182


pkg/management/postgres/configuration.go
```diff
@@ -253,15 +253,31 @@ func (instance *Instance) RefreshPGIdent(
...
 
 	return changed, createStandbySignal(pgData)
 }
 
+func configureAdditionalConfFiles(pgData, primaryConnInfo, slotName string, cluster *apiv1.Cluster) (bool, error) {
+	changed, err := configurePostgresOverrideConfFile(pgData, primaryConnInfo, slotName)
+	if err != nil {
+		return changed, err
+	}
+	extChanged, err := configureExtensionsConfFile(pgData, cluster)
```
armru, line 273: we might need to ignore extChanged

PR #8265, opened 2025-08-06
fix: always consider `WALArchiver` as a valid instance sidecar for metrics

armru:


leonardoce:
The metrics capability is only available from 1.27. I'm removing the backporting labels.

NiccoloFei (approved):

leonardoce (approved):

gbartolini (approved):

Issue #8189, opened 2025-07-31
[Bug]: Pod stuck when migrating from barman-in-tree to plugin


gabriele-wolfox:
### Is there an existing issue already for this bug?

- [x] I have searched for an existing issue, and could not find anything. I believe this is a new bug.

### I have read the troubleshooting guide

- [x] I have read the troubleshooting guide and I think this is a new bug.

### I am running a supported version of CloudNativePG

- [x] I have read the troubleshooting guide and I think this is a new bug.

### Version

1.26 (latest patch)

### What version of Kubernetes are you using?

1.32

### What is your Kubernetes environment?

Self-managed: kind (evaluation)

### How did you install the operator?

YAML manifest

### What happened?

Steps to reproduce it:
- creating a CNPG cluster with 3 pod instances and configure it to archive WALs using the in-tree Barman cloud support
- make a bit of traffic and ensuring WAL archiving really works
- interrupt the connection between the object storage and the primary, and make a bit of traffic
WAL archiving will begin to lag
- try to migrate this situation to the plugin

I used Kind and Azure storageaccount.

The reason is that we aren't able to archive it because there is no plugin, yet.

Deleting the stuck pod and let it be created again worked, as workaround.

### Cluster resource

```shell

```

### Relevant log output

```shell
This is the error I got:

{"level":"info","ts":"2025-07-31T09:03:41.252339614Z","msg":"Detected ready WAL files in a former primary, triggering WAL archiving","logger":"instance-manager","logging_pod":"pg14-2","controller":"instance-cluster","controllerGroup":"postgresql.cnpg.io","controllerKind":"Cluster","Cluster":{"name":"pg14","namespace":"default"},"namespace":"default","name":"pg14","reconcileID":"98136b34-46ab-410a-8550-6943e26f3baf","logging_pod":"pg14-2","readyWALCount":3}
[...]

PR #8236, opened 2025-08-04
fix: unblock rollout of former primary Pod when WAL archiving plugin is missing

leonardoce:
This patch addresses a rollout issue when migrating from the in-tree Barman Cloud support to the `barman-cloud` plugin, particularly when using the `switchover` primary update strategy.

During such a migration, a switchover is triggered to roll out the former primary instance. After the switchover, the former primary Pod restarts and attempts to archive any remaining WAL files in the queue before starting PostgreSQL. These files may still be present due to archiving or misconfiguration delays.

However, the previous version of the Pod does not include support for the plugin-based archiving system, causing all archiving attempts to fail during this pre-start phase. As a result, the Pod stalls indefinitely, blocking the rollout.

This fix introduces a mechanism to detect this specific failure scenario. If WAL archiving cannot proceed due to a missing plugin, the Pod exits with a distinct exit code. The operator recognises this exit code and replaces the Pod with one based on the updated container image, ensuring the plugin is available and WAL archiving can continue as expected.

Fixes: #8189

PR #8273, opened 2025-08-06
fix: unblock rollout of former primary Pod when WAL archiving plugin is missing (#8236)

leonardoce:
This patch addresses a rollout issue when migrating from the in-tree Barman Cloud support to the `barman-cloud` plugin, particularly when using the `switchover` primary update strategy.

During such a migration, a switchover is triggered to roll out the former primary instance. After the switchover, the former primary Pod restarts and attempts to archive any remaining WAL files in the queue before starting PostgreSQL. These files may still be present due to archiving or misconfiguration delays.

However, the previous version of the Pod does not include support for the plugin-based archiving system, causing all archiving attempts to fail during this pre-start phase. As a result, the Pod stalls indefinitely, blocking the rollout.

This fix introduces a mechanism to detect this specific failure scenario. If WAL archiving cannot proceed due to a missing plugin, the Pod exits with a distinct exit code. The operator recognises this exit code and replaces the Pod with one based on the updated container image, ensuring the plugin is available and WAL archiving can continue as expected.

Fixes: #8189

(cherry picked from commit 99550adf171efb72d806cea2210fee5698ef5386)

### DIFF
--- a/api/v1/cluster_types.go
+++ b/api/v1/cluster_types.go
@@ -112,6 +112,11 @@ const (
 	// MissingWALDiskSpaceExitCode is the exit code the instance manager
 	// will use to signal that there's no more WAL disk space
 	MissingWALDiskSpaceExitCode = 4
+
+	// MissingWALArchivePlugin is the exit code used by the instance manager
+	// to indicate that it started successfully, but the configured WAL
+	// archiving plugin is not available.
+	MissingWALArchivePlugin = 5
 )
 
 // SnapshotOwnerReference defines the reference type for the owner of the snapshot.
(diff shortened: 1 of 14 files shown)


leonardoce:
This is a cherry-pick to release-1.26 where I needed to partially back port the extension's patch.
Can I please have another pair of eyes?

flokoe:
Hi there,

it seems that one of my clusters has this exact problem. Is there a manual way to get them "unstuck"?

mnencia:
This PR has been closed because it was on the 1.26 branch, and 1.26 is not EOL.

PR #8253, opened 2025-08-05
docs: release notes for 1.27.0

gbartolini:
Closes #8231

PR #8281, opened 2025-08-07
docs: fix broken liveness-probe link in v1.27 release notes

archeearjun:
markdown Fixes an MkDocs warning: WARNING - Doc file 'release_notes/v1.27.md' contains a link 'instance_manager.md#liveness-probe', but the target 'release_notes/instance_manager.md' is not found. Changed the path to `../instance_manager.md#liveness-probe`, which renders correctly. Built locally with `mkdocs serve`; warning is gone.


gbartolini:
There are actually two links, not just one. But it has already been addressed in #8253 which will be merged before the release.

PR #8308, opened 2025-08-11
chore(controller_backup): add VolumeSnapshotReconciler to the struct

armru:
The volumesnapshot.Reconciler was previously created on-the-fly inside the reconcileSnapshotBackup method, worsening the overall performance and making the BackupReconciler harder to test in isolation.

Low priority chore


armru:
closing, integrating in the backup_controller refactor PR

PR #8320, opened 2025-08-11
refactor(backup_controller): simplify flow

armru:


internal/controller/backup_controller.go
```diff
@@ -362,28 +365,36 @@ func (r *BackupReconciler) getCluster(
...
+	if flagErr := resourcestatus.FlagBackupAsFailed(ctx, r.Client, backup, nil,
+		fmt.Errorf("while getting cluster %s: %w", clusterName, err)); flagErr != nil {
+		contextLogger.Error(flagErr, "while flagging backup as failed, retrying...")
+		return nil, flagErr
 	}
 
-	_ = resourcestatus.FlagBackupAsFailed(ctx, r.Client, backup, nil,
-		fmt.Errorf("while getting cluster %s: %w", clusterName, err))
 	r.Recorder.Eventf(backup, "Warning", "FindingCluster",
 		"Error getting cluster %v, will not retry: %s", clusterName, err.Error())
```
armru: previously this event emitter was very misleading, given that the backup could have potentially not be flagged as failed the reconciler would have retried (eventually)

mnencia (approved):

leonardoce:
After an informal discussion with the other maintainers, we're backporting this patch to release-1.27.
This happens because otherwise it would be difficult to backport subsequent bug fixes on this part of the controller.

leonardoce (approved):

PR #8325, opened 2025-08-12
Fixes Issue #8299 [Feature]: Unsupervised Pooler PAUSE during switchover

### LINKED ISSUE #8299: [Feature]: Unsupervised Pooler `PAUSE` during switchover

It is implied that the `Pooler` will not `PAUSE` during a switchover, potentially leading to transaction failures, instead of transaction delays.

For the `Pooler`s to `PAUSE` while a switchover is happening.

N/A

### Additional context

I saw this comment in the documentation:

> In future versions, the switchover operation will be fully integrated with the PgBouncer pooler and take advantage of the PAUSE/RESUME features to reduce the perceived downtime by client applications. Currently, you can achieve the same results by setting the paused attribute to true, issuing the switchover command through the [cnpg plugin](https://cloudnative-pg.io/documentation/current/kubectl-plugin/#promote), and then restoring the paused attribute to false.

However, I could not find a feature request associated with this, using keywords `Pooler`, `switchover`, and `pause`.

Thus, here I have a feature request to add this, as this interests me greatly, to achieve at least the illusion of "zero-downtime" in periodic switchover operations (during cluster upgrades, rollout of upgrades, etc.)

### Backport?

N/A

No

### Code of Conduct

omsuneri:
resolves #8299 
@sxd @gbartolini review please !!

### DIFF
--- a/internal/controller/cluster_controller.go
+++ b/internal/controller/cluster_controller.go
@@ -668,6 +668,10 @@ func (r *ClusterReconciler) handleSwitchover(
 		return &ctrl.Result{Requeue: true}, nil
 	}
 	if selectedPrimary != "" {
+		// If we selected a new primary, pause poolers to reduce perceived downtime
+		if err := r.pausePoolersDuringSwitchover(ctx, cluster); err != nil {
+			contextLogger.Error(err, "Failed to pause poolers during switchover, continuing anyway")
+		}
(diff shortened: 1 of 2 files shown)


kyrofa:
Would very much like this feature, I don't understand why it's so hard to get a review in this project.

Issue #8022, opened 2025-07-12
[Feature]: Add configurable replica backup method preference


martinohansen:
### Is there an existing issue already for this feature request/idea?

- [x] I have searched for an existing issue, and could not find anything. I believe this is a new feature request to be evaluated.

### What problem is this feature going to solve? Why should it be added?

Operator continues to use old volume snapshot backups for creating new replicas, even after I've removed the volume snapshot configuration from my cluster spec. When I try to scale up my cluster (increase replicas), the operator prioritizes existing volume snapshot backups over barman, that is not always preferable if the goal is to move to another storage cluster where the snapshots wont be available. 

### Describe the solution you'd like

I want to be able to explicitly configure which backup method the operator should prefer when creating new replicas through a new field in the cluster's backup configuration. The solution should:

1. Add a replicaMethodPreference field to the BackupConfiguration spec
2. Support two options: volumeSnapshot (default for backward compatibility) and barmanObjectStore
3. Include automatic fallback behavior - if the preferred method is unavailable, fall back to the other option
4. Only affect replica creation, not regular backup operations

Example usage:
```yaml
apiVersion: postgresql.cnpg.io/v1
kind: Cluster
spec:
  backup:
    replicaMethodPreference: "barmanObjectStore" 
    barmanObjectStore: {...}
    volumeSnapshot: {...}
```

### Describe alternatives you've considered

A workaround is to delete all the volumesnapshots before creating the replica

### Backport?

No

### Are you willing to actively contribute to this feature?

Yes

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #8328, opened 2025-08-12
feat: replica method preference option

martinohansen:
Add `replicaMethodPreference` to the  backup config in the cluster spec 
which allows users to control which backup method is preferred when 
creating new replicas. Useful if you are moving clusters between storage
backends where snapshots wont work. Like before, the default prefers 
VolumeSnapshot.

Signed-off-by: Martin Hansen <dontbeevilpls@gmail.com>


martinohansen:
Closes #8022

PR #8392, opened 2025-08-18
Add timeline metrics to pg_replication

### LINKED ISSUE #8391: [Feature]: Add timeline to default metrics

I've been running CNPG in production for about 6 months now. It's been excellent, though about four times now I have ended up with a replica falling out of sync with the primary (I assume due to drains/maintenance/etc.). In these situations I need to remove the pod/pvc and let the operator create a new one, but I'm usually alerted to the fact that I need to do that because the database goes down: the WAL fills up its disk.

I have finally found the thing to monitor to alert me to this situation BEFORE the database goes down: the timeline ID (`SELECT timeline_id FROM pg_control_checkpoint()`). I think it would be a useful thing to have as part of the default metrics. I imagine `SELECT received_tli FROM pg_stat_wal_receiver` would be useful as well. I'd be happy to propose a PR for this.

I am currently utilizing user-defined metrics to obtain these for all of my clusters. I'm happy with that solution, but I suspect I'm not the only one who would like to be monitoring this, which is why I argue for adding it to the default set.

### Additional context

_No response_

### Backport?

No

Yes

### Code of Conduct

kyrofa:
Today there aren't any default metrics that clearly indicate that a replica has fallen out of sync with the primary. This PR resolves #8391 by adding both `timeline_id` and `received_timeline_id` metrics to `pg_replication`.

### DIFF
--- a/config/manager/default-monitoring.yaml
+++ b/config/manager/default-monitoring.yaml
@@ -117,7 +117,12 @@ data:
               END AS lag,
               pg_catalog.pg_is_in_recovery() AS in_recovery,
               EXISTS (TABLE pg_stat_wal_receiver) AS is_wal_receiver_up,
-              (SELECT count(*) FROM pg_catalog.pg_stat_replication) AS streaming_replicas"
+              (SELECT count(*) FROM pg_catalog.pg_stat_replication) AS streaming_replicas,
+              (SELECT timeline_id FROM pg_control_checkpoint()) AS timeline_id,
+              (SELECT CASE WHEN EXISTS (TABLE pg_stat_wal_receiver)
+              THEN (SELECT received_tli FROM pg_stat_wal_receiver)
+              ELSE 0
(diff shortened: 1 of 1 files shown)


jsilvela (approved):
tested, and seems useful

xvzf:
Any chance to get this in @jsilvela?

jsilvela:
Eh, the maintainers need to decide ...
Feel free to upvote this.

kyrofa:
Your regex is wrong. This commit uses `resolve` singular, which is a [perfectly valid way to close issues](https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword). Not that it matters, the issue has already been closed because no one reviews anything around here :shrug: .

kyrofa:
I'm just going to close this. There's no point.

Issue #8263, opened 2025-08-06
[Feature]: Add image version info to backup CR


hh24k:
### Is there an existing issue already for this feature request/idea?

- [x] I have searched for an existing issue, and could not find anything. I believe this is a new feature request to be evaluated.

### What problem is this feature going to solve? Why should it be added?

In cases like we have multiple backups performed on different minor/major postgresql images available for a cnpg cluster, while choosing backups for restoration, we would like to know the image version information from the source cluster where the backup was taken so that we can choose a proper target image for restoration.

### Describe the solution you'd like

Probably, we can utilize the `cluster.status.PGDataImageInfo`, add it to the backup in a proper way

### Describe alternatives you've considered

Impl. an operator to add `cluster.status.PGDataImageInfo` properly to the backup

### Backport?

No

### Are you willing to actively contribute to this feature?

Yes

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

omsuneri:
@hh24k I will work on this issue

PR #8464, opened 2025-08-28
fix(backup,status): add `majorVersion` to backups

### LINKED ISSUE #7705: [Bug]: Errors scaling up after an in-place major upgrade when VolumeSnapshots backup are available

### Contact Details

_No response_

### Version

1.26 (latest patch)

1.33

YAML manifest

### What happened?

When valid volume snapshot backups are present, after a successful major version upgrade, the creation of the replicas tries to use the snapshots taken before the major upgrade, leading to a failure:

```
{
  "level": "info",
  "ts": "2025-05-29T20:16:42.497601464Z",
  "logger": "postgres",
  "msg": "2025-05-29 20:16:42 UTC FATAL:  database files are incompatible with server",
  "pipe": "stderr",
  "logging_pod": "cluster-example-4"
}
{
  "level": "info",
  "ts": "2025-05-29T20:16:42.497639304Z",
  "logger": "postgres",
  "msg": "2025-05-29 20:16:42 UTC DETAIL:  The data directory was initialized by PostgreSQL version 16, which is not compatible with this version 17.5.",
  "pipe": "stderr",
  "logging_pod": "cluster-example-4"
}
```

### Cluster resource

```shell

```

### Relevant log output

```shell

```

### Code of Conduct

armru:
Expose the PostgreSQL major version in the backup status and add a new label `cnpg.io/majorVersion` to volume snapshots.

This change also fixes a bug during major upgrades, preventing backups from being taken.

Partially closes #7705

### DIFF
--- a/api/v1/backup_types.go
+++ b/api/v1/backup_types.go
@@ -195,8 +195,8 @@ type BackupStatus struct {
 	// The potential credentials for each cloud provider
 	BarmanCredentials `json:",inline"`
 
-	// The major version of PostgreSQL of the cluster where the backup
-	// has been taken from.
+	// The PostgreSQL major version that was running when the
+	// backup was taken.
 	MajorVersion int `json:"majorVersion,omitempty"`
 
 	// EndpointCA store the CA bundle of the barman endpoint.
--- a/api/v1/backup_types.go
+++ b/api/v1/backup_types.go
@@ -195,6 +195,10 @@ type BackupStatus struct {
 	// The potential credentials for each cloud provider
 	BarmanCredentials `json:",inline"`
 
+	// The major version of PostgreSQL of the cluster where the backup
+	// has been taken from.
+	MajorVersion int `json:"majorVersion,omitempty"`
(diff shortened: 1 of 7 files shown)


hh24k:
Hi @armru @mnencia I have a quite related request here https://github.com/cloudnative-pg/cloudnative-pg/issues/8263, would be better if we can step a bit forward, adding the imageInfo which not only includes major version but also the imageRef where the backup was taken, WDYT?

mnencia (approved):

mnencia:
> Hi @armru @mnencia I have a quite related request here #8263, would be better if we can step a bit forward, adding the imageInfo which not only includes major version but also the imageRef where the backup was taken, WDYT?

The image alone is insufficient to determine the major version when using an ImageCatalog. I recommend omitting the image from this PR and discussing this topic further in the other PR.

mnencia (approved):

gbartolini (approved):

PR #8475, opened 2025-08-29
fix(instance,scaleup): avoid using snapshots if they could be of an older majorVersion

### LINKED ISSUE #7705: [Bug]: Errors scaling up after an in-place major upgrade when VolumeSnapshots backup are available

armru:
Fixes a bug during major upgrades where a volume snapshot backup from a previous minor version could have been incorrectly used to optimise the recreation of replicas.

Depends On #8464  
Closes #7705

### DIFF
--- a/docs/src/appendixes/backup_volumesnapshot.md
+++ b/docs/src/appendixes/backup_volumesnapshot.md
@@ -115,14 +115,6 @@ defined in your PostgreSQL cluster.
 Once a cluster is defined for volume snapshot backups, you need to define
 a `ScheduledBackup` resource that requests such backups on a periodic basis.
 
-!!! Note
-    When creating replicas from VolumeSnapshot backups, CloudNativePG enforces
-    PostgreSQL major version compatibility. Newer backups include the
-    `status.majorVersion` field; for older snapshot backups that don’t, set the
-    `cnpg.io/backupMajorVersion` annotation (integer) on the `Backup` so the operator
-    can select a compatible snapshot. See [Labels and annotations](../labels_annotations.md)
-    for details about `cnpg.io/backupMajorVersion`.
-
 ## Hot and cold backups
 
 !!! Warning
--- a/docs/src/appendixes/backup_volumesnapshot.md
+++ b/docs/src/appendixes/backup_volumesnapshot.md
@@ -115,6 +115,14 @@ defined in your PostgreSQL cluster.
 Once a cluster is defined for volume snapshot backups, you need to define
 a `ScheduledBackup` resource that requests such backups on a periodic basis.
 
+!!! Note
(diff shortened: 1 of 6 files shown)


mnencia (approved):

gbartolini (approved):

mofirouz:
Apologies for waking this up - however we've just begun testing and hit the issue where this ticket is attempting to fix. However, I don't think the fix here is sufficient specially for large databases.

When creating replicas it decides how to do it in following [code](https://github.com/cloudnative-pg/cloudnative-pg/blob/2770d1270db50795a0fd0466e5a29eeb8ddb43c7/internal/controller/cluster_create.go#L1248-L1254):

```
	job := specs.JoinReplicaInstance(*cluster, nodeSerial)

	// If we can bootstrap this replica from a pre-existing source, we do it
	storageSource := persistentvolumeclaim.GetCandidateStorageSourceForReplica(ctx, cluster, backupList)
	if storageSource != nil {
		job = specs.RestoreReplicaInstance(*cluster, nodeSerial)
	}
```

where `GetCandidateStorageSourceForReplica` looks for snapshot backups or other sources to init replica. Failing that (and this commit will make sure it fails to find previous snapshots due to PG version mismatch) it uses default `JoinReplicaInstance`  init. Default init is barebones postgresql replication setup where replica is given credentials of master, connects there and downloads whole database with `pg_basebackup`.

This default method won't work because of database sizes.

Proper way of doing it would be to take backup of primary after upgrade using snapshot method and only then create replica, but it doesn't seem to do it. So for big enough instances we'll need to have custom upgrade procedure regardless of CNPG version:

- scale cluster to 1
- PG upgrade
- take backup with snapshot method
- scale cluster back to original size

I believe the custom upgrade procedure described above should be the proper fix. Does this make sense?

PR #8476, opened 2025-08-29
chore(plugin): clarify primary promotion time in status command

pchovelon:
In the plugin status output, there is a line called "Primary start time". I found it quite confusing because it doesn't represent what it pretend to be. It's actually the time when the current primary instance became primary but when it's restart time.

Example : 
```
Primary instance:    postgresql-enabled-2
Primary start time:  2025-08-29 09:02:43 +0000 UTC (uptime 4m50s)
```

I understand that my instance postgresql-enabled-2 is primary and (re)started almost 5m ago.

But the actual restart time of the PostgreSQL instance/postmaster is not that at all (and can be days, weeks ago or even more) :
```
kubectl exec -it postgresql-enabled-2 -- psql -c "select pg_postmaster_start_time()"
Defaulted container "postgres" out of: postgres, bootstrap-controller (init)
   pg_postmaster_start_time    
-------------------------------
 2025-08-29 08:49:26.845241+00
(1 row)
```

So I suggest to change the word used in the output.

### DIFF
--- a/internal/cmd/plugin/status/status.go
+++ b/internal/cmd/plugin/status/status.go
@@ -281,10 +281,10 @@ func (fullStatus *PostgresqlStatus) printBasicInfo(ctx context.Context, k8sClien
 	case isPrimaryFenced:
 		summary.AddLine("Status:", aurora.Red("Primary instance is fenced"))
 	default:
-		// Avoid printing the start time when hibernated or fenced
-		primaryStartTime := getPrimaryStartTime(cluster)
-		if len(primaryStartTime) > 0 {
-			summary.AddLine("Primary since:", primaryStartTime)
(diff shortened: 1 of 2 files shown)


jsilvela:
You bring a good point. Now, looking at the "uptime" computation, and now it makes no sense to me.
If the timestamp is the time at which the current primary became primary, then the uptime computed as the time since then to current time, is measuring how long it's been primary, not how long it's been up.
Worth fixing too, I'd say. And perhaps worth a new line of status, since we're really talking of two different things?

pchovelon:
Yes, you are right uptime should be changed as well !

gbartolini (approved):

leonardoce (approved):

mnencia (approved):

Issue #7807, opened 2025-06-11
[Feature]: access to db from another namespace


gecube:
### Is there an existing issue already for this bug?

- [x] I have searched for an existing issue, and could not find anything. I believe this is a new bug.

### I have read the troubleshooting guide

- [x] I have read the troubleshooting guide and I think this is a new bug.

### I am running a supported version of CloudNativePG

- [x] I have read the troubleshooting guide and I think this is a new bug.

### Contact Details

gb12335@gmail.com

### Version

1.26 (latest patch)

### What version of Kubernetes are you using?

1.33

### What is your Kubernetes environment?

Self-managed: kind (evaluation)

### How did you install the operator?

YAML manifest

### What happened?

Good day !

I accidentally found that it is tricky to access db from another namespace as secret does not contain fqdn - preferably with .svc.cluster.local. suffix or appropriate db suffix.

The documentation supposes that the operator of the cluster should create postgres hostname manually which sounds ridiculous. 

https://cloudnative-pg.io/documentation/current/applications/

Expected solution:
- put an internal fqdn with k8s in postgresql secret in host and uri fields.
- for every additional lb - as I want to access the cluster with ease, generate two secrets - for internal access and for external access. So as an user of cnpg operator I would be able to replicate secret outside or just grab the jdbc uri and manually paste to the application config.

### Cluster resource

```shell

```

### Relevant log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

armru:
Hello, I've seen that this is labeled as bug, could you please clarify what the operator states do vs what is happening? 

If I understand correctly this is more of a feature request but I could be wrong

gecube:
@armru Hi! Thanks for the attention. I think it depends on what was mentioned by authors. If the namespace should be shown in secret - it means - bug. If no - feature request.

I found what is disappointing me.

Let's say, we have a secret from cnpg.

The `uri` would look like `postgresql://app:******@top1-rw.postgresql:5432/app` - the namespace is present

But if I would check the `host` field: `top1-rw` (without namespace).

If the application for different reasons could not consume `uri` or `jdbc-uri` I am in deadlock situation. I can't use directly `host` as I need to enrich it with namespace.

[...]

PR #8532, opened 2025-09-05
feat(secret): secret to expose fully qualified `host`

wingsofovnia:
https://github.com/cloudnative-pg/cloudnative-pg/issues/7807 brought a issue that I faced with my deployment: accessing the instance across namespaces. However, I find the resolution #7852 lacking as it does not directly address the main issue.

Use case:
* `Cluster` deployed in a namespace `A`.
* A resource `R` that needs to access it is in a namespace `B`.
* `R` uses `Cluster`'s secret to get connection details (via [`NamespacedName`](https://pkg.go.dev/k8s.io/apimachinery/pkg/types)).
* `R` reads `host`, `port`, `user`, and `password` to connect to the instance.

This doesn't work because `Secret#host` is not fully qualified and the connection doesn't get through. One might say to combine data from the secret myself but it becomes cumbersome with 7645988 enabled, it is no longer `${host}.svc.cluster.local` and there is no more transparent way to get the host (except parsing `fqdn-uri` or `fqdn-jdbc-uri` which I'd call transparent).

This change makes `Secret#host` to contain fully qualified host and keeps local hostname in `Secret#hostname`. 

The change should not have any effect on users that use this field directly, but might break some brittle code that have worked around the issue by adding `.svc.cluster.local` suffix.

Relates to 7645988
Relates to #7807

### DIFF
--- a/docs/src/applications.md
+++ b/docs/src/applications.md
@@ -64,18 +64,19 @@ every PostgreSQL cluster it deploys:
 
 Each secret contain the following:
 
-* username
-* password
-* hostname to the RW service
-* port number
-* database name
-* a working [`.pgpass file`](https://www.postgresql.org/docs/current/libpq-pgpass.html)
-* [uri](https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNSTRING)
-* [jdbc-uri](https://jdbc.postgresql.org/documentation/use/#connecting-to-the-database)
-* [fqdn-uri](https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNSTRING)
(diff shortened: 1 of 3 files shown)


ahachmann:
We would love to see this feature to be merged soon. We have to work arround this issue in our GitOps at the moment, and this solution would be very convenient! 

Thanks @wingsofovnia for your effort!!

mmoreiradj:
Would love to have this merged, we currently need this for a Kubernetes operator.

:pray:

wingsofovnia:
@jsilvela it's been almost a year, can someone from the team be assigned to this PR?

jasoncouture:
Please? I'm so sick of this spamming my DNS search path :|

PR #8536, opened 2025-09-05
feat: dedicated phase for hibernated clusters

### LINKED ISSUE #3390: [Feature]: Dedicated phase for hibernated clusters

The operator, when performs the hibernations on clusters using [Declarative Hibernation](https://cloudnative-pg.io/documentation/1.21/declarative_hibernation/) keeps the `status.phase` as `Cluster in healthy state` once the hibernation get completed successfully.

It would be useful to have a dedicated phase for it, like `Cluster hibernated`.

It can also be useful having a dedicated phase when the annotation is removed or set to `off` and the clusters get [rehydrated](https://cloudnative-pg.io/documentation/1.21/declarative_hibernation/#rehydration).

### Additional context

```
gabrielefedi cloudnative-pg >>  (main) k get cluster
NAME              AGE    INSTANCES   READY   STATUS                     PRIMARY
cluster-example   4m9s   3           3       Cluster in healthy state   cluster-example-1

k annotate cluster cluster-example "cnpg.io/hibernation=on"

gabrielefedi cloudnative-pg >>  (main) k get cluster cluster-example -o yaml | yq .status.conditions
...
- lastTransitionTime: "2023-11-20T21:15:38Z"
  message: Cluster has been hibernated
  reason: Hibernated
  status: "True"
  type: cnpg.io/hibernation

gabrielefedi cloudnative-pg >>  (main) k get cluster
NAME              AGE   INSTANCES   READY   STATUS                     PRIMARY
cluster-example   13m   3                   Cluster in healthy state   cluster-example-1
```

### Backport?

Yes

### Code of Conduct

GabriFedi97:
Sets the phase `Cluster is hibernated` for hibernated clusters when all instance Pods have been successfully removed

closes #3390

### DIFF
--- a/internal/controller/cluster_status.go
+++ b/internal/controller/cluster_status.go
@@ -266,12 +266,6 @@ func (r *ClusterReconciler) updateResourceStatus(
 		resources.instances.Items,
 	)
 
-	if meta.IsStatusConditionTrue(cluster.Status.Conditions, hibernation.HibernationConditionType) {
-		if err := r.RegisterPhase(ctx, cluster, apiv1.PhaseHibernated, "Instance Pods have been removed"); err != nil {
-			return err
-		}
-	}
-
 	// Count jobs
 	newJobs := int32(len(resources.jobs.Items)) //nolint:gosec
 	cluster.Status.JobCount = newJobs
--- a/internal/controller/cluster_status.go
+++ b/internal/controller/cluster_status.go
@@ -266,6 +266,12 @@ func (r *ClusterReconciler) updateResourceStatus(
 		resources.instances.Items,
 	)
 
+	if meta.IsStatusConditionTrue(cluster.Status.Conditions, hibernation.HibernationConditionType) {
+		if err := r.RegisterPhase(ctx, cluster, apiv1.PhaseHibernated, "Instance Pods have been removed"); err != nil {
+			return err
+		}
+	}
+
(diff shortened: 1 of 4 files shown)


internal/controller/cluster_status.go
```diff
@@ -266,6 +266,12 @@ func (r *ClusterReconciler) updateResourceStatus(
 		resources.instances.Items,
 	)
 
+	if meta.IsStatusConditionTrue(cluster.Status.Conditions, hibernation.HibernationConditionType) {
+		if err := r.RegisterPhase(ctx, cluster, apiv1.PhaseHibernated, "Instance Pods have been removed"); err != nil {
+			return err
+		}
+	}
+
```
armru: One considerations:

- I'm not sure that this is the correct place to declare this phase, kinda unexpected.

And a question:
- This doesn't get evaluated anywhere else in the code so it could be overwritten with another phase and then it could flip it back and forth with the next reconciliations, I'am I correct to assume that?

GabriFedi97: I think you're right. Even if currently there's nothing overriding that phase if the cluster is hibernated, we cannot guarantee that will still be the case as the codebase changes. 
Let me see if I can find a better place

GabriFedi97: @armru what about now?

PR #8604, opened 2025-09-19
chore(lint): lint import aliases

mnencia:
The main goal of this patch is to eliminate all unqualified `v1` aliases.

In multiple places within the code, we import API packages that end with `v1`, all of which are aliased as `v1`. When reading the code, encountering `v1.XxxYyyZzz(...)` makes it unclear which package's functions are being invoked.

In contrast, we have a more consistent naming model in other places, where we use more descriptive names such as `appsv1`, `metav1`, or `corev1`.

This patch implements a linter for the imports of these packages, ensuring that they are consistently imported using the longer, more explanatory names.

A comment in the patch highlights the relevant section.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -52,6 +52,40 @@ linters:
         - '-QF1007'
     lll:
       line-length: 120
+    importas:
+      alias:
+        - pkg: github.com/cloudnative-pg/cloudnative-pg/api/v1
+          alias: apiv1
+        - pkg: github.com/cloudnative-pg/cloudnative-pg/internal/webhook/v1
+          alias: webhookv1
+        - pkg: github.com/kubernetes-csi/external-snapshotter/client/v8/apis/volumesnapshot/v1
+          alias: volumesnapshotv1
+        - pkg: github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1
+          alias: monitoringv1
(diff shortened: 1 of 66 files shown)


.golangci.yml
```diff
@@ -52,6 +52,40 @@ linters:
...
+        - pkg: k8s.io/api/rbac/v1
+          alias: rbacv1
+        - pkg: k8s.io/api/storage/v1
+          alias: storagev1
+        - pkg: k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1
+          alias: apiextensionsv1
+        - pkg: k8s.io/apimachinery/pkg/apis/meta/v1
+          alias: metav1
+        - pkg: k8s.io/client-go/kubernetes/typed/core/v1
+          alias: corev1client
```
mnencia, line 88: This list is the TL;DR version of the patch ;-)

mnencia: Regarding the last item, it was originally named `typedcorev1`, but I renamed it to `corev1client` because I find it more descriptive and it is a widely used name in other projects.

mnencia: Regarding `volumesnapshotv1`, it was previously imported as `volumesnapshot` or `storagesnapshotv1`. I chose `volumesnapshotv1` for consistency with other imports.

leonardoce (approved):

armru (approved):

Issue #8508, opened 2025-09-01
[Bug]: Backups are still tried for hibernated clusters


simonapencea:
### Is there an existing issue already for this bug?

- [x] I have searched for an existing issue, and could not find anything. I believe this is a new bug.

### I have read the troubleshooting guide

- [x] I have read the troubleshooting guide and I think this is a new bug.

### I am running a supported version of CloudNativePG

- [x] I have read the troubleshooting guide and I think this is a new bug.

### Contact Details

simona@xata.io

### Version

1.26 (latest patch)

### What version of Kubernetes are you using?

1.33

### What is your Kubernetes environment?

Other

### How did you install the operator?

YAML manifest

### What happened?

We are creating Clusters and for each cluster we also create a ScheduledBackup to provide continuous backups.
Sometimes we want to hibernate those clusters. 

What happened:

* when we use the provided annotation here:
https://cloudnative-pg.io/documentation/1.26/declarative_hibernation/#hibernation
to hibernate the cluster, the ScheduledBackup remains active. 
This means it will still trigger a full backup according to the schedule and if at that time the cluster is hibernated, the full backup fails. It fails because it cannot find the pod, which is normal for a hibernated cluster. But it is still a failing backup that needs to be investigated - it will trigger an alarm, page someone, we see it in the logs etc.

What we expected to happen: 
* the ScheduledBackup gets suspended when the cluster gets hibernated. There is this annotation https://cloudnative-pg.io/documentation/1.26/backup/#pause-scheduled-backups - but it is not automatically set in tandem with the cluster hibernation.

### Cluster resource

```shell

```

### Relevant log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

simonapencea:
To be clear, we did implement our own automation around it, but it felt to us that this should be a part of the operator.

eminano:
>This is not currently documented as a bug or limitation, but rather as the designed separation between cluster hibernation and backup scheduling—there is no built-in automation to link the two features [source](https://cloudnative-pg.io/documentation/1.26/declarative_hibernation/).

[...]

PR #8656, opened 2025-09-24
Check if cluster is hibernated in scheduledbackup_controller

pchovelon:
Try to fix this issue https://github.com/cloudnative-pg/cloudnative-pg/issues/8508. But think also that it needs discussion :)

The idea is to check if the Cluster is hibernated or not before creating the backup.
If this is the case, backup is not created, a message appears in log ("Skipping scheduled backup of a hibernated cluster) and then a retry will be made 24 hours later.

No documentation yet. Waiting for discussion before making it.

### DIFF
--- a/internal/controller/scheduledbackup_controller.go
+++ b/internal/controller/scheduledbackup_controller.go
@@ -222,6 +222,19 @@ func ReconcileScheduledBackup(
 		return ctrl.Result{RequeueAfter: nextTime.Sub(now)}, nil
 	}
 
+	var cluster apiv1.Cluster
+	if err := cli.Get(ctx, client.ObjectKey{
+		Namespace: scheduledBackup.Namespace,
+		Name:      scheduledBackup.Spec.Cluster.Name,
+	}, &cluster); err != nil {
+		// if cluster is hibernated, don't create new scheduled backup
+		clusterHibernationAnnotation, ok := cluster.GetAnnotations()[utils.HibernationAnnotationName]
+		if !ok || clusterHibernationAnnotation == "" {
+			contextLogger.Info("Skipping scheduled backup of a hibernated cluster", "next", nextTime)
+			return ctrl.Result{RequeueAfter: 24 * time.Hour}, nil
+		}
+	}
+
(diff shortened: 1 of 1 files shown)


internal/controller/scheduledbackup_controller.go
```diff
@@ -222,6 +222,19 @@ func ReconcileScheduledBackup(
...
+		Name:      scheduledBackup.Spec.Cluster.Name,
+	}, &cluster); err != nil {
+		// if cluster is hibernated, don't create new scheduled backup
+		clusterHibernationAnnotation, ok := cluster.GetAnnotations()[utils.HibernationAnnotationName]
+		if !ok || clusterHibernationAnnotation == "" {
+			contextLogger.Info("Skipping scheduled backup of a hibernated cluster", "next", nextTime)
+			return ctrl.Result{RequeueAfter: 24 * time.Hour}, nil
+		}
+	}
+
```
armru, line 237: First of all thanks for the PR! :)

Some observations:

- I think there is a typo if I read correctly, the cluster is checked when err != nil
- even if the err would be nil it would still pose a change of behaviour given that we allow the creation of backup resources when a cluster does not exist
- the RequeueAfter of 24 hours poses a problem, what if I have a scheduledbackup running for example every hour? There is no guarantee that an hibernation will last 24hours (still the scheduledbackup could be retried earlier if any modification happens on the resource)
-
```
	if scheduledBackup.Status.LastCheckTime == nil && scheduledBackup.IsImmediate() {
		// we populate the status (lastCheckTime...) by following the same rules of the scheduled backup
		event.Eventf(scheduledBackup, "Normal", "BackupSchedule", "Scheduled immediate backup now: %v", now)
		return createBackup(ctx, event, cli, scheduledBackup, now, now, schedule, true)
	}
```
Is not covered by this logic.
- The info message would make it very sneaky to catch	
- The hibernation annotation to be enabled should be set to 'on'

pchovelon:
Thanks @armru .
Indeed my pull request was not the best one :) Thank for the time you took.
I saw what you did in backup_controller.go (version 1.27.1). Clever way for sure.

Need to practice and understand more the operator code :)

PR #8663, opened 2025-09-25
fix: wait for app user secret before starting initdb

### LINKED ISSUE #6069: [Bug]: initdb doesn't wait for secret

armru:
The scope of this PR is to address #6069 with minimally invasive changes

Closes #6069

### DIFF
--- a/internal/controller/cluster_create.go
+++ b/internal/controller/cluster_create.go
@@ -1153,6 +1153,21 @@ func (r *ClusterReconciler) createPrimaryInstance(
 		job = specs.CreatePrimaryJobViaPgBaseBackup(*cluster, nodeSerial)
 
 	default:
+		appSecretName := cluster.GetApplicationSecretName()
+		if appSecretName != "" && cluster.ShouldInitDBCreateApplicationDatabase() {
+			var appSecret corev1.Secret
+			key := client.ObjectKey{Namespace: cluster.Namespace, Name: appSecretName}
+			if err := r.Get(ctx, key, &appSecret); err != nil {
+				if apierrs.IsNotFound(err) {
+					contextLogger.Warning("Waiting for application secret to be created")
+					r.Recorder.Eventf(cluster, "Warning", "WaitingForSecret",
+						"Waiting for application secret %s to exist before initdb", appSecretName)
+					return ctrl.Result{RequeueAfter: 10 * time.Second}, nil
+				}
+				return ctrl.Result{}, fmt.Errorf("cannot get application secret: %w", err)
+			}
+		}
+
 		r.Recorder.Event(cluster, "Normal", "CreatingInstance", "Primary instance (initdb)")
 		job = specs.CreatePrimaryJobViaInitdb(*cluster, nodeSerial)
 	}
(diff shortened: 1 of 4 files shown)


internal/controller/cluster_create.go
```diff
@@ -1153,6 +1153,20 @@ func (r *ClusterReconciler) createPrimaryInstance(
...
 	default:
+		appSecretName := cluster.GetApplicationSecretName()
+		if appSecretName != "" && cluster.ShouldInitDBCreateApplicationDatabase() {
+			var appSecret corev1.Secret
+			key := client.ObjectKey{Namespace: cluster.Namespace, Name: appSecretName}
+			if err := r.Get(ctx, key, &appSecret); err != nil {
+				if apierrs.IsNotFound(err) {
+					r.Recorder.Eventf(cluster, "Normal", "WaitingForSecret",
+						"Waiting for application secret %s to exist before initdb", appSecretName)
+					return ctrl.Result{RequeueAfter: 10 * time.Second}, nil
```
armru: we could add a phase but I'm making this minimally invasive

pkg/specs/jobs.go
```diff
@@ -385,6 +386,22 @@ func CreatePrimaryJob(cluster apiv1.Cluster, nodeSerial int, role jobRole, initC
...
+			ValueFrom: &corev1.EnvVarSource{
+				SecretKeyRef: &corev1.SecretKeySelector{
+					LocalObjectReference: corev1.LocalObjectReference{Name: cluster.GetApplicationSecretName()},
+					Key:                  "username",
+					Optional:             ptr.To(false),
+				},
+			},
+		})
+	}
+
```
armru, line 404: Potentially we could use the container envfrom field to mount the whole secret, but I felt like for what we want to achieve this minimalistic approach that exposes only the username to the job is the best

dominik-niebuhr:
do we have any ETA for a release once this is merged? :) this would help us remove a very ugly workaround

leonardoce (approved):

mnencia (approved):

dominik-niebuhr:
@armru Will this fix also work for secrets of managed roles? If not should I open a new issue or do we keep the existing issue open for that?

PR #8671, opened 2025-09-26
feat(pgbouncer): simplify authentication using `auth_dbname`

### LINKED ISSUE #1383: Introduce support for auth_dbname in pgbouncer

leonardoce:
Previously, authentication was handled by creating a `user_search` function in every accessible database and granting `CONNECT` privileges to the PgBouncer `auth_user`.

This change leverages the `auth_dbname` option, so PgBouncer runs its authentication queries only against the `postgres` database.

PgBouncer 1.19+ is required (`auth_dbame` option needed).

Benefits:

- simpler reconciliation loop
- no extra connections required for other databases

Closes: #1383

### DIFF
--- a/api/v1/pooler_types.go
+++ b/api/v1/pooler_types.go
@@ -43,6 +43,9 @@ const (
 
 	// DefaultPgBouncerPoolerAuthQuery is the default auth_query for PgBouncer
 	DefaultPgBouncerPoolerAuthQuery = "SELECT usename, passwd FROM public.user_search($1)"
+
+	// PoolerAuthDBName is the database name used to run the auth_query
+	PoolerAuthDBName = "postgres"
 )
 
 // PgBouncerPoolMode is the mode of PgBouncer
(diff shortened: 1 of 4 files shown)


gbartolini (approved):

armru (approved):

hyat7:
This should perhaps be documented in release notes as being a Change instead of Enhancement, or something? Maybe I'm just some edge case, but I'm using externalClusters to behave as a read-only replica from a non-CNPG postgres, and then having CNPG Pooler against this CNPG-owned replica. As such, it couldn't automatically add this function to the `postgres` DB.

It was unclear what the problem was at first, because here is no mention in this new documentation about migration, and the previous documentation said just `GRANT CONNECT ON DATABASE { database name here } TO cnpg_pooler_pgbouncer;` -- so unless people thought "I'll add this function and privileges to `postgres` DB for no reason", they'll encounter the same thing.

Thanks!

Issue #8675, opened 2025-09-28
[Chore]: Deprecated `authQuerySecret`


gbartolini:
### Is there an existing issue already for this feature request/idea?

- [x] I have searched for an existing issue, and could not find anything. I believe this is a new feature request to be evaluated.

### What problem is this feature going to solve? Why should it be added?

#8672  will supersede `authQuerySecret`.

### Describe the solution you'd like

`authQuerySecret` should be marked as deprecated, with admission webhooks to guide users during the transition.

### Describe alternatives you've considered

N/A

### Additional context

See #8673 

### Backport?

No

### Are you willing to actively contribute to this feature?

Yes

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #8678, opened 2025-09-29
chore: rename `authQuerySecret` to `serverSecret`

leonardoce:
This patch introduces a new field, `.spec.pgbouncer.serverSecret`, which can reference either a TLS secret or a BasicAuth secret:

- TLS secret: mapped to the `server_tls_key_file` and `server_tls_cert_file` parameters in the PgBouncer configuration. These files are used by PgBouncer to authenticate to the PostgreSQL server and by PostgreSQL to validate the provided certificate.

- BasicAuth secret: provides the username and password that PgBouncer will use when running `auth_query` against the PostgreSQL server.

The goal of this change is to align the Pooler CRD naming with the corresponding PgBouncer settings, simplifying configuration for users already familiar with PgBouncer.

For backward compatibility, the existing `authQuerySecret` field is still accepted. The validation webhook now raises a warning to guide users toward migrating to the new name. If both fields are specified, `serverSecret` takes precedence.

See: #8675


leonardoce:
Closing this PR to propose an alternative approach.

PR #8753, opened 2025-10-06
chore: deprecate `monitoring.enablePodMonitor` in Cluster and Pooler

### LINKED ISSUE #8075: [Feature]: Deprecation of `.spec.monitoring.enablePodMonitor`

CloudNativePG currently provides automated support for a basic `PodMonitor` resource associated with a `Cluster` resource via the `.spec.monitoring.enablePodMonitor` option, as described in our ["Monitoring with the Prometheus Operator"](https://cloudnative-pg.io/documentation/current/monitoring/#monitoring-with-the-prometheus-operator) documentation.

Over time, we have received several requests to extend this feature by adding fields and capabilities from the upstream `PodMonitor` API for greater customisation.

However, this direction is at odds with our ongoing efforts to refactor and simplify the CloudNativePG codebase (see CNPG-I) and with our broader goal of providing clear guidance on integrating CloudNativePG with other CNCF projects now that we are part of the Sandbox.

Begin deprecating the `.spec.monitoring.enablePodMonitor` field and improve the documentation referenced above. The cluster Helm chart could take responsibility for enhancing the integration between CNPG and Prometheus.

Add a `PodMonitorTemplate` feature; however, it requires maintaining a dependency on the interface, resulting in minimal overall benefit.

### Additional context

N/A

### Backport?

No

Yes

### Code of Conduct

phisco:
Mark the `enablePodMonitor` field as deprecated in both `Cluster` and `Pooler` monitoring configuration. This field will be removed in future releases, and users should create `PodMonitor` resources manually if needed.

The deprecation notice has been added to:

- API type definitions (`cluster_types.go` and `pooler_types.go`)
- Generated CRD manifests for both Clusters and Poolers

Documentation has been updated accordingly.

Closes #8075

### DIFF
--- a/api/v1/cluster_types.go
+++ b/api/v1/cluster_types.go
@@ -2197,8 +2197,8 @@ type MonitoringConfiguration struct {
 	// Enable or disable the `PodMonitor`
 	// +kubebuilder:default:=false
 	//
-	// Deprecated: Will be removed in future releases, create a PodMonitor manually instead,
-	// if needed.
+	// Deprecated: This feature will be removed in an upcoming release. If
+	// you need this functionality, you can create a PodMonitor manually.
 	// +optional
 	EnablePodMonitor bool `json:"enablePodMonitor,omitempty"`
 
--- a/api/v1/cluster_types.go
+++ b/api/v1/cluster_types.go
@@ -2211,6 +2211,7 @@ type MonitoringConfiguration struct {
 	//
 	// Deprecated: This feature will be removed in an upcoming release. If
 	// you need this functionality, you can create a PodMonitor manually.
+	// +optional
 	PodMonitorMetricRelabelConfigs []monitoringv1.RelabelConfig `json:"podMonitorMetricRelabelings,omitempty"`
 
 	// The list of relabelings for the `PodMonitor`. Applied to samples before scraping.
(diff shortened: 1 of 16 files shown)


gbartolini (approved):

gbartolini:
I think we should backport this patch to give users more time to update.

phisco:
while working on the helm chart PR, I noticed I missed deprecating a few other related fields, given that this is still in flight I'll update this pr.

gbartolini:
@phisco I have added a draft of documentation. However, could you please add a warning/notice to users that still set the option?

phisco:
@gbartolini, just to be sure, you mean through the validating webhook, right?

gbartolini:
I think a webhook now is fine, as long as when we decomission support for enable pod monitor, we remove the ownership from the cluster and detach the pod monitor resource (unless you have better ideas). In the meantime we will just start telling users to manually configure them. The important thing for us is that it is clear that we won't invest in this feature anymore as it is not our domain.

armru (approved):

gbartolini (approved):

mnencia (approved):

rouke-broersma:
> I think a webhook now is fine, as long as when we decomission support for enable pod monitor, we remove the ownership from the cluster and detach the pod monitor resource (unless you have better ideas). In the meantime we will just start telling users to manually configure them. The important thing for us is that it is clear that we won't invest in this feature anymore as it is not our domain.

Personally I would much rather that CNPG removes the pod monitor instead of detaching ownership so it becomes clear immediately during testing of the new operator that you missed something important instead of silently continuing to receive metrics for existing resources but having unmanaged podmonitors all of a sudden.

PR #8816, opened 2025-10-09
fix(metrics): add cnpg_cluster additionally to cluster prometheus label

### LINKED ISSUE #2501: Proposal: Rename cluster label to cnpg_cluster

Preisschild:
Fixes: https://github.com/cloudnative-pg/cloudnative-pg/issues/2501

If this is merged, I will create a follow-up PR for the grafana-dashboard so that the new `cnpg_cluster` label is used there instead of the `cluster` label. The old `cluster` label can be deleted in the future, but we keep it for now.

This is essentially the same as prior PR https://github.com/cloudnative-pg/cloudnative-pg/pull/2608, but with the new pgVersion metric also having gotten the additional label and having added getClusterName() a second time to the `.Set(1)`, so it doesnt `panic` there like in the prior PR. 

I successfully ran `make lint` & `make test` & `FEATURE_TYPE=smoke,basic make e2e-test-kind`

### DIFF
--- a/pkg/management/postgres/webserver/metricserver/pg_collector.go
+++ b/pkg/management/postgres/webserver/metricserver/pg_collector.go
@@ -132,7 +132,8 @@ func newMetrics() *metrics {
 			Subsystem: subsystem,
 			Name:      "up",
 			Help:      "1 if PostgreSQL is up, 0 otherwise.",
-		}, []string{"cluster"}),
+			// "cluster" label for backwards-compatibility
+		}, []string{"cluster", "cnpg_cluster"}),
 		CollectionDuration: prometheus.NewGaugeVec(prometheus.GaugeOpts{
 			Namespace: PrometheusNamespace,
 			Subsystem: subsystem,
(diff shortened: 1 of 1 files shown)


Preisschild:
https://github.com/cloudnative-pg/cloudnative-pg/actions/runs/18385653940/job/52427662144?pr=8816
```
Error: Error verifying linked issue.
Error: HttpError: Resource not accessible by integration
Error: Error verifying linked issue.
```

Is this something I need to fix?

lieberlois:
We really need this to! In Azure Monitor, the "cluster" label is reserved for the Kubernetes cluster name.

PR #8857, opened 2025-10-14
fix: make sure certificate private key has 0o600 permission

### LINKED ISSUE #8838: [Bug]: private key file \"server.key\" has group or world access

### Contact Details

_No response_

### Version

1.27 (latest patch)

1.31

Other

YAML manifest

### What happened?

I was going to setup a standalone replica cluster via pg_basebackup from an external cluster
The cluster was created but it does not start because of wrong permissions of the server.key file.

On the source cluster, server.key and server.crt have the correct permissions 0600.
On the new cluster, server.key and server.crt have 0660

### Cluster resource

```shell
...

  replica:
    enabled: true
    source: xxx

  bootstrap:
    pg_basebackup:
      source: xxx

  externalClusters:
    - name: xxx
      connectionParameters:
        # Use the correct IP or host name for the source database
        host: hostname
        port: '5432'
        user: dba
        dbname: postgres 
      password:
        name: xxx-dba
        key: password
...
```

### Relevant log output

```shell
{
  "level": "info",
  "ts": "2025-10-13T08:04:56.671192805+02:00",
  "logger": "postgres",
  "msg": "2025-10-13 08:04:56.671 CEST [27] @ []FATAL:  private key file \"server.key\" has group or world access",
  "pipe": "stderr",
  "logging_pod": "xxx-1"
}
{
  "level": "info",
  "ts": "2025-10-13T08:04:56.671240952+02:00",
  "logger": "postgres",
  "msg": "2025-10-13 08:04:56.671 CEST [27] @ []DETAIL:  File must have permissions u=rw (0600) or less if owned by the database user, or permissions u=rw,g=r (0640) or less if owned by root.",
  "pipe": "stderr",
  "logging_pod": "xxx-1"
}
```

### Code of Conduct

mnencia:
At every certificate reconciliation, this patch makes sure the private keys
has permission 0o600, in case something external has changed it.

Closes #8838

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -13,7 +13,7 @@ require (
 	github.com/cheynewallace/tabby v1.1.1
 	github.com/cloudnative-pg/barman-cloud v0.3.3
 	github.com/cloudnative-pg/cnpg-i v0.3.0
-	github.com/cloudnative-pg/machinery v0.3.1
+	github.com/cloudnative-pg/machinery v0.3.2-0.20251014152802-0da5936168b6
 	github.com/davecgh/go-spew v1.1.2-0.20180830191138-d8f796af33cc
 	github.com/evanphx/json-patch/v5 v5.9.11
 	github.com/go-logr/logr v1.4.3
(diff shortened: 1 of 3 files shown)


mnencia:
As explained in #8838, this fix is not needed as the certificates generating the error were not the ones under the operator control, but a leftover in the datadir, referred by `postgresql.conf` before the location is overridden in `custom.conf`

PR #8867, opened 2025-10-15
feat: run `CHECKPOINT` before `smart` and `fast` shutdowns

### LINKED ISSUE #8865: [Feature]: Issue a CHECKPOINT before shutting down

Currently, before shutting down, we do not issue a checkpoint and leave it to the standard procedure of PostgreSQL.

Issuing a CHECKPOINT will ensure that those with a high checkpoint_timeout can reduce the time of shutdown and subsequent promotion of a replica.

N/A

### Additional context

Proposed by @zjworacle in #8846

### Backport?

Yes

Yes

### Code of Conduct

armru:
Run a `CHECKPOINT` before every PostgreSQL `smart` and `fast` shutdown request to reduce both shutdown duration and replica promotion time, especially in systems with a high `checkpoint_timeout`.

This follows the best-effort approach used in `instance_controller.go`: errors during the checkpoint are logged, but the shutdown proceeds regardless. The checkpoint step is skipped for `immediate` shutdown mode.

Closes #8865

### DIFF
--- a/internal/management/controller/instance_controller.go
+++ b/internal/management/controller/instance_controller.go
@@ -594,7 +594,7 @@ func (r *InstanceReconciler) reconcileOldPrimary(
 		return false, err
 	}
 
-	contextLogger.Info("This is an old primary node. Shutting it down to get it demoted to a replica")
+	contextLogger.Info("This is the former primary instance. Shutting it down to allow it to be demoted to a replica.")
 
 	// Perform a fast shutdown on the instance and wait for the instance manager to stop.
 	// The fast shutdown process will be preceded by a CHECKPOINT.
--- a/internal/management/controller/instance_controller.go
+++ b/internal/management/controller/instance_controller.go
@@ -594,23 +594,11 @@ func (r *InstanceReconciler) reconcileOldPrimary(
 		return false, err
 	}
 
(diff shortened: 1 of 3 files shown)


gbartolini:
This should be backported as it is a minor change that prevents interference with some existing Postgres configurations.

mnencia (approved):

internal/management/controller/instance_controller.go
```diff
@@ -594,23 +594,11 @@ func (r *InstanceReconciler) reconcileOldPrimary(
...
-	if err != nil {
-		contextLogger.Error(err, "Cannot connect to primary server")
-	} else {
-		_, err = db.Exec("CHECKPOINT")
-		if err != nil {
-			contextLogger.Error(err, "Error while requesting a checkpoint")
-		}
-	}
-
-	contextLogger.Info("This is an old primary node. Shutting it down to get it demoted to a replica")
```
gbartolini, line 609: @mnencia @armru I have removed this piece of code, as now the `CHECKPOINT` is embedded in the inner `Shutdown` procedure of the instance manager.

mnencia (approved):

gbartolini (approved):

PR #8880, opened 2025-10-18
fix(cache): return value copy when loading env

armru:
Make `LoadEnv` copy cached environment slices before returning so callers can't mutate the shared backing array. Add a regression test to ensure mutations to the returned slice don't leak back into the cache.

### DIFF
--- a/internal/management/cache/cache.go
+++ b/internal/management/cache/cache.go
@@ -20,6 +20,7 @@ SPDX-License-Identifier: Apache-2.0
 package cache
 
 import (
+	"slices"
 	"sync"
 )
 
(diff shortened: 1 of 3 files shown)


internal/management/cache/cache.go
```diff
@@ -43,7 +43,8 @@ func LoadEnv(c string) ([]string, error) {
 	}
 
 	if v, ok := value.([]string); ok {
-		return v, nil
+		clone := append([]string(nil), v...)
+		return clone, nil
```
armru: given that we use a sync.Map, I think this is the correct approach

leonardoce (approved):

mnencia (approved):

Issue #8909, opened 2025-10-22
[Bug]: Cluster operator does not assign serviceAccountTemplate.metadata.name as the ServiceAccount name.


calizarr:
### Is there an existing issue already for this bug?

- [x] I have searched for an existing issue, and could not find anything. I believe this is a new bug.

### I have read the troubleshooting guide

- [x] I have read the troubleshooting guide and I think this is a new bug.

### I am running a supported version of CloudNativePG

- [x] I have read the troubleshooting guide and I think this is a new bug.

### Contact Details

calizarr@gmail.com

### Version

1.27 (latest patch)

### What version of Kubernetes are you using?

1.34

### What is your Kubernetes environment?

Cloud: Other

### How did you install the operator?

YAML manifest

### What happened?

When defining a Cluster with the Cluster CR, the `name` key exists in the `serviceAccountTemplate.metadata` field. However, the operator itself does not use the name and uses the default `cluster.Name` instead.

I expected the `name` in `serviceAccountTemplate.metadata.name` to be used as the name of the ServiceAccount created by the operator.

### Cluster resource

```shell
apiVersion: postgresql.cnpg.io/v1
kind: Cluster
metadata:
  name: postgresql-auto-generated
  namespace: default
spec:
  instances: 3
  enableSuperuserAccess: true

  serviceAccountTemplate:
    metadata:
      name: test-serviceaccount
      labels:
        wheee: whooo
      annotations:
        what: nooooo

  postgresql:
    parameters:
      log_checkpoints: "on"
      log_lock_waits: "on"
      log_min_duration_statement: '1000'
      log_statement: 'ddl'
      log_temp_files: '1024'
      log_autovacuum_min_duration: '1s'
      log_replication_commands: 'on'

  bootstrap:
    initdb:
      database: appdb
      owner: appuser
  storage:
    size: 1Gi
    storageClass: standard
```

### Relevant log output

```shell
Not sure I found any logs related to this issue.
```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

calizarr:
I would like to work on this as I have a working fork that passes tests etc.

Relevant PR: https://github.com/cloudnative-pg/cloudnative-pg/pull/8921

calizarr:
As far as I know the Pull Request has yet to be reviewed, but it is still releveant.

lambcode-unified:
[...]

PR #8921, opened 2025-10-22
fix: Allow use of arbitrary ServiceAccount name from the template (#8909)

calizarr:
When the key `serviceAccountTemplate.metadata.name` is filled in override the ServiceAccount name to be created with that name instead of using `cluster.Name`

Relevant issue: https://github.com/cloudnative-pg/cloudnative-pg/issues/8909


calizarr:
I'm also working on a fix to be able to use an existing ServiceAccount that I can include in this PR once I've got it done.

calizarr:
I've added the ability to use a pre-existing ServiceAccount. This is ready to go.

Issue #168, opened 2022-05-19
PG Sensible setting decrease sometimes cause cluster switchover


litaocdl:
Reproduce steps 
1. Install the operator 
2. create the cluster use this yaml
```
kubectl apply -f $CNPG/tests/e2e/fixtures/config_update/04-mixed-params.yaml
```
3. descrease the max_connection from 110 to 105

```
k -n test apply -f $CNPG/tests/e2e/fixtures/config_update/07-restart-decrease.yaml
```
Expected: everytime switchover should not happen, only pod restart to pickup the  change.

Actually: sometime, switch over happens, cluster changed to `Waiting for user action` and then `Upgrading cluster` status. pod turns to crashloopbackoff as switch over happens. 

```
NAME                       AGE     INSTANCES   READY   STATUS              PRIMARY
postgresql-storage-class   4m34s   3           3       Waiting for user action   postgresql-storage-class-3
postgresql-storage-class   5m53s   3           2       Upgrading cluster   postgresql-storage-class-3
```

litaocdl:
This is the log of of primary pod for a failing case,  the cluster is $CNPG/tests/e2e/fixtures/config_update/04-mixed-params.yaml, when primary pod is started up, we can see message of 
```
"msg":"Instance is still down, will retry in 1 second"
```


```
{"level":"info","ts":1653289140.558103,"logger":"setup","msg":"Starting CloudNativePG Instance Manager","logging_pod":"postgresql-storage-class-1","version":"1.15.0","build":{"Version":"1.15.0+dev41","Commit":"d4aad1ef","Date":"2022-05-11"}}
{"level":"info","ts":1653289140.7627873,"logger":"setup","msg":"starting controller-runtime manager","logging_pod":"postgresql-storage-class-1"}
{"level":"info","ts":1653289140.7629292,"logger":"controller.cluster","msg":"Starting EventSource","reconciler group":"postgresql.cnpg.io","reconciler kind":"Cluster","source":"kind source: *v1.Cluster"}
{"level":"info","ts":1653289140.7629762,"logger":"controller.cluster","msg":"Starting Controller","reconciler group":"postgresql.cnpg.io","reconciler kind":"Cluster"}
{"level":"info","ts":1653289140.763299,"msg":"Starting webserver","logging_pod":"postgresql-storage-class-1","address":":9187"}
[...]

PR #8937, opened 2025-10-24
fix(targetTime): clarify timestamp formats and UTC interpretation

### LINKED ISSUE #8928: [Docs]: Incorrect syntax for targetTime in documented example

Under documentation for "Point-in-Time Recovery (PITR) from VolumeSnapshot Objects", the example for `targetTime` is expressed as `"2023-07-06T08:00:39"`, which is the incorrect syntax and will not work.  Issue #7894 was possibly opened because the user referred to this documented example.  It should be updated to the proper syntax.

_No response_

Recovery (https://cloudnative-pg.io/documentation/1.27/recovery)

Specifically under the "Point-in-Time Recovery (PITR) from VolumeSnapshot Objects" section

_No response_

### Additional context

_No response_

### Backport?

Yes

### Code of Conduct

pchovelon:
Update the machinery dependency to v0.3.3 to ensure timestamps without timezone (both RFC3339 and PostgreSQL formats) are interpreted as UTC.

The documentation has been updated to clarify that targetTime accepts both RFC 3339 and PostgreSQL timestamp formats.

Closes #8928

Related: cloudnative-pg/machinery#167

### DIFF
--- a/docs/src/recovery.md
+++ b/docs/src/recovery.md
@@ -405,12 +405,19 @@ Here are the recovery target criteria you can use:
 
 targetTime
 :  Time stamp up to which recovery proceeds, expressed in
-   [RFC 3339](https://datatracker.ietf.org/doc/html/rfc3339) format, or as a timestamp.
+   [RFC 3339](https://datatracker.ietf.org/doc/html/rfc3339) format, or as a
+   [timestamp](https://www.postgresql.org/docs/current/runtime-config-wal.html#GUC-RECOVERY-TARGET-TIME).
    (The precise stopping point is also influenced by the `exclusive` option.)
 
 :::note
-    RFC 3339 timestamps without an explicit timezone suffix
-    (e.g., `2023-07-06T08:00:39`) are interpreted as UTC.
+    Timestamps without an explicit timezone suffix
+    (e.g., `2023-07-06 08:00:39`) are interpreted as UTC.
+:::
+
+:::warning
+    Always specify an explicit timezone in your timestamp to avoid ambiguity.
+    For example, use `2023-07-06T08:00:39Z` or `2023-07-06T08:00:39+02:00`
+    instead of `2023-07-06 08:00:39`.
 :::
 
 :::warning
--- a/docs/src/recovery.md
+++ b/docs/src/recovery.md
@@ -367,7 +367,7 @@ spec:
           kind: VolumeSnapshot
           apiGroup: snapshot.storage.k8s.io
       recoveryTarget:
-        targetTime: "2023-06-11 08:00:39.00000+02"
+        targetTime: "2023-07-06T08:00:39Z"
   externalClusters:
     - name: origin
       plugin:
@@ -404,7 +404,8 @@ recovery target.
 Here are the recovery target criteria you can use:
 
 targetTime
-:  Time stamp up to which recovery proceeds, expressed in the form of a timestamp.
+:  Time stamp up to which recovery proceeds, expressed in
+   [RFC 3339](https://datatracker.ietf.org/doc/html/rfc3339) format, or as a timestamp.
    (The precise stopping point is also influenced by the `exclusive` option.)
 
 :::warning
--- a/docs/src/recovery.md
+++ b/docs/src/recovery.md
@@ -367,7 +367,7 @@ spec:
           kind: VolumeSnapshot
           apiGroup: snapshot.storage.k8s.io
       recoveryTarget:
-        targetTime: "2023-07-06T08:00:39"
+        targetTime: "2023-06-11 08:00:39.00000+02"
   externalClusters:
     - name: origin
       plugin:
@@ -404,8 +404,7 @@ recovery target.
 Here are the recovery target criteria you can use:
 
 targetTime
-:  Time stamp up to which recovery proceeds, expressed in
-   [RFC 3339](https://datatracker.ietf.org/doc/html/rfc3339) format.
+:  Time stamp up to which recovery proceeds, expressed in the form of a timestamp.
    (The precise stopping point is also influenced by the `exclusive` option.)
 
 :::warning
--- a/docs/src/recovery.md
+++ b/docs/src/recovery.md
@@ -408,6 +408,11 @@ targetTime
    [RFC 3339](https://datatracker.ietf.org/doc/html/rfc3339) format, or as a timestamp.
    (The precise stopping point is also influenced by the `exclusive` option.)
 
+:::note
+    RFC 3339 timestamps without an explicit timezone suffix
+    (e.g., `2023-07-06T08:00:39`) are interpreted as UTC.
+:::
+
 :::warning
     PostgreSQL recovery will stop when it encounters the first transaction that
     occurs after the specified time. If no such transaction exists after the
--- a/api/v1/cluster_types.go
+++ b/api/v1/cluster_types.go
@@ -2021,7 +2021,8 @@ type RecoveryTarget struct {
 	// +optional
 	TargetLSN string `json:"targetLSN,omitempty"`
 
-	// The target time as a timestamp in the RFC3339 standard
+	// The target time as a timestamp in RFC3339 format or PostgreSQL timestamp format.
+	// Timestamps without an explicit timezone are interpreted as UTC.
 	// +optional
 	TargetTime string `json:"targetTime,omitempty"`
 
--- a/config/crd/bases/postgresql.cnpg.io_clusters.yaml
+++ b/config/crd/bases/postgresql.cnpg.io_clusters.yaml
@@ -1978,8 +1978,9 @@ spec:
                               integer)
                             type: string
                           targetTime:
-                            description: The target time as a timestamp in the RFC3339
-                              standard
+                            description: |-
+                              The target time as a timestamp in RFC3339 format or PostgreSQL timestamp format.
+                              Timestamps without an explicit timezone are interpreted as UTC.
                             type: string
                           targetXID:
                             description: The target transaction ID
--- a/docs/src/cloudnative-pg.v1.md
+++ b/docs/src/cloudnative-pg.v1.md
@@ -2232,7 +2232,7 @@ _Appears in:_
 | `targetXID` _string_ | The target transaction ID |  |  |  |
 | `targetName` _string_ | The target name (to be previously created<br />with `pg_create_restore_point`) |  |  |  |
 | `targetLSN` _string_ | The target LSN (Log Sequence Number) |  |  |  |
-| `targetTime` _string_ | The target time as a timestamp in the RFC3339 standard |  |  |  |
+| `targetTime` _string_ | The target time as a timestamp in RFC3339 format or PostgreSQL timestamp format.<br />Timestamps without an explicit timezone are interpreted as UTC. |  |  |  |
 | `targetImmediate` _boolean_ | End recovery as soon as a consistent state is reached |  |  |  |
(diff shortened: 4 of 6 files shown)


mnencia (approved):

jsilvela:
Was looking at this, and have a concern.

The code still says RFC3339. See `cluster_types.go`

``` go
type RecoveryTarget struct {
	<------->
	// The target time as a timestamp in the RFC3339 standard
	// +optional
	TargetTime string `json:"targetTime,omitempty"`
```

And in the `machinery` repo, in `time.go`, the below function takes RFC3339, RFC3339Micro, or timestamp

``` go
func ConvertToPostgresFormat(timestamp string) string {
```

So, if the documentation is indeed wrong, and only timestamps are to be allowed, shouldn't the code change too?
The doc and code are definitely out of sync.

docs/src/recovery.md
```diff
@@ -351,7 +351,7 @@ spec:
           kind: VolumeSnapshot
           apiGroup: snapshot.storage.k8s.io
       recoveryTarget:
-        targetTime: "2023-07-06T08:00:39"
+        targetTime: "2023-06-11 08:00:39.00000+02"
```
jsilvela: Been digging a bit in the code.
The problem with the existing targetTime in the doc,  `"2023-07-06T08:00:39"`, is that it's not valid RFC3339. To have a zero timezone offset, you need a final "Z".
This worked in a unit test.

```
recoveryTarget := RecoveryTarget{
	TargetTime: "2023-07-06T08:00:39Z",
}

	Expect(recoveryTarget.BuildPostgresOptions()).To(ContainSubstring("recovery_target_time = '2023-07-06 08:00:39.000000Z'"))
```

jsilvela:
```suggestion
        targetTime: "2023-07-06T08:00:39Z"
```

jsilvela (dismissed):
I think the issue with the targetTime in the document is the missing `Z` at the end.
The targetTime does admit RFC3339, RFC3339Micro, and timestamp.
That said, the code for handling is a bit convoluted and could use some updating.

docs/src/recovery.md
```diff
@@ -385,8 +385,7 @@ recovery target.
 Here are the recovery target criteria you can use:
 
 targetTime
-:  Time stamp up to which recovery proceeds, expressed in
-   [RFC 3339](https://datatracker.ietf.org/doc/html/rfc3339) format.
+:  Time stamp up to which recovery proceeds, expressed in the form of a timestamp.
```
jsilvela: Forgot to add in the review. Timestamps are admitted, definitely, but so is RFC3339, see my previous comment on the missing `Z`.

jsilvela:
```suggestion
:  Time stamp up to which recovery proceeds, expressed in
   [RFC 3339](https://datatracker.ietf.org/doc/html/rfc3339) format, or as a timestamp.
```

jsilvela:
And, since @mnencia approved previously: I'd like your take on this.
There is a problem here in the code in that we have asymmetry.
The webhook admits the targetTime formated in as in the documentation  (`"2023-07-06T08:00:39"`, which is not valid RFC3339 because of the missing `Z`), but the processing will not generate a proper Postgres timestamp.

The webhook uses func ParseTargetTime at `/cloudnative-pg/machinery@v0.3.1/pkg/types/time.go`, which specifically admits the above format.

BUT, in func ConvertToPostgresFormat `/cloudnative-pg/machinery@v0.3.1/pkg/postgres/time/time.go`, that format is passed through, and it is NOT a valid Postgres format, whereas RFC3339 and RFC3339Micro are regenerated into Postgres format.

Why do we have the special format allowed in ParseTargetTime?

pchovelon:
After your comments @jsilvela, I tried again an unit test to confirm there is an issue, event with Z at the end of timestamp.
Here are steps : 

First create a cluster with backup section using plugin : 

```yaml
---
apiVersion: v1
kind: Secret
metadata:
  name: minio-creds
type: Opaque
data:
  ACCESS_KEY_ID: bWluaW9hZG1pbg==
  ACCESS_REGION: ZnItcGFy
  ACCESS_SECRET_KEY: bWluaW9hZG1pbg==
---
kind: Cluster
metadata:
  name: cluster-alice
spec:
  imageName: ghcr.io/cloudnative-pg/postgresql:17.5
  instances: 1
  storage:
    size: 1Gi
  # Backup properties
  plugins:
  - name: barman-cloud.cloudnative-pg.io
    isWALArchiver: true
    parameters:
      barmanObjectName: minio-store
---
apiVersion: barmancloud.cnpg.io/v1
kind: ObjectStore
metadata:
  name: minio-store
spec:
  configuration:
    destinationPath: s3://demo/pierrick
    endpointURL: http://172.18.0.3:9000
    s3Credentials:
      accessKeyId:
        name: minio-creds
        key: ACCESS_KEY_ID
      secretAccessKey:
        name: minio-creds
        key: ACCESS_SECRET_KEY
      region:
        name: minio-creds
        key: ACCESS_REGION

```

Create backup 
`kubectl cnpg backup cluster-alice --method=plugin --plugin-name=barman-cloud.cloudnative-pg.io

```
k describe backups.postgresql.cnpg.io cluster-alice-20251027172908
...
  Plugin Metadata:
    Cluster UID:   def9ae28-9aea-4a99-833e-e05a61c03901
    Display Name:  BarmanCloudInstance
    Name:          barman-cloud.cloudnative-pg.io
    Plugin Name:   barman-cloud.cloudnative-pg.io
    Timeline:      1
    Version:       0.7.0
  Started At:      2025-10-27T16:29:09Z
  Stopped At:      2025-10-27T16:29:10Z
Events:
  Type    Reason     Age   From                   Message
  ----    ------     ----  ----                   -------
  Normal  Starting   10m   cloudnative-pg-backup  Starting backup for cluster cluster-alice
  Normal  Starting   10m   local-webserver        Backup started
  Normal  Completed  10m   local-webserver        Backup completed
```

Create a cluster with bootstrap restore method

```yaml                        
kind: Cluster
metadata:
  name: cluster-bob
spec:
  imageName: ghcr.io/cloudnative-pg/postgresql:17.5
  instances: 1
  storage:
    size: 1Gi
  bootstrap:
    recovery:
      source: backup-cluster-alice
        targetTime: "2025-10-27T16:30:10Z"
  externalClusters:
    - name: backup-cluster-alice
      plugin:
        parameters:
          barmanObjectName: minio-store
          serverName: cluster-alice
```

The bob cluster start with a recovery Pod : 

```json
{
  "level": "info",
  "ts": "2025-10-27T16:32:59.56105952Z",
  "msg": "Starting webserver",
  "logging_pod": "cluster-bob-1-full-recovery",
  "address": "localhost:8010",
  "hasTLS": false
}

```
... but get this error : 

```json
{
  "level": "info",
  "ts": "2025-10-27T16:33:03.221008034Z",
  "msg": "Starting up instance",
  "logging_pod": "cluster-bob-1-full-recovery",
  "pgdata": "/var/lib/postgresql/data/pgdata",
  "options": [
    "start",
    "-w",
    "-D",
    "/var/lib/postgresql/data/pgdata",
    "-o",
    "-c port=5432 -c unix_socket_directories=/controller/run",
    "-t 40000000",
    "-o",
    "-c listen_addresses='127.0.0.1'"
  ]
}
{
  "level": "info",
  "ts": "2025-10-27T16:33:03.238768322Z",
  "logger": "pg_ctl",
  "msg": "waiting for server to start....2025-10-27 16:33:03.238 UTC [46] LOG:  invalid value for parameter \"recovery_target_time\": \"2025-10-27 16:30:10.000000Z\"",
  "pipe": "stdout",
  "logging_pod": "cluster-bob-1-full-recovery"
}
{
  "level": "info",
  "ts": "2025-10-27T16:33:03.238788644Z",
  "logger": "pg_ctl",
  "msg": "2025-10-27 16:33:03.238 UTC [46] FATAL:  configuration file \"/var/lib/postgresql/data/pgdata/custom.conf\" contains errors",
  "pipe": "stdout",
  "logging_pod": "cluster-bob-1-full-recovery"
}
{
  "level": "info",
  "ts": "2025-10-27T16:33:03.33110249Z",
  "logger": "pg_ctl",
  "msg": "pg_ctl: could not start server",
  "pipe": "stderr",
  "logging_pod": "cluster-bob-1-full-recovery"
}

```

`LOG:  invalid value for parameter \"recovery_target_time\": \"2025-10-27 16:30:10.000000Z\""`

docs/src/recovery.md
pchovelon: See test below :)

jsilvela:
@pchovelon huh! The conversion from RFC3339 to postgres was partial. I was expecting full or pass-through, from the code ... I'll try to zero on this tomorrow.

jsilvela:
OK, @pchovelon , pretty stumped.
The conversion above was not partial, but complete and correct. The final `Z` for UTC is in both the RFC3339 format, and ISO 8601, which Postgres is compatible with.

And, the string you show is a valid Postgres timestamp:

```
select timestamp with time zone  '2025-10-27 16:30:10.000000Z';
      timestamptz       
------------------------
 2025-10-27 16:30:10+00
```

So, may be a specific issue with `recovery_target_time`.
I'm getting more confused with time (pun intended).

jsilvela:
According to postgres, if it's valid as a `timestamp with time zone`, it ought to work. :thinking:

> The value of this parameter is a time stamp in the same format accepted by the timestamp with time zone data type

https://www.postgresql.org/docs/17/runtime-config-wal.html#RUNTIME-CONFIG-WAL-RECOVERY-TARGET

jsilvela:
Howdy @pchovelon . I've made some experiments.
The problem is with times formatted with a trailing `Z` (those in UTC zone).
In the Postgres doc for `recovery_target_time` you can see

> The value of this parameter is a time stamp in the same format accepted by the timestamp with time zone data type, except that you cannot use a time zone abbreviation 

If the RFC3339 timestamp input in the manifest is instead converted to use `+00` for UTC, instead of `Z`, PITR works fine.

Your doc fix, IMO, while relevant, should be changed:

- on one hand, if we say we accept RFC3339 in the manifest, let's make it work rather than give up on it.
- on the other hand, the docstrings in the code make it seem like ONLY RFC3339 is supported, which is not accurate. 

I've opened https://github.com/cloudnative-pg/machinery/issues/167 to fix the conversion function.
IMO while the code is easy and I'll get it out today, this requires pondering from the maintainers.

pchovelon:
> Howdy @pchovelon . I've made some experiments.

Great, thank you @jsilvela . Running after time here ... I try to answer asap.

pchovelon (commented):
Change doc to include RFC3339 as the bug will be fixed with [#168](https://github.com/cloudnative-pg/machinery/pull/168)

pchovelon:
>     * on one hand, if we say we accept RFC3339 in the manifest, let's make it work rather than give up on it.

You are right, let's make it works.

mnencia:
This PR has been updated to include the machinery dependency bump that ensures RFC3339-like timestamps without timezone (e.g., `2023-07-06T08:00:39`) are consistently interpreted as UTC.

Previously, there was an inconsistency where `ParseTargetTime` (used by barman-cloud for backup selection) interpreted them as UTC, while `ConvertToPostgresFormat` output them without timezone, letting PostgreSQL use server local time. This could cause backup selection and recovery to target different points in time if the server wasn't in UTC.

The documentation has also been updated to clarify this behavior.

Related PRs: cloudnative-pg/machinery#168 and cloudnative-pg/plugin-barman-cloud#700

mnencia (approved):

armru (approved):

NiccoloFei (approved):

jsilvela (approved):

NiccoloFei (approved):

PR #6323, opened 2024-12-12
feat: support customizable pod patches via annotations

armru:
This patch enables precise customization of pod specifications through the `cnpg.io/podPatch` annotation on clusters. By applying valid JSON patches, users can fine-tune pod configurations, including areas not directly configurable within the Cluster resource, offering greater flexibility for advanced use cases.

**⚠️ WARNING:** This feature may introduce discrepancies between the operator’s expectations and Kubernetes behavior. Use with caution and only as a last resort.

Closes #6234 

## Release notes

Added support for custom pod patches using the `cnpg.io/podPatch` annotation, enabling user-defined pod adjustments.

PR #8943, opened 2025-10-25
fix: disable service links in bootstrap job pods

### LINKED ISSUE #6234: [Bug]: "exec /manager: argument list too long" in  bootstrap-controller

### Contact Details

maxim.i.makarov@gmail.com

### Version

1.24.0

1.28

Cloud: Other

Helm

### What happened?

I have a k8s managed cluster and a namespace with thousands of pods with different VENV. My cnpg worked fine for over a year and suddenly, newly created clusters or rebooted old ones start hanging in pending status due to sidecar container error.
Log:
` exec /manager: argument list too long`
Exactly same cluster manifest worked fine in other namespaces, so the problem relays to bash problem with large amount of VENVs in one namespace.

### Proposed solution:

Add an ability to set Pod parameter: `enableServiceLinks: false` to any created CNPG pod.

Current working solution is to use kyverno: 

```
apiVersion: kyverno.io/v1
kind: ClusterPolicy
metadata:
  name: disable-service-links-for-database
spec:
  rules:
    - name: set-enable-service-links-false
      match:
        resources:
          kinds:
            - Pod
          namespaces:
            - <some namespace>
      mutate:
        patchesJson6902: |
          - op: replace
            path: /spec/enableServiceLinks
            value: false
```
and with `enableServiceLinks=false` clusters are able to start again 

### Cluster resource

```shell

```

### Relevant log output

```shell

```

### Code of Conduct

usiegj00:
## Summary

This PR fixes the "argument list too long" error in bootstrap job pods that occurs in namespaces with many services, completing the fix started in PR #6323.

## Problem

Issue #6234 reported that bootstrap controller containers fail with `exec /manager: argument list too long` in namespaces with many services (reported in production with 465 PostgreSQL databases). 

**Critical limitation:** Even with the `cnpg.io/podPatch` annotation setting `enableServiceLinks: false` for regular cluster instance pods (fixed in PR #6323), clusters still **cannot launch** at this scale because bootstrap job pods were not covered by that fix.

Why job pods are affected:

1. Bootstrap jobs are created separately in `pkg/specs/jobs.go` via `CreatePrimaryJob()`
2. These jobs create their own pod specs that don't go through the `buildInstance()` function where the patch annotation is applied
3. The bootstrap-controller init container in these job pods receives the same excessive environment variables from Kubernetes service links
4. Since jobs must complete before the cluster can start, this failure prevents cluster creation entirely

## Solution

Set `enableServiceLinks: false` directly in the job pod spec in `pkg/specs/jobs.go`. This affects all bootstrap job types:
- `initdb` jobs (initialize new cluster)
- `join` jobs (join replica to cluster)  
- `pg_basebackup` jobs
- Recovery/restore jobs

## Alternative Solutions Considered

We considered several approaches:
- Extending the `cnpg.io/podPatch` annotation to cover job pods
- Adding a new CRD field for `enableServiceLinks`
- Using policy-based approaches (Kyverno, etc.)

However, setting `enableServiceLinks: false` directly in the job spec appears to have no negative side-effects. Bootstrap jobs do not require service discovery via environment variables as they use DNS or explicit connection strings from secrets, making this a safe and simple default that resolves this critical aspect of large-cluster scaling challenges.

## Testing

- Added unit test to verify `EnableServiceLinks` is set to `false`
- All existing tests pass

## Related Issues

Fixes #6234

### DIFF
--- a/pkg/specs/jobs.go
+++ b/pkg/specs/jobs.go
@@ -369,6 +369,7 @@ func CreatePrimaryJob(
 				},
 				Spec: corev1.PodSpec{
 					Hostname: jobName,
+					EnableServiceLinks: ptr.To(false),
 					InitContainers: []corev1.Container{
 						createBootstrapContainer(cluster, extList),
 					},
--- a/pkg/specs/jobs.go
+++ b/pkg/specs/jobs.go
@@ -368,7 +368,7 @@ func CreatePrimaryJob(
 					},
 				},
 				Spec: corev1.PodSpec{
-					Hostname: jobName,
+					Hostname:           jobName,
 					EnableServiceLinks: ptr.To(false),
 					InitContainers: []corev1.Container{
 						createBootstrapContainer(cluster, extList),
--- a/pkg/specs/jobs_test.go
+++ b/pkg/specs/jobs_test.go
@@ -245,4 +245,17 @@ var _ = Describe("Job created via InitDB", func() {
 			utils.KubernetesAppManagedByLabelName: utils.ManagerName,
 		}))
 	})
+	It("disables service links to prevent argument list too long errors", func() {
+		cluster := apiv1.Cluster{
+			Spec: apiv1.ClusterSpec{
+				Bootstrap: &apiv1.BootstrapConfiguration{
+					InitDB: &apiv1.BootstrapInitDB{},
+				},
+			},
+		}
+		job := CreatePrimaryJobViaInitdb(cluster, 0)
+
+		Expect(job.Spec.Template.Spec.EnableServiceLinks).NotTo(BeNil())
+		Expect(*job.Spec.Template.Spec.EnableServiceLinks).To(BeFalse())
+	})
 })
--- a/pkg/specs/pods.go
+++ b/pkg/specs/pods.go
@@ -189,7 +189,8 @@ func createClusterPodSpec(
 	enableHTTPS bool,
 ) corev1.PodSpec {
 	return corev1.PodSpec{
-		Hostname: podName,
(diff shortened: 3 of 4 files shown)


gbartolini:
Hi @usiegj00, thank you so much for your contribution and apologies for the delay.

In today's office hours, we agreed to set `enableServiceLinks: false` by default on both jobs and pods just in 1.31.0, without backporting the feature. Could you please update the PR?

usiegj00:
Thanks @gbartolini, and no problem on the delay. I've updated the PR to match the office-hours decision:

- Extended the `enableServiceLinks: false` default to the instance pods as well (in `createClusterPodSpec`), so it now applies to both jobs and pods by default for 1.31.0 — no backport.
- Added a `NewInstance` test asserting the instance PodSpec has `EnableServiceLinks: false`.
- Also fixed the gofmt alignment on the `Hostname`/`EnableServiceLinks` fields in the bootstrap job spec (the previous commit's formatting would otherwise fail the format check).

kfcss:
@gbartolini could this get merged? I really want to see this in the new release. It fixes bugs and actually also fixes performance issues for wal-restore and backups. We have 25k environment variables and https://github.com/cloudnative-pg/plugin-barman-cloud/blob/2db8cc37074254633293762a3a1b33b88334f5ca/internal/cnpgi/common/common.go#L76-L99 is an O(n^2) algorithm. There is a fix for the algorithm here: https://github.com/cloudnative-pg/plugin-barman-cloud/pull/1114, but it would not have been a problem if the 25k environment variables was not there in the first place.

PR #8964, opened 2025-10-26
fix(controller,backup): select Postgres container by name instead of assuming index

Joda89:
Ensure backups identify the Postgres container by its name rather than assuming the Postgres container is the first container in the Pod spec/status. Pods may include sidecars or other containers which can shift ordering; relying on index 0 is brittle and caused incorrect containerID/image checks.

Closes #8966

### DIFF
--- a/internal/controller/backup_controller.go
+++ b/internal/controller/backup_controller.go
@@ -688,7 +688,9 @@ func (r *BackupReconciler) getBackupTargetPod(ctx context.Context,
 			return false
 		}
 
-		if pod.Spec.Containers[0].Image != cluster.Status.Image {
+		idx := slices.IndexFunc(pod.Spec.Containers, func(container Container) bool { return container.name == "postgres" })
+
+		if pod.Spec.Containers[idx].Image != cluster.Status.Image {
 			contextLogger.Debug("Instance not having expected image, discarded as target for backup")
 			return false
 		}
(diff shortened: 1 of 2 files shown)


internal/controller/backup_controller.go
```diff
@@ -688,7 +688,9 @@ func (r *BackupReconciler) getBackupTargetPod(ctx context.Context,
 			return false
 		}
 
-		if pod.Spec.Containers[0].Image != cluster.Status.Image {
+		idx := slices.IndexFunc(pod.Spec.Containers, func(container Container) bool { return container.name == "postgres" })
+
+		if pod.Spec.Containers[idx].Image != cluster.Status.Image {
```
vishnuitta: we should validate for 'idx' (like idx >= 0) before using it.

armru: ~~we are doing it on the previous 3-4 lines unless I'm missing something~~

Oh got it, git diff wasn't exactly clear

armru:
Proposed commit message: 

Ensure backups identify the Postgres container by its name rather than assuming the Postgres container is the first container in the Pod spec/status. Pods may include sidecars or other containers which can shift ordering; relying on index 0 is brittle and caused incorrect containerID/image checks.

Joda89:
Thanks, @armru, for the help.

armru (approved):

jbattiato (approved):

mnencia (approved):

PR #9089, opened 2025-11-06
Instance manager: surface bind errors for status port 8000; improve local dev registry handling

Samyak2605:
Explicitly surface socket bind errors when the instance manager starts the status webserver (port 8000), preventing silent failures on certain K3s control-plane nodes.
Improve local developer experience by hardening the local registry/containerd configuration for kind-based workflows.


Samyak2605:
I updated the fork’s main with signed-off commits and rebased onto upstream. I also force-pushed so the PR should now have DCO satisfied and be up to date.

mnencia:
@Samyak2605, please check your PR. It appears to contain a large number of unrelated commits and changes.

PR #9132, opened 2025-11-11
Execute db init in parallel

exekias:
This change executes `postgres` & `template` DB bootstraps in parallel, helping to reduce overall cluster provision times.

### DIFF
--- a/internal/controller/cluster_controller.go
+++ b/internal/controller/cluster_controller.go
@@ -739,7 +739,7 @@ func (r *ClusterReconciler) reconcileResources(
 
 	if len(runningJobs) > 0 {
 		contextLogger.Debug("A job is currently running. Waiting", "runningJobs", runningJobs)
-		return ctrl.Result{RequeueAfter: 5 * time.Second}, nil
+		return ctrl.Result{RequeueAfter: 2 * time.Second}, nil
 	}
 
 	if result, err := r.deleteTerminatedPods(ctx, cluster, resources); err != nil {
--- a/pkg/management/postgres/initdb.go
+++ b/pkg/management/postgres/initdb.go
@@ -62,6 +62,14 @@ type connectionProvider interface {
 	ConnectionPool() pool.Pooler
 }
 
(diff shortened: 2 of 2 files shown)


exekias:
Run tests locally with `FEATURE_TYPE=smoke,basic make e2e-test-kind`
```
Ran 36 of 212 Specs in 554.179 seconds
SUCCESS! -- 36 Passed | 0 Failed | 0 Pending | 176 Skipped
```

internal/controller/cluster_controller.go
```diff
@@ -739,7 +739,7 @@ func (r *ClusterReconciler) reconcileResources(
 
 	if len(runningJobs) > 0 {
 		contextLogger.Debug("A job is currently running. Waiting", "runningJobs", runningJobs)
-		return ctrl.Result{RequeueAfter: 5 * time.Second}, nil
+		return ctrl.Result{RequeueAfter: 2 * time.Second}, nil
```
exekias, line 742: 5 seconds wait may be too long for jobs that are expected to run quickly, I went for reducing this a bit, ensuring more responsiveness in these cases

pkg/management/postgres/initdb.go
```diff
@@ -396,6 +413,62 @@ func (info InitInfo) ConfigureNewInstance(instance connectionProvider) error {
 	return nil
 }
 
+// executeTasksInParallel executes multiple database initialization tasks in parallel
+func (info InitInfo) executeTasksInParallel(tasks []dbTask) error {
```
sxd, line 417: You're missing the "*" it should be like `func (info *InitInfo)`

exekias: thanks! it feels like all the other functions in this files follow the same pattern? (without the *)

exekias:
I got good feedback on this PR:

This change as it is changes the behavior, as there could be race conditions between bootstrap scripts. One idea that came to me would be to add a new paramter to actually configure this new behavior, defaulting to the current one.

PR #9148, opened 2025-11-12
fix(probe,readiness): improve resilience to transient API server connectivity issues

armru:
This change enhances the resilience of all probe types (liveness, readiness, and startup) when facing transient Kubernetes API server connectivity issues. Previously, readiness and startup probes would fail immediately if unable to reach the API server, potentially causing unnecessary pod restarts or preventing pods from becoming ready.

The improvement introduces a unified cluster caching mechanism that:
- Creates a **single shared cache** instance used across all three probe types (liveness, readiness, startup) to reduce memory usage and ensure consistency
- Implements **thread-safe** cache operations with proper mutex locking to support concurrent probe execution
- Attempts to fetch the cluster definition with a **500ms timeout** to avoid blocking the probe for too long
- **Falls back to a cached cluster definition** if the API server is temporarily unreachable
- **Falls back to default probe configuration** if no cached cluster is found
- Maintains probe functionality during brief network interruptions or API server unavailability
- Uses optimized memory allocation patterns to avoid unnecessary `DeepCopy` operations

This ensures consistent behavior across all probe types and reduces false positives during transient network issues, while also improving performance through shared resources and optimized memory usage.

### DIFF
--- a/pkg/management/postgres/webserver/probes/cache.go
+++ b/pkg/management/postgres/webserver/probes/cache.go
@@ -30,10 +30,10 @@ import (
 	apiv1 "github.com/cloudnative-pg/cloudnative-pg/api/v1"
 )
 
-// clusterCache provides a resilient way to fetch cluster definitions with caching
+// ClusterCache provides a resilient way to fetch cluster definitions with caching
 // to handle transient API server connectivity issues.
-// This cache is thread-safe and can be shared across concurrent probe requests.
-type clusterCache struct {
+// This cache is thread-safe and can be shared across multiple probe types.
+type ClusterCache struct {
 	cli     client.Client
 	key     client.ObjectKey
 	timeout time.Duration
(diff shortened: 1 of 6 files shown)


leonardoce (approved):

pkg/management/postgres/webserver/probes/cache.go
```diff
@@ -0,0 +1,85 @@
...
+
+		// Return the current cached value on failure
+		c.mu.RLock()
+		cached := c.latestKnownCluster
+		c.mu.RUnlock()
+		return cached, false
+	}
+
+	c.mu.Lock()
+	c.latestKnownCluster = &cluster
```
armru: this should be a deep copy, this small optimization (of passing the pointer) could lead to several hard to debug behaviors

mnencia: The reason I removed it is that we are in a probe, which, by definition, does not modify anything. The Cluster is now passed by value to the probe function (previously passed by pointer to readiness and by value to liveness), which makes it clear that we do not expect any change to the content of the cluster. Also, when the API server is reachable, the cached Cluster is replaced at every invocation.

mnencia (approved):

kamikaze:
suddenly I'm flooded with:

readiness probe using cached cluster definition due to API server connectivity issue


but my tests barely reach 200ms max

PR #9222, opened 2025-11-18
fix(cnpgi): support custom plugin certificate DNS names

### LINKED ISSUE #9218: [Feature]: Allow customizing plugin certificate dnsName
By default the plugin's `barman-cloud-server` Certificate has `dnsNames` set to `['barman-cloud']` https://github.com/cloudnative-pg/plugin-barman-cloud/blob/main/kubernetes/server-certificate.yaml#L9.

This naming convention is not allowed in our environment.  It must end in `.svc`, like `barman-cloud.svc`.  If I do this though I then get an error when I try to bring up the cluster:

>  Phase:                   Cluster cannot proceed to reconciliation due to an error while interacting with plugins
>  Phase Reason:            Error while discovering plugins: while getting plugin connection: while querying plugin identity: rpc error: code = Unavailable desc = connection error: desc = "transport: authentication handshake failed: tls: failed to verify certificate: x509: certificate is valid for barman-cloud.svc, not barman-cloud"

Is there somewhere else I can change it to use `barman-cloud.svc` to avoid this error?

armru:
Add a new annotation `cnpg.io/pluginServerName` that allows customizing the DNS name used for TLS certificate verification when connecting to CNPG-I plugins.

Previously, the operator always used the Service name as the ServerName in the TLS configuration. This caused issues in environments where the plugin's certificate was issued with a different DNS name (e.g., `barman-cloud.svc` instead of `barman-cloud`).

With this change, users can specify the expected DNS name via the new annotation, enabling the operator to verify the plugin's certificate against the custom name while still connecting to the Service.

Closes #9218

### DIFF
--- a/internal/controller/plugin_controller.go
+++ b/internal/controller/plugin_controller.go
@@ -215,11 +215,17 @@ func (r *PluginReconciler) reconcile(
 
 	pluginAddress := fmt.Sprintf("%s:%d", service.Name, pluginPort)
 
+	// Use custom server name if provided, otherwise default to service name
+	serverName := service.Annotations[utils.PluginServerNameAnnotationName]
+	if len(serverName) == 0 {
+		serverName = service.Name
+	}
+
 	err = r.Plugins.RegisterRemotePlugin(
 		pluginName,
 		pluginAddress,
 		&tls.Config{
-			ServerName: service.Name,
+			ServerName: serverName,
 			RootCAs:    serverCertificatePool,
 			Certificates: []tls.Certificate{
 				clientKeyPair,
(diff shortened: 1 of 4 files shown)


internal/controller/plugin_controller.go
```diff
@@ -215,11 +215,17 @@ func (r *PluginReconciler) reconcile(
 
 	pluginAddress := fmt.Sprintf("%s:%d", service.Name, pluginPort)
 
+	// Use custom server name if provided, otherwise default to service name
+	serverName := service.Annotations[utils.PluginServerNameAnnotationName]
```
mateusoliveira43, line 219: does it make sense to allow a comma separate list of server names?

mnencia: No, it doesn't. This is the server name the operator uses to contact the plugin. It's only one.

mnencia (approved):

jbattiato (approved):

gbartolini (approved):

PR #9309, opened 2025-11-24
test(e2e): retry port-forward connection on failure

armru:
The E2E tests were failing because the port-forward connection was getting stuck. This commit introduces a retry mechanism with a timeout for the port-forward connection establishment. It modifies `StartAndWait` to accept a context and `startForwardConnection` to retry the connection attempt with a shorter timeout (60s) using `k8s.io/client-go/util/retry`.

### DIFF
--- a/tests/utils/forwardconnection/forwardconnection.go
+++ b/tests/utils/forwardconnection/forwardconnection.go
@@ -120,7 +120,7 @@ func NewDialer(
 }
 
 // StartAndWait begins the port-forwarding and waits until it's ready
-func (fc *ForwardConnection) StartAndWait() error {
+func (fc *ForwardConnection) StartAndWait(ctx context.Context) error {
 	var err error
 	go func() {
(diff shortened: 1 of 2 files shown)


tests/utils/forwardconnection/forwardconnection.go
```diff
@@ -120,7 +120,7 @@ func NewDialer(
 }
 
 // StartAndWait begins the port-forwarding and waits until it's ready
-func (fc *ForwardConnection) StartAndWait() error {
+func (fc *ForwardConnection) StartAndWait(ctx context.Context) error {
```
armru, line 123: if err != nil {
		return fmt.Errorf("error starting port-forward: %w", err)
	}
	
contained in this function is a bug given that it is checking an err that is executed in a goroutine.
	
It deserves a separate fix and probably a rethinking of the function

jbattiato (approved):

NiccoloFei (approved):

PR #9332, opened 2025-11-27
feat(pgbouncer): add per-database configuration support

aryklein:
## Summary
- Add per-database configuration support for PgBouncer Pooler
- New `databases` field in `PgBouncerSpec` allows overriding connection settings per database
- Wildcard entry is automatically added to ensure all databases remain accessible

## Motivation
Currently, PgBouncer Pooler only supports a single wildcard (`*`) database entry with global settings. This PR adds the ability to configure per-database connection parameters, which is a native PgBouncer feature that allows fine-tuning connection pooling behavior for specific databases.

## Changes
- Added `PgBouncerDatabaseConfig` struct in `api/v1/pooler_types.go` with fields:
  - `name`: database name as seen by clients
  - `dbname`: actual database name on PostgreSQL server (optional)
  - `poolMode`: per-database pool_mode override (optional)
  - `parameters`: additional per-database parameters like `pool_size`, `reserve_pool`, etc.
- Added `databases` field to `PgBouncerSpec`
- Updated pgbouncer.ini template generation to support multiple database entries
- Database entries are sorted alphabetically for stable output
- Wildcard (`*`) entry is always added automatically at the end
- Added webhook validation to reject `*` in the databases list (to prevent confusion)
- Added unit tests for the new functionality

## Example Usage
```yaml
kind: Pooler
metadata:
  name: my-pooler
spec:
  cluster:
    name: my-cluster
  pgbouncer:
    poolMode: transaction  # default for all databases (including wildcard)
    databases:
      - name: mydb
        poolMode: session
        parameters:
          pool_size: "20"
      - name: analytics
        parameters:
          pool_size: "50"
          reserve_pool: "5"
```

This generates:
```ini
[databases]
analytics = host=my-cluster-rw pool_size=50 reserve_pool=5
mydb = host=my-cluster-rw pool_mode=session pool_size=20
* = host=my-cluster-rw
```

**Note:** The wildcard entry is automatically added - users don't need to specify it. If a user tries to add `name: "*"` in the databases list, they'll receive a validation error.

## Testing
- Added unit tests in `pkg/management/pgbouncer/config/database_entries_test.go`
- Added webhook validation tests in `internal/webhook/v1/pooler_webhook_test.go`
- All existing tests pass


aryklein:
## Update: Automatic Wildcard Fallback

I've added an improvement to automatically include a wildcard (`*`) fallback entry when the user specifies custom databases but doesn't explicitly include a wildcard.

### Why?

Without this, if a user only specifies certain databases:

```yaml
pgbouncer:
  databases:
    - name: mydb
      poolMode: session
```

Then only `mydb` would be routable through the pooler - all other databases would fail to connect. This could lead to unexpected behavior.

### New Behavior

The operator now automatically adds `* = host=<cluster>-<type>` if no wildcard is present:

```ini
[databases]
mydb = host=test-cluster-rw pool_mode=session
* = host=test-cluster-rw
```

Users can still provide their own wildcard with custom parameters if needed:

```yaml
pgbouncer:
  databases:
    - name: mydb
      poolMode: session
    - name: "*"
      parameters:
        pool_size: "10"
```

This ensures backward compatibility and prevents accidental database lockout while still allowing full customization.

aryklein:
## Update: Simplified Wildcard Handling

After further consideration, I've simplified the wildcard behavior:

### Changes
1. **Wildcard is always added automatically** - The default `* = host=<cluster>-<type>` entry is always appended to ensure all databases are accessible
2. **Wildcard (`*`) is now rejected in the databases list** - Users cannot specify `*` in the `databases` field. This prevents confusion about whether the user's wildcard overrides the default behavior.

### Rationale
- The global `poolMode` in `PgBouncerSpec` already controls the default pool mode for all databases
- Allowing users to customize the wildcard entry would be confusing since it doesn't really "override" anything - it's the same as not specifying per-database configs
- This makes the API cleaner: users only specify databases they want to customize

### Example
```yaml
pgbouncer:
  poolMode: transaction  # applies to wildcard (all other databases)
  databases:
    - name: mydb
      poolMode: session  # override for this specific database
```

Generates:
```ini
[databases]
mydb = host=cluster-rw pool_mode=session
* = host=cluster-rw
```

If a user tries to add `name: "*"` in the databases list, they'll get a validation error:
```
wildcard (*) is not allowed in databases list; a default wildcard entry is automatically added by the operator
```

aryklein:
@mnencia , @gbartolini do you think it's going to be merged? I think this is a good feature

PR #9344, opened 2025-12-01
feat(majorupgrade): Allow image rollbacks on failed major upgrades

### LINKED ISSUE #9128: [Bug]: Major upgrade rollback doesn't work as expected

### Contact Details

lipatov@mindbox.cloud

### Version

1.26 (latest patch)

1.31

Cloud: Other

Helm

### What happened?

Due to misconfiguration a couple of my postgres clusters started to upgrade from 16.2-bullseye to 17.5-bookworm. Upgrade failed as expected (you can't upgrade to a different debian distribution). As docs says in [Post upgrade actions](https://cloudnative-pg.io/documentation/1.26/postgres_upgrades/#post-upgrade-actions) it should be possible to rollback by reverting image version change and deleting major upgrade job. But after I did this operator continued to create postgres pods with the new image version (17.5-bookworm) despite cluster spec already had imageName: 16.2-bullseye. This was happening because the cluster status had image: 17.5-bookworm and it seems that operator couldn't update it or do anything with that case without my manual actions (extra actions according to the docs).

I upgraded cluster spec with:
```
kubectl patch clusters.postgresql.cnpg.io db-<my-db> --subresource=status --patch "$(cat status.yaml)" --type=merge
```
```
status:
  image: <my-registry>/postgresql:16.2
```
And only after that my cluster recovered.

I expect that I never should manually patch cluster's status. I expect that major upgrade rollback works as docs says (in simple situations like mine - revert image change and delete upgrade job).

### Cluster resource

```shell

```

### Relevant log output

```shell

```

### Code of Conduct

redbaron:
Unless PGData version was updated (as recorded by the status), allow image rollback to the previous version.

Fixes #9128

### DIFF
--- a/internal/controller/cluster_image.go
+++ b/internal/controller/cluster_image.go
@@ -77,6 +77,16 @@ func (r *ClusterReconciler) reconcileImage(ctx context.Context, cluster *apiv1.C
 
 	// Case 2: nothing to be done.
 	if !imageChanged && !extensionsChanged {
+		// In case user rolls back failed major upgrade.
+		if cluster.Status.Image != requestedImageInfo.Image {
+			return nil, status.PatchWithOptimisticLock(
+				ctx,
+				r.Client,
+				cluster,
+				status.SetImage(requestedImageInfo.Image),
+				status.SetPGDataImageInfo(&requestedImageInfo),
+			)
+		}
 		return nil, nil
 	}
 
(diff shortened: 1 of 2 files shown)


mnencia:
The E2E tests added by @NiccoloFei highlighted that the initial approach was not working due to the race between the job recreating the pod and the rollback. I experimented with a more radical approach by making the job run only once (BackoffLimit=0) and automatically handling the job deletion when the user reverts the image: #10104

mnencia:
Hi @redbaron, thank you for your work on this — it was instrumental in shaping the final solution. As mentioned above, the E2E tests revealed a race condition with the default BackoffLimit that led us to take a different approach in #10104: disabling job retries entirely (BackoffLimit=0) and having the operator handle the cleanup automatically. Since that PR supersedes this one, I'm going to close it. Thanks again for the contribution!

PR #9351, opened 2025-12-02
fix(backup): add reconciliation timestamps to backup status

### LINKED ISSUE #9352: [Bug]: Backup status does not set startedAt until backup completes

### Contact Details

_No response_

### Version

1.27 (latest patch)

1.34

YAML manifest

### What happened?

Currently, CloudNativePG does not set the .status.startedAt field on Backup resources until the end of the backup when it also sets the .status.stoppedAt field.

This makes it challenging to effectively monitor long-running backups. Users have no way to know when a backup actually started processing.

# Root Cause
The startedAt and stoppedAt fields are a mix of when the operator starts processing and when the backup tool finishes. There is no clear separation between:

- When the operator starts reconciling the backup
- When the actual backup tool started/stopped execution

# Proposed Solution

Separate the two concepts by introducing new fields:

- reconciliationStartedAt: Set when the operator begins processing the backup
- reconciliationTerminatedAt: Set when the operator finishes processing (either successfully or with failure)

The existing startedAt and stoppedAt fields should be reserved for the backup tool's actual execution times.

### Cluster resource

```shell

```

### Relevant log output

```shell

```

### Code of Conduct

armru:
Add `reconciliationStartedAt` and `reconciliationTerminatedAt` fields to the BackupStatus to track when the operator started and terminated the backup reconciliation process.

This separates the operator's reconciliation timing from the backup tool's execution timing (`startedAt`/`stoppedAt`), which allows for better monitoring of long-running backups.

Previously, `startedAt` was not set until the backup completed, making it difficult to monitor backup progress. Now:

- `reconciliationStartedAt`: Set when the operator begins processing the backup (in `SetAsStarted`)
- `reconciliationTerminatedAt`: Set when the operator finishes processing (in `SetAsCompleted` or `SetAsFailed`)
- `startedAt`/`stoppedAt`: Reserved for the backup tool's actual execution times

For volume snapshot backups, `startedAt` and `stoppedAt` are now aligned with the reconciliation timestamps since the backup execution is managed by the operator.

Fixes monitoring of long-running backups by providing accurate timing information throughout the backup lifecycle. 

Closes #9352

### DIFF
--- a/api/v1/backup_types.go
+++ b/api/v1/backup_types.go
@@ -237,13 +237,21 @@ type BackupStatus struct {
 	// +optional
 	Phase BackupPhase `json:"phase,omitempty"`
 
-	// When the backup was started
+	// When the backup execution was started by the backup tool
 	// +optional
 	StartedAt *metav1.Time `json:"startedAt,omitempty"`
 
-	// When the backup was terminated
+	// When the backup execution was terminated by the backup tool
 	// +optional
 	StoppedAt *metav1.Time `json:"stoppedAt,omitempty"`
+
+	// When the backup process was started by the operator
+	// +optional
+	ReconciliationStartedAt *metav1.Time `json:"reconciliationStartedAt,omitempty"`
+
+	// When the reconciliation was terminated by the operator (either successfully or not)
+	// +optional
+	ReconciliationTerminatedAt *metav1.Time `json:"reconciliationTerminatedAt,omitempty"`
 
 	// The starting WAL
 	// +optional
--- a/.wordlist-en-custom.txt
+++ b/.wordlist-en-custom.txt
@@ -1057,13 +1057,13 @@ malcolm
 mallocs
 managedRoleSecretVersion
 managedRolesStatus
-matchExpressions
-matchLabels
-mateusoliveira
 managedconfiguration
 managedroles
 managedservice
 managedservices
+matchExpressions
+matchLabels
+mateusoliveira
 maxClientConnections
 maxParallel
 maxStandbyNamesFromCluster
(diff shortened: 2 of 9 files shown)


api/v1/backup_types.go
```diff
@@ -237,14 +237,22 @@ type BackupStatus struct {
...
 	// +optional
 	StoppedAt *metav1.Time `json:"stoppedAt,omitempty"`
 
+	// When the backup process was started by the operator
+	// +optional
+	ReconciliationStartedAt *metav1.Time `json:"reconciliationStartedAt,omitempty"`
+
+	// When the reconciliation was terminated by the operator (either successfully or not)
+	// +optional
+	ReconciliationTerminatedAt *metav1.Time `json:"reconciliationTerminatedAt,omitempty"`
```
jsilvela, line 254: "terminated" jumps at me. It's not a word you use on success (unless the success is that the termination command was carried out).
So, for a term that works with either success or failure, how about "ended", "finished", "concluded"?
I particularly think "end" can also applied in the negative.

```suggestion
	// When the reconciliation was ended by the operator (either successfully or not)
	// +optional
	ReconciliationEndedAt *metav1.Time `json:"reconciliationTerminatedAt,omitempty"`
```

armru: In my view there are many reconciliations cycle which start -> end.

Termination is about there won't be any other reconciliations for that given resource.

I don't see any problems in changing it to `end` instead of terminated anyway 👍

jsilvela: You know, I was finding "terminate" aggressive, but I checked the dictionary, and it's fine to use it as you have.
I think it was a false false friend. Ignore my suggestion :)

gabriele-wolfox (approved):
LGTM

jsilvela (approved):

leonardoce (approved):

PR #9364, opened 2025-12-03
fix(backup): set pending phase when backup is waiting to be processed

armru:
When a backup is created but cannot be immediately processed (e.g., cluster not found, another backup is running, or target pod is not ready), the backup's phase should be set to "pending" to indicate its state to users.

Previously, the backup phase could remain empty in these scenarios, which was misleading as users couldn't distinguish between a backup that hadn't started processing and one that was actively waiting.

This change:
- Adds a SetAsPending() method to BackupStatus for consistency with other SetAs* methods (SetAsStarted, SetAsCompleted, SetAsFailed)
- Updates getCluster() to set pending when the target cluster is not found
- Updates waitIfOtherBackupsRunning() to set pending when another backup for the same cluster is already running or waiting
- Ensures startBackupManagedByInstance() and reconcileSnapshotBackup() use the new SetAsPending() method instead of direct phase assignment

### DIFF
--- a/api/v1/backup_funcs.go
+++ b/api/v1/backup_funcs.go
@@ -36,6 +36,11 @@ import (
 	"github.com/cloudnative-pg/cloudnative-pg/pkg/utils"
 )
 
+// SetAsPending marks a certain backup as pending
+func (backupStatus *BackupStatus) SetAsPending() {
+	backupStatus.Phase = BackupPhasePending
+}
+
 // SetAsFailed marks a certain backup as invalid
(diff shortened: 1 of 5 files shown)


gabriele-wolfox (approved):
LGTM

leonardoce:
I tested it with the in core barman-cloud support and with the plugin and the patch works beautifully.
It looks fine to me.

leonardoce (approved):

mnencia (approved):

PR #9466, opened 2025-12-16
fix: remove `Backup` owner field indexer from `ScheduledBackup` controller

andrew-farries:
Remove the indexer on `Backup` resources that allowed finding all `Backups` owned by a `ScheduledBackup`.

The indexer never worked because it was checking `Backups` for owner references of kind `Backup` instead of `ScheduledBackup`:

https://github.com/cloudnative-pg/cloudnative-pg/blob/e2d6bd090cc17c123299ec109056fa510c72ddf6/internal/controller/scheduledbackup_controller.go?plain=1#L350-L352

This check meant that the indexer never returned any results and so the `GetChildBackups` method always returned an empty slice.

Even if it had correctly looked for owners of kind `ScheduledBackup`, it would only have worked for the cases where the `ScheduledBackup` was configured with:

```yaml
backupOwnerReference: self
```

When `backupOwnerReference` was set to `none` (the default) or `cluster` this indexer would also not have returned anything.

The intended purpose of the `GetChildBackups` method that used the indexer was to prevent concurrent `Backups` for a `ScheduledBackup`.  The check never worked and the `Backup` controller handles concurrency itself by checking for running `Backups` before starting a new one. So rather than fixing this indexer it is better to simply remove it.

### DIFF
--- a/internal/cmd/manager/controller/controller.go
+++ b/internal/cmd/manager/controller/controller.go
@@ -249,7 +249,7 @@ func RunController(
 		Client:   mgr.GetClient(),
 		Scheme:   mgr.GetScheme(),
 		Recorder: mgr.GetEventRecorderFor("cloudnative-pg-scheduledbackup"),
-	}).SetupWithManager(ctx, mgr, maxConcurrentReconciles); err != nil {
+	}).SetupWithManager(mgr, maxConcurrentReconciles); err != nil {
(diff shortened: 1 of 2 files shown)


armru:
Given that this patch fixes the issue but also alter the behaviour (feat) I would propose to use the more conservative: #9489  approach

mnencia:
I agree with @armru, let's close this PR in favor of #9489

PR #9468, opened 2025-12-16
feat: support DefaultAzureCredential authentication

armru:
Adds support for the DefaultAzureCredential authentication mechanism

### DIFF
--- a/api/v1/backup_types.go
+++ b/api/v1/backup_types.go
@@ -66,8 +66,9 @@ type BarmanCredentials = barmanApi.BarmanCredentials
 //
 // - storageKey
 // - storageSasToken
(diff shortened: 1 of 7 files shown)


api/v1/backup_types.go
```diff
@@ -66,8 +66,9 @@ type BarmanCredentials = barmanApi.BarmanCredentials
 //
 // - storageKey
 // - storageSasToken
+// - inheritFromAzureAD: inheriting the credentials from the pod environment
```
GabriFedi97: What this actually does is enforcing the managed identity authentication mechanism, right? Should we make that more explicit?

mnencia (approved):

fcanovai (approved):

PR #9473, opened 2025-12-16
ci: add modernize linter

PascalBourdier:
and follow advice in code

I just launch:
```bash
golangci-lint run --enable-only modernize --fix
```

Refers #11212


internal/controller/cluster_status.go
```diff
@@ -24,6 +24,7 @@ import (
 	"fmt"
 	"reflect"
 	"runtime"
+	slices0 "slices"
```
PascalBourdier, line 27: maybe we should replace `k8s.io/utils/strings/slices` by the native `slices` here (and everywhere too)

sxd:
hi @PascalBourdier 

Many of the changes proposed here may or may not affect a lot of things, so I think that before adding this linter:
* we should have an issue explaining why and open a discussion on this.
* some changes makes a lot of sense, so they requires their own PR like the slices change, maps.Copy instead of range, interface, etc
The change related to maps.Copy() requires a full PR just for that because there's so many place where it's changed

PascalBourdier:
> hi @PascalBourdier
> 
> Many of the changes proposed here may or may not affect a lot of things, so I think that before adding this linter:
> 
>     * we should have an issue explaining why and open a discussion on this.
> 
>     * some changes makes a lot of sense, so they requires their own PR like the slices change, maps.Copy instead of range, interface, etc
>       The change related to maps.Copy() requires a full PR just for that because there's so many place where it's changed

ok, I can split this PR

PR #9507, opened 2025-12-19
fix(monitoring): respect user-managed PodMonitor and expand docs

simonebenati:
- skip deleting PodMonitor without the operator label when monitoring is disabled (internal/controller/cluster_create.go)
 - add guidance and examples for manual PodMonitor/Pooler monitoring, relabeling, and RBAC setups (docs/src/monitoring.md)


simonebenati:
@sxd Suggested that it might be better instead of checking for "operator reserved" label, to check for ownership of podmonitor. I'm open to that.
My thought process was that "reserving" a label, we are sure that this podmonitor belongs to operator.
But I'm open to implement also another solution :)

simonebenati:
I will close this pr as the fix was ported into production. Thanks anyway :)

PR #9593, opened 2025-12-29
fix(replica_cluster): check condition for designated primary transition completion

### LINKED ISSUE #9591: [Bug]: Race condition in replica cluster switch prevents designated primary transition completion

### Contact Details

_No response_

### Version

trunk (main)

1.34

YAML manifest

### What happened?

 During a replica cluster switch operation, a rare race condition (observed in ~1-5% of test runs) occurs where the operator and instance manager concurrently modify the cluster object, resulting in an optimistic lock conflict that prevents the designated primary transition from completing.

From E2E test failure: https://github.com/cloudnative-pg/cloudnative-pg/actions/runs/20547048546

### Relevant log output

**Attempted transition:**
```json
{"ts":"2025-12-28T02:52:54.810548Z",
 "msg":"Setting myself as the current designated primary"}
```

**Optimistic lock conflict:**
```json
{"ts":"2025-12-28T02:52:55.372302Z",
 "error":"Operation cannot be fulfilled... the object has been modified"}
```

**Test timeout:**
```
Timed out after 300.000s.
Expected pod to be in recovery mode
```

### Code of Conduct

mnencia:
When a replica cluster switch is initiated, a race condition occurs where the operator and instance manager concurrently modify the cluster object, resulting in an optimistic lock conflict that prevents the designated primary transition condition from being set.

The root cause was in the early return check of reconcileDesignatedPrimary, which used transient instance state (RequiresDesignatedPrimaryTransition) instead of checking the actual cluster condition status. This caused the function to return early on subsequent reconciliation loops, never retrying the failed status update.

This fix makes the early return check more robust by verifying the actual condition status using IsDesignatedPrimaryTransitionRequested. If the condition is still requested (not yet completed), the function will retry setting it, even if CurrentPrimary already matches this pod.

Closes #9591

### DIFF
--- a/internal/management/controller/instance_controller.go
+++ b/internal/management/controller/instance_controller.go
@@ -1212,8 +1212,10 @@ func (r *InstanceReconciler) reconcileDesignatedPrimary(
 	ctx context.Context,
 	cluster *apiv1.Cluster,
 ) (changed bool, err error) {
-	// If I'm already the current designated primary everything is ok.
-	if cluster.Status.CurrentPrimary == r.instance.GetPodName() && !r.instance.RequiresDesignatedPrimaryTransition {
+	// Return early if this pod is already the designated primary and
+	// the transition condition is not requested.
+	if cluster.Status.CurrentPrimary == r.instance.GetPodName() &&
+		!externalcluster.IsDesignatedPrimaryTransitionRequested(cluster) {
 		return false, nil
 	}
 


armru (approved):

mnencia:
## This Fix Does Not Work

Thanks to @NiccoloFei for identifying the issue: this fix **does not resolve the race condition**.

While it prevents the early return, the condition is never actually set in subsequent loops. At line 1242, the condition is only set when `RequiresDesignatedPrimaryTransition` is TRUE. After the first loop creates `standby.signal`, `IsPrimary()` returns FALSE, making the sentinel FALSE, so subsequent loops skip setting the condition.

Closing this PR. I'll open another PR with the correct fix.

PR #9599, opened 2025-12-29
fix(podlogs): Follow mode exit bug and test improvements

mnencia:
The test "should catch extra logs if given the follow option" was flaky in CI because it tested implementation details (counting loop iterations with tight timing) rather than actual behavior.

Redesigned the test to verify what Follow=true actually does: it keeps streaming until the context is cancelled. The test now uses Eventually/Consistently patterns that handle timing variations gracefully, making it robust across different environments.

The improved test exposed a bug in the production code: the streaming function would exit when all current log streams completed, even when Follow=true was set. This caused premature exit instead of continuing to poll for new or restarted pods. Fixed by changing the exit condition to only apply when Follow=false.

### DIFF
--- a/pkg/podlogs/cluster_writer.go
+++ b/pkg/podlogs/cluster_writer.go
@@ -220,7 +220,7 @@ func (csr *ClusterWriter) SingleStream(ctx context.Context, writer io.Writer) er
 				}
 			}
 		}
-		if streamSet.isZero() {
+		if !csr.Options.Follow && streamSet.isZero() {
 			return nil
 		}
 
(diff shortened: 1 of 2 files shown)


armru (approved):

mnencia:
One of the executions of the improved test failed in CI: https://github.com/cloudnative-pg/cloudnative-pg/actions/runs/20791171388/job/59713483662?pr=9599

The failure exposed a bug in `cluster_writer.go`: the streaming function would exit when `streamSet.isZero()` returned true, even with `Follow=true` enabled. This violated the Follow semantics which should keep polling for new/restarted pods.

The race occurred because log streams could complete before the main loop checked for active streams, causing premature exit. Fixed by adding `!csr.Options.Follow` to the exit condition, ensuring Follow mode continues looping as intended.

leonardoce (approved):

PR #9622, opened 2026-01-03
prevents snyk and sync-api workflows on forks

DrowsyWings:
Fixes #9619 

## Description
- Added a if check before running the workflow

## Changes
- Added `if: ${{ github.repository == 'cloudnative-pg/cloudnative-pg' }}` condition to `snyk.yml` and `sync-api.yml`

PR #9623, opened 2026-01-03
fix(ci): add repository condition to Snyk and sync API workflows

### LINKED ISSUE #9619: [Bug]: Avoid running GitHub actions in forks

### Contact Details

_No response_

### Version

1.28 (latest patch)

1.34

YAML manifest

### What happened?

The Snyk scanning (https://github.com/cloudnative-pg/cloudnative-pg/blob/main/.github/workflows/snyk.yml) and the Sync API workflows (https://github.com/cloudnative-pg/cloudnative-pg/blob/main/.github/workflows/sync-api.yml)  should not be triggered in forks.

This helps to avoid notifying maintainers about errors in other forks repositories

### Cluster resource

```shell

```

### Relevant log output

```shell

```

### Code of Conduct

gaurangkudale:
Fix #9619 Added a job-level condition so the workflow runs only in the upstream repository.

### DIFF
--- a/.github/workflows/snyk.yml
+++ b/.github/workflows/snyk.yml
@@ -12,6 +12,7 @@ permissions: read-all
 
 jobs:
   security:
+    if: github.repository == 'cloudnative-pg/cloudnative-pg'
     name: Security scan
     runs-on: ubuntu-24.04
     permissions:
--- a/.github/workflows/sync-api.yml
+++ b/.github/workflows/sync-api.yml
@@ -9,6 +9,7 @@ permissions: read-all
 
 jobs:
   trigger-sync:
+    if: github.repository == 'cloudnative-pg/cloudnative-pg'
     runs-on: ubuntu-latest
     steps:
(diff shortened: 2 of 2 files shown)


.github/workflows/snyk.yml
```diff
@@ -12,6 +12,7 @@ permissions: read-all
 
 jobs:
   security:
+    if: github.repository == 'cloudnative-pg/cloudnative-pg'
```
sxd, line 15: We need to check the owner, the repository owner, just like we do on the CI

sxd:
Hi @gaurangkudale 

There's another PR opened before this one, would you consider add your comments there? https://github.com/cloudnative-pg/cloudnative-pg/pull/9622

Regards,

.github/workflows/snyk.yml
gaurangkudale: Closing this PR as there is another same PR exists

PR #9636, opened 2026-01-06
fix: ensure postgresql.auto.conf is writable during restore

### LINKED ISSUE #9569: [Bug]: postgres 16 image has file permission 400 for postgres.auto.conf

### Contact Details

_No response_

### Version

older in 1.27.x

1.33

YAML manifest

### What happened?

While restoring new cluster from previous backup (taken by [pgbackrest plugin](https://github.com/operasoftware/cnpg-plugin-pgbackrest)) we came across issue of failing restore due to file `postgres.auto.conf` restrictive permissions (400).

Comparing to supplied cnpg image of PostgreSQL v17 we can see that this file has less restrictive permissions (600) and restores are working fine there. It is impossible to chnage permissions of this file due to "no ALTER SYSTEM" policy in cnpg. Hence the only way is to alter supplied cnpg PostgreSQL image.

Since PostgreSQL 16 and even 14 are still supported versions I believe such subtle change in image build process can help many people.

Best regards

### Cluster resource

```shell

```

### Relevant log output

```shell
{"level":"info","ts":"2025-12-25T09:03:12.798950572+07:00","logger":"pgbackrest restore","msg":"ERROR: [041]: unable to open file '/var/lib/postgresql/data/pgdata/postgresql.auto.conf' for write: [13] Permission denied","pipe":"stderr","logging_pod":"db-is-adv-dev-16-1-full-recovery"}
```

### Code of Conduct

DrowsyWings:
PostgreSQL 16 and earlier restores fail when migratePostgresAutoConfFile()
tries to write to postgresql.auto.conf because the file is read-only (0400).
The operator sets these restrictive permissions to prevent ALTER SYSTEM usage.

This fix makes the migration function ensure the file is writable before
attempting to write to it. This approach catches all restore paths (barman-cloud,
volume snapshots, and plugins) automatically without needing separate fixes in
each path.

The reconciler will reset permissions after migration based on the cluster's
enableAlterSystem configuration.

Closes #9569

### DIFF
--- a/pkg/management/postgres/configuration.go
+++ b/pkg/management/postgres/configuration.go
@@ -331,6 +331,15 @@ var cleanupAutoConfOptions = []string{
 func (instance *Instance) migratePostgresAutoConfFile(ctx context.Context) (changed bool, err error) {
 	contextLogger := log.FromContext(ctx).WithName("migratePostgresAutoConfFile")
 
+	// Ensure postgresql.auto.conf is writable before attempting to write to it.
+	// On PostgreSQL ≤16, this file may be read-only (0400) to prevent ALTER SYSTEM,
+	// but we need to write to it during migration. The reconciler will reset
+	// permissions later based on the cluster configuration.
+	if err := instance.SetPostgreSQLAutoConfWritable(true); err != nil && !os.IsNotExist(err) {
+		contextLogger.Debug("Could not ensure postgresql.auto.conf is writable, continuing anyway",
+			"error", err)
+	}
+
 	// this is an idempotent operation. Ensures that we always include the override import.
 	// See: #5747
 	if changed, err = configfile.EnsureIncludes(path.Join(instance.PgData, "postgresql.conf"),
--- a/pkg/management/postgres/restore.go
+++ b/pkg/management/postgres/restore.go
@@ -166,6 +166,14 @@ func (info InitInfo) RestoreSnapshot(ctx context.Context, cli client.Client, imm
 		return err
 	}
 
+	// Volume snapshots may contain postgresql.auto.conf with read-only (0400)
+	autoConfPath := filepath.Join(info.PgData, "postgresql.auto.conf")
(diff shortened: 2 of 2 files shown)


mnencia:
After deep investigation, this PR does not actually fix issue #9569.

## The Real Problem

The reported issue occurs when **pgBackRest itself** (not the plugin wrapper) tries to write recovery configuration to `postgresql.auto.conf` during restore. The file has 0400 permissions from the source cluster backup (where `enableAlterSystem=false`), causing pgBackRest to fail with "Permission denied."

**Critical timing:**
1. pgBackRest extracts backup files (including postgresql.auto.conf with 0400 permissions)
2. pgBackRest tries to write recovery config to postgresql.auto.conf
3. **pgBackRest fails with Permission denied** ← Issue happens here
4. Restore never completes
5. Our CNPG code (`concludeRestore()`, `migratePostgresAutoConfFile()`) never runs

## Why This PR Doesn't Help

Both the original fix in `restore.go` and our improved version in `configuration.go` run **after** the restore completes. Since pgBackRest fails **during** its own execution, our code never gets a chance to run.

## Root Cause in pgBackRest

In `src/command/restore/restore.c`, pgBackRest uses `.noAtomic = true` when writing to postgresql.auto.conf:

```c
storagePutP(
    storageNewWriteP(..., .noAtomic = true, ...),
    BUFSTR(content));
```

This makes it try to directly open the file with `O_WRONLY`, which fails on read-only files (0400). If pgBackRest used atomic writes instead, it would write to a temp file and rename it, which would work even with 0400 permissions.

This is tracked in the plugin repository: https://github.com/operasoftware/cnpg-plugin-pgbackrest/issues/23

## Why E2E Tests Don't Catch It

Our tests don't trigger this because modern CNPG backups (v1.25+) already use `override.conf` for replication settings, so `postgresql.auto.conf` is empty and no write is needed.

Closing this PR as it doesn't address the root cause of #9569.

PR #9685, opened 2026-01-12
refactor(instance): eliminate cached cluster configuration fields

armru:
The Instance struct was caching cluster configuration values in separate fields, hiding their original source and making the code harder to understand. This refactoring improves clarity by accessing cluster configuration directly through the cluster object.

The Cluster field is now private with GetClusterOrDefault() and SetCluster() methods providing encapsulation and lazy initialization. The getter returns an empty cluster when nil, providing safe defaults without explicit initialization. This eliminates the need for cached timeout fields (PgCtlTimeoutForPromotion, MaxSwitchoverDelay, MaxStopDelay, SmartStopDelay) while making it obvious where configuration values originate.

RequiresDesignatedPrimaryTransition has been converted from a cached field to a method on Instance, keeping the logic close to where it's used. The conditions package has been moved to replicaclusterswitch/conditions for better import organization.

### DIFF
--- a/internal/management/controller/instance_controller.go
+++ b/internal/management/controller/instance_controller.go
@@ -1209,12 +1209,7 @@ func (r *InstanceReconciler) reconcileDesignatedPrimary(
 	if r.instance.RequiresDesignatedPrimaryTransition() {
 		conditions.SetDesignatedPrimaryTransitionCompleted(cluster)
 	}
-
-	if err := r.client.Status().Update(ctx, cluster); err != nil {
-		return changed, err
-	}
-
-	return changed, nil
+	return changed, r.client.Status().Update(ctx, cluster)
 }
 
 // waitForWalReceiverDown wait until the wal receiver is down, and it's used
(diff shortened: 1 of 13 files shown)


internal/management/controller/instance_controller.go
```diff
@@ -1226,25 +1203,17 @@ func (r *InstanceReconciler) reconcileDesignatedPrimary(
...
+	}
 
-		cluster.Status = updatedCluster.Status
+	if err := r.client.Status().Update(ctx, cluster); err != nil {
+		return changed, err
+	}
 
-		return r.client.Status().Update(ctx, updatedCluster)
-	})
+	return changed, nil
```
gabriele-wolfox:
```suggestion
	return changed, r.client.Status().Update(ctx, cluster)
```

gabriele-wolfox (approved):
Left a suggestion that is just a cosmetic change.
Definetely the PR is improving the code readability of the instance reconciliation.

LGTM

mnencia (approved):

leonardoce (approved):

PR #9686, opened 2026-01-12
feat(api): add initContainerResources field to configure init contain…

briquier:
…er resources

This change allows users to specify separate resource requirements (CPU/memory requests and limits) for init containers, independently from the main PostgreSQL container resources.

If initContainerResources is not specified, init containers will use the same resources as the main container (existing behavior).

- Add InitContainerResources field to ClusterSpec
- Update createBootstrapContainer to use the new field
- Add webhook validation for initContainerResources
- Add unit tests for the new validation


briquier:
Apologies, this was pushed to the wrong repo by mistake.

PR #9720, opened 2026-01-14
feat(rollout): honor primaryUpdateMethod when adding PVCs

### LINKED ISSUE #5484: [Bug]: primary pod is not switchover when add new wal volume

### LINKED ISSUE #5564: [Feature]: Tablespace addition should work with Primary Switchover

In the current implementation of the tablespace reconciler, adding a new tablespace to a cluster will force a restart of the primary without a switchover, regardless of the `PrimaryUpdateMethod` setting for the cluster.

The reason is that the tablespace reconciler looks only at the primary instance, without knowledge of the status of the replicas.
In Postgres, the `CREATE TABLESPACE` command is issued in the primary only, and the tablespace mounts/PVCs need to be present for both the primary and the replicas.

If we added a new tablespace, and had `PrimaryUpdateMethod=switchover` with the current tablespace reconciler, the sequence would be:

1. upgrade and restart of the replica pods to mount the new tablespace volumes
2. initiate switchover to a new primary
3. new primary is promoted
4. the new primary will `CREATE TABLESPACE`
5. old primary is restarted in place to become a replica
6. when the old primary starts streaming from new primary, it will crash because it's missing the mount for the tablespace
7. old primary will CrashLoop

It is for this reason that there is an argument `forceRecreate` in the `updatePrimaryPod` method in `internal/controller/cluster_upgrade.go` (see below).
With this argument set to true, the primary will be updated without switchover, no matter the value of `PrimaryUpdateStrategy`.

``` go
func (r *ClusterReconciler) updatePrimaryPod(
	ctx context.Context,
	cluster *apiv1.Cluster,
	podList *postgres.PostgresqlStatusList,
	primaryPod corev1.Pod,
	inPlacePossible bool,
	forceRecreate bool,
```

The tablespace controller should verify that ALL the instances have the mounts for a given tablespace (MyTbs), before issuing the `CREATE TABLESPACE`.

The necessary information is in the cluster status, which keeps a list of  healthy PVCs:

``` yaml
healthyPVC:
  - cluster-example-with-tablespaces-1
  - cluster-example-with-tablespaces-1-tbs-another-tablespace
  - cluster-example-with-tablespaces-1-tbs-atablespace
  - cluster-example-with-tablespaces-1-tbs-tablespacea1
```

If the controller can verify that the mounts for MyTbs are there for all the instances before issuing the `CREATE TABLESPACE`, then the extra logic in `func (r *ClusterReconciler) updatePrimaryPod(` could be eliminated.

There is a viable alternative in doing nothing. The setting for Primary Switchover will be subverted when adding new tablespaces, but will otherwise be in place.

### Additional context

_No response_

### Backport?

Yes

Yes

### Code of Conduct

jsilvela:
Currently, when upgrading a cluster to add a new tablespace or set
a dedicated WAL volume, the configured choice of
`primaryUpdateMethod` is ignored, and primary restart is forced.
This PR enables cluster upgrade for these cases to proceed with
primary switchover if so configured.

Closes #5564 
Closes #5484


internal/controller/cluster_upgrade.go
```diff
@@ -542,9 +562,8 @@ func checkClusterHasDifferentRestartAnnotation(
 		podRestart := pod.Annotations[utils.ClusterRestartAnnotationName]
 		if clusterRestart != podRestart {
 			return rollout{
-				required:     true,
-				reason:       "cluster has been explicitly restarted via annotation",
-				canBeInPlace: false,
+				required: true,
+				reason:   "cluster has been explicitly restarted via annotation",
```
jsilvela, line 562: setting `canBeInPlace` here is meaningless. Default is already false, and fields is true only if pod diff is negative but there is a restart annotation

internal/controller/cluster_upgrade.go
```diff
@@ -141,7 +141,32 @@ func (r *ClusterReconciler) rolloutRequiredInstances(
...
+		return false, err
+	}
+	if err := r.setPrimaryInstance(ctx, cluster, targetInstance.Pod.Name); err != nil {
+		return false, err
+	}
+	// since we need rollout, we can do the pod upgrade rather than wait for
+	// the old primary to restart, and then the outdated PodSpec to trigger pod upgrade
+	// if err := r.upgradePod(ctx, cluster, &primaryPod, reason); err != nil {
+	// 	return false, err
+	// }
```
jsilvela: Taking this out to keep the current process. I have gotten full-green E2E with this code active.
I would like to understand if this would be feasible.
In switchover provoked by a rolling update, the current behavior issues a restart on the old primary, I think for the benefit a CHECKPOINT and a clean shutdown.
Later the `isPodNeedingRollout` will delete the old primary and force a second restart of that instance.
Seems it could be done in one shot, and that might enable switchover with simultaneous image and config change.

jsilvela:
```suggestion
```

internal/management/controller/tablespaces/controller_test.go
```diff
@@ -282,6 +289,60 @@ var _ = Describe("Tablespace synchronizer tests", func() {
...
+				clusterStatus: apiv1.ClusterStatus{
+					InstanceNames: []string{"cluster-example-1", "cluster-example-2"},
+					HealthyPVC: []string{
+						"cluster-example-1-tbs-foo", "cluster-example-1-tbs-bar",
+						"cluster-example-2-tbs-foo",
+					},
+				},
+				postgresExpectations: func(mock sqlmock.Sqlmock) {
+					// we expect the reconciler to list the tablespaces on DB, and to
+					// create a new tablespace
```
jsilvela:
```suggestion
					// we expect the reconciler to list the tablespaces on DB, and NOT to
					// create a new tablespace
```

armru (approved):

mnencia (approved):

mnencia:
/ok-to-merge E2E failed for a network error.

Issue #6109, opened 2024-11-14
[Bug]: Operator deletes PodMonitor that it didn't create due to lack of owner reference check


ari-becker:
### Is there an existing issue already for this bug?

- [x] I have searched for an existing issue, and could not find anything. I believe this is a new bug.

### I have read the troubleshooting guide

- [x] I have read the troubleshooting guide and I think this is a new bug.

### I am running a supported version of CloudNativePG

- [x] I have read the troubleshooting guide and I think this is a new bug.

### Version

1.24.0

### What version of Kubernetes are you using?

1.30

### What is your Kubernetes environment?

Cloud: Google GKE

### How did you install the operator?

YAML manifest

### What happened?

1. https://github.com/cloudnative-pg/cloudnative-pg/issues/6108 , decide to set `monitoring.enablePodMonitor: false`. In the same commit, create a `PodMonitor` to be applied manually, with the same name (i.e. the cluster name) and namespace.
2. Apply both changes simultaneously
3. Wonder why I don't have a `PodMonitor`
4. Go spelunking in the code: https://github.com/cloudnative-pg/cloudnative-pg/blob/3c2c3f695eafbf1a0db96954f1796ff37cbd0733/internal/controller/cluster_create.go#L959 gets a reference to the `PodMonitor` based on namespace and name only, *not* looking at the owner reference. Therefore https://github.com/cloudnative-pg/cloudnative-pg/blob/3c2c3f695eafbf1a0db96954f1796ff37cbd0733/internal/controller/cluster_create.go#L973 indeed finds the `PodMonitor` (that it doesn't have an owner reference on) and deletes it.

I worked around the issue by adding a short suffix to the `PodMonitor` name so that it wouldn't be deleted, but this isn't ideal because it changes the job label that Prometheus attaches to it.

### Cluster resource

```shell
N/A
```

### Relevant log output

```shell
N/A
```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

Jean-Daniel:
I'm also impacted by this issue.

Note that if you need to create your own PodMonitor but want to keep the same job label, you can override it using relabelings:

```yaml
apiVersion: monitoring.coreos.com/v1
kind: PodMonitor
metadata:
  name: clustername-suffix
spec:
  podMetricsEndpoints:
  - metricRelabelings:
    - replacement: clustername
      targetLabel: job
    port: metrics
  selector:
    matchLabels:
[...]

PR #9723, opened 2026-01-15
docs: Update PodMonitor examples to use distinct names from clusters

ardentperf:
For documentation of current CNPG versions, update all manual PodMonitor
examples in the documentation to use names that differ from the cluster
name (e.g., 'cluster-example-podmonitor' instead of 'cluster-example').

This change addresses a known issue where manually created PodMonitors
with the same name as the cluster are deleted by the operator during
reconciliation when .spec.monitoring.enablePodMonitor is disabled.

The pattern of using a suffix like '-podmonitor' ensures clear
separation between operator-managed and user-managed resources, and
prevents unintended deletion during operator reconciliation.

In future versions of CNPG once the deletion behavior is removed, we can
update the language here to indicate it applies only to older versions,
or we can remove this text.

Related to: #8075 (deprecation of enablePodMonitor)
Fixes naming conflict described in: #6109

Signed-off-by: Jeremy Schneider <schneider@ardentperf.com>

### DIFF
--- a/docs/src/monitoring.md
+++ b/docs/src/monitoring.md
@@ -99,7 +99,7 @@ find `PodMonitor` resources.
 apiVersion: monitoring.coreos.com/v1
 kind: PodMonitor
 metadata:
-  name: cluster-example
+  name: cluster-example-podmonitor
 spec:
   selector:
     matchLabels:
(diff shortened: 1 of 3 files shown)


mnencia:
@ardentperf PR #9340 has been merged

ardentperf:
> @ardentperf PR #9340 has been merged

@mnencia  Yes but not yet released right? Maybe a “back patch only” PR to update existing docs? Not required, was just a suggestion… seems a good number of people hit this (including me)

with docs telling people to create podmonitors now, I think more people might hit this as they reference latest docs but aren’t yet running latest versions of CNPG. It’s not ideal for docs to recommend something with examples that we know are broken & will trigger the bug in existing versions.

PR #9774, opened 2026-01-19
feat:Add jobPatch annotation functionality

### LINKED ISSUE #10712: [Feature]: Add support for job patches

I have K8s cluster where the administrator restricts security policy for all pods, it requires to disable hostUsers on all pods in cluster, so I can not create any postgresql cluster on such K8s environment. For pods created by operator I can use `cnpg.io/iPodPatch` annotation, but there is no such options for jobs used for many processes by operator.

Added support for all types of jobs patches:

* `cnpg.io/initdbJobPatch`
* `cnpg.io/importJobPatch`
* `cnpg.io/pgbasebackupJobPatch`
* `cnpg.io/fullRecoveryJobPatch`
* `cnpg.io/joinJobPatch`
* `cnpg.io/snapshotRecoveryJobPatch`
* `cnpg.io/majorUpgradeJobPatch`

This feature mirrors the existing cnpg.io/podPatch annotation functionality, extending customization capabilities to Job resources used for cluster initialization, cloning, recovery, and joining operations.

I prepare the change with hostUsers directly and one of maintainers suggest to make it with annotations way, the issue is here #9668 

### Additional context

_No response_

### Backport?

Yes

Yes

### Code of Conduct

czqrny:
Closes #10712


armru:
hello @czqrny, first of all thanks for the contribution.

I gave a quick look and I would say that there is a problem with the patch, not all the jobs are the same, there should be a way to differentiate to which kind of job the patch should be applied (example: initdb, restore).

The problem can be potentially solved in two way: a) we stick with a single annotation but the contained value has to specify also the type of job, b) we add a different annotation for each job type

Also the PR needs to have also E2E test coverage

czqrny:
Hi @mnencia @armru, it's been quite a long time, but I think this PR is ready. I tested the new e2e tests locally and they passed. The PR is now ready for review!

Issue #9771, opened 2026-01-19
[Bug]: Logical replication - subscriber database created/imported with wrong name


plaes:
### Is there an existing issue already for this bug?

- [x] I have searched for an existing issue, and could not find anything. I believe this is a new bug.

### I have read the troubleshooting guide

- [x] I have read the troubleshooting guide and I think this is a new bug.

### I am running a supported version of CloudNativePG

- [x] I have read the troubleshooting guide and I think this is a new bug.

### Contact Details

plaes@plaes.org

### Version

1.28 (latest patch)

### What version of Kubernetes are you using?

1.33

### What is your Kubernetes environment?

Self-managed: k3s

### How did you install the operator?

Helm

### What happened?

I've been following the "Example of live migration and major Postgres upgrade with logical replication" examples, but the subscriber cluster ends up with wrong database name called `app`.
Basically:
1. Primary cluster (`partdb-cluster`) contains database `partdb`
2. Updated cluster (`partdb-cluster-v17`) ends up with data migrated into databased named `app`, though internally  postgresql attempts to "bootstrap" `partdb` database.

Everything seems to work initially:
- Both clusters are "green"
- Database schema is created in the new "upgraded" cluster, but in wrong database called `app`.

But I can see following in logs:
[...]

PR #9775, opened 2026-01-20
Docs: Add default keys for cluster bootstrap values

plaes:
Update docs to be a bit more clear about the default values for the replication cases when using `microservice`-based bootstrap.

Hopefully this makes logical replication-based upgrades more clear - #9771


jsilvela (commented):
Not sure your changes are widely useful. Different strokes for different folks etc.
And the convention of `app` as a default is worth learning and keeping present.

PR #9781, opened 2026-01-20
feat: support PostgreSQL extensions in image catalogs

### LINKED ISSUE #8495: [Feature]: Dynamic loading of PostgreSQL extensions via image catalogs

In issue #7991, we completed the first part of the job described in issue #7188 by implementing dynamic loading of extensions through the `.spec.postgresql.extensions` stanza. The second part of the initial proposal involves adding support for dynamic loading of extensions via image catalogs. This issue is focused on that aspect.

Consider the following catalog:

```yaml
kind: ImageCatalog
metadata:
  name: postgresql
  namespace: default
spec:
  images:
    # ...
    - major: 18
      image: # ... PostgreSQL 18 container image
      extensions: # this should be a list of ExtensionConfiguration objects
        - name: pgvector
          image:
            reference: # registry path for the "pgvector" PostgreSQL 18 extension image
```

A possible cluster definition could be:

```yaml
kind: Cluster
metadata:
  name: pg18
spec:
  instances: 1
  imageCatalogRef:
    apiGroup: postgresql.cnpg.io
    kind: ImageCatalog
    name: postgresql
    major: 18
  postgresql:
    extensions:
      - name: pgvector
  storage:
    size: 1Gi
```

Given that the `Cluster` already points to a catalog definition for version 18 and requires the `pgvector` extension, the operator will point to the only `pgvector` extension image declared in the catalog for PostgreSQL 18. If no such extension exists, the change fails.

The catalog should enable the default configuration of the following options for an extension:

- `extension_control_path`
- `dynamic_library_path`
- `ld_library_path`

We should also consider the possibility of exposing dependencies on other extensions by potentially including the `requires` field in the control file of PostgreSQL extensions.

## Override

The image catalog will basically introduce a middle layer in the "convention over configuration" process of extension definition. Right now, the `.spec.postgresql.extensions` stanza changes the value from an empty set. When a catalog is present, the `.spec.postgresql.extensions` stanza will change the values of a given extension from the catalog ones.

For example:

```yaml
kind: Cluster
metadata:
  name: pg18-direct-extension
spec:
  instances: 1
  imageCatalogRef:
    apiGroup: postgresql.cnpg.io
    kind: ImageCatalog
    name: postgresql
    major: 18
  postgresql:
    extensions:
      - name: pgvector
        image:
          reference: # ... alternative pgvector container image for PostgreSQL 18
  storage:
    size: 1Gi
```

The override here will request a different image from the one in the catalog. The following options should be overridden based on the name of the extension:

- `extension_control_path`
- `dynamic_library_path`
- `ld_library_path`

See #7188 

### Additional context

_No response_

### Backport?

No

Yes

### Code of Conduct

NiccoloFei:
Extend `ImageCatalog` and `ClusterImageCatalog` CRDs to allow defining
PostgreSQL extension images via the `.spec.images[].extensions` stanza.

This enables a "convention over configuration" approach where extension
details (image references, paths, etc.) can be centrally managed in a
catalog and inherited by Clusters.

Key behaviors:

- Uses the same API schema as the Cluster's `.spec.postgresql.extensions`.
- Supports merging: Cluster-level definitions are merged with catalog
  definitions.
- Supports overrides: If an extension name exists in both, the Cluster's
  configuration takes precedence.
- Acts as a default configuration layer for referenced catalogs.

Closes #8495


internal/controller/cluster_image.go
```diff
@@ -104,16 +111,18 @@ func (r *ClusterReconciler) reconcileImage(ctx context.Context, cluster *apiv1.C
 		status.SetPGDataImageInfo(&requestedImageInfo))
 }
 
-func getImageInfoFromImage(image string) (apiv1.ImageInfo, error) {
+func getImageInfoFromImage(cluster *apiv1.Cluster) (apiv1.ImageInfo, error) {
```
GabriFedi97: Should we rename this `getImageInfoFromCluster`?

pkg/utils/extensions/extensions.go
```diff
@@ -0,0 +1,127 @@
...
+	apiv1 "github.com/cloudnative-pg/cloudnative-pg/api/v1"
+)
+
+// ResolveFromCatalog returns a list of requested Extensions from a Catalog for a given
+// Postgres major version.
+// If present, extension entries present in the catalog are always used as a starting base.
+// If additional configuration is passed via `cluster.Spec.PostgresConfiguration.Extensions`
+// for an extension that's already defined in the catalog, the values defined in the Cluster
+// take precedence.
+func ResolveFromCatalog(
```
GabriFedi97, line 36: Should we rename this function from `ResolveFromCatalog` to just `Resolve`? I feel the current naming doesn't fit very well with how the function behaves, like resolving extensions from the catalog first but then overriding with cluster spec definition.

pkg/utils/extensions/extensions.go
```diff
@@ -0,0 +1,127 @@
...
+	cluster *apiv1.Cluster,
+	catalog apiv1.GenericImageCatalog,
+	requestedMajorVersion int,
+) ([]apiv1.ExtensionConfiguration, error) {
+	requestedExtensions := cluster.Spec.PostgresConfiguration.Extensions
+	resolvedExtensions := make([]apiv1.ExtensionConfiguration, 0, len(requestedExtensions))
+
+	// Build a map of extensions coming from the catalog
+	catalogExtensionsMap := make(map[string]apiv1.ExtensionConfiguration)
+	if extensions, ok := catalog.GetSpec().FindExtensionsForMajor(requestedMajorVersion); ok {
```
GabriFedi97: We can allocate the `catalogExtensionsMap` once

```suggestion
var catalogExtensionsMap map[string]apiv1.ExtensionConfiguration
	if extensions, ok := catalog.GetSpec().FindExtensionsForMajor(requestedMajorVersion); ok {
		catalogExtensionsMap = make(map[string]apiv1.ExtensionConfiguration, len(extensions))
```

pkg/utils/extensions/extensions.go
```diff
@@ -0,0 +1,127 @@
...
+		var resultExtension apiv1.ExtensionConfiguration
+		if found {
+			// Found the extension in the catalog, so let's use the catalog entry as a base
+			// and eventually override it with Cluster Spec values
+			resultExtension = catalogExtension
+		} else {
+			// No catalog entry, rely fully on the Cluster Spec
+			resolvedExtensions = append(resolvedExtensions, extension)
+			continue
+		}
```
GabriFedi97: Is this more conventional?
```suggestion
		if !found {
			// No catalog entry, rely fully on the Cluster Spec
			resolvedExtensions = append(resolvedExtensions, extension)
			continue
		}

		// Let's use the catalog entry as a base
		// and eventually override it with Cluster Spec values
		resultExtension := catalogExtension
```

pkg/reconciler/majorupgrade/reconciler.go
```diff
@@ -237,13 +239,38 @@ func majorVersionUpgradeHandleCompletion(
 		return nil, err
 	}
 
+	// Resolve extensions for the new major version.
+	// This ensures that extension images match the upgraded PostgreSQL version.
+	// If extension resolution fails (e.g., catalog doesn't have extensions for new version),
+	// we fail the major upgrade completion to avoid leaving the cluster in an inconsistent state.
+	extensions, err := resolveExtensionsForMajorVersion(ctx, c, cluster, requestedMajor)
```
GabriFedi97: This is currently colliding with the `github.com/cloudnative-pg/cloudnative-pg/pkg/utils/extensions` import
```suggestion
	exts, err := resolveExtensionsForMajorVersion(ctx, c, cluster, requestedMajor)
```

pkg/utils/extensions/extensions.go
NiccoloFei: I'm not sure about this, because while it's true that we are potentially merging and/or overriding catalog-defined entries with cluster-defined entries, you still must be providing a Catalog to this function (meaning that it only gets used when a Catalog is being referenced by a Cluster). So it doesn't generically "resolve" extensions, it does that when a Catalog is being used

armru (approved):

internal/controller/cluster_image.go
NiccoloFei: Fixed with https://github.com/cloudnative-pg/cloudnative-pg/pull/9781/changes/cdd76024c765d18e1dd1b2a3914553b9f333a130

pkg/utils/extensions/extensions.go
NiccoloFei: Fixed with https://github.com/cloudnative-pg/cloudnative-pg/pull/9781/changes/cdd76024c765d18e1dd1b2a3914553b9f333a130

pkg/reconciler/majorupgrade/reconciler.go
NiccoloFei: Fixed with https://github.com/cloudnative-pg/cloudnative-pg/pull/9781/changes/cdd76024c765d18e1dd1b2a3914553b9f333a130

pkg/utils/extensions/extensions.go
```diff
@@ -0,0 +1,126 @@
...
+	cluster *apiv1.Cluster,
+	catalog apiv1.GenericImageCatalog,
+	requestedMajorVersion int,
+) ([]apiv1.ExtensionConfiguration, error) {
+	requestedExtensions := cluster.Spec.PostgresConfiguration.Extensions
+	resolvedExtensions := make([]apiv1.ExtensionConfiguration, 0, len(requestedExtensions))
+
+	// Build a map of extensions coming from the catalog
+	catalogExtensionsMap := make(map[string]apiv1.ExtensionConfiguration)
+	if extensions, ok := catalog.GetSpec().FindExtensionsForMajor(requestedMajorVersion); ok {
```
GabriFedi97: We can allocate the `catalogExtensionsMap` once

```suggestion
var catalogExtensionsMap map[string]apiv1.ExtensionConfiguration
	if extensions, ok := catalog.GetSpec().FindExtensionsForMajor(requestedMajorVersion); ok {
		catalogExtensionsMap = make(map[string]apiv1.ExtensionConfiguration, len(extensions))
```

GabriFedi97 (approved):

gbartolini:
I have done further tests to perform "simultaneous" changes to:

- the extension image in the catalog, and
- the Postgres configuration for shared buffers and then max connections (both raising and lowering the value)

I tested on a 3 instance cluster with both restart and switchover primary update method.

Everything worked correctly.

I have also rewritten entirely both the image catalog and image volumes pages.

Assigning it to @mnencia

gbartolini (approved):

yttasdfghjk (approved):

mnencia (approved):

leonardoce (approved):

PR #9788, opened 2026-01-22
Collect WAL stats metric from primary only.

### LINKED ISSUE #9106: [Bug]: 1.27.1 stats query bug

### Contact Details

_No response_

### Version

1.27 (latest patch)

1.33

Cloud: Amazon EKS

Helm

### What happened?

I upgraded from 1.27.0 to 1.27.1 on a cluster and I'm seeing the output below in a postgres cluster.

### Cluster resource

```shell

```

### Relevant log output

```shell
{"level":"error","ts":"2025-11-08T16:41:01.349452822Z","msg":"while collecting pg_stat_wal","logging_pod":"postgres-cluster-application-3","error":"sql: Scan error on column index 4, name \"stats_reset\": converting NULL to string is unsupported","stacktrace":"github.com/cloudnative-pg/machinery/pkg/log.(*logger).Error\n\tpkg/mod/github.com/cloudnative-pg/machinery@v0.3.1/pkg/log/log.go:125\ngithub.com/cloudnative-pg/machinery/pkg/log.Error\n\tpkg/mod/github.com/cloudnative-pg/machinery@v0.3.1/pkg/log/log.go:163\ngithub.com/cloudnative-pg/cloudnative-pg/pkg/management/postgres/webserver/metricserver.(*Exporter).collectPgMetrics\n\tpkg/management/postgres/webserver/metricserver/pg_collector.go:454\ngithub.com/cloudnative-pg/cloudnative-pg/pkg/management/postgres/webserver/metricserver.(*Exporter).Collect\n\tpkg/management/postgres/webserver/metricserver/pg_collector.go:321\ngithub.com/prometheus/client_golang/prometheus.(*Registry).Gather.func1\n\tpkg/mod/github.com/prometheus/client_golang@v1.23.2/prometheus/registry.go:456"}
```

### Code of Conduct

redbaron:
pg_stat_wal view shows WAL generation stats and only primary generates WAL.

This resolves problem where pg_stat_wal can be empty on a replica which was never a primary before, which causes metric collection to fail on a NULL conversion.

fixes #9106

### DIFF
--- a/pkg/management/postgres/webserver/metricserver/pg_collector.go
+++ b/pkg/management/postgres/webserver/metricserver/pg_collector.go
@@ -454,6 +454,14 @@ func (e *Exporter) updateInstanceMetrics() {
 		e.collectFromPrimaryLastAvailableBackupTimestamp()
 
 		e.collectFromPrimaryLastFailedBackupTimestamp()
+
+		if version, _ := e.instance.GetPgVersion(); version.Major() >= 14 {
+			if err := collectPGStatWAL(e); err != nil {
+				log.Error(err, "while collecting pg_stat_wal")
+				e.Metrics.Error.Set(1)
+				e.Metrics.PgCollectionErrors.WithLabelValues("Collect.PGWALStat").Inc()
+			}
+		}
 	}
 
 	if err := collectPGWalArchiveMetric(e); err != nil {
@@ -475,14 +483,6 @@ func (e *Exporter) updateInstanceMetrics() {
 		e.Metrics.Error.Set(1)
 		e.Metrics.PgCollectionErrors.WithLabelValues("Collect.PGVersion").Inc()
 		e.Metrics.PgVersion.Reset()
-	}
-
-	if version, _ := e.instance.GetPgVersion(); version.Major() >= 14 {
-		if err := collectPGStatWAL(e); err != nil {
-			log.Error(err, "while collecting pg_stat_wal")
-			e.Metrics.Error.Set(1)
-			e.Metrics.PgCollectionErrors.WithLabelValues("Collect.PGWALStat").Inc()
-		}
 	}
 }
 


jsilvela (approved):

redbaron:
There was 1.29 release without this fix. Could you please consider merging it for 1.29.1 or 1.30 at least?

redbaron:
@gbartolini , @jsilvela , what would it take to merge this in?

jsilvela:
a second approval, @redbaron. You already have mine

honi-at-simspace:
we are having this problem too.  Can this be merged?

honi-at-simspace:
I also put another PR to fix the underlying issue (https://github.com/cloudnative-pg/cloudnative-pg/pull/10624) which allows for NULL values for `pg_stat_wal.stats_reset`, which would resolve this issue as well as other times that it can be NULL on a primary as well.

mnencia:
Thanks for digging into this, and sorry it sat for a while.

Looking closer, I don't think gating on `isPrimary` is the right fix here. `IsPrimary()` checks for the absence of `standby.signal` on disk, but designated primaries in a replica-cluster topology run with `standby.signal` present (they're PG standbys) and `archive_mode = always`, so `isPrimary()` returns false for them permanently. Gating this collector on it would silently stop collecting `pg_stat_wal` on those nodes for the life of the cluster, not just fix the reported crash.

The actual defect is that `PgStatWal.StatsReset` (`probes.go`) scans `pg_stat_wal.stats_reset` into a non-nullable Go `string`, but that column is nullable in Postgres. The fix belongs there, at the scan site, whether that's a `COALESCE` in the query or a nullable scan target, rather than trying to predict which instances will and won't have a populated value.

I'm going to close this in favor of a fix at the scan site. Appreciate you tracking down the original crash, that part of the diagnosis was right.

redbaron:
> designated primaries in a replica-cluster topology run with standby.signal present (they're PG standbys) and archive_mode = always, so isPrimary() returns false for them permanently.

If they are standby they don't generate WAL on their own and querying `pg_stat_wal` there should return empty values there.

Are you saying there can be a situation where `IsPrimary() == false` , but pod still generates own WAL?

HoniSanders:
@mnencia I put a new PR here that makes the Null string bug directly: https://github.com/cloudnative-pg/cloudnative-pg/pull/11207. 
Can you review that?

Thank you!

mnencia:
> > designated primaries in a replica-cluster topology run with standby.signal present (they're PG standbys) and archive_mode = always, so isPrimary() returns false for them permanently.
> 
> If they are standby they don't generate WAL on their own and querying `pg_stat_wal` there should return empty values there.
> 
> Are you saying there can be a situation where `IsPrimary() == false` , but pod still generates own WAL?

Good point, you're right.

`archive_mode = always` only re-archives WAL the standby already received. It does not make the standby create new WAL. A designated primary stays in recovery mode as long as it follows the replica cluster, so it does not generate WAL either. `pg_stat_wal` would be empty there too, just like on any other standby. My archive_mode argument was wrong, I'm dropping it.

One smaller concern remains. `IsPrimary()` just checks if `standby.signal` is missing. Right after promotion, there may be a short moment where this is already true, but Postgres has not fully started as primary yet. I have not confirmed this happens, so it's only a guess for now.

Either way, checking primary vs replica is just a proxy for the real condition (is `stats_reset` set). Proxies like this often cause problems later. #11207 fixes the real problem directly, so I prefer that approach.

PR #9802, opened 2026-01-26
test(e2e): increase timeout for recovery CHECKPOINT in disk space test

mnencia:
After crash recovery on network-attached storage, PostgreSQL may need more time to complete CHECKPOINT operations. The default 10-second timeout was insufficient on cloud environments like AKS with Rook-Ceph storage, causing the recovery phase of the disk space test to fail.

The test failure occurred in the recovery phase, not the detection phase. After resizing the PVC and waiting for the pod to become ready, the recovery CHECKPOINT command timed out after only 10 seconds. On network-attached storage with higher latency, PostgreSQL needs more time to complete CHECKPOINT operations after crash recovery.

This change uses the configurable ClusterIsReadyQuick timeout (300 seconds by default) to give PostgreSQL enough time to complete the recovery CHECKPOINT. The detection CHECKPOINT remains unchanged as it expects an error regardless of whether it fails immediately or times out.

### DIFF
--- a/tests/e2e/disk_space_test.go
+++ b/tests/e2e/disk_space_test.go
@@ -176,9 +176,8 @@ var _ = Describe("Volume space unavailable", Label(tests.LabelStorage), func() {
 			}).WithTimeout(10 * time.Minute).Should(BeTrue())
 		})
 		By("writing some WAL", func() {
-			// After crash recovery on network storage, PostgreSQL may need
-			// more time to complete CHECKPOINT operations.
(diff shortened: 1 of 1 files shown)


mnencia:
This test is only executed on cloud providers, so I started an E2E on the EDB fork with d=schedule: https://github.com/EnterpriseDB/cloudnative-pg/actions/runs/21357920163

NiccoloFei (approved):

mnencia:
/ok-to-merge The specific e2e test has passed in all the jobs

armru:
/ok-to-merge no substantial changes, just comment removal

armru (approved):

PR #9823, opened 2026-01-30
Cross namespace `Database` object

### LINKED ISSUE #6043: [Feature]: allow cross-namespace Database and Role configuration

werwolfby:
# feat(database): allow cross-namespace Database configuration

Closes #6043

## Summary

This PR adds support for cross-namespace Database resources, allowing Database objects to reference Clusters in different namespaces.
This is useful for organizations that want to share a PostgreSQL cluster among applications deployed in separate namespaces.

### API Changes

- Added `EnableCrossNamespaceDatabases` field to `ClusterSpec` to opt-in for cross-namespace Database management
- Changed `ClusterRef` in `DatabaseSpec` from `LocalObjectReference` to new `ClusterObjectReference` type that includes an optional `Namespace` field

### RBAC Changes

When `EnableCrossNamespaceDatabases` is enabled, a `ClusterRole` and `ClusterRoleBinding` are created to grant the instance manager cross-namespace access to Database resources

> [!NOTE]
> **Implementation note**:
> To simplify the implementation, a single ClusterRole is created per cluster (not per CloudNativePG installation).
> Since we already need to create a ClusterRoleBinding for each cluster, creating the ClusterRole alongside it adds minimal overhead and will only be created once per cluster

### Validation

**Cluster webhook:**
- Prevents disabling `enableCrossNamespaceDatabases` once it has been enabled

**Database webhook:**
- When creating or updating a Database that references a Cluster in a different namespace, the webhook fetches the target Cluster and verifies that `enableCrossNamespaceDatabases: true` is set
- If the Cluster does not exist, the validation fails with a "not found" error
- If the Cluster exists but does not have `enableCrossNamespaceDatabases` enabled, the validation fails with a "forbidden" error

> [!NOTE]
> **Important**:
> Once `enableCrossNamespaceDatabases` is enabled on a Cluster, it cannot be disabled.
> This is intentional to prevent orphaning Database resources in other namespaces.
> Atomically verifying that no cross-namespace Database resources exist before allowing the change is not feasible,
> as new Database objects could be created after the check but before the update completes

### Files Changed

- `api/v1/cluster_types.go` - Added `EnableCrossNamespaceDatabases` field
- `api/v1/common_types.go` - Added `ClusterObjectReference` type
- `api/v1/database_types.go` - Changed `ClusterRef` to use `ClusterObjectReference`
- `api/v1/database_funcs.go` - Updated cluster reference handling
- `pkg/specs/roles.go` - Added ClusterRole and ClusterRoleBinding creation functions
- `internal/controller/cluster_create.go` - Create/delete RBAC resources based on feature flag
- `internal/webhook/v1/cluster_webhook.go` - Prevent disabling cross-namespace databases
- `internal/webhook/v1/database_webhook.go` - Validate cross-namespace references
- `internal/management/controller/database_controller.go` - Support cross-namespace cluster lookups

## Usage Example

1. Enable cross-namespace databases on the Cluster:

```yaml
kind: Cluster
metadata:
  name: my-cluster
  namespace: postgres
spec:
  enableCrossNamespaceDatabases: true
  instances: 3
  # ... other spec fields
```

2. Create a Database in a different namespace:

```yaml
kind: Database
metadata:
  name: my-database
  namespace: my-app
spec:
  cluster:
    name: my-cluster
    namespace: postgres
  name: mydb
  owner: app
```

## Security Considerations

- Cross-namespace access is opt-in and disabled by default
- The feature requires explicit configuration on the Cluster resource
- Only clusters with `enableCrossNamespaceDatabases: true` will accept Database resources from other namespaces

## Testing

- Added unit tests for ClusterRole and ClusterRoleBinding creation
- Added webhook validation tests for cross-namespace database references:
  - Validates that same-namespace Database references work without checking the cluster
  - Validates that cross-namespace references are allowed when cluster has `enableCrossNamespaceDatabases: true`
  - Validates that cross-namespace references are rejected when cluster has `enableCrossNamespaceDatabases: false`
  - Validates that cross-namespace references are rejected when cluster does not exist
- Added tests for preventing `enableCrossNamespaceDatabases` from being disabled
- Updated existing Database controller tests for cross-namespace support
- Tested manually with `setup-cluster.sh` script in Kind cluster


werwolfby:
Tested like this:

```bash
$> ./hack/setup-cluster.sh create load deploy
... redacted for brivity...

$> kubectl create ns postgres
namespace/postgres created

$> kubectl create ns app1
namespace/app1 created

$> kubectl create ns app2
namespace/app2 created

$> kubectl apply -f - <<'EOF'
kind: Cluster
metadata:
  name: my-cluster
  namespace: postgres
spec:
  instances: 1
  enableCrossNamespaceDatabases: true
  storage:
    size: 1Gi
EOF
cluster.postgresql.cnpg.io/my-cluster created

$> kubectl apply -f - <<'EOF'
kind: Database
metadata:
  name: mydb
  namespace: app1
spec:
  name: mydb
  cluster:
    name: my-cluster
    namespace: postgres
EOF
database.postgresql.cnpg.io/mydb created

$> kubectl apply -f - <<'EOF'
kind: Database
metadata:
  name: mydb
  namespace: app2
spec:
  name: mydb
  cluster:
    name: my-cluster
    namespace: postgres
EOF
database.postgresql.cnpg.io/mydb created

$> kubectl get database -A
NAMESPACE   NAME   AGE   CLUSTER      CLUSTER NAMESPACE   PG NAME   APPLIED   MESSAGE
app1        mydb   11s   my-cluster   postgres            mydb
app2        mydb   3s    my-cluster   postgres            mydb

$> "After some waiting... around 6m..."

$> kubectl get cluster -A
NAMESPACE   NAME         AGE    INSTANCES   READY   STATUS                     PRIMARY
postgres    my-cluster   6m7s   1           1       Cluster in healthy state   my-cluster-1

$> kubectl get database -A
NAMESPACE   NAME   AGE     CLUSTER      CLUSTER NAMESPACE   PG NAME   APPLIED   MESSAGE
app1        mydb   6m4s    my-cluster   postgres            mydb      true
app2        mydb   5m56s   my-cluster   postgres            mydb      true
```

jmealo (changes requested):
Thanks for working on this. The use case is real, and the opt-in approach makes sense. I reviewed this with a focus on runtime behavior and lifecycle safety. I had assistance from a review panel composed of Claude, Codex, and Gemini in preparing this review.

I currently see three must-fix issues before I’d be comfortable merging. (I am a contributor, not a maintainer, so I do not speak for the project; these are my own opinions.)

## Must-fix (blocking)

### 1. Cluster-scoped RBAC resources are not cleaned up on Cluster deletion

`ClusterRole`/`ClusterRoleBinding` created by cross-namespace RBAC reconcile are not deleted in the deletion path.

- Deletion short-circuit path:  
  https://github.com/cloudnative-pg/cloudnative-pg/blob/c3fd638aaf62f75ceefe05d439f8e92ceb3b9dd7/internal/controller/cluster_controller.go#L166-L197
- RBAC create/delete logic exists but is called from create/reconcile flow, not deletion branch:  
  https://github.com/cloudnative-pg/cloudnative-pg/blob/c3fd638aaf62f75ceefe05d439f8e92ceb3b9dd7/internal/controller/cluster_create.go#L1053-L1120

Why this matters:
- Cluster-scoped resources cannot rely on namespace-scoped owner references for GC.
- Deleting opted-in clusters leaks cluster-scoped RBAC objects over time.

Suggested fix:
- Add explicit RBAC cleanup in cluster deletion flow (or a finalizer path that guarantees `deleteCrossNamespaceDatabaseRBAC` runs before cluster removal).

---

### 2. Instance manager watches `Database` across all namespaces unconditionally

Current cache config removes namespace scoping for `Database` for all instance managers:

- https://github.com/cloudnative-pg/cloudnative-pg/blob/c3fd638aaf62f75ceefe05d439f8e92ceb3b9dd7/internal/cmd/manager/instance/run/cmd.go#L193-L213

`Publication`/`Subscription` remain namespace-scoped, but `Database` is opened cluster-wide always.

Why this matters:
- **Possible crash-loop due to missing cluser-wide RBAC when this feature is not enabled.**
- Possible memory/event-load regression at scale.
- Cross-tenant churn in one namespace can wake reconcilers in others.
- Least-privilege/rbac assumptions become harder to reason about for clusters that never opted in.

Suggested fix:
- Keep namespaced watch by default, and only widen scope when cross-namespace support is effectively enabled for that cluster.

---

### 3. Cross-namespace `Database` finalizers are not handled on cluster delete

Owned-resource deletion helper lists `Database` only in cluster namespace:

- https://github.com/cloudnative-pg/cloudnative-pg/blob/c3fd638aaf62f75ceefe05d439f8e92ceb3b9dd7/internal/controller/finalizers_delete.go

With cross-namespace databases, dependent `Database` CRs may live outside cluster namespace and won’t be processed during cluster deletion.

Why this matters:
- Can leave cross-namespace `Database` resources stuck with finalizers / stale status during teardown.

Suggested fix:
- Include cross-namespace `Database` resources in deletion cleanup (cluster-wide list with precise matching by referenced cluster name+namespace, or indexed query if available).

## Follow-up / design concerns (non-blocking for this PR)

### A. Authorization model granularity / reversibility

I think there is a worthwhile follow-up discussion around a `ReferenceGrant`-like pattern (target-namespace grants to specific source namespaces) instead of a coarse boolean gate.

Relevant references:
- https://gateway-api.sigs.k8s.io/api-types/referencegrant/
- https://gateway-api.sigs.k8s.io/geps/gep-709/

I’m not asking to block this PR on that design change, but I’d appreciate maintainer guidance on intended direction.

### B. `validateCrossNamespaceCluster` fail-open on non-`NotFound` errors

- https://github.com/cloudnative-pg/cloudnative-pg/blob/c3fd638aaf62f75ceefe05d439f8e92ceb3b9dd7/internal/webhook/v1/database_webhook.go#L150-L180

Probably an availability trade-off, but worth documenting clearly (or re-validating in controller path).

### C. CRD schema detail: `x-kubernetes-map-type: atomic` on `.spec.cluster`

Worth verifying whether switching to `ClusterObjectReference` unintentionally changed server-side apply semantics.

## Request for maintainer guidance

Could maintainers confirm preferred scope for this PR?

1. Fix the three runtime/lifecycle blockers here and keep broader authorization design for follow-up.
2. Expand this PR to include stronger source-namespace authorization semantics.
3. Should this be handled once `DatabaseRole` lands, so we can apply one cohesive authorization strategy for both?

hypery2k:
@werwolfby any chance to get this PR ready? I really appreciate this feature. If you don't have the time to incooperate the changes I can takeover if you give me permission to write in your fork

tiran133:
Is this still being worked on? I'm interested in this too!

mostpinkest:
Worth pointing out this PR doesn't actually close #6043 because it only covers Databases. The Roles will still need to be addressed separately

werwolfby:
Sorry, at some point I gave up to check the status of the PR 😢 

And just decided to check today, I'll address the comments, will rebases to latest, etc. Sorry for delay.

And yes, I'm waiting for this feature as well 😏

werwolfby:
@jmealo thanks for the thorough review — the runtime/lifecycle focus caught real problems, and all three blockers plus the two verifiable follow-ups are now addressed. The branch has also been rebased onto the latest `main`.

Up front, for transparency: I used **Claude Opus 5** to work through this review, and I reviewed every change manually before committing. Each finding was re-verified against the current code rather than taken at face value — worth noting because the review was written against `c3fd638`, and the rebase moved the branch a long way forward.

---

## 1. Cluster-scoped RBAC not cleaned up on Cluster deletion — fixed

**`fix(database): delete cross-namespace RBAC on cluster deletion`**

Confirmed: `reconcileCrossNamespaceDatabaseRBAC` was only reachable from the create/patch flow, and the deletion branch returned before anything cleaned up the `ClusterRole`/`ClusterRoleBinding`. Being cluster-scoped, they can't be garbage collected through the namespaced owner reference of the Cluster, so they leaked on every deletion of an opted-in cluster.

They're now deleted explicitly while reconciling the deletion. One detail worth flagging: there is no finalizer on the `Cluster` object, so by the time that reconcile runs the Cluster is usually already gone (`cluster == nil`). The cleanup therefore derives the role name from the namespaced name of the request rather than from the object, and recording an event on the Cluster is optional for the same reason.

Covered by tests for the cleanup without a recorder object and for the no-op when the resources were never created.

## 2. Instance manager watching `Database` across all namespaces — fixed

**`fix(database): widen the Database watch only when opted in`**

Confirmed, including the crash-loop you predicted. Cluster-wide `list`/`watch` on `databases` is granted only by the `ClusterRole` created for opted-in clusters; the namespaced `Role` grants it just in the cluster namespace. So every instance manager of a cluster that never opted in was being asked to establish a watch it had no permission for, and would have failed to start its cache.

The manager now reads its Cluster before it is built and keeps the `Database` watch scoped to the cluster namespace unless cross-namespace support is enabled.

**`fix(database): restart the instance manager when the watch scope changes`**

This surfaced a follow-on gap: the cache scope of a controller-runtime manager is fixed when the manager is built, and the setting is not part of the pod specification, so it does not trigger a rollout. Enabling the feature on a running cluster would have left the cross-namespace `Database` objects unreconciled, with no signal, until the pods happened to be recreated.

The instance manager now watches its own Cluster and, when the setting no longer matches the one it started with, replaces its own process so it rebuilds the cache with the new scope. This reuses the existing online-upgrade mechanism, so **PostgreSQL keeps running across the restart** — no downtime, no switchover.

## 3. Cross-namespace `Database` finalizers on cluster delete — fixed

**`fix(database): notify the deletion to cross-namespace databases`**

The owned resources were listed only in the namespace of the cluster, so cross-namespace `Database` objects were never notified and were left with their finalizer and a stale status.

They are now looked up in every namespace. Widening the lookup alone would have been worse than the original bug, so the match also compares the **namespace of the referenced cluster**, not just its name — otherwise deleting `ns-a/cluster` would have stripped finalizers from a `Database` in another namespace pointing at its own local `ns-b/cluster`. Both behaviours are pinned by tests; I checked the negative one actually fails without the namespace comparison rather than passing vacuously.

**`perf(database): look up the databases of a cluster through an index`**

Following your "or indexed query if available" note: `Database` objects are indexed by the namespaced name of the cluster they reference, so the deletion path does a lookup instead of scanning every cached `Database`. Kept as a separate commit so it can be dropped independently if you consider it noise.

## A. `ReferenceGrant`-like authorization — deferred

Agreed this deserves its own discussion, and I'd rather not fold a design change of that size into this PR. Deferring pending maintainer guidance on your scope question.

## B. `validateCrossNamespaceCluster` failing open — fixed

**`fix(database): reject the cross-namespace databases that cannot be verified`**

Looking at it closely, the inconsistency was the real problem: a `Database` naming a cluster that does not exist was rejected, while one whose cluster could not be read at all was admitted. The more uncertain case got the more permissive outcome, on a check that authorizes one namespace to attach to a cluster in another.

It now rejects. The availability cost is smaller than the original comment implied: the validation returns early for `Database` objects referencing a cluster in their own namespace, so it can only ever affect cross-namespace admission — a deliberate and infrequent operation — and never ordinary database operations. There's a test asserting exactly that, i.e. same-namespace databases are still admitted even when every `Get` fails.

Worth noting the two fixes compose: with the watch scope corrected in (2), a wrongly admitted cross-namespace `Database` is no longer reconciled by a cluster that never opted in. So this change closes a consistency gap rather than a live hole in the current code.

## C. `x-kubernetes-map-type: atomic` on `.spec.cluster` — confirmed and fixed

Good catch, this was a real regression. `corev1.LocalObjectReference` is marked `+structType=atomic`, and switching to `ClusterObjectReference` silently dropped `x-kubernetes-map-type: atomic` from the schema, so server-side apply would have treated the reference as a granular struct and let two field managers own `name` and `namespace` separately. The type now carries `+structType=atomic` and the manifests are regenerated.

While verifying it, one other intentional schema change is worth calling out explicitly: `name` is now **required**, and the inherited `default: ""` is gone. `LocalObjectReference` carries that default only for backwards compatibility, and it would have defeated the `required` constraint, since CRD defaulting runs before validation. Stored objects are not re-validated until they are next written, and an empty cluster name could never have worked, so the practical impact should be nil — but it is a validation tightening on a shipped v1 API and I did not want it to pass unnoticed.

api/v1/generic_funcs.go
```diff
@@ -38,10 +37,15 @@ func ensureManagedResourceExclusivity[T managedResourceComparer](t1 T, list []T)
 			continue
 		}
 
+		// Check if both resources reference the same cluster (name and namespace)
 		if t1.GetClusterRef().Name != t2.GetClusterRef().Name {
 			continue
 		}
 
+		if t1.GetClusterNamespace() != t2.GetClusterNamespace() {
```
jmealo, line 45: This namespace comparison is correct, but the caller still lists candidates only from `resource.GetNamespace()`. A `Database` in `app1` therefore cannot see a `Database` in `app2` targeting the same Cluster and `spec.name`; both can become `applied: true`, as the manual test in the PR thread already demonstrates. With `databaseReclaimPolicy: delete`, deleting either object can drop the shared PostgreSQL database out from under the other tenant. Please register an equivalent target-Cluster index in the instance manager and use an all-namespace indexed lookup for Database exclusivity.

internal/cmd/manager/instance/run/watchscope.go
```diff
@@ -0,0 +1,125 @@
...
+				"the upgrade will apply the new configuration")
+		return ctrl.Result{}, nil
+	}
+
+	contextLogger.Info(
+		"Cross-namespace databases support changed, restarting the instance manager to apply it",
+		"enableCrossNamespaceDatabases", cluster.Spec.EnableCrossNamespaceDatabases)
+
+	// This does not return when it succeeds, as the process is replaced
+	if err := r.restart(r.cancelFunc, r.exitedCondition); err != nil {
```
jmealo, line 111: The `false -> true` transition races the operator's RBAC reconciliation. If this process replacement happens first, the new manager's cluster-wide Database informer receives `403` and cannot sync. The expected controller-runtime path is repeated reflector retries while cache startup remains blocked; because the remaining runnables include the probe webserver, a sufficiently long RBAC delay can cause liveness failures and a container/PostgreSQL restart. Please verify that cluster-wide Database access is usable before replacing the process—via `SelfSubjectAccessReview` or a live cluster-wide list—and requeue until it is. An e2e test should characterize and prevent the live failure mode.

internal/controller/cluster_create.go
```diff
@@ -1121,6 +1126,134 @@ func (r *ClusterReconciler) createRoleBinding(ctx context.Context, cluster *apiv
...
+// needed for cross-namespace Database support based on the cluster's configuration
+func (r *ClusterReconciler) reconcileCrossNamespaceDatabaseRBAC(ctx context.Context, cluster *apiv1.Cluster) error {
+	contextLogger := log.FromContext(ctx)
+
+	if cluster.Spec.EnableCrossNamespaceDatabases {
+		return r.createCrossNamespaceDatabaseRBAC(ctx, cluster)
+	}
+
+	// Feature is disabled, clean up any existing resources
+	if err := r.deleteCrossNamespaceDatabaseRBAC(ctx, cluster); err != nil {
```
jmealo, line 1139: When the feature is off, this still enters the deletion helper, whose cached `Get`s dynamically start cluster-wide informers for every ClusterRole and ClusterRoleBinding. That makes every operator cache all cluster RBAC even when nobody enables this feature. Because `true -> false` is forbidden and this feature has not shipped, please return without those reads when disabled; also put both RBAC types in `Client.Cache.DisableFor` so enabled-cluster reconciliation uses direct reads without introducing global caches.

pkg/specs/roles.go
```diff
@@ -321,6 +321,74 @@ func CreateRole(opts RoleOptions) rbacv1.Role {
...
+func GetCrossNamespaceDatabaseRoleName(cluster apiv1.Cluster) string {
+	return GetCrossNamespaceDatabaseRoleNameFor(cluster.Namespace, cluster.Name)
+}
+
+// GetCrossNamespaceDatabaseRoleNameFor returns the name of the ClusterRole used
+// for cross-namespace Database access by the cluster having the passed namespace
+// and name. It allows computing the name when the Cluster object is no longer
+// available, such as while reconciling its deletion.
+func GetCrossNamespaceDatabaseRoleNameFor(namespace, name string) string {
+	return "cnpg-" + namespace + "-" + name + "-cross-ns-db"
```
jmealo, line 335: This encoding is ambiguous because `-` is valid in both components: `app-db/one` and `app/db-one` both produce `cnpg-app-db-one-cross-ns-db`. This fails closed rather than granting cross-tenant access: the second Cluster finds the first Cluster's binding, receives no binding for its own ServiceAccount, and its instance manager takes `403`s. Deleting either Cluster can then remove the survivor's shared-name RBAC. Please use an unambiguous encoding or a stable hash of the namespaced name, and verify the ownership labels before reusing or deleting an existing object.

internal/controller/cluster_controller.go
```diff
@@ -292,6 +272,67 @@ func (r *ClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+
+	// The RBAC resources supporting cross-namespace Database objects are
+	// cluster-scoped, so they are not garbage collected together with the
+	// Cluster and need to be deleted explicitly. The Cluster may already be
+	// gone at this point, in which case no event can be recorded on it.
+	var recorderObject client.Object
+	if cluster != nil {
+		recorderObject = cluster
+	}
+	if err := r.deleteCrossNamespaceDatabaseRBACByKey(ctx, req.NamespacedName, recorderObject); err != nil {
```
jmealo, line 323: This cleans up correctly when this reconciliation runs, but no Cluster finalizer guarantees that it will. If the operator is down while an opted-in Cluster with no surviving dependent resources is deleted, startup has no Cluster or dependent-resource event to enqueue this key, so both cluster-scoped RBAC objects leak. Please add a Cluster finalizer or a startup/periodic sweep of the labeled RBAC objects. A direct unit call to this function is not sufficient coverage; the failure is the absence of any reconcile request after restart.

internal/management/controller/common.go
```diff
@@ -158,7 +157,7 @@ func markAsReady(
 // object reference.
 type clusterScopedResource interface {
 	client.Object
-	GetClusterRef() corev1.LocalObjectReference
+	GetClusterRef() apiv1.ClusterObjectReference
```
jmealo, line 160: Changing the reference type makes cross-namespace Databases possible, but `mapClusterToManagedResources` below still lists only `client.InNamespace(obj.GetNamespace())`. A remote Database is therefore not enqueued when its Cluster is demoted or promoted and can retain stale `applied: true` state until an unrelated event or informer resync. Please use the same all-namespace target-Cluster index needed for the exclusivity fix while keeping the other resource kinds namespace-scoped.

internal/controller/cluster_controller.go
```diff
@@ -149,6 +154,8 @@ var ErrNextLoop = utils.ErrNextLoop
 // +kubebuilder:rbac:groups=postgresql.cnpg.io,resources=clusters/status,verbs=get;watch;update;patch
 // +kubebuilder:rbac:groups=rbac.authorization.k8s.io,resources=rolebindings,verbs=create;patch;update;get;list;watch
 // +kubebuilder:rbac:groups=rbac.authorization.k8s.io,resources=roles,verbs=create;patch;update;get;list;watch
+// +kubebuilder:rbac:groups=rbac.authorization.k8s.io,resources=clusterroles,verbs=create;delete;patch;update;get;list;watch
+// +kubebuilder:rbac:groups=rbac.authorization.k8s.io,resources=clusterrolebindings,verbs=create;delete;patch;update;get;list;watch
```
jmealo, line 158: This permanently grants every installation authority to create, mutate, delete, and bind cluster-scoped RBAC, regardless of whether the feature is enabled. That is a meaningful increase in operator compromise blast radius and needs explicit maintainer acceptance. A statically installed shared ClusterRole, with the operator managing only per-Cluster bindings, would at least remove the need to mint and update ClusterRoles dynamically.

internal/controller/cluster_create.go
```diff
@@ -1121,6 +1126,134 @@ func (r *ClusterReconciler) createRoleBinding(ctx context.Context, cluster *apiv
...
+		patchedRole.Rules = clusterRole.Rules
+		if err := r.Patch(ctx, patchedRole, client.MergeFrom(&existingRole)); err != nil {
+			return fmt.Errorf("while patching ClusterRole: %w", err)
+		}
+	}
+
+	// Create ClusterRoleBinding
+	clusterRoleBinding := specs.CreateCrossNamespaceDatabaseRoleBinding(*cluster)
+	var existingBinding rbacv1.ClusterRoleBinding
+	if err := r.Get(ctx, client.ObjectKey{Name: clusterRoleBinding.Name}, &existingBinding); err != nil {
```
jmealo, line 1177: An existing binding is accepted without reconciling its Subjects or labels. Please patch those fields back to the desired values. `roleRef` is immutable, so if it differs the safe repair is delete-and-recreate rather than patch. The same ownership-label verification should protect the deletion path from removing an object that merely happens to have the computed name.

internal/webhook/v1/database_webhook.go
```diff
@@ -40,7 +43,7 @@ var databaseLog = log.WithName("database-resource").WithValues("version", "v1")
 // SetupDatabaseWebhookWithManager registers the webhook for Database in the manager.
 func SetupDatabaseWebhookWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewWebhookManagedBy(mgr, &apiv1.Database{}).
-		WithValidator(newBypassableValidator[*apiv1.Database](&DatabaseCustomValidator{})).
+		WithValidator(newBypassableValidator[*apiv1.Database](&DatabaseCustomValidator{client: mgr.GetClient()})).
```
jmealo, line 46: This is an authorization decision, so the cached client can produce stale rejection immediately after enabling the flag. Under `WATCH_NAMESPACE`, it can also return an opaque `unknown namespace for the cache` error for an existing Cluster outside the configured cache. Please use `mgr.GetAPIReader()` here and type the validator dependency as `client.Reader`.

internal/webhook/v1/database_webhook.go
```diff
@@ -119,7 +126,78 @@ func (v *DatabaseCustomValidator) ValidateUpdate(
 		database.Name, allErrs)
 }
 
-func (v *DatabaseCustomValidator) validateDatabaseChanges(_ *apiv1.Database, _ *apiv1.Database) field.ErrorList {
+func (v *DatabaseCustomValidator) validateDatabaseChanges(newDB, oldDB *apiv1.Database) field.ErrorList {
+	var allErrs field.ErrorList
+
+	// Prevent changing cluster.namespace after creation
+	if oldDB.Spec.ClusterRef.Namespace != "" &&
```
jmealo, line 133: The CRD's `self == oldSelf` CEL rule already makes the entire Cluster reference immutable before validating admission runs, so this branch is unreachable through the API. It is also weaker on its own because it allows empty-to-non-empty namespace changes. I would remove it; if it is intentionally retained as defense against a stale CRD, drop the `oldDB.Spec.ClusterRef.Namespace != ""` condition so it matches the CEL semantics.

api/v1/database_types.go
```diff
@@ -61,7 +60,7 @@ const (
 type DatabaseSpec struct {
 	// The name of the PostgreSQL cluster hosting the database.
 	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="cluster reference is immutable after creation"
-	ClusterRef corev1.LocalObjectReference `json:"cluster"`
+	ClusterRef ClusterObjectReference `json:"cluster"`
```
jmealo, line 63: Changing this public field type is source-breaking for consumers constructing `DatabaseSpec`; the accompanying `GetClusterRef()` signature changes also break callers for Database, DatabaseRole, Publication, and Subscription. Please get an explicit maintainer compatibility decision and remove the patch-release backport labels. If feasible, preserve the existing method signatures and expose the target namespace through a new method.

internal/webhook/v1/database_webhook.go
```diff
@@ -119,7 +126,78 @@ func (v *DatabaseCustomValidator) ValidateUpdate(
...
+	ctx context.Context,
+	db *apiv1.Database,
+) field.ErrorList {
+	// If not cross-namespace, no validation needed
+	if !db.IsCrossNamespace() {
+		return nil
+	}
+
+	// If no client is available (e.g., in unit tests), skip this validation
+	if v.client == nil {
```
jmealo, line 157: The production admission guard constructs this validator without a client, so this validation is skipped when the webhook is absent. The corrected watch scope means an opted-out Cluster still cannot observe or reconcile that remote object, so I do not see this as an active authorization bypass. It is nevertheless inconsistent to admit an object the webhook would reject; please thread a `client.Reader` into the guard or document why the guard intentionally omits this check.

jmealo (changes requested):
Thanks for the quick turnaround and the detailed response. I re-reviewed the current head. The unconditional cluster-wide watch and cross-namespace deletion-notification blockers are fixed, as are the fail-open API-read behavior and the lost atomic-map semantics. The RBAC cleanup now works when a deletion reconciliation occurs, but it is not yet durable across a missed deletion event.

I still see several blockers in the inline comments, most importantly the cross-namespace exclusivity gap: two `Database` resources in different namespaces can claim the same PostgreSQL database, and deleting either with `databaseReclaimPolicy: delete` can drop it out from under the other tenant.

Before merge, I also think this needs:

- an explicit maintainer decision on permanently granting the operator create/delete/update authority over cluster-scoped RBAC;
- removal of the backport labels—the public Go API change is source-breaking, and #6043 explicitly says not to backport;
- e2e coverage for enabling the feature on a running Cluster and for RBAC lifecycle across deletion/operator downtime;
- user-facing documentation for the security model, irreversible opt-in, and cluster-scoped RBAC;
- `Partially addresses #6043` instead of `Closes #6043`, because cross-namespace `DatabaseRole` support remains out of scope; and
- the disclosure requested by [section 5 of the project's AI policy](https://github.com/cloudnative-pg/governance/blob/main/AI_POLICY.md#5-disclosure): a note in the PR description and `Assisted-by: Claude Opus 5` trailers on the commits where that assistance applied.

Please add failing regression tests before fixing the inline correctness findings, and where practical land the red tests separately so each is shown to fail for the stated reason. In particular, the deletion/restart test must use an opted-in Cluster with no surviving dependent resources, otherwise the existing initial-list recovery path masks the leak. The watch-scope transition and deletion-during-operator-downtime paths also need e2e coverage because fake clients cannot exercise their process-lifecycle behavior.

I cross-checked this follow-up with Claude and Codex. As before, I am a contributor rather than a maintainer, so the architecture and compatibility calls are maintainer decisions; the inline correctness findings are the reasons I am keeping the review at Changes Requested.

werwolfby:
@jmealo thanks for the feedback. I will take a while to fix all this.

PR #9826, opened 2026-01-30
ci: disable provenance in metadata to avoid size limits

mnencia:
Docker Buildx 0.31.0 includes full provenance attestations in the metadata output for multi-platform builds, resulting in 444KB+ JSON that exceeds bash argument list limits.

Setting BUILDX_METADATA_PROVENANCE=disabled excludes provenance from the metadata file while keeping attestations attached to images in the registry.

### DIFF
--- a/.github/workflows/continuous-delivery.yml
+++ b/.github/workflows/continuous-delivery.yml
@@ -381,6 +381,7 @@ jobs:
         uses: docker/bake-action@5be5f02ff8819ecd3092ea6b2e6261c31774f2b4 # v6
         id: bake-push
         env:
+          BUILDX_METADATA_PROVENANCE: disabled
           environment: "testing"
           buildVersion: ${{ env.VERSION }}
           tag: ${{ env.IMAGE_TAG }}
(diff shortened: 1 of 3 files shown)


armru (approved):

mnencia:
@sxd jonathis is weird, the tests are running on the wrong PR/branch

mnencia:
@sxd jonathis is weird, the tests are running on the wrong PR/branch

mnencia:
using /test is not enough to test this PR, because the E2E tests would run using the workflow definition on main that doesn't work.

I manually triggered https://github.com/cloudnative-pg/cloudnative-pg/actions/runs/21542717415, which has already passed the Build Containers step.

mnencia:
I'm writing the reason for the change here for further reference: the large metadata issue was introduced in Docker Buildx v0.31.0 by docker/buildx#3536, which fixed a bug where provenance metadata was missing for multi-platform builds. The fix correctly includes per-platform provenance entries in the metadata output, but this results in 400KB+ JSON that exceeds bash argument limits.

Setting `BUILDX_METADATA_PROVENANCE=disabled` excludes provenance from the metadata file while keeping attestations attached to images in the registry.

See https://docs.docker.com/reference/cli/docker/buildx/bake/#metadata-file for further references

sxd:
Maybe we should open an issue for possible related issue on https://github.com/cloudnative-pg/postgres-containers ? what do you think @mnencia ?

sxd (approved):

PR #9827, opened 2026-01-30
fix: handle OCI CSI 409 Conflict error

### LINKED ISSUE #8608: [Bug] VolumeSnapshot on OKE/OCI fails with HTTP 409 “backup in progress”

### Version

1.27 (latest patch)

1.32

Cloud: Other

Helm

### What happened?

We’re validating CloudNativePG (CNPG) for production on Oracle Cloud Infrastructure (OCI) via Oracle Kubernetes Engine (OKE) and we’re hitting a problem where VolumeSnapshot backups intermittently fail with HTTP 409 Conflict from the OCI Block Volume CSI — even though the CSI then finds the same snapshot, waits, and reports it ready moments later.

Per the docs and recent changes, timeouts and some driver-specific errors are treated as retryable during volumeSnapshot backups; we believe this 409 case should be handled the same way. 

**What happened**

Triggering a backup with `method: volumeSnapshot` makes the CNPG Backup go to `status.phase=failed` as soon as the CSI’s `CreateVolumeBackup` returns 409 Conflict:

> “There is currently a backup operation in progress for Volume …”.

However, the OCI CSI immediately detects that the same snapshot already exists, polls GetVolumeBackup until it is Available, and then returns `ready_to_use: true`.

So the 409 is a transient provider state, and the snapshot is actually created and usable, but the CNPG Backup fails earlier because the 409 is not considered retryable within the `volumeSnapshotDeadline`.

**OCI CSI controller log** 
```
# 1) CreateVolumeBackup -> 409 Conflict (another/own backup in progress)
... ERROR ... Could not create snapshot "snapshot-7ab0b69a-...":
Error returned by Blockstorage Service. Http Status Code: 409. Error Code: Conflict.
Message: There is currently a backup operation in progress for Volume ocid1.volume.oc1.iad....

# 2) Driver recognizes snapshot exists; starts polling for Available
... INFO  Snapshot already created, checking if lifecycleState is Available
... INFO  Checking if backup ocid1.volumebackup.oc1.iad.... has become available
... INFO  GetVolumeBackup ... statusCode: 200 (repeated every few seconds)

# 3) Snapshot becomes ready
... INFO  Snapshot is created and available.
... gRPC response ... "ready_to_use": true, "snapshot_id": "ocid1.volumebackup.oc1.iad...."
```

Meanwhile, the CNPG Backup ends as failed with a message similar to:

```
can't execute snapshot backup: failed to take snapshot of the volume ...:
"rpc error: code = Internal desc = Could not create snapshot "...":
Http Status Code: 409. Error Code: Conflict.
Message: There is currently a backup operation in progress for Volume ...
```

**What we expected**

The VolumeSnapshot reconciler should treat HTTP 409 (Conflict) from the OCI driver (when the message indicates “backup operation in progress”) as retryable, honoring the `backup.cnpg.io/volumeSnapshotDeadline` window before marking the Backup as failed. This aligns with CNPG’s documented retry behavior for recoverable errors during VolumeSnapshot backups, and with the intent of recent fixes that already made timeouts during snapshot creation retryable.

**How to reproduce (on OKE/OCI)**

Install CNPG on OKE and configure a `VolumeSnapshotClass` for OCI Block Volume (blockvolume.csi.oraclecloud.com).

In a CNPG Cluster with:
```
spec:
  backup:
    volumeSnapshot:
      className: oci-bv-snapshot-class
```

create a Backup with `method: volumeSnapshot`.

Watch the CSI logs: CreateVolumeBackup returns 409, then the driver lists/gets the backup by name and waits for Available, then returns ready_to_use=true.

Observe that the CNPG Backup is marked failed promptly after the first 409, before the CSI surfaces the snapshot as ready.

**Hypothesis & suggested fix**

- Mark HTTP 409 Conflict from OCI Block Volume as retryable, when the error text matches this provider state (e.g., “There is currently a backup operation in progress for Volume …”, Http Status Code: 409, Error Code: Conflict, CreateVolumeBackup).

- This matches existing logic that classifies timeouts and some driver-specific messages as retryable and would prevent false negatives for a snapshot that does become ready moments later. 

(Optional) Implementation hint: in the VolumeSnapshot backup reconciler’s error classification (the same area that was extended to treat timeouts as retryable), add patterns like:

- HTTP Status Code:\s*409
- Error Code:\s*Conflict
- There is currently a backup operation in progress
- optionally coupled with CreateVolumeBackup / volumeBackups

So the reconciler keeps retrying (every ~10s) until the snapshot becomes Ready or volumeSnapshotDeadline elapses.

Happy to attach the full CSI logs if useful. Thank you!

### Code of Conduct

silvio89:
## Description

This PR addresses a race condition encountered with the OCI (Oracle Cloud Infrastructure) CSI driver during VolumeSnapshot backups.

The OCI CSI driver may return a `409 Conflict` error when a backup operation is requested while another implementation-level backup is already in progress (e.g. due to internal retries). The error message typically contains:
`Error returned by Blockstorage Service. Http Status Code: 409. Error Code: Conflict.`

Previously, this error was treated as fatal, causing the backup to fail immediately (and preventing the `VolumeSnapshot` mechanism from succeeding). This change identifies this specific error signature and classifies it as **retryable**, allowing the operator to wait and retry the operation instead of failing.

## Checklist

- [x] Code compiles correctly
- [x] Unit tests added and passed (`errors_test.go`)
- [x] Verified manually on OCI environment

Fixes #8608

### DIFF
--- a/pkg/reconciler/backup/volumesnapshot/errors.go
+++ b/pkg/reconciler/backup/volumesnapshot/errors.go
@@ -79,6 +79,7 @@ func isCSIErrorMessageRetriable(msg string) bool {
 		isRetryableHTTPError,
 		isConflictError,
 		isContextDeadlineExceededError,
+		isOCIConflictError,
 	}
 
 	for _, isRetryableFunc := range isRetryableFuncs {
@@ -135,4 +136,12 @@ func isRetryableHTTPError(msg string) bool {
 	}
 
 	return false
+}
+
+// isOCIConflictError detects the 409 Conflict returned by the OCI Blockstorage
+// Service while a snapshot is still being created.
+func isOCIConflictError(msg string) bool {
+	return strings.Contains(msg, "Error returned by Blockstorage Service") &&
+		strings.Contains(msg, "Http Status Code: 409") &&
+		strings.Contains(msg, "Error Code: Conflict")
 }
--- a/pkg/reconciler/backup/volumesnapshot/errors_test.go
+++ b/pkg/reconciler/backup/volumesnapshot/errors_test.go
@@ -50,6 +50,7 @@ var _ = Describe("Retriable error messages", func() {
 		Entry("context deadline exceeded - retriable", "context deadline exceeded waiting for snapshot creation", true),
 		Entry("deadline exceeded - retriable", "deadline exceeded during Azure snapshot creation", true),
 		Entry("timed out - retriable", "operation timed out for csi-disk-handler", true),
+		Entry("OCI 409", "Error returned by Blockstorage Service. Http Status Code: 409. Error Code: Conflict.", true),
 	)
 
 	Describe("isContextDeadlineExceededError", func() {
--- a/pkg/reconciler/backup/volumesnapshot/reconciler_test.go
+++ b/pkg/reconciler/backup/volumesnapshot/reconciler_test.go
@@ -716,3 +716,94 @@ var _ = Describe("unmarshalMetadata", func() {
 		Expect(data).To(BeNil())
 	})
 })
+
+var _ = Describe("createSnapshot AlreadyExists handling", func() {
+	var (
+		ctx       context.Context
+		backup    *apiv1.Backup
+		cluster   *apiv1.Cluster
+		targetPod *corev1.Pod
+		pvc       *corev1.PersistentVolumeClaim
+	)
+
+	BeforeEach(func() {
+		ctx = context.TODO()
+		backup = &apiv1.Backup{
+			ObjectMeta: metav1.ObjectMeta{
+				Namespace: "test-namespace",
+				Name:      "test-backup",
+			},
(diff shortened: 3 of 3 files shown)


igorGevaerd (approved):

mnencia:
Closing in favor of #11132 (now merged, closing #8608). Rather than adding an OCI-specific `409 Conflict` pattern to the classifier, #11132 takes the more general route: it removes the CSI error-message classification entirely and retries every volume snapshot error until `volumeSnapshotDeadline` elapses. That covers the OCI conflict you reported here along with every other driver's own wording of a transient error, so a per-driver signature is no longer needed. The `AlreadyExists` handling in `createSnapshot` was also since reworked to be label-aware.

Thanks for the report and the fix, the OCI reproduction here was what pinned down the classifier gap.

PR #9842, opened 2026-02-03
ci(setup-cluster): add support for k3d

### LINKED ISSUE #9836: [Feature]: Add native k3d cluster setup support to CI

The current CI cluster setup script is intended to support different Kubernetes engines, but at the moment, it only supports Kind. It would be useful to add k3d support to allow for easy validation of the operator on k3s.

Add k3d support to setup-cluster.sh script.

N/A

### Additional context

_No response_

### Backport?

Yes

Yes

### Code of Conduct

danishedb:
Add k3d as a supported engine in setup-cluster.sh alongside kind, bringing back support for running e2e tests on a Rancher-based setup.

Closes #9836


hack/e2e/run-e2e-k3d.sh
```diff
@@ -0,0 +1,110 @@
...
+  # Run E2E tests
+  "${E2E_DIR}/run-e2e.sh" || RC=$?
+
+  ## Export logs
+  "${HACK_DIR}/setup-cluster.sh" export-logs
+
+  exit $RC
+}
+
+main
```
gbartolini: I think this is a bad approach. This file should not exist and ideally we should refactor the kind one to be generic enough to run on different environments, letting the 'testing-tools' part manage each engine's specificities.

hack/testing-tools/k8s-engines/k3d/load-helper-images.sh
```diff
@@ -0,0 +1,62 @@
...
+    for IMG in "${HELPER_IMGS[@]}"; do
+        echo -e "${bright}Pulling '${IMG}'${reset}"
+        docker pull "${IMG}"
+
+        echo -e "${bright}Loading '${IMG}' into k3d nodes for ${cluster_name}${reset}"
+        load_image_k3d "${cluster_name}" "${IMG}"
+    done
+
+    echo -e "${bright}Done loading helper images on cluster ${cluster_name}${reset}"
+}
```
gbartolini: Similar situation here. If every implementation of a specific engine is basically copy/paste of the other one, we end up with an increase of technical debt - which is the opposite goal of what we are trying to achieve with this initiative.

jbattiato (approved):

hack/e2e/run-e2e-suite.sh
```diff
@@ -0,0 +1,109 @@
...
+
+# renovate: datasource=github-releases depName=onsi/ginkgo
+go install github.com/onsi/ginkgo/v2/ginkgo@v2.28.1
+
+
+# Unset DEBUG to prevent k8s from spamming messages
+unset DEBUG
+
+# Build kubectl-cnpg and export its path
+make build-plugin
```
gbartolini, line 82: I am not convinced about this ... but I might have mistaken the goal of the script to run on an existing cluster.

mnencia: You are right. Existing environment: the plugin should already be available.

mnencia (approved):

PR #9650, opened 2026-01-08
fix(walrestore): prevent replicas from downloading future timeline history files

mnencia:
Replicas can crash-loop when orphaned "future timeline" .history files
exist in the WAL archive. This can occur during split-brain scenarios
or other conditions where timeline history files are created for timelines
that the cluster never officially adopts.

This fix adds validation during WAL restore to prevent replicas from
downloading timeline history files with timeline IDs greater than the
cluster's current timeline. Primary instances retain full access.

The validation works by:
- Parsing timeline ID from .history filenames (e.g., 00000022.history)
- Checking if the instance is a primary or replica
- For replicas, rejecting files where fileTimeline > clusterTimeline
- Returning "file not found" to PostgreSQL for rejected files

This prevents PostgreSQL from ever seeing the problematic history file,
allowing normal recovery to proceed. Combined with PR #9637 (auto-recovery
via re-cloning), this provides complete coverage of timeline divergence
scenarios.

Closes #4188

PR #9849, opened 2026-02-03
fix(walrestore): apply timeline validation before plugin WAL restore

mnencia:
Move timeline history file validation to execute before any WAL restore attempt (plugin or in-tree) rather than only for in-tree restores. This prevents replicas from downloading timeline history files with timeline IDs higher than the cluster's current timeline when plugins handle WAL restore.

The timeline protection added in #9650 was only applied to in-tree WAL restore, but not to plugin-based restore. This allowed the protection to be completely bypassed when using plugins, causing replicas to download future timeline history files and fail with timeline mismatch errors.

Fixes the timeline protection introduced in ef73994.

### DIFF
--- a/docs/src/cloudnative-pg.v1.md
+++ b/docs/src/cloudnative-pg.v1.md
@@ -43,14 +43,14 @@ _Appears in:_
 
 | Field | Description | Required | Default | Validation |
 | --- | --- | --- | --- | --- |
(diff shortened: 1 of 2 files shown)


mnencia:
@dosu never add the ok-to-merge. It is reserved for the CI automation, and you must never add it.

leonardoce (approved):

armru (approved):

PR #9873, opened 2026-02-04
feat: Add automatic pooler pause/resume during switchover or failover 

### LINKED ISSUE #8299: [Feature]: Unsupervised Pooler `PAUSE` during switchover

jakubhajek:
# Summary

Add automatic Pgbouncer pooler pause/resume during switchover or failover operations to minimize client connection failures. The operator drives the pause entirely through operator-owned state (status + an annotation) and never touches the user-owned Pooler `spec`, so it is safe to use with GitOps.

Fix: #8299

# Disclaimer

This feature was developed with assistance from Claude Code. Significant effort was invested in manual testing to validate the behavior across multiple scenarios including manual switchover, automatic failover, multiple poolers, and edge cases.

# Changes

## API Changes

Add the following fields to `spec.pgbouncer` in the Pooler CRD, since each pooler independently manages its
  own pause/resume lifecycle:

  - `pauseDuringSwitchover` - opt-in boolean to enable the feature (default: `false`)
  - `pauseDuringSwitchoverTimeout` - safety timeout (default: `120s`)

  Add status fields to Pooler:

  - `status.pausedForSwitchover` - tracks if this pooler is currently paused for a switchover
  - `status.pausedForSwitchoverTimestamp` - when this pooler was paused (RFC3339Micro)

  Add `cnpg.io/pausedDuringSwitchover` annotation to track auto-paused poolers.

## Controller Changes

- Pooler controller now watches Cluster objects via a predicate that fires when `CurrentPrimary` or `TargetPrimary` changes
- When a switchover/failover is detected (`currentPrimary != targetPrimary`), the pooler is marked paused via `status.pausedForSwitchover` and the `cnpg.io/pausedDuringSwitchover` annotation; the operator never writes to `spec.pgbouncer.paused`
- The pooler instance manager issues the PgBouncer `PAUSE`/`RESUME` command when either the user requested it via spec or the operator paused it for switchover via status, so a manual pause and an automatic switchover pause coexist
- Resume pooler when switchover completes (`currentPrimary == targetPrimary`)
- Timeout check runs each reconcile to force-resume if stuck, and the reconcile requeues at the timeout so the force-resume fires even if no further cluster event arrives
- Each pooler independently manages its own pause/resume state

## Documentation Changes

- Updated `docs/src/connection_pooling.md` - new "Automatic pause during switchover/failover" section explaining the feature, configuration, and behavior, including the GitOps-safe design
- Updated `docs/src/labels_annotations.md` - added `cnpg.io/pausedDuringSwitchover` annotation documentation
- Updated `docs/src/cloudnative-pg.v1.md` - updated PgBouncerSpec and PoolerStatus API reference tables

## Default Behaviour

- **Opt-in**: Feature disabled by default, no behavior change without explicit configuration
- **GitOps-safe**: The operator only writes operator-owned fields (`status.pausedForSwitchover` and the annotation). It never changes the user-owned `spec.pgbouncer.paused`, so Argo CD or Flux see no drift and do not fight the operator during a switchover
- **Selective**: Only poolers with automated integration are paused (poolers without the built-in integration are skipped)
- **Respects manual pause**: Poolers a user has paused via `spec.pgbouncer.paused` are not modified
- **Safe resume**: Only poolers we paused (identified by the annotation and status) are resumed (preserves user intent)
- **Per-pooler**: Each pooler tracks its own pause state independently
- **Timeout safety**: Force-resume after timeout (default 120s) prevents indefinite service disruption

## Configuration example

```yaml
  ---
  kind: Cluster
  metadata:
    name: test-switchover-pause
  spec:
    instances: 3
    postgresql:
      parameters:
        max_connections: "500"
    storage:
      size: 1Gi
  ---
  kind: Pooler
  metadata:
    name: test-switchover-pause-rw
  spec:
    cluster:
      name: test-switchover-pause
    type: rw
    pgbouncer:
      poolMode: session
      pauseDuringSwitchover: true
      pauseDuringSwitchoverTimeout: "30s"  # shorter for testing
  ```

## Events

  | Event | Type | Description |
  |-------|------|-------------|
  | `PoolerPaused` | Normal | Pooler paused for switchover/failover |
  | `PoolerResumed` | Normal | Pooler resumed after the operation is completed |
  | `PoolerPauseTimeout` | Warning | Force resumed pooler after timeout - switchover may not have completed
  successfully |

# How it Works

**Manual Switchover** (via `cnpg.io/targetPrimary` annotation or `kubectl cnpg promote`):
1. Cluster status changes (`currentPrimary != targetPrimary`)
2. Pooler controller is triggered via cluster watch predicate
3. Each eligible pooler is marked paused via `status.pausedForSwitchover` + the annotation (never `spec`); the pooler instance manager reads that status and issues the PgBouncer `PAUSE` command
4. Switchover proceeds
5. Once complete (`currentPrimary == targetPrimary`), the status and annotation are cleared and the instance manager issues `RESUME`

**Automatic Failover** (primary pod failure):
1. Cluster controller selects a new primary, updating `targetPrimary`
2. Pooler controller is triggered via cluster watch predicate
3. Each eligible pooler is marked paused via status + the annotation, and PgBouncer is paused
4. Failover proceeds
5. Once the new primary is healthy, each pooler resumes itself

**Timeout Protection**:
1. Each reconcile checks if the pooler has been paused beyond its configured timeout, and requeues so the check runs even without a new cluster event
2. Pooler is force-resumed to restore service
3. Warning event emitted on the pooler

# Limitations

- Each Pooler references exactly one Cluster (`spec.cluster` is a single reference). Users needing poolers for multiple clusters must create separate Pooler resources.
- Pause happens after switchover/failover is detected (when `currentPrimary != targetPrimary`), not before the primary change is initiated. Pausing before the switchover starts (seamless for planned switchover) is possible future work; it is not achievable for failover, where the old primary is already gone by the time the operator reacts.
- Only poolers with automated integration are eligible. Poolers with a custom `authQuery`, `authQuerySecret` or `serverTLSSecret` are skipped.
- In-flight queries on existing connections may still fail when the backend connection is severed.
- Some transient connection errors during the switchover window are expected.
- The pooler controller relies on the cluster watch predicate firing on status changes. If the predicate event is missed, the pooler will catch up on its next reconcile.

# Testing

## Automated tests

- **Unit tests** pass - covers: feature disabled, non-automated poolers skipped, manually paused poolers skipped, pause on switchover, resume on completion, no double-pause, timeout force-resume (keeps a marker so it does not re-pause the same stuck switchover), resume when the feature is disabled while paused, and a partial-write leftover cleanup
- **Instance manager tests** - `synchronizePause` pauses from status without touching spec, resumes when the status flag is cleared, and still honors a user pause via spec
- **E2E tests** (`tests/e2e/pgbouncer_pause_switchover_test.go`):
    - Creates cluster + pooler with `pauseDuringSwitchover: true`, triggers a switchover, proves the pause through the persisted PoolerPaused event, and verifies the pooler resumes after completion
    - Verifies manually paused poolers are not affected by the automatic pause/resume mechanism
    - Verified on Kind: the `service-connectivity` suite passed
- Each pooler independently pauses and resumes based on its own configuration
- Events emitted on individual pooler resources (`PoolerPaused`, `PoolerResumed`, `PoolerPauseTimeout`)
- `make generate manifests` - CRDs updated, no leftover Cluster CRD fields

## Manual testing scenario

Apply the following manifests:

```yaml
  ---
  kind: Cluster
  metadata:
    name: test-switchover-pause
  spec:
    instances: 3
    postgresql:
      parameters:
        max_connections: "500"
    storage:
      size: 1Gi

  ---
  kind: Pooler
  metadata:
    name: pooler-switchover-pause-rw
  spec:
    cluster:
      name: test-switchover-pause
    instances: 3
    type: rw
    pgbouncer:
      poolMode: transaction
      pauseDuringSwitchover: true
      pauseDuringSwitchoverTimeout: "30s"
      parameters:
        max_client_conn: "100"
        default_pool_size: "2"
        query_wait_timeout: "120"
        server_connect_timeout: "30"
  ---
  kind: Pooler
  metadata:
    name: pooler-switchover-pause-ro
  spec:
    cluster:
      name: test-switchover-pause
    instances: 3
    type: ro
    pgbouncer:
      poolMode: transaction
      pauseDuringSwitchover: true
      pauseDuringSwitchoverTimeout: "30s"
      parameters:
        max_client_conn: "100"
        default_pool_size: "2"
        query_wait_timeout: "120"
        server_connect_timeout: "30"
  ```

* Deploy a test pod with Python on the same Kubernetes cluster to ensure you have access to the pooler endpoints (services).
* Run the following Python script, ensure that the user password is exported as the `PGPASSWORD` environment variable.

```python
  import psycopg2
  import os
  import time

  while True:
      try:
          conn = psycopg2.connect(
              host="pooler-switchover-pause-rw",
              user="app",
              password=os.environ["PGPASSWORD"],
              dbname="app",
              connect_timeout=30,
              keepalives=1,
              keepalives_idle=30,
              keepalives_interval=2,
              keepalives_count=3
          )
          cur = conn.cursor()
          cur.execute("SELECT 1")
          ts = time.strftime("%H:%M:%S")
          print(f"OK: {ts}")
          conn.close()
      except Exception as e:
          ts = time.strftime("%H:%M:%S")
          print(f"ERR: {ts} - {e}")
      time.sleep(0.5)
  ```

* Manually trigger switchover using the CNPG plugin.
* You should notice that the script has been frozen for the time when the switchover is happening, but none of the errors were thrown. Once the switchover is complete, the script should resume execution. Once completed, the script should continue to work.
* Run another test by killing the primary to simulate failover.
* The limited number of connection errors should be thrown.

### Additional tests

On the same pod install Sysbench and run the following testing scenario:

  **Initialize a test table:**

  ```bash
  sysbench oltp_read_write \
    --db-driver=pgsql \
    --pgsql-host=pooler-switchover-pause-rw \
    --pgsql-port=5432 \
    --pgsql-user=app \
    --pgsql-password=$PGPASSWORD \
    --pgsql-db=app \
    --tables=4 \
    --table-size=10000 \
    prepare
  ```

  **Run benchmark that tolerates some errors rate:**

  ```bash
  sysbench oltp_read_write \
    --db-driver=pgsql \
    --pgsql-host=pooler-switchover-pause-rw \
    --pgsql-port=5432 \
    --pgsql-user=app \
    --pgsql-password=$PGPASSWORD \
    --pgsql-db=app \
    --tables=4 \
    --table-size=10000 \
    --threads=10 \
    --time=120 \
    --report-interval=5 \
    run
  ```


ardentperf:
nb. this feature is referenced in docs as something we want to do - https://cloudnative-pg.io/docs/1.28/connection_pooling#pausing-connections

Would be good to add some E2E tests around this

Would be good to check types (things added on cluster, should they be on pooler?)  For future direction, we don't want to close the door to having multiple cnpg clusters behind single poolers.  So pooler should know about clusters, versus other way around. Probably want to move logic from cluster controller to pooler controller. Should be possible to "watch" cluster and it will dispatch notices when it does switchovers.

Having pooler know more about clusters, it also knows more about pods, might enable other useful features in the future. It should also simplify E2E tests since there's a section dedicated to pooler.

Completely separate comment - there is interest in integrating sysbench this into cnpg plugin as another testing tool.

jakubhajek:
@ardentperf @sxd  Jeremy, Jonathan - thanks for the feedback - it seems that I have addressed all your points:

- Moved configuration and logic from Cluster CRD/controller to Pooler CRD/controller
- Pooler now watches Cluster via a predicate on CurrentPrimary/TargetPrimary changes
- Added E2E tests (pause/resume  + manually paused poolers not affected)
- Updated docs (connection_pooling.md, labels_annotations.md, API reference)

Manually tested: planned switchover, pod kill failover, multiple poolers (RW + RO), timeout force-resume, and manual pause not being overwritten.

Ready for review when you get a chance. Thank you for the opportunity to contribute to the CNPG!

charlesomer:
> - Pause happens after switchover/failover is detected (when currentPrimary != targetPrimary), not before the primary change is initiated.
> - In-flight queries on existing connections may still fail when the backend connection is severed.
> - Some transient connection errors during the switchover window are expected.

Just curious if the longer term goal for CNPG is to do the PAUSE before the switchover is initiated to prevent disconnects - to me this would be the expected behaviour. This PR is definitely a step in the right direction though. Thank you 😄

jakubhajek:
@charlesomerThe goal is to automate the process of upgrading Kubernetes worker nodes. Currently, we manually pause and resume the nodes using an additional Ansible playbook to minimise disruption to our customers. With this feature, we can automate the process, eliminating the need for any extra steps. Simply move the primary role to a replica, and the operator will handle the entire task, reducing the number of disruptions for our customers.

ardentperf:
i'm concerned with this approach. i do think that the finally merged code needs to pause connections before starting a switchover, making it seamless, and i'm worried that this code would have to be completely refactored to accomplish that

also - what is the use case for `pauseDuringSwitchoverTimeout`? i'm curious whats the reasoning about why we'd want the pause to timeout

jakubhajek:
Thanks @ardentperf — both fair questions, let me try to address them.

## On "pause before switchover"

I think it's worth separating the two paths this PR covers, because "pause-first" is only meaningful for one of them:

1. `Planned switchover` (the main use case for us - K8S node maintenance, rolling worker upgrades). The primary is healthy and we control the timing, so a `pause-first` design is feasible: signal intent ->  poolers pause and ack -> cluster controller initiates the primary change ->  resume. That seems to be feasible. 

2. `Failover` (primary unhealthy/dead). There is no "before" - by the time CNPG decides to fail over, the primary is already gone, in-flight backend connections are already broken, and pausing PgBouncer after the fact only protects new queries from  hitting a known-dead backend during the gap. Pause-first isn't really achievable here; the best you can do is what this PR does: pause as soon as the operator picks a new target.

So I read your concern as primarily applying to the switchover path (1), and I agree the seamless ideal there is `pause-first`.

For context: an earlier revision of this PR placed the configuration and logic on the Cluster CRD/controller. Based on your earlier feedback ("Cluster CRD is large enough, things should be on the Pooler, pooler should know about clusters rather than the other way around"), I moved everything to the Pooler side - that's why the current shape is pooler-driven, watching cluster status. Pause-first requires the cluster controller to coordinate with poolers (signal intent -> wait for ack -> proceed), which means either:

- A new switchover-phase signal on the Cluster status (puts coordination state back on the Cluster side), or
- The cluster controller reads pooler status and waits before swapping targetPrimary (puts pooler-awareness into the cluster controller).

Either way, some cluster-side awareness of poolers comes back. That's not necessarily wrong - but it does pull against the "keep it on the pooler" direction, so I'd want alignment on that tradeoff before refactoring.

That said, I don't think a pause-first evolution would require throwing this code away. The pooler-side machinery - annotation as ownership marker, status fields, watch + predicate, spec patching, timeout fail-safe, E2E scaffolding — all are already implemented. The change is mostly on the cluster-controller side: introducing the phase/wait coordination and keying the pooler's `reconcileSwitchoverPause` off that signal instead of currentPrimary != targetPrimary.

##  On pauseDuringSwitchoverTimeout

Short version: it's an auto-resume fail-safe. If the switchover hasn't completed (`currentPrimary == targetPrimary` again) within the configured window, the pooler force-resumes itself even though the switchover technically isn't done, and emits a Warning event (PoolerPauseTimeout) so the operator knows it ended in the fallback path rather than the happy path.

 Why it has to be automatic: a paused PgBouncer queues all client queries indefinitely. If the resume signal never arrives, clients block until their own timeouts fire - effectively a full outage on that pooler, which is strictly worse than letting the switchover-window errors happen. So the design choice is: "if something went wrong with the switchover, prefer letting clients see real errors over silently blocking them forever."

The fail-safe covers cases where the normal resume edge is missed:

- Operator pod crash/restart mid-switchover
- Network partition between operator and apiserver after pause but before resume
-  Switchover aborted by external action without normal cleanup
- A missed predicate event for the resume edge (the watch is the primary trigger; the periodic pooler reconcile is the
  backstop)

`120s` is conservative - a healthy switchover is typically well under 30s. The fail-safe stays useful (arguably more so) under a pause-first design: if the cluster controller tells poolers to pause and dies before initiating the switchover, the pooler still needs to self-recover.

Happy to hear your thoughts - and if `pause-first` orchestration is the direction you want to go, I can sketch a design for the cluster-controller side as a separate proposal.

jakubhajek:
@ardentperf thank you for the feedback. I reworked the design. Here is what the recent changes include.

* The pause is now status-driven. The operator writes only the `status.pausedForSwitchover` fields and the `cnpg.io/pausedDuringSwitchover`annotation. The operator never changes `spec.pgbouncer.paused`. 
The pooler instance manager reads the status and runs the PgBouncer `PAUSE` and `RESUME` commands. This keeps the feature safe for GitOps. Argo CD and Flux see no drift, so they do not revert the pause during a switchover.

* All the logic stays on the Pooler side, in the controller and the instance manager. 

* I also added reliability fixes. The reconcile requeues at the timeout, so the force-resume always runs, even when no new cluster event arrives. The operator marks a timed-out switchover, so it does not flap between pause and resume. If you disable the feature while a pooler is paused, or a partial write leaves state behind, the operator resumes the pooler.

* I verified the feature on Kind. The `service-connectivity` e2e suite passed, and the switchover test now asserts the `PoolerPaused` event, so it proves the pause happened and not only the resume.

* On pause before switchover: I kept "pause after detection" for now. The pooler-side machinery here carries over. A pause-first version for planned switchover mostly adds a cluster-side signal. It does not throw this work away.

* Failover has no "before", because the old primary is already gone when the operator reacts. So this is the best result for failover. I can sketch pause-first as a separate follow-up if you want it.

* On `pauseDuringSwitchoverTimeout`: it is an auto-resume fail-safe. A paused PgBouncer queues clients for an unlimited time. The resume signal can be missed. Examples are an operator restart, a missed event, or an aborted switchover. If the resume signal never arrives, the pooler resumes itself after the timeout and emits a Warning. This is better than blocking clients forever.

Issue #9149, opened 2025-11-12
[Feature]: Implement MutateCluster callback in CNPG


maxlengdell:
### Is there an existing issue already for this feature request/idea?

- [x] I have searched for an existing issue, and could not find anything. I believe this is a new feature request to be evaluated.

### What problem is this feature going to solve? Why should it be added?

The gRPC spec specifies a MutateCluster spec. But it is currently not implemented in the CNPG operator.
Such a plugin would greatly increase the customisation possibilities of the CNPG operator and thus fit more use cases.
Is there any time plan for this feature? Any desired implementation / additional details for such a PR to complete?

### Describe the solution you'd like

I'd like the MutateCluster grpc to behave as described in the spec. I.e the possibility for a plugin to modify the Cluster spec as needed. This Mutation callback shall be executed before any resource is reconciled. 

For example if a Postgres Major version upgrade is requested, force the number of replicas to 1. 

### Describe alternatives you've considered

Alternative solutions would be to have a k8 api access in the cluster to patch the cluster resource directly. Though that would interfere with the state machine of the reconciliation and risk double reconciliation / unwanted behavior.

### Backport?

No

### Are you willing to actively contribute to this feature?

Yes

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

Issue #9278, opened 2025-11-21
[Feature]: Make webhook dependency optional


maxlengdell:
### Is there an existing issue already for this feature request/idea?

- [x] I have searched for an existing issue, and could not find anything. I believe this is a new feature request to be evaluated.

### What problem is this feature going to solve? Why should it be added?

In a deployment with strict security requirements, cluster wide resources and namespaced resources are packaged separately. Cluster wide resources like CRDs which are deployed by a cluster admin and namespaced resources can then be installed with lower permissions. CNPG requires cluster-wide access by the k8 admin deploying the operator.

CNPG is currently dependent on the webhooks at startup. Deploying admission webhooks **MutatingWebhookConfiguration** and **ValidatingWebhookConfiguration** require cluster admin rights. 

- Organizations with strict RBAC policies don't grant cluster-admin permissions to deployment engineers.
- A webhook has access to all resources in the cluster of the subscribed type.
- While webhooks can be limited to specific namespaces using namespace selectors, running multiple CNPG operators on the same cluster requires complex customization to avoid webhook configuration name conflicts and still **requires cluster-admin rights for each deployment**
- Security-conscious deployments require RBAC policies set to absolute minimum. 
- This needs to work on other flavors of k8s than OpenShift

Disabling the admission controller is possible today with annotation *cnpg.io/validation: "disabled"*, but the startup dependency still exists. 

The behavior can be witnessed by deleting the webhooks, e.g **ValidatingWebhookConfiguration**, and then restart the operator pod. This makes the startup fail for the operator

The current rbac for admission registration is too permissive, which gives access to patch access to the webhook: 
```yaml
- apiGroups:
  - admissionregistration.k8s.io
  resources:
  - mutatingwebhookconfigurations
  - validatingwebhookconfigurations
  verbs:
  - get
  - patch
```

### Describe the solution you'd like

I would like CNPG to be deployable without reequiring cluster-wide access from the k8 admin.
Make webhook dependency not part of the startup check currently enforced in ensurePKI(..) at startup.
The dependency can be controlled via deployment parameter so CNPG does not check webhook availability.

The cluster validation behavior can be achieved at reconcile time via plugins or by using the existing validation logic to make up for the lack of admission control. This would mean that if someone provides a faulty cluster spec, the reconciliation would fail and set a appropriate status. 

[...]

PR #9942, opened 2026-02-09
feat: add support for CNPG-I `ValidateClusterCreate` capability

maxlengdell:
This PR adds ValidateClusterCreate callback for CNPG-I. This callback already exist in the gRPC spec but is not fully implemented. This will allow plugins to validate the cluster spec before reconciliation. For example to validate plugin configurations, or individual restrictions that one might want to impose on the cluster spec.

* Added source code changes needed to call ValidateClusterCreate before reconciliation. 
* Added unit tests to verify expected plugin behavior. 

This implementation will block reconciliation if the validation fails, or if the plugin fails to respond successfully.
The corresponding feature request: #9149.

### DIFF
--- a/api/v1/cluster_types.go
+++ b/api/v1/cluster_types.go
@@ -691,6 +691,9 @@ const (
 
 	// PhaseCannotCreateClusterObjects is set by the operator when is unable to create cluster resources
 	PhaseCannotCreateClusterObjects = "Unable to create required cluster objects"
+
+	// PhasePluginValidationFailed is set when cluster validation by a plugin fails
+	PhasePluginValidationFailed = "Cluster validation failed due to validation in plugin"
 )
 
 // EphemeralVolumesSizeLimitConfiguration contains the configuration of the ephemeral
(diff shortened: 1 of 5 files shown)


armru:
AFAIK, ValidateClusterCreate, given that is a validator should be injected in the webhook and not in the reconciliation loop, given that the cluster has been already created and can't be rejected.

maxlengdell:
Wouldnt it be nice that the plugin has a possibility to reject clusters that it doesnt approve of as well? Indeed the primary admission control should be done with webhooks. But for deployments where webhooks can not be used ( see #9278 ). 
For example; 
if CNPG is a backup & restore agnostic operator, I want to validate that the `cluster.spec.plugins[..].parameters` is correct so that any issues are caught during validation. Critical feature failures like B&R must not be caught during trial and error. How else can a plugin developer ensure that the plugin is configured correctly without creating a new webhook & admission controller?

armru:
My opinion on the matter:

First problem is that the resource is already created(ValidateClusterCreate), so the api would be end up being misleading and acting very differently based on the context.

Second problem is that even if you validate only once during the reconcile (cluster.Status.Instances == 0) it will still be vulnerable to subsquent parameters changes

What you are suggesting is more on the line of validating the cluster during every reconciliation loop, so i would say it is more of a new feature

maxlengdell:
Indeed the logic to verify the cluster is not reconciled should be improved. But the original intention with `ValidateCreateCluster` would be very useful. For example, we want to have a plugin that validates all clusters in our system contain a server certificate. So that CNPG doesnt issue one itself. By having a plugin verifying the cluster, I can prevent such a cluster being created, and the problem would be caught at deployment. Would you be interested in such a solution? If I update the PR to more cleverly check that the cluster hasnt been reconciled

PR #9970, opened 2026-02-12
chore(lint): suppress warnings for conflicting pkg with stdlib names 

armru:
Extend the golangci-lint exclusion rule to cover all packages whose names conflict with Go standard library package names, fixing lint failures introduced by the revive linter update.


mnencia (approved):

.golangci.yml
```diff
@@ -107,10 +107,10 @@ linters:
           - revive
         path: /(utils|common)/[^/]+.go
         text: avoid meaningless package names
-      # TODO: fix by renaming or removing pkg/utils/context and pkg/utils/hash packages
+      # TODO: fix by renaming packages that conflict with Go standard library names
       - linters:
           - revive
-        path: pkg/utils/(context|hash)/
+        path: (pkg/utils/(context|hash)|internal/(cmd|cnpi)/plugin|internal/cmd/manager/debug|tests/utils/exec|pkg/management/(postgres/metrics|url)|pkg/postgres/plugin)/
```
sxd: We talked about this some time ago, either we remove the linter or we just ended up having a big list was the talk back then, now we have a big list, shouldn't we just remove the linter or something that disable this specific check for everything?
Another option is to prefix the packages with something like cnpg and cnpg-i instead of adding a big exclusion, also useful when reading the code

mnencia: I don't want to disable the linter. It is valuable to have it active, preventing us from adding further packages with conflicting names. However, I don't think we should rename what we already have. That's why I find Armando's proposed solution to exclude the existing packages perfectly reasonable.

sxd: I don't find reasonable to add packages to a long list full of or and directories that isn't; easy to read to begin with and on the other side, this has been enabled for a long time, and didn't prevent to have those names.
On the other side, adding those directories doesn't exclude the linter to run any rule too? meaning that other rules will be excluded? 
Also, the same comments mention to fix by renaming, we should start to rename instead of just continue patching

mnencia: Good point about readability. I also checked the concern about the exclusion disabling other revive rules, and it doesn't happen. When a rule specifies both path and text, the exclusion only selects findings that match both conditions, so all other revive rules keep working normally on those paths.

That said, I totally agree that the single regex is becoming hard to follow. What do you think about splitting it into separate rules, one per path group?

        # TODO: fix by renaming packages that conflict with Go standard library names
        - linters:
            - revive
          path: pkg/utils/(context|hash)/
          text: avoid package names that conflict with Go standard library package names
        - linters:
            - revive
          path: internal/(cmd|cnpi)/plugin/
          text: avoid package names that conflict with Go standard library package names
        - linters:
            - revive
          path: internal/cmd/manager/debug/
          text: avoid package names that conflict with Go standard library package names
        - linters:
            - revive
          path: tests/utils/exec/
          text: avoid package names that conflict with Go standard library package names
        - linters:
            - revive
          path: pkg/management/(postgres/metrics|url)/
          text: avoid package names that conflict with Go standard library package names
        - linters:
            - revive
          path: pkg/postgres/plugin/
          text: avoid package names that conflict with Go standard library package names

Each entry is self-contained and easy to read, and adding a new one later is straightforward. Would this work for you?

mnencia: I also agree that we should start addressing the renames, but I think that deserves its own issue to be done properly. This PR is just about making main compatible with the latest golangci-lint.

PR #10002, opened 2026-02-17
chore: use bigger runners for the E2E tests

### LINKED ISSUE #10226: [Bug]: use bigger runners for heavy tasks

### Contact Details

_No response_

### Version

1.28 (latest patch)

1.35

YAML manifest

### What happened?

Some CI/CD work like building containers and running e2e tests require a lot of CPU, we can make use of the 16 cores runners available in the organization 

### Cluster resource

```shell

```

### Relevant log output

```shell

```

### Code of Conduct

sxd:
Closes #10226

### DIFF
--- a/.github/workflows/continuous-delivery.yml
+++ b/.github/workflows/continuous-delivery.yml
@@ -312,6 +312,8 @@ jobs:
       bundle_img: ${{ env.BUNDLE_IMG }}
       catalog_img: ${{ env.CATALOG_IMG }}
     steps:
+      - name: Check CPU count
+        run: nproc
       -
         name: Checkout
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
--- a/.github/workflows/continuous-delivery.yml
+++ b/.github/workflows/continuous-delivery.yml
@@ -648,7 +648,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix: ${{ fromJSON(needs.generate-jobs.outputs.kindMatrix) }}
-    runs-on: ${{ needs.evaluate_options.outputs.runs_on }}
+    runs-on: "${{ needs.evaluate_options.outputs.runs_on }}"
     env:
       # TEST_DEPTH determines the maximum test level the suite should be running
       TEST_DEPTH: ${{ needs.evaluate_options.outputs.test_level }}
--- a/.github/workflows/continuous-delivery.yml
+++ b/.github/workflows/continuous-delivery.yml
@@ -274,13 +274,16 @@ jobs:
           fi
 
       - name: Set runner based on organization
+        env:
+          REPO_OWNER: ${{ github.repository_owner }}
         id: set-runner
         run: |
-          if [[ "${{ github.repository_owner }}" == "cloudnative-pg" ]]; then
+          if [[ "${REPO_OWNER}" == "cloudnative-pg" ]]; then
             echo "runs_on=ubuntu-latest-16-cores" >> $GITHUB_OUTPUT
           else
             echo "runs_on=ubuntu-24.04" >> $GITHUB_OUTPUT
           fi
+
   buildx:
     name: Build containers
     needs:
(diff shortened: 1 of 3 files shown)


mnencia (approved):

hack/e2e/run-e2e.sh
```diff
@@ -160,7 +160,7 @@ mkdir -p "${ROOT_DIR}/tests/e2e/out"
 # would create CPUs-1 nodes and saturate the testing server
 RC_GINKGO2=0
 export TEST_SKIP_UPGRADE=true
-ginkgo --nodes=4 --timeout 3h --poll-progress-after=1200s --poll-progress-interval=150s \
+ginkgo --nodes=6 --timeout 3h --poll-progress-after=1200s --poll-progress-interval=150s \
```
mnencia: Shouldn't we just remove the number and call it with -p, leaving ginkgo to decide how many parallel executions to run? I believe that with 16 cores, 6 is a bit low.

sxd: I'm thinking the same but I'm testing I want to try to find the sweet spot for 16 cores

jmealo:
Looks good to me :+1:

gbartolini:
Not much of a difference :(

jmealo:
I sped up my AKS E2E time to tighten the iteration loop on dynamic sizing/pvc-resizing.

The Azure CSI Driver doesn't like too many concurrent operations per-node. Tests got slow, timed out or became generally flaky. Limiting the PVCs managed by each node/driver seemed to help storage related tests more than anything else I tried.

I'm not sure if there's a point of diminishing returns with packing nodes or if there's an opportunity to tune the CSI driver itself.

I'm on my phone but I'll try to isolate what sped things up. Bumping the gingko node count alone didn't help.

I was only running the suite for dynamic sizing, directly on AKS, bypassing kind, pushing a new operator each run. Adding a pre-check for leftover artifacts from the last run made things reasonably fast/reliable.

Happy to share if it's helpful. Not sure if it's AKS-specific.

gbartolini (approved):

hack/e2e/run-e2e-local.sh
```diff
@@ -90,7 +90,7 @@ echo "E2E tests are running with the following filters: ${LABEL_FILTERS}"
 mkdir -p "${ROOT_DIR}/tests/e2e/out"
 RC_GINKGO=0
 export TEST_SKIP_UPGRADE=true
-ginkgo --nodes=4 --timeout 3h --poll-progress-after=1200s --poll-progress-interval=150s \
+ginkgo --nodes=12 --timeout 3h --poll-progress-after=1200s --poll-progress-interval=150s \
```
mnencia: I think we should leave this file alone. It is only used by devs on their local machines, and this parallelism is too much for any dev machine.

sxd: Yes I need to rollback, I touched by mistake (a sed :D )

jmealo:
It turns out my speed fixes were AKS specific. The only thing I can suggest is testing if limiting parallelism on AKS is faster than using Rook. You basically need to make sure that only one PVC is provisioning at a time on each node and then it's fast(er). Do more than that and things go poorly.

.github/workflows/continuous-delivery.yml
danishedb:
```suggestion
    runs-on: ubuntu-latest-16-cores
```

danishedb (changes requested):

.github/workflows/continuous-delivery.yml
danishedb:
```suggestion
    runs-on: ubuntu-latest-16-cores
```

danishedb (dismissed):

sxd:
Manual runs:
* https://github.com/cloudnative-pg/cloudnative-pg/actions/runs/22899469429
* https://github.com/cloudnative-pg/cloudnative-pg/actions/runs/22899492865

NiccoloFei (approved):

mnencia (approved):

PR #10004, opened 2026-02-17
fix: handle failed jobs and stuck scaling phases (#7793)

### LINKED ISSUE #7793: [Bug]: Cluster reconciliation loop gets stuck when scaling replicas due to failed snapshot recovery jobs

jmealo:
## Summary

When a job fails or is manually deleted during replica creation, the reconciler could get stuck indefinitely in a scaling phase. This PR fixes that by:

- Ensuring failed jobs don't block reconciliation (they are excluded from `runningJobNames()`)
- Preserving failed jobs for troubleshooting (only completed/successful jobs are cleaned up)
- Adding a safety net to clear stuck scaling phases when the cluster is actually healthy

All signals are deterministic: no timeouts.

## Changes

| File | Change |
|------|--------|
| `pkg/utils/job_conditions.go` | Add `IsJobFailed()` helper |
| `api/v1/cluster_funcs.go` | Add `IsScalingPhase()` helper |
| `internal/controller/cluster_status.go` | `runningJobNames()` excludes failed jobs |
| `internal/controller/cluster_cleanup.go` | `cleanupCompletedJobs()` preserves failed jobs (only deletes successful ones) |
| `internal/controller/cluster_controller.go` | Safety net in `reconcileResources` clears stuck scaling phase |

## What This Fixes

1. **Failed jobs blocking reconciliation**: Previously, `runningJobNames()` only excluded completed jobs. A failed job would still be considered "running" and block the reconciler.

2. **Stuck scaling phase after manual job deletion**: The workaround for #7793 was to manually delete the stuck job, but the cluster phase would remain stuck on "Creating a new replica". The new safety net detects when the cluster is in a scaling phase but has the correct instance count with no running jobs, and clears the phase to Healthy.

## Design Decision: Failed Jobs Are Preserved

Failed jobs are intentionally left in place (not cleaned up) for two reasons:
- They serve as evidence for troubleshooting why the job failed
- They act as markers for fallback logic (see #10005 - snapshot recovery fallback)

Only completed/successful jobs are cleaned up by `cleanupCompletedJobs()`.

## Test Plan

- [x] Unit tests for `IsJobFailed()` helper
- [x] Unit tests for `IsScalingPhase()` helper
- [x] Unit tests for `runningJobNames()` excluding failed jobs
- [x] Unit tests verifying `cleanupCompletedJobs()` does NOT delete failed jobs
- [x] All existing tests pass (562 specs across affected packages)
- [x] E2E: Failed join job does not block scaling (AKS)
- [x] E2E: Safety net clears stuck phase when instances match desired count (AKS)

## Related

- Fixes #7793
- Supersedes #7794 (comprehensive fix split into focused PRs)
- Complemented by: #10005 which adds fallback from snapshot recovery to pg_basebackup when snapshot recovery fails


gbartolini:
For maintainers' attention to evaluate inclusion in 1.29.0.

armru:
Hello Jeff, some of my considerations after having spent a bit of time on the patch:

Drawbacks of the current approach                                                                                                                                                                                
                                                                                                                                                                                                                   
  1. Safety net complexity — We needed a 4-condition guard just to recover from a stuck phase. That's a symptom, not a fix for the root cause.                                                                     
  2. Failed jobs accumulate — No automatic cleanup, relies on manual kubectl delete.               
  
**3. Dual-purpose objects — A K8s Job is being used both as infrastructure (run a task) AND as application state (don't retry this snapshot).** 

  4. Global blast radius — The runningJobNames() change applies to ALL job types (join, initdb, pgbasebackup, snapshot-recovery), not just the snapshot case.


  Alternative design

  Store the failure state where  on the Cluster instead of using jobs as a state:

  1. When a job fails, record the failure on the Cluster (e.g., a status condition like SnapshotRecoveryFailed with the snapshot name, or an annotation like cnpg.io/failed-snapshots: snap-1,snap-2).
  2. Clean up failed jobs in cleanupCompletedJobs and record the failure state  on the Cluster.
  3. Snapshot fallback logic reads the Cluster state, not the job list. It skips snapshots listed as failed and tries alternatives.
  4. User use plugin to re-enable a snapshot. More discoverable than knowing to delete a specific Job. CNPG status can be enhanced to list excluded snapshots

  This eliminates:
  - The safety net entirely (no stuck phases because failed jobs are cleaned up)
  - The runningJobNames() change (failed jobs don't linger)
  - Failed job accumulation
  - The dual-purpose object pattern

  And it gains:
  - Explicit, visible failure state on the Cluster (shows up in kubectl describe cluster OR kubectl cnpg status)
  - Cleaner separation between job lifecycle and application retry logic
  - The fallback decision is made by the reconciler based on Cluster state, not by inferring from the presence/absence of Job objects

jmealo:
Thanks for the thorough review, @armru. Let me work through each point.

### `runningJobNames()` — "Global blast radius" (Point 4)
- A failed job isn't a running job — this seems like a correctness fix for all job types, not just snapshot recovery. If an initdb or pg_basebackup job fails, the reconciler shouldn't spin waiting for it either.
- Is there a job type where we'd want a failed job treated as still running? I couldn't think of one but may be missing something.

### Safety net (Point 1)
- The scenario beyond failed jobs is manual job deletion: that's the current workaround for #7793 (`kubectl delete job`), and it leaves the phase stuck. External cleanup (TTL controllers, etc.) would do the same.
- The conditions express: *"if we're supposedly scaling but have the right instance count and no jobs in flight, we're not actually scaling."*

Would you still see value keeping it as defense-in-depth, or does it feel unnecessary? Open to it just logging instead of correcting if that's preferable.

### Failed job accumulation (Point 2) and dual-purpose objects (Point 3)
- These seem to apply more to #10005's design than this PR — in #10004, failed jobs are preserved for troubleshooting only; no application logic reads them. The dual-purpose pattern (jobs as retry markers) is introduced in #10005.

That said, the Cluster annotation/condition idea is interesting for #10005. A few things I'd want to think through:

- **Observability tradeoff** — Failed jobs carry pod logs, events, and exit codes. An annotation like `cnpg.io/failed-snapshots: snap-1` doesn't. Would we need to build additional surfacing in `cnpg status`?
- **CRD impact** — Would a status condition require API review, or would an annotation be lighter weight?
- **Re-enabling** — Is there existing plugin infrastructure for this, or would that need to be built?

### Context on the approach
The original PR (#7794) actually took a similar direction — deleting failed jobs and building detection/recovery machinery. The scope grew partly because deleting jobs meant needing mechanisms to surface failure info back to users.

That's why I closed it and split into these two targeted PRs (#1004 and #1005)

Preserving failed jobs was a deliberate simplification: the job itself *is* the record, and existing K8s tooling already knows how to surface it. Happy to explore Cluster-level state as a follow-up once the core fix is in.

---
### How would you like to proceed?

I see a few options for this PR:

1. **Keep as-is** — `runningJobNames()` fix + safety net + preserve failed jobs. Rework state management in #10005 per your proposal.
2. **Clean up failed jobs here too**: but that means #10005 needs the Cluster annotation approach from the start (scope increase).
3. **Something else?** open to alternatives.

jmealo:
One thing I didn't articulate well earlier: the retained failed job isn't just for troubleshooting, it's a user-facing control mechanism.

If CNPG falls back to `pg_basebackup` after a snapshot failure, and the user fixes the underlying issue, they can `kubectl delete job <failed-job>` to re-enable the snapshot fast path. They can also kill the slow recovery job if they
  don't want to wait. This uses standard K8s operations: no CNPG-specific knowledge, no plugin, no patching annotations.

With Cluster-level state, the equivalent recovery requires knowing how to clear a CNPG-specific condition or annotation. At 3am, kubectl delete job is a lot more intuitive.

jmealo:
### [Slack discussion summary ](https://cloud-native.slack.com/archives/C08MW1HKF40/p1771364718265939)(@armru + @jmealo)

## Should failed jobs be retained or cleaned up?

  @armru's position:
  - Jobs are infrastructure, not state: failure tracking belongs in Cluster status
  - It's reasonable to expect a logging stack in production for troubleshooting
  - Retaining failed jobs sets a precedent: the same argument could apply to failed instances or any other resource
  - The operator can't cover all cases; need to draw a line on what it's responsible for surfacing

  @jmealo's position:
  - Failed jobs carry rich context (pod logs, events, exit codes) that existing K8s tooling already surfaces
  - Not all users have logging configured for operator namespaces/jobs: deleting the evidence shifts the burden to CNPG to surface it another way
  - The retained failed job also serves as a user-facing control mechanism: if CNPG falls back to `pg_basebackup` (slow path) after a snapshot failure, the user can fix the underlying issue and `kubectl delete job` to re-enable the
  snapshot fast path: no CNPG-specific knowledge, no plugin, no patching annotations. Standard K8s operations that anyone debugging at 3am already knows.
  - With Cluster-level state, the equivalent recovery requires knowing how to clear a CNPG-specific condition or annotation
  - The original PR (#7794) took the delete-and-detect approach; scope grew partly because of this, leading to the split into #10004/#10005

  **Where we landed:**
  - Both agreed this is a broader project-wide question about how CNPG handles failed resources generally, not specific to this job type
  - @jmealo offered to go back to cleaning up failed jobs, with possible retention until cluster is healthy or a configurable maxAge/maxTTL
  - @armru noted he won't oppose the patch if other maintainers are OK with it, and encouraged hearing other opinions
  - Open question: should changes to failed job lifecycle happen in this PR or a follow-up?
  
  **Summarized by:** Claude 4.5 Opus

PR #10005, opened 2026-02-17
feat: fallback to pg_basebackup on snapshot recovery failure (#7793)

### LINKED ISSUE #7793: [Bug]: Cluster reconciliation loop gets stuck when scaling replicas due to failed snapshot recovery jobs

jmealo:
## Summary

When a snapshot recovery job fails during replica creation, the operator now automatically falls back to using `pg_basebackup` instead of remaining stuck or repeatedly retrying broken snapshot recovery.

This provides self-healing behavior: instead of just unblocking (#10004), the cluster recovers by trying an alternative join method.

## Changes

| File | Change |
|------|--------|
| `internal/controller/cluster_create.go` | Add `getFailedSnapshotDataSources()`, integrate fallback logic in `joinReplicaInstance`, and annotate snapshot-recovery jobs with their data source |
| `internal/controller/cluster_create_test.go` | Unit tests for `getFailedSnapshotDataSources()` and fallback behavior |
| `pkg/utils/labels_annotations.go` | Add `SnapshotDataSourceAnnotationName` constant |
| `tests/e2e/failed_job_handling_test.go` | E2E test: snapshot-based replica creation and graceful fallback |

## How It Works

1. When `joinReplicaInstance` is called and a snapshot source is available:
2. Check if the specific snapshot's recovery job has previously failed
3. If that snapshot failed:
   - Emit a warning event
   - Fall back to pg_basebackup instead of attempting snapshot recovery
4. If the snapshot hasn't failed, annotate the job with the snapshot name for tracking

This design means:
- Only the specific failed snapshot is skipped; new snapshots are still attempted
- The failed job is preserved as both troubleshooting evidence and a marker to prevent retrying that snapshot
- To re-enable snapshot recovery for a specific snapshot, a user can manually delete the failed job (`kubectl delete job <job-name>`)

## Why Create PVCs Before Job

The order of operations was changed from "create job, then PVCs" to "create PVCs, then job" because:
- If the job is created first and PVC creation fails, we have an orphaned job
- Creating PVCs first is safer: if it fails, no job is created
- This matches the pattern in `createPrimaryInstance`

Note: The `ctrl.Result` from `CreateInstancePVCs` is intentionally ignored in both `createPrimaryInstance` and `joinReplicaInstance`. PVC creation returns a requeue request, but the init/join job must be created in the same reconciliation loop. Returning early would leave dangling PVCs without pods, causing the operator to get stuck.

## Test Plan

- [x] Unit tests for `getFailedSnapshotDataSources()`
- [x] Unit tests for snapshot fallback logic
- [x] All existing tests pass
- [x] E2E (AKS): Snapshot-based replica creation succeeds
- [x] E2E (AKS): Phase 1 tests still pass on Phase 2 branch (backward compatible)

## Related

- Builds on #10004
- Fixes #7793

### DIFF
--- a/api/v1/cluster_funcs.go
+++ b/api/v1/cluster_funcs.go
@@ -1310,6 +1310,12 @@ func (cluster *Cluster) IsInplaceRestartPhase() bool {
 		cluster.Status.Phase == PhaseInplaceDeletePrimaryRestart
 }
 
+// IsScalingPhase returns true if the cluster is in a phase related to instance creation
+func (cluster *Cluster) IsScalingPhase() bool {
+	return cluster.Status.Phase == PhaseFirstPrimary ||
+		cluster.Status.Phase == PhaseCreatingReplica
+}
+
 // GetTablespaceConfiguration returns the tablespaceConfiguration for the given name
 // otherwise return nil
 func (cluster *Cluster) GetTablespaceConfiguration(name string) *TablespaceConfiguration {
--- a/api/v1/cluster_funcs_test.go
+++ b/api/v1/cluster_funcs_test.go
@@ -1788,3 +1788,23 @@ var _ = Describe("Failover quorum", func() {
 		Entry("with failover quorum disabled", clusterWithFailoverQuorumDisabled, false),
 	)
 })
+
+var _ = Describe("Scaling phase detection", func() {
+	DescribeTable(
+		"IsScalingPhase",
+		func(phase string, expected bool) {
+			cluster := &Cluster{
+				Status: ClusterStatus{
+					Phase: phase,
+				},
+			}
+			Expect(cluster.IsScalingPhase()).To(Equal(expected))
+		},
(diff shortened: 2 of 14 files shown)


gbartolini:
For maintainers' attention to evaluate inclusion in 1.29.0.

Issue #9969, opened 2026-02-12
[Bug]: Orphaned logical replication slots on demoted primary cause WAL accumulation after switchover


lambcode-unified:
### Is there an existing issue already for this bug?

- [x] I have searched for an existing issue, and could not find anything. I believe this is a new bug.

### I have read the troubleshooting guide

- [x] I have read the troubleshooting guide and I think this is a new bug.

### I am running a supported version of CloudNativePG

- [x] I have read the troubleshooting guide and I think this is a new bug.

### Version

1.27 (latest patch)

### What version of Kubernetes are you using?

1.35

### What is your Kubernetes environment?

Cloud: Google GKE

### How did you install the operator?

Helm

### What happened?

When using synchronizeLogicalDecoding: true with PostgreSQL 17's native sync_replication_slots, a switchover leaves orphaned logical replication slots on the demoted primary (now replica) that
cause unbounded WAL accumulation.

The slot sync worker on the new replica continuously fails with:
exiting from slot synchronization because same name slot "firenote_reporting_sub" already exists on the standby

This occurs because the demoted primary's original slot has synced=false (it was created locally when that node was the primary), and PostgreSQL's slot sync worker refuses to overwrite slots
that aren't marked as synced=true.

How to reproduce

1. Create a CNPG cluster with instances: 2 and synchronizeLogicalDecoding: true
2. Create a logical replication subscription that connects to the primary (creates a slot with synced=false on primary, synced=true on replica)
3. Perform a switchover (kubectl cnpg promote <cluster> <replica>)
4. Observe:
  - The new primary has the slot with synced=true (was the synced copy)
  - The new replica has the slot with synced=false (was the original)
  - The slot sync worker fails repeatedly on the new replica
  - WAL files accumulate on the new replica because the stale slot's restart_lsn is never advanced

### Expected behavior

After a switchover, logical replication slots should be properly synchronized, and WAL files should not accumulate on replicas due to stale slot positions.

### Actual behavior

The new replica retains its original slot with synced=false. The slot sync worker cannot overwrite it, so:
- The slot's restart_lsn is never updated
- WAL files accumulate (we observed 17GB+ on replicas vs ~600MB on primary)
- The slot sync worker logs errors every 10 seconds indefinitely

Slot state comparison after switchover:
```
 Node    Role     synced  restart_lsn
───────────────────────────────────────────────────────────
 node-1  Primary  true    Current (8/88045520)
[...]

PR #10006, opened 2026-02-18
fix: cleanup orphaned logical replication slots with synced=false aft…

bramhanandlingala:
[Bug]: Orphaned logical replication slots on demoted primary cause WAL accumulation after switchover #9969 - fix
…er switchover

- Update slot listing to include logical slots and synced column (PostgreSQL 17+)
- Drop orphaned logical slots with synced=false on replicas after switchover
- Prevents WAL accumulation due to stale logical slots

Closes: WAL accumulation issue after switchover with synchronizeLogicalDecoding=true


bramhanandlingala:
could you review this fix and approve for merge

PR #10008, opened 2026-02-18
chore(lint): fix warnings from golangci-lint v2.10.1

mnencia:
Bump golangci-lint to v2.10.1 and address the new warnings it introduced. Gosec taint analysis (G703) and integer overflow (G115) false positives are suppressed with targeted nolint comments, and G304 path traversal warnings are resolved with filepath.Clean rather than suppression. Replace WriteString(fmt.Sprintf(...)) with fmt.Fprintf, and fix path.Join to filepath.Join for OS file paths. Enable nolintlint to enforce well-formed directives and clean up stale suppressions. Add revive exclusion rules for packages that conflict with Go standard library names.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -107,10 +107,30 @@ linters:
           - revive
         path: /(utils|common)/[^/]+.go
         text: avoid meaningless package names
-      # TODO: fix by renaming or removing pkg/utils/context and pkg/utils/hash packages
+      # TODO: fix by renaming packages that conflict with Go standard library names
       - linters:
           - revive
         path: pkg/utils/(context|hash)/
+        text: avoid package names that conflict with Go standard library package names
+      - linters:
+          - revive
+        path: internal/(cmd|cnpi)/plugin/
+        text: avoid package names that conflict with Go standard library package names
+      - linters:
+          - revive
+        path: internal/cmd/manager/debug/
+        text: avoid package names that conflict with Go standard library package names
+      - linters:
+          - revive
+        path: tests/utils/exec/
+        text: avoid package names that conflict with Go standard library package names
+      - linters:
+          - revive
+        path: pkg/management/(postgres/metrics|url)/
+        text: avoid package names that conflict with Go standard library package names
+      - linters:
+          - revive
+        path: pkg/postgres/plugin/
         text: avoid package names that conflict with Go standard library package names
     paths:
       - zz_generated.*
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -27,6 +27,7 @@ linters:
     - misspell
     - nakedret
     - nestif
+    - nolintlint
     - prealloc
     - predeclared
     - revive
--- a/internal/cmd/plugin/psql/psql.go
+++ b/internal/cmd/plugin/psql/psql.go
@@ -183,7 +183,7 @@ func (psql *Command) Run() error {
 		return err
 	}
 
-	cmd := exec.Command(psql.kubectlPath, kubectlExec[1:]...) // nolint:gosec
+	cmd := exec.Command(psql.kubectlPath, kubectlExec[1:]...) //nolint:gosec
 	cmd.Stdout = os.Stdout
 	cmd.Stderr = os.Stderr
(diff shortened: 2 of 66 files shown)


armru (approved):

armru (approved):

internal/cmd/plugin/psql/psql.go
```diff
@@ -197,7 +197,7 @@ func (psql *Command) Output() ([]byte, error) {
 		return nil, err
 	}
 
-	cmd := exec.Command(psql.kubectlPath, kubectlExec[1:]...) // nolint:gosec
+	cmd := exec.Command(psql.kubectlPath, kubectlExec[1:]...) //nolint:gosec
 	cmd.Stderr = os.Stderr
```
sxd, line 201: This sounds like we should parse and make sure that the arguments are what we expect? or at least sanitize  what we're passing to the exec.Command() function, that's the right thing to do right?

pkg/executablehash/executablehash.go
```diff
@@ -36,12 +36,12 @@ var (
 
 // Stream opens a stream reading from the executable of the current process binary (os.Args[0] after path cleaning).
 func Stream() (io.ReadCloser, error) {
-	return os.Open(filepath.Clean(os.Args[0]))
+	return os.Open(filepath.Clean(os.Args[0])) //nolint:gosec // reading our own binary
```
sxd, line 39: Isn't enough to keep the fliepath.Clean() outside the return just to keep this error out ?

.golangci.yml
```diff
@@ -107,11 +108,31 @@ linters:
...
+      - linters:
+          - revive
+        path: tests/utils/exec/
+        text: avoid package names that conflict with Go standard library package names
+      - linters:
+          - revive
+        path: pkg/management/(postgres/metrics|url)/
+        text: avoid package names that conflict with Go standard library package names
+      - linters:
+          - revive
```
sxd, line 133: Besides that I don't; agree with having this, I don't think `postgres` is in the Go standard library

.golangci.yml
```diff
@@ -107,11 +108,31 @@ linters:
...
       - linters:
           - revive
         path: pkg/utils/(context|hash)/
         text: avoid package names that conflict with Go standard library package names
+      - linters:
+          - revive
+        path: internal/(cmd|cnpi)/plugin/
+        text: avoid package names that conflict with Go standard library package names
+      - linters:
+          - revive
```
sxd, line 121: We have three packages with the same name `plugin` inside the project, plus, one in the standard library, we should definitely rename this package to something descriptive

.golangci.yml
```diff
@@ -107,11 +108,31 @@ linters:
...
+        path: internal/cmd/manager/debug/
+        text: avoid package names that conflict with Go standard library package names
+      - linters:
+          - revive
```
sxd, line 125: The name of this package here is `debug` but does anything bug debug something, it's actually gathering system data

.golangci.yml
mnencia: it is metrics not postgres

.golangci.yml
mnencia: see https://github.com/cloudnative-pg/cloudnative-pg/pull/10099

.golangci.yml
mnencia: see https://github.com/cloudnative-pg/cloudnative-pg/pull/10099

pkg/executablehash/executablehash.go
mnencia: `filepath.Clean` satisfies G304 but not G703, so the nolint is still required.

internal/cmd/plugin/psql/psql.go
mnencia: kubectlExec content is generated from already parsed Kubernetes objects by `getKubectlInvocation()`. Gosec's G204 flags any exec.Command with non-constant args, which is a fundamental false positive for this pattern.

sxd (approved):

PR #10014, opened 2026-02-18
feat(controller): detect failed jobs and exclude bad snapshots

armru:
When a Kubernetes Job fails during replica creation (e.g. a snapshot- recovery job), the reconciler could get stuck indefinitely because it waits for running jobs to complete before proceeding.

This commit introduces handleFailedJobs() which runs early in the reconcile loop to:

1. Detect failed jobs via the standard JobFailed condition
2. Emit a Kubernetes Warning event with failure details
3. For snapshot-recovery jobs, record the VolumeSnapshot name from the PVC's spec.dataSource into cluster.status.failedSnapshots
4. Delete the failed job so the reconciler can proceed

Additionally, failed jobs are now excluded from runningJobNames() so the reconciler does not treat them as in-progress work.

The storage source selection logic (storagesource.go) now checks the failedSnapshots list, skipping snapshots that previously caused recovery failures and falling back to pg_basebackup when no viable snapshots remain.

PVC creation in joinReplicaInstance is reordered to happen before job creation, preventing orphaned jobs when PVC creation fails. 

Assisted-By: Opus 4.6


armru:
branch name not respecting dev/issue-number closing and reopening

PR #10015, opened 2026-02-18
feat(controller): detect failed jobs and exclude bad snapshots

### LINKED ISSUE #7793: [Bug]: Cluster reconciliation loop gets stuck when scaling replicas due to failed snapshot recovery jobs

armru:
# This PR is a POC and should not be merged 'as-is' and needs to be merged in #10004 #10005 for the relative changes that we want to retain

## This is a branch off the PRs : #10004 #10005 @jmealo 

When a Kubernetes Job fails during replica creation (e.g. a snapshot-recovery job hits BackoffLimitExceeded), the reconciler gets stuck: it waits for running jobs to complete before proceeding, but a failed job will never complete.

Flow:

```
reconcileResources()
  └─ handleFailedJobs()          # detect failures, emit events, record snapshots
       │
       ├─ failed job found?
       │    ├─ emit Warning event
       │    └─ role == "snapshot-recovery"?
       │         └─ find VolumeSnapshot from PVC dataSource
       │              └─ add to cluster.status.excludedSnapshots
       │
       └─ runningJobNames()       # excludes failed jobs → reconciler proceeds
            └─ storagesource      # skips excluded snapshots, falls back to pg_basebackup
```

`handleFailedJobs()` runs early in the reconcile loop to:

1. Detect failed jobs via the standard `JobFailed` condition
2. Emit a Kubernetes Warning event with failure details (job name, role, instance, reason)
3. For snapshot-recovery jobs, read the VolumeSnapshot name from the instance's PGDATA PVC `spec.dataSource` and record it in `cluster.status.excludedSnapshots`

Failed jobs are excluded from `runningJobNames()` so the reconciler no longer treats them as in-progress work.

The storage source selection logic (`storagesource.go`) checks `excludedSnapshots`, skipping snapshots that previously caused recovery failures and falling back to `pg_basebackup` when no viable snapshots remain.

PVC creation in `joinReplicaInstance()` is reordered to happen before job creation, preventing orphaned jobs when PVC creation fails.

Failed jobs are intentionally kept in place for troubleshooting. `TTLSecondsAfterFinished: 172800` (2 days) is set on all operator-created jobs so the Kubernetes TTL controller handles cleanup automatically.

### Plugin commands 

- `kubectl cnpg snapshot enable CLUSTER SNAPSHOT` — removes a snapshot from the excluded list
- `kubectl cnpg snapshot disable CLUSTER SNAPSHOT` — manually excludes a snapshot
- `kubectl cnpg status` now shows excluded snapshots when present

If a snapshot was excluded by mistake or the underlying issue is fixed:

```bash
# Check what's excluded
kubectl cnpg status <cluster>

# Re-enable a snapshot
kubectl cnpg snapshot enable <cluster> <snapshot-name>

# Manually exclude a snapshot
kubectl cnpg snapshot disable <cluster> <snapshot-name>
```

Assisted-By: Opus 4.6 (for test coverage)
Closes #7793


jmealo:
Thanks for putting this together @armru. The approach is clean and `Status.FailedSnapshots` fills the needs the CNPG operator to track failures, but, I think it leaves a gap for troubleshooting with your Cloud provider.

# Recovery control
This permanently records failed snapshots in Status.FailedSnapshots. If a snapshot failure is transient (cloud provider hiccup, temporary CSI issue), the cluster either falls back to a potentially much older snapshot (meaning more WAL replay) or to `pg_basebackup` entirely - with no automatic way to retry the latest, fastest snapshot path. Clearing it requires manually editing the cluster status or a future plugin command.

With #10004/#10005, the failed job itself serves as the marker. An operator can `kubectl delete job <failed-job>` to re-enable snapshot recovery, or cancel a slow recovery in favor of retrying the latest snapshot. **This gives operators control over the recovery method and timeframe using standard K8s operations.**

# Externally deleted jobs
If a job disappears before handleFailedJobs() runs (manual `kubectl delete`, TTL controller, external cleanup):

1. The snapshot name never gets recorded in FailedSnapshots — the operator retries the same broken snapshot
2. There's no safety net to clear a stuck scaling phase

#10004/#10005 handle both sub-cases: the safety net clears stuck phases when instances == desired with no running jobs, and external deletion of a failed job intentionally re-enables the snapshot path.

# Failed job retention
Deleting failed jobs means the debugging information is definitively gone. A FailedSnapshots: `["snap-1"]` status field tells you what failed but not why.

**From experience:** our volume snapshot failures are frequent enough to have a dedicated Slack channel. When engaging Azure support, timestamps alone are insufficient - they require correlation IDs at minimum, which live in pod logs and events attached to the job.

Even with a director of engineering involved, and a year of data (100MB of alerts from Victoria Metrics) we couldn't get help without that level of detail.

If failed jobs are accumulating to the point that cleanup is a concern, that's a signal worth surfacing, not masking.

# E2E tests and PR size
Production code is 169 lines — 18% of the PR. Tests (unit + E2E) account for 737 lines / 80%:
| Category | Lines | % |
  |---|---|---|
  | Production code (.go) | 169 | 18% |
  | Unit tests | 519 | 56% |
  | E2E tests | 218 | 24% |
  | CRD/generated/docs | 16 | 2% |
  
@sxd @gbartolini If we're encouraging E2E coverage in PRs (which I think we should), it might be worth excluding tests/e2e/ from the size label calculation or tallying it under a dedicated test label (that excludes YAML 😆).

armru:
added cnpg plugin commands to interact with excluded backups and job retention ttl (so they automatically delete after 2 days or so)

jmealo:
@armru I think keeping the failed jobs around for a short time makes sense, but, if someone is having a lot of trouble, and needs to retain them longer, perhaps we should have a config flag:

```
RETAIN_FAILED_JOBS=true                           # global
RETAIN_FAILED_JOBS=job_type                   # single type
RETAIN_FAILED_JOBS=*volume-snapshot* # glob support
````

This is a nice pattern for diagnostics/debugging/logging opt-in. I think that seems to cover what I'm most concerned about.

---

My remaining concern probably deserves [to be part of a larger discussion](https://github.com/cloudnative-pg/cloudnative-pg/discussions/10036)... How should users be able to steer CNPG to take a certain path during disaster recovery?

- Do we want to make things apparent to experienced Kubernetes users? (they clear failed job, slow to restore pods/pvcs manually)
- Do we want to make things CNPG specific and require documentation/kubectl plugin around known issues? [And ideally provide an imperative plugin approach/declarative manifest approach]?

We want CNPG to work correctly, but, due to environmental issues, things can get borked... and having established ways to recover so everyone doesn't need to maintain their own runbooks would be a huge win for the community.

jmealo:
Opened up a poll in [Slack](https://cloud-native.slack.com/archives/C08MAUJ7NPM/p1771606102640489) and on [GitHub](https://github.com/cloudnative-pg/cloudnative-pg/discussions/10036) to try to get a feel if there's any community consensus on a preferred approach for manual intervention/steering.

I didn't have a lot of hands on K8s experience with `kubectl` when I got started. I mostly wrote manifests and deployed via CI/CD. I actually have nearly zero opinion on this. 

I leaned towards k8s primitives because the odds you have someone on your team with that expertise is higher, and somewhat a requirement to be successful with CNPG (more so than just CNPG knowledge).

I would like to shift some of the burden on documentation/runbooks/procedures for when things go wrong off of the organizations using CNPG without burdening maintainers.

The most obvious way to do that to me is to pick the choice of least surprise and to use k8s primitives for the unplanned, and declarative/plugin for everything else.

PR #10020, opened 2026-02-19
feat(backup): support reconciliation-disabled annotation

### LINKED ISSUE #10019: [Feature]: Stop Reconciliation for Backups

The Cluster Controller supports skipping the reconciliation of resources having the annotation `cnpg.io/reconciliationLoop=disabled`.

It would be nice to add this feature to Backup resources as well. 

Skip the reconciliation in the Backup Controller for resources having the `cnpg.io/reconciliationLoop=disabled` annotation set

Adding a specific boolean field in the spec `suspend` and do not perform any action in the controller if set to true.  

This would imitate the behavior of a `Job` being created with the `suspend: true` field but not what happen when the same field value is flipped from false to true on an active `Job`, where existing Pods are terminated. 

### Additional context

_No response_

### Backport?

Yes

Yes

### Code of Conduct

GabriFedi97:
Add support for the `cnpg.io/reconciliationLoop=disabled` annotation on Backup resources, matching the existing Cluster behavior.

Closes #10019

### DIFF
--- a/docs/src/failure_modes.md
+++ b/docs/src/failure_modes.md
@@ -60,17 +60,10 @@ may be required.
 
 ### Disabling Reconciliation
 
-To temporarily disable the reconciliation loop for a PostgreSQL cluster, use
-the `cnpg.io/reconciliationLoop` annotation:
-
-```yaml
-metadata:
-  name: cluster-example-no-reconcile
-  annotations:
-    cnpg.io/reconciliationLoop: "disabled"
-spec:
-  # ...
-```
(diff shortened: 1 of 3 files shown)


leo-jin-edb:
@mnencia or @armru can one of you review this PR from @GabriFedi97?   This is blocking us from completing PG4K plugin migration, see details here https://enterprisedb.atlassian.net/browse/CNP-7630 as to why.

armru (approved):

gabriele-wolfox (approved):

mnencia (approved):

PR #10067, opened 2026-02-25
ci(release): sign operator YAML manifest during release

### LINKED ISSUE #10065: [Bug]: Missing signature of operator yaml file during release

### Contact Details

_No response_

### Version

1.28 (latest patch)

1.35

YAML manifest

### What happened?

During the release process, we're not signing the file cnpg-{verison}.yaml that we publish with the release.
The signature is part of the OpenSSF compliant process and should be done during the release process, after the file is generated, and it should be uploaded within the release

### Cluster resource

```shell

```

### Relevant log output

```shell

```

### Code of Conduct

sxd:
The release workflow was not signing the cnpg-{version}.yaml manifest
with cosign. Added blob signing with Sigstore bundle upload and
documented verification steps for all release assets (YAML, operator
images, operand images).

Closes #10065

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,6 +77,34 @@ CloudNativePG manages additional Kubernetes resources to enhance PostgreSQL
 management, including: `Backup`, `ClusterImageCatalog`, `Database`,
 `ImageCatalog`, `Pooler`, `Publication`, `ScheduledBackup`, and `Subscription`.
 
+## Verifying releases
+
+During the release process CloudNativePG uses [cosign](https://www.sigstore.dev/)
+to sign some of the assest's like the YAML and container images, these can be verified
+with the following commands:
+
+### Container images
+
+The following command requires only the image to be verified
+```shell
+cosign verify OPERATOR_IMAGE \
+  --certificate-identity-regexp="^https://github.com/cloudnative-pg/cloudnative-pg/" \
+  --certificate-oidc-issuer="https://token.actions.githubusercontent.com"
+```
+
+### Operator YAML deployment
+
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -76,6 +76,34 @@ CloudNativePG manages additional Kubernetes resources to enhance PostgreSQL
...
+```
+
+### Operator YAML deployment
+
+This step requires to have the `.yaml` file and the `sigstore.json` file from the
+release page:
+
+```shell
+cosign verify-blob cnpg-{version}.yaml \
+--bundle cnpg-{verison}.sigstore.json \
```
gbartolini:
```suggestion
--bundle cnpg-{version}.sigstore.json \
```

README.md
```diff
@@ -76,6 +76,34 @@ CloudNativePG manages additional Kubernetes resources to enhance PostgreSQL
 management, including: `Backup`, `ClusterImageCatalog`, `Database`,
 `ImageCatalog`, `Pooler`, `Publication`, `ScheduledBackup`, and `Subscription`.
 
+## Verifying releases
```
gbartolini: I think we should add this to the actual documentation, not in GitHub. What do you think?

sxd: I think it should be in both places, in the README.md for now it should be enough, but in the documentation we need to add the documentation to verify operator release, charts, containers, etc

gbartolini: I think it should be in one place only, otherwise we need to update it twice. Ok if for now you want to keep it here.

gbartolini (approved):

mnencia (approved):

PR #10069, opened 2026-02-25
fix(switchover) bypass spec.unschedulable when checking if a node is cordoned

### LINKED ISSUE #7799: [Bug]: Node cordon (unschedulable taint) causes a switchover

### Contact Details

_No response_

### Version

1.26 (latest patch)

1.31

Other

Helm

### What happened?

Cordoning a kubernetes node causes the operator to perform a switchover, even if the primaryUpdateStrategy is set to "supervised". While I believe this is intended, I would still consider this "feature" as a bug, since you do not normally expect in a kubernetes environment, that a cordon affects existing workloads.

I'm aware, that you can turn on the maintenance window for a given database, but we manage a lot of database clusters, and manually turning these on every time we conduct some form of maintenance that requires a node cordon is suboptimal.

I'm also aware, that instead of cordoning (which is equivalent to adding the node.kubernetes.io/unschedulable:NoSchedule taint), you could add a custom taint with the NoSchedule effect, so you could achieve the same thing, while it being ignored by the operator. While this is a good workaround, I do not believe this should be the way to go, since it's not a thing people are generally aware of. Therefore, leaving taints on the nodes unintentionally could be common, or even worse, cordoning a node running a primary instance, and performing an unintended switchover.

While a switchover normally causes little to negligible downtime, in some cases, this could very much disturb sensitive workloads, so I believe at least there should be an option (ideally, on the operator level) to disable automated switchover, if a cordon is issued on a node.

### Cluster resource

```shell

```

### Relevant log output

```shell
{"level":"info","ts":"2025-06-10T07:19:30.066361654Z","msg":"Primary is running on an unschedulable node, will try switching over","controller":"cluster","controllerGroup":"postgresql.cnpg.io","controllerKi
nd":"Cluster","Cluster":{"name":"cluster","namespace":"cluster-namespace"},"namespace":"cluster-namespace","name":"cluster","reconcileID":"9b6e086d-48a6-4e62-8799-9a2634a2c419","node":"node1","primary":"cluster-1"}
```

### Code of Conduct

enneitex:
Fixes https://github.com/cloudnative-pg/cloudnative-pg/issues/7799

The goal is to not trigger a switchover when a node is cordoned

When a user runs `kubectl cordon`, it adds two things to a node:
- `spec.unschedulable: true`
- a taint with the key `node.kubernetes.io/unschedulable`

Even when adding `DRAIN_TAINTS="ToBeDeletedByClusterAutoscaler,karpenter.sh/disrupted,karpenter.sh/disruption"` whitout `node.kubernetes.io/unschedulable`, a switchover will happen because of line https://github.com/cloudnative-pg/cloudnative-pg/blob/main/internal/controller/replicas.go#L193 who checks for `spec.unschedulable`.

The solution is to ignore this field as it is redundant with the taint.
I'm not aware of a scenario where `spec.unschedulable` is set without the associated taint but if there is a valid case, then I can rework this PR to add a new env var so a user can chose to bypass or not `spec.unschedulable`


legrandlegrand:
Hello,
I would like it for prod 😬
This may help bypassing replica PDB non respect, and limit downtime during K8S rollouts to a single switchover (after replicas drain)
Regards
PAscal

enneitex:
There is an issue linked.

legrandlegrand:
Fixes #7799

Melchior01:
Hello, we’d really benefit from getting this change merged. Is there anything I can do to help move it forward?

Issue #10058, opened 2026-02-25
[Security Slam 26] Achievement 2 - Chronicler Badge (Baseline Documentation)


gbartolini:
### Is there an existing issue already for this task?

- [x] I have searched for an existing issue, and could not find anything. I believe this is a new request.

### Is this a new minor release?

No

### Is this a preview release?

No

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

gbartolini:
WIth the new https://github.com/cloudnative-pg/cloudnative-pg/blob/main/SECURITY-INSIGHTS.yml file we should meet the Chronicler badge. I have submitted for it.

PR #10091, opened 2026-02-27
chore(security-insights): OSPS baseline action and asset verification

gbartolini:
Relates #10058

Brings `SECURITY-INSIGHTS.yml` in line with the OSSF Security Insights v2.2.0 schema and adds CI to keep it that way.

It adds the OSPS Baseline Scanner tooling entry and repoints `signature-verification` to the documented release-asset verification procedure. The self-assessment evidence, which the schema models as a single URL, now points to a new `.github/security-self-assessment.md` that links both the threat and capability catalogs rather than listing the two files directly. It also corrects the SLSA tool entry, which declared a `type` (`automated-tooling`) that is not part of the schema enum, to `other`.

A new `Security Insights Linter` job validates the file with `cue vet` against the schema version declared in its header, so future drift is caught in CI.

### DIFF
--- a/SECURITY-INSIGHTS.yml
+++ b/SECURITY-INSIGHTS.yml
@@ -219,9 +219,7 @@ repository:
 
     assessments:
       self:
-        evidence:
-          - https://github.com/cloudnative-pg/cloudnative-pg/blob/main/.github/threat-catalog.yaml
-          - https://github.com/cloudnative-pg/cloudnative-pg/blob/main/.github/capability-catalog.yaml
+        evidence: https://github.com/cloudnative-pg/cloudnative-pg/blob/main/.github/security-self-assessment.md
         comment: >-
           Gemara-compatible threat self-assessment covering capabilities
           and threats mapped to FINOS Common Cloud Controls (CCC) Core
--- a/SECURITY-INSIGHTS.yml
+++ b/SECURITY-INSIGHTS.yml
@@ -207,7 +207,7 @@ repository:
           ci: true
           release: true
       - name: OSPS Baseline Scanner
-        type: Policy-as-Code
+        type: other
         rulesets:
           - Maturity Level 2
         results:
--- a/SECURITY-INSIGHTS.yml
+++ b/SECURITY-INSIGHTS.yml
@@ -223,7 +223,7 @@ repository:
 
     assessments:
       self:
-        evidence-url:
+        evidence:
           - https://github.com/cloudnative-pg/cloudnative-pg/blob/main/.github/threat-catalog.yaml
           - https://github.com/cloudnative-pg/cloudnative-pg/blob/main/.github/capability-catalog.yaml
         comment: >-
--- a/SECURITY-INSIGHTS.yml
+++ b/SECURITY-INSIGHTS.yml
@@ -79,7 +79,7 @@ project:
     quickstart-guide: https://cloudnative-pg.io/docs/current/quickstart
     release-process: https://github.com/cloudnative-pg/cloudnative-pg/blob/main/contribute/release_procedure.md
     # OSPS-DO-03.01
-    signature-verification: https://github.com/cloudnative-pg/postgres-containers?tab=readme-ov-file#security
+    signature-verification: https://cloudnative-pg.io/docs/devel/installation_upgrade/#verifying-release-assets
     # OSPS-DO-04.01, OSPS-DO-05.01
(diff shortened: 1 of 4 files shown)


SECURITY-INSIGHTS.yml
```diff
@@ -203,6 +203,20 @@ repository:
           adhoc: true
           ci: true
           release: true
+      - name: OSPS Baseline Scanner
+        type: Policy-as-Code
+        rulesets:
+          - Maturity Level 1
+        results:
```
sxd: Why not add Maturity Level 2 right away ?

gbartolini: Sure

SECURITY-INSIGHTS.yml
```diff
@@ -203,6 +203,20 @@ repository:
           adhoc: true
           ci: true
           release: true
+      - name: OSPS Baseline Scanner
+        type: Policy-as-Code
+        rulesets:
+          - Maturity Level 1
```
gbartolini:
```suggestion
          - Maturity Level 2
```

sxd:
@gbartolini I think this is still good to have but it has issues with the rebase, can you fix that please?

SECURITY-INSIGHTS.yml
```diff
@@ -210,6 +210,20 @@ repository:
...
+          - Maturity Level 2
+        results:
+          adhoc:
+            name: OSPS Baseline Assessment
+            location: https://github.com/cloudnative-pg/cloudnative-pg/actions/workflows/osps_security_assessment.yml
+            comment: |
+              Evaluates the repository against the Open Source Project Security Baseline
+        integration:
+          adhoc: true
+          ci: true
```
mnencia: Per the SECURITY-INSIGHTS 2.2.0 schema, `adhoc` is "a scheduled process or on-demand" and `ci` is specifically "the continuous integration process". This workflow triggers only on `schedule` and `workflow_dispatch`, so the daily run is already captured by `adhoc: true`. This should be `ci: false` unless we also wire the scanner into the push/PR pipeline.

```suggestion
          ci: false
```

SECURITY-INSIGHTS.yml
```diff
@@ -210,6 +210,20 @@ repository:
           adhoc: false
           ci: true
           release: true
+      - name: OSPS Baseline Scanner
+        type: Policy-as-Code
```
mnencia: `type` is a closed enum in the SECURITY-INSIGHTS 2.2.0 schema: `fuzzing | container | secret | SCA | SAST | other`. `Policy-as-Code` is not a member, so this fails schema validation. `other` is the closest fit for a baseline/policy scanner.

```suggestion
        type: other
```

SECURITY-INSIGHTS.yml
```diff
@@ -210,6 +210,20 @@ repository:
...
+        rulesets:
+          - Maturity Level 2
+        results:
+          adhoc:
+            name: OSPS Baseline Assessment
+            location: https://github.com/cloudnative-pg/cloudnative-pg/actions/workflows/osps_security_assessment.yml
+            comment: |
+              Evaluates the repository against the Open Source Project Security Baseline
```
mnencia: `results.adhoc` is typed as an in-toto-style `#Attestation`, which requires a `predicate-uri`. This scanner only emits SARIF (to code scanning) and a results artifact, not a signed attestation, so there is no meaningful predicate type to point at. Cleaner to move the description to the tool-level `comment` field and use `results: {}`, like every other tool entry in this file.

```suggestion
        comment: Evaluates the repository against the Open Source Project Security Baseline
        results: {}
```

mnencia:
A couple of non-blocking things while you're in here.

The `signature-verification` URL points at `/docs/devel/`, but the doc links right next to it (`detailed-guide`, `quickstart-guide`, `support-policy`) all use `/docs/current/`. I'd normally say make it consistent, except the "Verifying release assets" section only exists on `devel` right now (`current` has no such anchor yet), so `devel` is actually the only link that works today. Once it lands in a release, worth flipping to `current` so it stays stable and matches the others.

Also `last-updated`/`last-reviewed` in the header are still `2026-02-25`. Might as well bump them since the PR touches the file.

jbattiato (approved):

mnencia (approved):

jbattiato (approved):

PR #10115, opened 2026-03-03
feat: add ReplicaFailingReplication alert and improve monitoring documentation

### LINKED ISSUE #10026: [Docs]: Provide working equivalent alert for ReplicaFailingReplication when using Distributed CNPG.

The alert definition from this repo's [`prometheusRules.yaml` file](https://github.com/cloudnative-pg/cloudnative-pg/blob/main/docs/src/samples/monitoring/prometheusrule.yaml#L63), titled `ReplicaFailingReplication` has a simple query `cnpg_pg_replication_in_recovery > cnpg_pg_replication_is_wal_receiver_up`, which seems to fire on the primary instance of a failover cluster in a dual-region distributed setup. I believe this is intended, because the 'primary' instance on the 'failover' region would have the WAL receiver down, but also be in recovery mode. At least, while using barman cloud plugin, as I do. Do I understand this correctly?

If so, is there a better alert for this that would be recommended?

Just guessing at a possible solution - Append a label to metrics provided by cnpg to expose the primary or non-primary status while in a distributed setup. Then tune this alert expression to exclude clusters that aren't primary.

_No response_

_No response_

### Additional context

_No response_

### Backport?

Yes

### Code of Conduct

Denyme24:
fixes #10026

this pull request improves the monitoring documentation and alerting logic for CNPG clusters, mainly around replication failure alerts.

it adds a new alert to reduce false positives in distributed CNPG setups using Barman Cloud, clarifies how the existing alert behaves, and expands the troubleshooting guidance. i also made a few small formatting fixes for better consistency and readability.

Monitoring and alerting improvements:

-> expanded the documentation for `ReplicaFailingReplication`, explaining its expected behavior in standard CNPG clusters and why it can produce false positives in distributed setups with Barman Cloud. added clearer troubleshooting steps and guidance on interpreting replication lag metrics.

-> introduced a new alert, `ReplicaFailingReplicationDistributed`, in both `prometheusrule.yaml` and `alerts.yaml`. this version requires significant and increasing replication lag before firing, which helps avoid false positives during normal Barman Cloud operations. added comments in both files to clarify when to use each alert.

-> added inline comments to the existing `ReplicaFailingReplication` samples, warning about possible false positives in distributed CNPG setups and recommending the new alert in those cases.


Denyme24:
@gbartolini  , @jsilvela

hi, whenever you get a chance, could you please take a look at this PR?

it addresses #10026 by introducing a new alert rule for distributed CNPG setups and adding related troubleshooting documentation. the changes are backward compatible and have been tested.

thanks in advance.

Denyme24:
Hi @gbartolini @jsilvela, just a gentle reminder to review this PR. Thanks.

PR #10131, opened 2026-03-04
feat(catalog): accept extensions field in image catalogs

### LINKED ISSUE #10130: [Feature]: Allow release-1.27 and release-1.28 to accept image catalogs containing extensions

The `extensions` field in `CatalogImage` was introduced in #9781 and targets 1.29. On 1.27 and 1.28, the CRD schema does not include the `extensions` field, so the API server rejects any catalog manifest that contains it.

This means users who want to share a single image catalog across clusters running different CNPG versions cannot include extensions in their catalog without breaking older clusters. They are forced to maintain separate catalog manifests per CNPG version, which is unnecessary friction since older operators can safely ignore unknown fields.

Backport the CRD schema change (adding the `extensions` field to `CatalogImage`) to the release-1.27 and release-1.28 branches. The operator code on those branches would simply ignore the field since the Go struct has no corresponding field to unmarshal into — older operators would continue working as before, just silently skipping extensions they don't support.

Users could maintain separate catalog manifests for each CNPG version (one with extensions for 1.29+, one without for older versions), but this adds unnecessary operational overhead when the older versions could simply accept and ignore the extra field.

### Additional context

_No response_

### Backport?

Yes

Yes

### Code of Conduct

mnencia:
Allow the CRD schema for ImageCatalog and ClusterImageCatalog to include the extensions field introduced in 1.29. The operator ignores the field since the Go struct simply skips unknown data during deserialization, but accepting it in the schema lets users share a single catalog manifest across clusters running different CNPG versions.

To keep the CRD schema consistent across versions, relax the ExtensionConfiguration image field from required to optional, matching the 1.29 schema. A webhook validation ensures Cluster extensions still require an image reference on 1.28 and 1.27.

Closes #10130


mnencia:
The same patch applies cleanly to release-1.27 since `ExtensionConfiguration` is already defined there as well. This should be manually backported to release-1.27 once merged.

mnencia:
Linter failure will be fixed in https://github.com/cloudnative-pg/cloudnative-pg/pull/10136

armru (approved):

NiccoloFei (approved):

PR #10134, opened 2026-03-04
chore(ci): read Go version from go.mod toolchain directive

### LINKED ISSUE #10133: Read Go version from go.mod toolchain directive in CI workflows
All GitHub Actions workflows currently hardcode a `GOLANG_VERSION` env var (kept in sync across 6 files via Renovate's custom regex manager). Since we already use setup-go v6, which supports reading the `toolchain` directive from `go.mod` via `go-version-file`, we can make `go.mod` the single source of truth for the Go version used in CI.

This eliminates the need to maintain `GOLANG_VERSION` in every workflow and removes a potential source of version drift between go.mod and CI.

mnencia:
Replace hardcoded GOLANG_VERSION env vars across all workflows with setup-go v6's go-version-file parameter, which reads the toolchain directive from go.mod. This makes go.mod the single source of truth for the Go version used in CI, removing the need to keep 6 workflow files in sync via Renovate's custom regex manager.

Closes #10133 

Assisted-by: Claude Opus 4.6

### DIFF
--- a/.github/workflows/backport.yml
+++ b/.github/workflows/backport.yml
@@ -10,10 +10,6 @@ on:
 
 permissions: read-all
 
-env:
-  # renovate: datasource=golang-version depName=golang versioning=loose
-  GOLANG_VERSION: "1.26.0"
-
 jobs:
   # Label the source pull request with 'backport-requested' and all supported releases label, the goal is, by default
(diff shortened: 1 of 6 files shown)


sxd (approved):

NiccoloFei (approved):

sxd:
/ok-to-merge updating go detection in github actions

PR #10135, opened 2026-03-04
feat: conditionally watch node resources

### LINKED ISSUE #9346: [Feature]: CNPG operator should not crash when nodes & clusterimagecatalog resource access not permitted

CNPG currently requires cluster-wide permissions to access `Node` and `ClusterImageCatalog` resources. This creates a security concern for hardened Kubernetes environments where RBAC should follow the principle of least privilege. Removing these permissions causes the operator to fail at startup.

Two independent environment variables to control cluster-wide object access:

- `WATCH_NODES` (default: `true`) - Enable/disable Node resource access
- `WATCH_CLUSTER_IMAGE_CATALOGS` (default: `true`) - Enable/disable ClusterImageCatalog resource access

When disabled, the operator:
1. Does not attempt to list or watch the corresponding cluster-wide resources
2. Allows the corresponding ClusterRole permissions to be removed by deployment tools (Helm, ArgoCD, FluxCD, etc.)
3. Continues to function normally for namespace-scoped resources

Similar deployment parameters exist in other open source projects. For example CN Cert Manager which implemented this: https://github.com/cert-manager/cert-manager/pull/7678/

Removing the ClusterRoles for the relevant cluster resources results in CNPG Operator to fail startup.
`"failed to list *v1.Node: nodes is forbidden: User \"system:serviceaccount:cnpg-system:cnpg-manager\" cannot list resource \"nodes\" in API group \"\" at the cluster scope"`

### Additional context

_No response_

### Backport?

No

Yes

### Code of Conduct

maxlengdell:
CNPG currently relies on access to cluster nodes. This PR introduces the possibility to opt out of this dependency by not watching for node events or invoking the k8 api to request node info. 

This PR also introduces a new utils package for configuring the RBAC for the operator at runtime. This is used to split the default cluster role policies into role and cluster role. This intends to mimic a deployment of the operator where the operator and operand are deployed in the same namespace. This package also has the possibility to remove some resources from the rbac policies, such as nodes

**Note**
This is a modification & split out of #9350 

Fixes #9346

Signed-off-by: Max Lengdell max.a.lengdell@ericsson.com

### DIFF
--- a/docs/src/e2e.md
+++ b/docs/src/e2e.md
@@ -58,6 +58,10 @@ and the following suite of E2E tests are performed on that cluster:
      * Operator upgrade
      * Operator High Availability
 
+* **Reduced RBAC deployment:**
+  * Operator configured with reduced RBAC
+  * Node drain while operator is configured without node RBAC
+
 * **Observability:**
      * Metrics collection
      * PgBouncer Metrics
--- a/docs/src/e2e.md
+++ b/docs/src/e2e.md
@@ -59,8 +59,8 @@ and the following suite of E2E tests are performed on that cluster:
      * Operator High Availability
 
 * **Reduced RBAC deployment:**
-  * Operator configured with reduced RBAC
-  * Node drain while operator is configured without node RBAC
+  * Operator Deployment
+  * Node drain
 
 * **Observability:**
      * Metrics collection
--- a/docs/src/security.md
+++ b/docs/src/security.md
@@ -429,6 +429,44 @@ to viewing resources. However, if an unauthorized user gains access to the
 Therefore, it's crucial to prevent users from accessing the operator's
 `ServiceAccount` and any other `ServiceAccount` with elevated permissions.
 
+### Restricted RBAC Mode
+
+You can restrict the operator's RBAC permissions by configuring `WATCH_NAMESPACES`
+to the same namespace as the operator deployment. This way all namespaced resource
+policies can be created in a `Role`, and the `ClusterRole` only needs to contain
+`admissionregistration.k8s.io`, `clusterImageCatalogs` and `Nodes` resources.
+
+#### Disable Node Watching
+
+Set `WATCH_NODES` to `false` to disable node access:
+
+When `WATCH_NODES` is set to `false`:
+- The operator will not require `nodes` RBAC permissions
+
+### Known Limitations with Restricted Nodes RBAC
+
+#### SyncReplicaElectionConstraint
+
+`cluster.Spec.PostgresConfiguration.SyncReplicaElectionConstraint` is used to enforce deployment constraints for synchronous replicas during election. This feature will not work with restricted RBAC due to lack of node access.
+
+#### PodDisruptionBudget
+
+:::warning
+    When running with restricted RBAC and PodDisruptionBudget (PDB) enabled, node drain operations may fail.
+:::
+
+CloudNativePG creates PodDisruptionBudgets to protect PostgreSQL instances during voluntary disruptions. However, with restricted RBAC, a critical limitation exists:
+
+- The Kubernetes node drain process requires cluster-level permissions to evict pods protected by PDBs
+- With restricted RBAC, the operator lacks the necessary cluster-wide permissions to properly coordinate with the node drain controller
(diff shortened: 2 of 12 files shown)


docs/src/security.md
```diff
@@ -338,6 +338,44 @@ to viewing resources. However, if an unauthorized user gains access to the
...
+Set `WATCH_NODES` to `false` to disable node access:
+
+When `WATCH_NODES` is set to `false`:
+- The operator will not require `nodes` RBAC permissions
+
+### Known Limitations with Restricted Nodes RBAC
+
+#### SyncReplicaElectionConstraint
+
+`cluster.Spec.PostgresConfiguration.SyncReplicaElectionConstraint` is used to enforce deployment constraints for synchronous replicas during election. This feature will not work with restricted RBAC due to lack of node access.
```
maxlengdell: Will fix so that all doc follow 80col rule in later commit once this has been reviewed

docs/src/e2e.md
```diff
@@ -58,6 +58,10 @@ and the following suite of E2E tests are performed on that cluster:
      * Operator upgrade
      * Operator High Availability
 
+* **Reduced RBAC deployment:**
+  * Operator configured with reduced RBAC
+  * Node drain while operator is configured without node RBAC
```
sxd: for both options we know is "reduced RBAC" and "whout node RBAC" probably no need to mention that since it's the section

docs/src/security.md
```diff
@@ -338,6 +338,44 @@ to viewing resources. However, if an unauthorized user gains access to the
 Therefore, it's crucial to prevent users from accessing the operator's
 `ServiceAccount` and any other `ServiceAccount` with elevated permissions.
 
+### Restricted RBAC Mode
+
+You can restrict the operator's RBAC permissions by configuring `WATCH_NAMESPACES`
+to the same namespace as the operator deployment. This way all namespaced resource
+policies can be created in a `Role`, and the `ClusterRole` only needs to contain
+`admissionregistration.k8s.io`, `clusterImageCatalogs` and `Nodes` resources.
+
```
sxd, line 441: This requires the especific permission that it's used to get the nodes information and which set of actions are required, otherwise, no one knows about which permissions are we talking about

docs/src/security.md
```diff
@@ -338,6 +338,44 @@ to viewing resources. However, if an unauthorized user gains access to the
...
+#### Disable Node Watching
+
+Set `WATCH_NODES` to `false` to disable node access:
+
+When `WATCH_NODES` is set to `false`:
+- The operator will not require `nodes` RBAC permissions
+
```
sxd, line 448: Again, the permission that is not required for this task

docs/src/security.md
sxd: this is missing the "why it will not work"

docs/src/security.md
```diff
@@ -338,6 +338,44 @@ to viewing resources. However, if an unauthorized user gains access to the
...
+`cluster.Spec.PostgresConfiguration.SyncReplicaElectionConstraint` is used to enforce deployment constraints for synchronous replicas during election. This feature will not work with restricted RBAC due to lack of node access.
+
+#### PodDisruptionBudget
+
+:::warning
+    When running with restricted RBAC and PodDisruptionBudget (PDB) enabled, node drain operations may fail.
+:::
+
+CloudNativePG creates PodDisruptionBudgets to protect PostgreSQL instances during voluntary disruptions. However, with restricted RBAC, a critical limitation exists:
```
sxd:
```suggestion
CloudNativePG creates `PodDisruptionBudget` to protect PostgreSQL instances during voluntary disruptions. However, with restricted RBAC, a critical limitation exists:
```

docs/src/security.md
maxlengdell: Good point, I added the explicit verbs for the ClusterRole. 
I didnt add any details about the Role, so that the automatically created RBAC settings from kubebuilder doesnt get duplicated here. But I can add those verbs here as well for the Role.

docs/src/security.md
maxlengdell: True, added a explanation

maxlengdell:
Thank you for the review @sxd , I added a commit with updates. I force pushed to resolve the DCO issue :)

PR #10148, opened 2026-03-05
feat: add `podSelectorRefs` for dynamic `pg_hba` address resolution

### LINKED ISSUE #10087: [Feature]: Dynamic `pg_hba` entries via `spec.podSelectorRefs`

Currently, `spec.postgresql.pg_hba` accepts only raw string entries that are passed verbatim into `pg_hba.conf`. Users who want to restrict access to specific application pods must manually determine pod IPs and hardcode them, which is impractical in dynamic Kubernetes environments where pod IPs change constantly.

Introduce a mechanism that allows users to define named pod label selectors and reference them inside `pg_hba` lines, enabling dynamic address resolution based on pod labels.

**New API field: `spec.podSelectorRefs**`
A new field at the `spec` level, structured as a list of objects (idiomatic Kubernetes style):

```yaml
spec:
  podSelectorRefs:
    - name: app-pods
      selector:
        matchLabels:
          app: myapp
    - name: monitoring
      selector:
        matchLabels:
          role: monitoring
  postgresql:
    pg_hba:
      - "hostssl mydb myuser @app-pods scram-sha-256"
      - "hostssl postgres monitor @monitoring scram-sha-256"
```

The `@ref_name` syntax in the address field of a `pg_hba` line references a named pod selector ref. While `@` is used for file inclusion in `pg_hba`'s database and user fields, it has no special meaning in the address field. PostgreSQL would reject an unresolved `@ref_name` as an invalid address, providing a fail-safe if the operator fails to expand it.

An EDB customer reached out to us to find a way to gain finer control over access from pgBouncer pods. Initially, we considered creating an integration between the pooler and the cluster, which is contrary to the direction we've recently taken. However, a more generic approach might be more effective. Hence, this idea.

### Additional context

This is an initial draft of the idea. Please continue the analysis in the comments of this thread.

### Backport?

No

Yes

### Code of Conduct

armru:
Introduces a declarative way to manage `pg_hba` rules by resolving pod label selectors into IP addresses, eliminating the need for static CIDR ranges or manual updates when client pods restart.

Users define named label selectors in `.spec.podSelectorRefs` and reference them with `${podselector:NAME}` in `pg_hba` address fields. The operator resolves matching pod IPs and the instance manager expands each reference into individual `/32` or `/128` entries at render time, with automatic reload on pod lifecycle changes.

Closes #10087


sxd:
@mnencia probably this should be tested with a pg_hba config like the one here https://github.com/cloudnative-pg/postgres-keycloak-oauth-validator?tab=readme-ov-file#example-cloudnativepg-configuration
Also I don't see a case of dual-stack in the testing, for order probably that case should be only one line?
On the other hand, instead of using a hard coded string for the mask is better to use something like this https://pkg.go.dev/net#IPMask.String but just an idea,
That was a quick check but the dual-stack thing is something that needs to be tested for sure

gbartolini:
There is a problem with the markdown code for `docs/src/cloudnative-pg.v1.md`.

gbartolini:
I created the `angus` cluster with a pgbouncer pooler (`marshall`):

```yaml
kind: Cluster
metadata:
  name: angus
spec:
  instances: 3

  podSelectorRefs:
    - name: marshall-pooler
      selector:
        matchLabels:
          cnpg.io/podRole: pooler
          cnpg.io/cluster: angus
          cnpg.io/poolerName: marshall

  postgresql:
    pg_hba:
      - "hostssl all all ${podselector:marshall-pooler} scram-sha-256"
      - "host all all all reject"

  storage:
    size: 1Gi
---
kind: Pooler
metadata:
  name: marshall
spec:
  cluster:
    name: angus

  instances: 3
  type: rw
  pgbouncer:
    poolMode: session
    parameters:
      max_client_conn: "1000"
      default_pool_size: "10"
```

The cluster defines the `pooler` podSelectorRef that points to the pooler's pods via three labels. Then it enables access to the database via the pooler and then adds a fallback rule to reject any access.

I then create these two simple jobs with `pgbench`, one going through the pooler (`PGHOST=marshall`) and one going directly to the `angus-rw` service.

```yaml
apiVersion: batch/v1
kind: Job
metadata:
  name: marshall-pgbench
spec:
  template:
    spec:
      containers:
      - args:
        - -i
        command:
        - pgbench
        env:
        - name: PGHOST
          value: marshall
        - name: PGDATABASE
          value: app
        - name: PGPORT
          value: "5432"
        - name: PGUSER
          valueFrom:
            secretKeyRef:
              key: username
              name: angus-app
        - name: PGPASSWORD
          valueFrom:
            secretKeyRef:
              key: password
              name: angus-app
        image: ghcr.io/cloudnative-pg/postgresql:18.1-system-trixie
        name: pgbench
      restartPolicy: Never
---
apiVersion: batch/v1
kind: Job
metadata:
  name: angus-pgbench
spec:
  template:
    spec:
      containers:
      - args:
        - -i
        command:
        - pgbench
        env:
        - name: PGHOST
          value: angus-rw
        - name: PGDATABASE
          value: app
        - name: PGPORT
          value: "5432"
        - name: PGUSER
          valueFrom:
            secretKeyRef:
              key: username
              name: angus-app
        - name: PGPASSWORD
          valueFrom:
            secretKeyRef:
              key: password
              name: angus-app
        image: ghcr.io/cloudnative-pg/postgresql:18.1-system-trixie
        name: pgbench
      restartPolicy: Never
```

This is the result:

```console
$ kubectl get pods
NAME                        READY   STATUS      RESTARTS   AGE
angus-1                     1/1     Running     0          17m
angus-2                     1/1     Running     0          17m
angus-3                     1/1     Running     0          14m
angus-pgbench-7q56c         0/1     Error       0          7s
csi-hostpathplugin-0        8/8     Running     0          6h21m
marshall-6796d674ff-czmkj   1/1     Running     0          17m
marshall-6796d674ff-gzmf9   1/1     Running     0          17m
marshall-6796d674ff-wmlwk   1/1     Running     0          17m
marshall-pgbench-blm7c      0/1     Completed   0          7s

$ kubectl logs jobs/angus-pgbench
Found 2 pods, using pod/angus-pgbench-7q56c
pgbench: error: connection to server at "angus-rw" (10.96.155.110), port 5432 failed: FATAL:  pg_hba.conf rejects connection for host "10.244.1.20", user "app", database "app", SSL encryption
connection to server at "angus-rw" (10.96.155.110), port 5432 failed: FATAL:  pg_hba.conf rejects connection for host "10.244.1.20", user "app", database "app", no encryption
pgbench: error: could not create connection for initialization

$ kubectl logs jobs/marshall-pgbench
dropping old tables...
creating tables...
generating data (client-side)...
100000 of 100000 tuples (100%) of pgbench_accounts done (elapsed 0.01 s, remaining 0.00 s)
vacuuming...
creating primary keys...
done in 0.10 s (drop tables 0.03 s, create tables 0.00 s, client-side generate 0.03 s, vacuum 0.02 s, primary keys 0.01 s).
```

The `pg_hba` rule we created using the pod selector is able to allow access through the pooler. The job that tries to connect directly through the `angus-rw` primary service is prevented from accessing the database.

gbartolini:
Moreover, the `status` of the cluster contains:

```yaml
  podSelectorRefs:
  - ips:
    - 10.244.1.16
    - 10.244.2.16
    - 10.244.3.24
    name: pooler
```

If I delete a pgbouncer pod, a new one is recreated with a different IP and the status is immediately updated:

```yaml
  podSelectorRefs:
  - ips:
    - 10.244.1.16
    - 10.244.2.20
    - 10.244.3.24
    name: pooler
```

leonardoce (approved):

gbartolini (approved):

gbartolini:
```sql
postgres=# SELECT database, user_name, address, netmask FROM pg_catalog.pg_hba_file_rules WHERE auth_method = 'scram-sha-256' and type = 'hostssl';
 database | user_name |   address   |     netmask
----------+-----------+-------------+-----------------
 {all}    | {all}     | 10.244.1.16 | 255.255.255.255
 {all}    | {all}     | 10.244.2.20 | 255.255.255.255
 {all}    | {all}     | 10.244.3.24 | 255.255.255.255
(3 rows)
```

gbartolini:
> @mnencia probably this should be tested with a pg_hba config like the one here https://github.com/cloudnative-pg/postgres-keycloak-oauth-validator?tab=readme-ov-file#example-cloudnativepg-configuration

Although I have not tested the validator, I have introduced the row in the `pg_hba.conf`, and it was parsed correctly, which to me was enough.

> Also I don't see a case of dual-stack in the testing, for order probably that case should be only one line? On the other hand, instead of using a hard coded string for the mask is better to use something like this https://pkg.go.dev/net#IPMask.String but just an idea, That was a quick check but the dual-stack thing is something that needs to be tested for sure

The dual-stack issue should be addressed globally, as it is not tied to this specific feature (as far as I understand). There is an ongoing discussion about this (#7116), but my response aligns with what I’ve been saying: considering everything, I believe there are more critical areas for the project to focus on at this stage, particularly regarding technical debt.

I believe the feature, as it stands, provides significant value and should be considered a major feature for version 1.29. However, it would be great to get your review if you have time.

mnencia (approved):

PR #10169, opened 2026-03-06
chore(renovate): registry version updated to 3

### LINKED ISSUE #10168: [Chore]: Add renovate for container registry in setup scripts
The hack/setup.sh currently uses version 2 of the registry. We should make sure we use the latest stable version.

gbartolini:
Closes #10168

### DIFF
--- a/hack/testing-tools/common/40-utils-registry.sh
+++ b/hack/testing-tools/common/40-utils-registry.sh
@@ -27,7 +27,7 @@ registry_net=registry
 registry_port='5000'
 builder_name=cnpg-builder
 # renovate: datasource=docker depName=registry versioning=docker
-registry_version=3
+REGISTRY_VERSION=3
 
 # ensure_registry: Sets up the local Docker registry container and network.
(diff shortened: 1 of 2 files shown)


hack/testing-tools/common/40-utils-registry.sh
```diff
@@ -26,6 +26,8 @@ registry_name=registry.dev
 registry_net=registry
 registry_port='5000'
 builder_name=cnpg-builder
+# renovate: datasource=docker depName=registry versioning=docker
+registry_version=3
```
GabriFedi97: This file and format don't seem to be referenced anywhere in the renovate.json config.
We probably have to add an entry [here](https://github.com/cloudnative-pg/cloudnative-pg/blob/b61918c5a0259ecc0cab6ef5272daf0660968693/.github/renovate.json5#L47) to reference this file and to capitalize the var name `registry_version`

jbattiato (approved):

mnencia (approved):

PR #10192, opened 2026-03-06
fix: validate VolumeSnapshot existence before using as storage source

### LINKED ISSUE #10029: [Bug]: Replica creation fails when bootstrap volume no longer exists

### Contact Details

simona@xata.io

### Version

older in 1.28.x

1.33

Cloud: Amazon EKS

Helm

### What happened?

 When a CNPG cluster is bootstrapped from a `VolumeSnapshot` (via `spec.bootstrap.recovery.volumeSnapshots`), the operator stores that snapshot reference in the cluster spec. Later, when adding a replica, the operator attempts to use the same `VolumeSnapshot` as a data source for the new replica's PVC. 

If the original `VolumeSnapshot` has been deleted (e.g. due to retention policies, manual cleanup, or infrastructure changes), the replica PVC is created with a `spec.dataSource` pointing to a non-existent           
`VolumeSnapshot` and it remains in `Pending` indefinitely with no automatic recovery.

  ### Steps to reproduce                                                                                                                                                                                                  
                                                                                                                                                                                                                      
  1. Create a cluster bootstrapped from a VolumeSnapshot: 
 ```                                                                                                                                                            
  spec:
    bootstrap:
      recovery:
        volumeSnapshots:
          storage:
            name: my-snapshot
    instances: 1
```
  2. Wait for the primary to be running.
  3. Delete the `VolumeSnapshot` my-snapshot.
  4. Scale the cluster: set `spec.instances: 2.`
  5. Observe the replica PVC stuck in `Pending`, referencing the deleted snapshot.

### Root cause

In `pkg/reconciler/persistentvolumeclaim/storagesource.go`, the function `GetCandidateStorageSourceForReplica()` determines the storage source for new replicas. It has two candidate paths:
 1. From the backup list (`getCandidateSourceFromBackupList`) - picks the most recent completed `VolumeSnapshot` backup.
  2. From the cluster definition (`getCandidateSourceFromClusterDefinition`) - falls back to the original bootstrap snapshot at `cluster.Spec.Bootstrap.Recovery.VolumeSnapshots`.

 Neither path validates that the referenced `VolumeSnapshot(s)` still exist in the namespace. The returned `StorageSource` is passed directly to `CreateInstancePVCs`(), which sets `pvc.Spec.DataSource` to the (possibly deleted) snapshot reference.

 ### Expected behaviour

When the referenced `VolumeSnapshot` does not exist, the operator should fall back to creating the replica by copying from the primary pod (by treating the storage source as nil) rather than creating a PVC with an unresolvable data source.

### Cluster resource

```shell

```

### Relevant log output

```shell

```

### Code of Conduct

armru:
When a cluster is bootstrapped from a VolumeSnapshot that is later deleted, adding replicas would fail because the operator referenced the deleted snapshot as data source for new PVCs, leaving them stuck in Pending state indefinitely.

Add VolumeSnapshot existence validation in GetCandidateStorageSourceForReplica and getCandidateSourceFromBackupList. When a referenced snapshot no longer exists, the operator now skips it and tries the next candidate or falls back to pg_basebackup for replica creation.

Closes #10029

### DIFF
--- a/internal/controller/cluster_create.go
+++ b/internal/controller/cluster_create.go
@@ -1253,7 +1253,7 @@ func (r *ClusterReconciler) joinReplicaInstance(
 	job := specs.JoinReplicaInstance(*cluster, nodeSerial)
 
 	// If we can bootstrap this replica from a pre-existing source, we do it
-	storageSource := persistentvolumeclaim.GetCandidateStorageSourceForReplica(ctx, cluster, backupList)
+	storageSource := persistentvolumeclaim.GetCandidateStorageSourceForReplica(ctx, r.Client, cluster, backupList)
 	if storageSource != nil {
 		job = specs.RestoreReplicaInstance(*cluster, nodeSerial)
 	}
--- a/pkg/reconciler/persistentvolumeclaim/storagesource.go
+++ b/pkg/reconciler/persistentvolumeclaim/storagesource.go
@@ -25,7 +25,9 @@ import (
 	"github.com/cloudnative-pg/machinery/pkg/log"
 	volumesnapshotv1 "github.com/kubernetes-csi/external-snapshotter/client/v8/apis/volumesnapshot/v1"
 	corev1 "k8s.io/api/core/v1"
+	apierrs "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/utils/ptr"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	apiv1 "github.com/cloudnative-pg/cloudnative-pg/api/v1"
 	"github.com/cloudnative-pg/cloudnative-pg/pkg/utils"
@@ -60,6 +62,7 @@ func GetCandidateStorageSourceForPrimary(
 // to be used to create a replica PVC
 func GetCandidateStorageSourceForReplica(
 	ctx context.Context,
+	c client.Client,
 	cluster *apiv1.Cluster,
 	backupList apiv1.BackupList,
 ) *StorageSource {
@@ -94,6 +97,7 @@ func GetCandidateStorageSourceForReplica(
 
 	if result := getCandidateSourceFromBackupList(
 		ctx,
+		c,
 		cluster,
 		backupList,
 	); result != nil {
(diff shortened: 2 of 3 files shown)


armru:
Hello @simonapencea, crediting you as co-author for your initial analysis, feel free to add suggestions or feedbacks and so on to the patch!

simonapencea:
This is very close to what I had in mind as well — thanks for taking this on!

One thing that might still be worth noting is that the check and the creation of the replica are not a single atomic operation. Because of that, there is technically a small window where the volume could exist when the check happens but be deleted before it is actually used as the data source.

That said, this change strengthens the behavior quite a bit and should cover the vast majority of cases. It also makes recovery from that edge case much easier if it ever happens. 
Right now, recovering from that situation can be fairly involved, whereas with this change I would expect that deleting the pending PVC together with the snapshot-recovery job should allow the next reconciliation loop to create the correct type of job again.

Overall this looks like a solid improvement to me 👍

armru:
Hello @simonapencea! I want to make sure I'm not missing something. Are you referring to:

- The time window between the snapshot existence check and the actual PVC creation within the same reconciliation loop? Would you suggest a different approach to address this? CNPG in general (not just the snapshots pkg) follows the eventual consistency pattern, like many Kubernetes operators where multiple components interact and report state concurrently. That said, if we want to handle recovery for the case where a snapshot is deleted during that narrow execution window, I'm personally open to an eventual contribution on that matter (would love to hear other maintainers opinion on this) but I think it belongs in a separate PR as an incremental improvement.

- Or some other behavior I might be overlooking? If so, I'd appreciate the additional context so we can evaluate whether it warrants a change.

armru:
I've added a commit that handles the case where a VolumeSnapshot is deleted between the operator's existence check and the PVC creation, leaving the PVC stuck in Pending and the restore Job blocked indefinitely. The new code detects these orphaned PVCs and cleans them up along with their associated Jobs, allowing the operator to retry via pg_basebackup on the next reconciliation. I'll let the other maintainers decide whether this warrants a separate PR or can be merged with this one.

simonapencea:
Sorry i missed this, yes, that was exactly the case i was referring to. 

As to including it here or in a separate PR, i have no opinion.

The PR was super valuable even without considering this edge case.
So i am happy either way :)

leonardoce:
Hi! Following a discussion with @armru, we've decided to move the second commit into a separate PR (#10223). This will allow us to discuss and review each change independently.

leonardoce (approved):

gbartolini (approved):

gbartolini:
I approve this, but the current process lacks transparency and control. In the long term, we need to address #9176 to provide clarity to users.

PR #10283, opened 2026-03-12
feat(plugins): handle requeue behavior from lifecycle hooks

### LINKED ISSUE #10290: [Feature]: Support BEHAVIOR_REQUEUE from plugin lifecycle hooks

 Add support for the new BEHAVIOR_REQUEUE behavior (see https://github.com/cloudnative-pg/cnpg-i/issues/293) from plugin lifecycle hooks,
 allowing plugins to request reconciliation retries without causing cluster
 failures.

  Plugins may need to wait for external dependencies (custom resources, secrets,
   etc.) before they can properly configure cluster resources. Currently, if a
  plugin cannot proceed, it must return an error, which puts the cluster in a
  failure state. This is problematic for recovery clusters where the plugin
  configuration may be created after the cluster.

1. Lifecycle client: When plugin returns `BEHAVIOR_REQUEUE`, return a           
  `RequeueError` (not a failure) with the requested duration.
  2. Cluster controller: Convert `RequeueError` to graceful requeue               
  (ctrl.Result{RequeueAfter: duration}) with info-level logging.
  3. Instance creation: Call lifecycle hook before PVC readiness checks to
  prevent deadlocks with `WaitForFirstConsumer` storage class.

 1. Skip PVC check for `WaitForFirstConsumer`: Storage class info not always     
  available, changes behavior globally.                                         
  2. Track requeue state via annotation: Adds state management complexity,      
  annotation cleanup logic.
  3. Always retry plugin errors: Cannot distinguish real errors from "not ready
  yet".
  4. Require plugins configured before cluster: Poor UX, breaks GitOps
  workflows.

### Additional context

_No response_

### Backport?

Yes

Yes

### Code of Conduct

gabriele-wolfox:
Closes #10290 . 
Needs https://github.com/cloudnative-pg/cnpg-i/issues/293, then a commit to update go.mod reference to CNPGI module, removing the replace directives.

Handle the `BEHAVIOR_REQUEUE` response from plugins in lifecycle hooks, allowing plugins to signal that a dependency is not yet available without treating it as an error condition.

Changes:
- Add `RequeueError` type and helper functions in plugin client
- Handle `BEHAVIOR_REQUEUE` in lifecycle hook response processing
- Handle `RequeueError` gracefully in cluster controller without setting failure phase


armru (approved):

armru (approved):

mnencia:
Closing this PR. After further discussion, we decided that this change is not needed. The cnpg-i lifecycle requeue API change has been reverted in cloudnative-pg/cnpg-i#297.

PR #10310, opened 2026-03-16
fix: Preserve LoadBalancerClass if unset in the servicespec

sharifmshaker:
When a LoadBalancerClass is not defined in a proposed service template spec it causes an error with the Cluster's Service reconciliation. 
```
phase: Unable to create required cluster objects
phaseReason: 'Service "test-rw-external" is invalid: spec.loadBalancerClass:
      Invalid value: null: may not change once set'
```
I believe this is due to the recent enhancement https://github.com/cloudnative-pg/cloudnative-pg/commit/c16f7483. The proposed change simply ensures that LoadBalancerClass is also configured with the `living` Service spec value if it is not set on the `proposed` service. This means that we will get the same error phase if a LB class is explicitly changed in the template spec, but we will avoid the error if the proposed template does not explicitly set a class.


armru (approved):

mnencia:
Superseeded by https://github.com/cloudnative-pg/cloudnative-pg/pull/10311

PR #10312, opened 2026-03-16
feat: allow targetImmediate without backupID

### LINKED ISSUE #8031: [Feature]: Allow omitting backupId with targetImmediate true

Right now `backupId` is required when setting `targetImmediate`. Why not allow omitting `backupId`, and use the latest one from the catalog?

Right now `backupId` is required when setting `targetImmediate`. Why not allow omitting `backupId`, and use the latest one from the catalog?

Setting the backupId...

### Additional context

_No response_

### Backport?

No

No

### Code of Conduct

dnplkndll:
## Summary

When `targetImmediate` is set in `bootstrap.recovery.recoveryTarget`, allow omitting `backupID` — the operator automatically selects the latest available backup from the catalog.

**Problem:** Currently, `backupID` is required for all non-time-based recovery targets (`targetImmediate`, `targetName`, `targetXID`). While this requirement makes sense for `targetName` and `targetXID` (where the operator cannot determine which base backup is appropriate), `targetImmediate` has no such ambiguity — it simply means "stop at the first consistent state after restoring the base backup."

This forces users who want a simple "restore latest backup without WAL replay" to first look up the backup ID externally (via kubectl or scripts), making what should be a declarative operation into a two-step imperative process.

**Solution:** Remove `targetImmediate` from the webhook validation that requires `backupID`. The barman-cloud catalog's `FindBackupInfo()` already handles this case correctly — when `backupID` is empty and no time/LSN target is set, it falls through to selecting the latest available backup via `findLatestBackupFromTimeline()`.

## Changes

- `internal/webhook/v1/cluster_webhook.go`: Exclude `TargetImmediate` from the `labelBasedPITR` check that requires `backupID`
- `internal/webhook/v1/cluster_webhook_test.go`: Add test cases for `targetImmediate` without and with `backupID`

## Use case

```yaml
spec:
  bootstrap:
    recovery:
      source: prod
      recoveryTarget:
        targetImmediate: true
        # No backupID needed — operator picks latest automatically
```

This enables fully declarative "clone from latest backup" workflows (e.g., staging refresh from production) without requiring external scripts to look up backup IDs.

## Test plan

- [x] Existing webhook validation tests pass (397/397)
- [x] New test: `targetImmediate: true` without `backupID` is accepted
- [x] New test: `targetImmediate: true` with explicit `backupID` still works
- [x] `targetName` and `targetXID` still require `backupID` (unchanged behavior)

Closes #8031

### DIFF
--- a/internal/webhook/v1/cluster_webhook.go
+++ b/internal/webhook/v1/cluster_webhook.go
@@ -1432,15 +1432,17 @@ func (v *ClusterCustomValidator) validateRecoveryTarget(r *apiv1.Cluster) field.
 	}
 
 	// When using a backup catalog, we can identify the backup to be restored
-	// only if the PITR is time-based. If the PITR is not time-based, the user
-	// need to specify a backup ID.
+	// only if the PITR is time-based or targetImmediate is set (which uses
+	// the latest available backup). If the PITR is label-based (targetName
+	// or targetXID), the user must specify a backup ID.
 	// If we use a dataSource, the operator will directly access the backup
 	// and a backupID is not needed.
 
-	// validate BackupID is defined when TargetName or TargetXID or TargetImmediate are set
+	// validate BackupID is defined when TargetName or TargetXID are set
+	// (targetImmediate is excluded: when no backupID is given, the operator
+	// automatically selects the latest available backup from the catalog)
 	labelBasedPITR := recoveryTarget.TargetName != "" ||
-		recoveryTarget.TargetXID != "" ||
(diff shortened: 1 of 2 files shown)


dnplkndll:
@mnencia what is the process to review review? 
this looks like maybe a bug. but my use case is the k8s reviews should bootstrap the most recent prod, quickly as PR might be short lived and this binary backup is fast and effective at just that. WAL restore more that doubles the time on smallish DBs
sanitization usually not needed but something I was wondering about as an option somewhere. but that is out of scope.

PR #10324, opened 2026-03-18
test(e2e): fix AfterAll cleanup failure when minio test is skipped

mnencia:
The backup_restore_minio AfterAll handler attempts to delete cluster resources even when BeforeAll was skipped on non-kind/k3d clusters, causing a spurious failure because the namespace was never created.

Skip the cleanup when namespace is empty.

### DIFF
--- a/tests/e2e/backup_restore_minio_test.go
+++ b/tests/e2e/backup_restore_minio_test.go
@@ -114,6 +114,9 @@ var _ = Describe("MinIO - Backup and restore", Label(tests.LabelBackupRestore), 
 		})
 
 		AfterAll(func() {
+			if namespace == "" {
+				return
+			}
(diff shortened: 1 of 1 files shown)


tests/e2e/backup_restore_minio_test.go
```diff
@@ -114,6 +114,9 @@ var _ = Describe("MinIO - Backup and restore", Label(tests.LabelBackupRestore),
 		})
 
 		AfterAll(func() {
+			if namespace == "" {
+				return
+			}
```
jsilvela:
```suggestion
			if CurrentSpecReport().State.Is(types.SpecStateSkipped) {
				return
			}
```

jsilvela (commented):
For clarity, I think this should be a check for Skip, not the lack of a namespace

jsilvela (approved):

PR #10352, opened 2026-03-22
Fix WAL data loss on instance hibernation/shutdown (#10341)

### LINKED ISSUE #10341: [Bug]: WAL archive gap after cluster hibernation with plugin-barman-cloud

### Contact Details

andrew_gleeson@external.mckinsey.com

### Version

older in 1.28.x

1.34

Other

Helm

### What happened?

We're experiencing data loss when restoring from WAL archives after hibernating a CNPG cluster that uses `plugin-barman-cloud`. The last WAL segment does not appear to be archived to S3 during shutdown. The same scenario works as expected with in-tree `barmanObjectStore`.

## Steps to reproduce

1. Create a CNPG cluster with `plugin-barman-cloud` configured for WAL archiving to S3, let it run for a while, WAL archiving is working correctly
2. Backup the database to S3
3. Insert data into the database (for timestamping purposes)
4. Hibernate the cluster
5. Create a new cluster restoring from the WAL archive of the hibernated one
6. The most recent data written before hibernation is missing — the last WAL segment was never archived to S3

The same steps using in-tree `barmanObjectStore` instead of the plugin work correctly — all data is recovered.

### Observations

1. **Native sidecar ordering:** We've confirmed from the actual pod spec that plugin-barman-cloud is deployed as a Kubernetes native sidecar (initContainers with restartPolicy: Always). Our clusters run K8s 1.33.8 and 1.34.4, where native sidecar lifecycle guarantees are GA. The kubelet will not send SIGTERM to the sidecar until after the postgres container has fully exited.

2. **preStop hooks:** We've seen recommendations to look at adding a hook to run `pg_switch_wal` via  `cnpg.io/podPatch` annotation. This feels a bit fragile as dependent on index of containers to attach this to the correct container.

3. **Possible Root cause:** We believe the actual root cause is that `ArchiveAllReadyWALs()` _(in archiver.go:63-120)_ is only called during switchover/demotion _(instance_startup.go:129)_, not during normal pod termination. After `TryShuttingDownSmartFast()` returns in _lifecycle.go:148_, the instance manager exits without sweeping remaining `.ready` WAL files -- even though the native sidecar is still alive and could handle the gRPC calls. Can you confirm whether this gap has been identified and whether there are plans to address it?

## Environment 

- CNPG Operator version: v1.28.0
- Plugin barman-cloud version: v0.10.0
  - running as a native sidecar *(therefore doesn't get SIGTERM on initial shutdown of the pod)*
- Kubernetes version: v1.34.4
- PostgreSQL 16

### Cluster resource

```shell
kind: Cluster
metadata:
  annotations:
    argocd.argoproj.io/sync-app: "-50"
    argocd.argoproj.io/tracking-id: appinstance-0123456789a:postgresql.cnpg.io/Cluster:appinstance-0123456789a/postgresql
  labels:
    app.kubernetes.io/component: db
    app.kubernetes.io/instance: appinstance-0123456789a
    app.kubernetes.io/managed-by: Helm
    app.kubernetes.io/name: postgresql
    app.kubernetes.io/part-of: helm-app
    app.kubernetes.io/version: 1.0.0
    app.acmefirm.com/manifest: "manifest-0123456789a"
    app.acmefirm.com/appCategory: internal
    app.acmefirm.com/appName: "0123456789a"
    app.acmefirm.com/appUuid: 0123456789a
    app.acmefirm.com/tier: dev
    helm.sh/chart: postgresql-3.0.0
  name: postgresql
  namespace: appinstance-0123456789a
spec:
  affinity:
    nodeSelector:
      node-asg: apps
    podAntiAffinityType: required
    tolerations:
    - effect: NoSchedule
      key: node-asg
      operator: Equal
      value: apps
    topologyKey: topology.kubernetes.io/zone
  bootstrap:
    initdb:
      database: app
      encoding: UTF8
      localeCType: C
      localeCollate: C
      owner: app
      postInitApplicationSQLRefs:
        configMapRefs:
        - key: sql
          name: postgresql-init-script
  description: The database cluster of the app application
  enablePDB: true
  enableSuperuserAccess: false
  failoverDelay: 0
  imageCatalogRef:
    apiGroup: postgresql.cnpg.io
    kind: ImageCatalog
    major: 16
    name: postgresql-image-catalog
  imagePullPolicy: IfNotPresent
  inheritedMetadata:
    annotations:
      cluster-autoscaler.kubernetes.io/safe-to-evict: "false"
    labels:
      app.kubernetes.io/component: db
      app.kubernetes.io/instance: appinstance-0123456789a
      app.kubernetes.io/managed-by: Helm
      app.kubernetes.io/name: postgresql
      app.kubernetes.io/part-of: helm-app
      app.kubernetes.io/version: 1.0.0
      app.acmefirm.com/manifest: "manifest-0123456789a"
      app.acmefirm.com/node-group: apps
      app.acmefirm.com/appCategory: internal
      app.acmefirm.com/appName: "0123456789a"
      app.acmefirm.com/appUuid: 0123456789a
      app.acmefirm.com/tier: dev
      helm.sh/chart: postgresql-3.0.0
      monitoring.acmefirm.com/enabled: "true"
      networking.acmefirm.com/allow-api-server-access: "true"
      networking.acmefirm.com/allow-external-egress-access: "true"
      networking.acmefirm.com/allow-postgresql-access: "true"
  instances: 3
  logLevel: debug
  managed:
    roles:
    - comment: App Read only user
      connectionLimit: -1
      ensure: absent
      inRoles:
      - pg_read_all_data
      inherit: true
      login: true
      name: app_ro
  maxSyncReplicas: 1
  minSyncReplicas: 1
  monitoring:
    customQueriesConfigMap:
    - key: queries
      name: cnpg-default-monitoring
    - key: custom-metrics
      name: postgresql-custom-metrics
    disableDefaultQueries: false
    enablePodMonitor: true
    podMonitorRelabelings:
    - action: replace
      sourceLabels:
      - __meta_kubernetes_pod_label_cnpg_io_instanceName
      targetLabel: cnpg_instance_name
    - action: replace
      sourceLabels:
      - __meta_kubernetes_pod_label_cnpg_io_instanceRole
      targetLabel: cnpg_instance_role
  plugins:
  - enabled: true
    isWALArchiver: true
    parameters:
      barmanObjectName: postgresql-backup
  postgresGID: 26
  postgresUID: 26
  postgresql:
    parameters:
      archive_mode: "on"
      archive_timeout: 5min
      auth_delay.milliseconds: "500"
      dynamic_shared_memory_type: posix
      full_page_writes: "on"
      hot_standby_feedback: "true"
      log_destination: csvlog
      log_directory: /controller/log
      log_filename: postgres
      log_rotation_age: "0"
      log_rotation_size: "0"
      log_truncate_on_rotation: "false"
      logging_collector: "on"
      max_connections: "1000"
      max_parallel_workers: "32"
      max_replication_slots: "32"
      max_slot_wal_keep_size: 3891MB
      max_wal_size: 1GB
      max_worker_processes: "32"
      pg_failover_slots.synchronize_slot_names: slot_for_data_gateway_changefeed_replication
      pg_stat_statements.max: "5000"
      pg_stat_statements.save: "false"
      pg_stat_statements.track: all
      pg_stat_statements.track_planning: "false"
      pg_stat_statements.track_utility: "true"
      shared_memory_type: mmap
      shared_preload_libraries: ""
      ssl_max_protocol_version: TLSv1.3
      ssl_min_protocol_version: TLSv1.3
      track_io_timing: "on"
      wal_keep_size: 512MB
      wal_level: logical
      wal_log_hints: "on"
      wal_receiver_timeout: 5s
      wal_sender_timeout: 5s
    pg_hba:
    - hostssl app streaming_replica all cert
    promotionTimeout: 8
    shared_preload_libraries:
    - auth_delay
    syncReplicaElectionConstraint:
      enabled: false
  primaryUpdateMethod: restart
  priorityClassName: app-non-premptible
  probes:
    liveness:
      isolationCheck:
        connectionTimeout: 1000
        enabled: true
        requestTimeout: 1000
  replicationSlots:
    highAvailability:
      enabled: true
      slotPrefix: _cnpg_
    synchronizeReplicas:
      enabled: true
    updateInterval: 30
  resources:
    limits:
      cpu: "6"
      memory: 2400Mi
    requests:
      cpu: 100m
      memory: 800Mi
  serviceAccountTemplate:
    metadata:
      annotations:
        eks.amazonaws.com/role-arn: arn:aws:iam::987654321234:role/dev-0123456789a-role
  smartShutdownTimeout: 180
  startDelay: 3600
  stopDelay: 1800
  storage:
    pvcTemplate:
      accessModes:
      - ReadWriteOnce
      resources:
        requests:
          storage: 10Gi
      storageClassName: postgresql-data-gp3-appinstance-0123456789a
    resizeInUseVolumes: true
  switchoverDelay: 3600
```

### Relevant log output

```shell

```

### Code of Conduct

abhayclasher:
While looking at the WAL archiving logic, I noticed that when a CNPG instance shuts down normally, any final WAL segments sitting in .ready state never get pushed to S3. The archiver only runs during switchover/demotion, not during regular shutdown. This is a problem especially during cluster hibernation — you lose the last batch of transactions.

The fix is straightforward: call ArchiveAllReadyWALs() after the PostgreSQL shutdown completes, same way we do for switchover. Ensures those final segments don't sit orphaned on local storage.

Fixes #10341

### DIFF
--- a/internal/cmd/manager/instance/run/lifecycle/lifecycle.go
+++ b/internal/cmd/manager/instance/run/lifecycle/lifecycle.go
@@ -31,6 +31,7 @@ import (
 
 	"github.com/cloudnative-pg/cloudnative-pg/pkg/concurrency"
 	"github.com/cloudnative-pg/cloudnative-pg/pkg/management/postgres"
+	"github.com/cloudnative-pg/cloudnative-pg/pkg/management/postgres/archiver"
 )
 
 // PostgresLifecycle implements the manager.Runnable interface for a postgres.Instance
@@ -132,6 +133,9 @@ func (i *PostgresLifecycle) Start(ctx context.Context) error {
 				if err := i.instance.TryShuttingDownSmartFast(ctx); err != nil {
 					contextLogger.Error(err, "error shutting down instance, proceeding")
 				}
+				if err := archiver.ArchiveAllReadyWALs(ctx, i.instance.GetClusterOrDefault(), i.instance.PgData); err != nil {
+					contextLogger.Error(err, "error archiving WALs before exiting, proceeding")
+				}
 				return nil
 
 			case sig := <-signals:
@@ -145,6 +149,9 @@ func (i *PostgresLifecycle) Start(ctx context.Context) error {
 				)
 				if err := i.instance.TryShuttingDownSmartFast(ctx); err != nil {
 					contextLogger.Error(err, "error while shutting down instance, proceeding")
+				}
+				if err := archiver.ArchiveAllReadyWALs(ctx, i.instance.GetClusterOrDefault(), i.instance.PgData); err != nil {
+					contextLogger.Error(err, "error archiving WALs before exiting, proceeding")
 				}
 				return nil
 


aglees:
👋 I have a feeling that the difference with #10345, is important to efficacy of the solution. 

I think the `<-ctx.Done()` path *(line 122)*, `ctx` is already cancelled. This means that in this PR context will propagate it through to gRPC calls to the plugin sidecar. Those calls will fail immediately with context canceled.

Due the `select` block in `lifecycle.go` it might be that this works when the `select` happens to pick the `SIGTERM` case, which is non-deterministic. For our hibernation scenario specifically, it would silently fail to archive WALs roughly half the time, with no indication in the logs beyond a context-cancelled error that could easily be overlooked.

WDYT?

abhayclasher2:
Hey @aglees — that's a solid catch. I'm at work right now so haven't had a chance to dig in properly, but I'll update this from my main account tonight with a proper fix (using a fresh context instead of the cancelled one). Will push the changes then. Thanks for flagging it!

PR #10362, opened 2026-03-23
ci(build): skip prime build when main build fails

mnencia:
Remove `always() && !cancelled()` from the prime build steps so they only run when prior steps, including the main build, succeed.

The prime image build steps used `always() && !cancelled()` in their conditions, causing them to run even when the main container build failed. This resulted in an invalid docker tag (`-prime`) because the `CONTROLLER_IMG` env var was never set by the skipped main build.


jsilvela (approved):

NiccoloFei (approved):

NiccoloFei:
E2E run: https://github.com/cloudnative-pg/cloudnative-pg/actions/runs/23753802899

PR #10373, opened 2026-03-26
Update SECURITY.md

PavithraR15:


gbartolini:
This is not the correct place for CVEs.

PR #10409, opened 2026-04-02
fix(cluster): strip stale primary label during failover

### LINKED ISSUE #10403: [Bug]: Primary pod label not removed during failover, allowing stale -rw service routing

### Contact Details

_No response_

### Version

trunk (main)

_No response_

Other

_No response_

### What happened?

During failover, the old primary's pod retains the `cnpg.io/instanceRole: primary` label longer than necessary. Since the `-rw` service routes traffic based on this label, the old primary can continue receiving connections until the label is eventually reconciled. This is a defense-in-depth issue that could compound with other problems (e.g. premature failover triggered by transient network loss).

Two label reconciliation gaps have been identified:

**1. Primary label not removed when failover starts**

When the operator initiates failover (`replicas.go:139`, setting `targetPrimary` to `PendingFailoverMarker`), it does not remove the primary label from the old primary's pod. The old primary should be cut off from the `-rw` service immediately at this point, since the operator has already committed to failing over.

**2. Primary labels not reconciled when dual primaries are detected**

When the old primary comes back and both pods report as primary, the dual-primary detection at `cluster_controller.go:417` returns early **before** `ReconcileMetadata` at line 438 is reached. This means pod labels are never updated during the dual-primary state, and the `-rw` service keeps routing to both primaries until the old one self-heals.

#### Proposed fix

1. When failover is initiated (setting `targetPrimary` to `PendingFailoverMarker`), immediately strip the primary label from the old primary's pod. This cuts it off from the `-rw` service before the new primary is even elected.
2. When dual primaries are detected, remove the primary label from all pods declaring themselves primary, instead of returning early and skipping `ReconcileMetadata`.

### Cluster resource

```shell

```

### Relevant log output

```shell

```

### Code of Conduct

armru:
When the operator initiates a failover, the old primary's pod keeps its `cnpg.io/instanceRole=primary` label until `ReconcileMetadata` runs. But `ReconcileMetadata` is skipped during the entire failover window (the `CurrentPrimary != TargetPrimary` guard returns early), so the `-rw` service keeps routing to the old primary. If the old primary comes back (e.g. after a temporary network partition), replicas reconnect through the `-rw` service, satisfy the sync quorum, and writes committed on the stale primary are lost to `pg_rewind`.

Introduce a third value for the instance role label, `unhealthy`, and apply it to the old primary as soon as failover starts. Since neither the `-rw` nor the `-ro` service selector matches `unhealthy`, the pod is immediately isolated from all service traffic for the duration of the failover window. `ReconcileMetadata` restores the `replica` label once `CurrentPrimary == TargetPrimary`.

The label is applied best-effort at the point where failover is initiated, and re-applied on every pass of the reconcile loop while the failover is in progress so transient API errors are retried automatically.

Note: stripping the label removes the pod from the service Endpoints, but does not drop TCP connections already established by a replica's walreceiver. This fix closes the reconnection window; established connections must still be terminated by the Postgres-level promotion on the new primary.

Closes #10403

### DIFF
--- a/contribute/technical-architecture.md
+++ b/contribute/technical-architecture.md
@@ -214,9 +214,11 @@ Networking is purely label-driven. The Operator manages the
 
 2. **Replica:** Labeled as `replica`. The `-ro` Service selects these Pods.
 
-3. **Failover:** When a new primary is promoted via `pg_ctl promote`, the
-   Operator updates the labels. The Kubernetes API Server then automatically
-   updates the **Endpoints** for the respective Services.
+3. **Unhealthy:** Transient value applied to the old primary during a
+   failover or switchover, so that neither the `-rw` nor the `-ro`
+   Service selects it. Once the transition completes, the Operator
+   relabels the demoted instance as `replica` and the Kubernetes API
+   Server updates the Service **Endpoints** accordingly.
 
 ## Source Code Reference
 
--- a/docs/src/labels_annotations.md
+++ b/docs/src/labels_annotations.md
@@ -102,15 +102,21 @@ This label is available only on `VolumeSnapshot` resources.
   default users created by CloudNativePG (typically `postgres` and `app`).
 
 `role` - **deprecated**
-:  Whether the instance running in a pod is a `primary` or a `replica`.
-   This label is deprecated, you should use `cnpg.io/instanceRole` instead.
+:  Role of the instance running in a pod: `primary`, `replica`, or
+   `unhealthy`. The `unhealthy` value is transient: the operator sets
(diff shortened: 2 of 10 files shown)


kfcss:
This is great progress. However, stripping the label will not immediately ensure that replicas don't connect to it as there is a delay in these updates. Could be nice if the primary would itself reject replicas connecting to it or tell replicas to stop connecting to anyone until a failover is complete.

leonardoce (approved):

mnencia (approved):

PR #10425, opened 2026-04-08
feat(services): add cluster-wide service template for default services

### LINKED ISSUE #10424: [Feature]: Support IPv6 dual-stack configuration for default cluster services
## Summary

Add the ability to configure `ipFamilyPolicy` and `ipFamilies` on the default cluster services (rw, ro, r, any) via a new `defaults` service template in `spec.managed.services`.

Relates to: https://github.com/cloudnative-pg/cloudnative-pg/discussions/7116

## Motivation

With the [US government mandate (M-21-07)](https://www.whitehouse.gov/wp-content/uploads/2020/11/M-21-07.pdf) requiring 80% IPv6-only assets by end of FY2025, and growing enterprise adoption of dual-stack Kubernetes, CNPG needs first-class support for IPv6/dual-stack service configuration.

**Current state:**
- Default services (rw, ro, r, any) are created without `ipFamilyPolicy` or `ipFamilies`, inheriting the Kubernetes cluster default (typically SingleStack IPv4)
- PostgreSQL itself already listens on all interfaces (`listen_addresses = '*'`) and `pg_hba.conf` generation correctly handles IPv6 addresses with `/128` CIDR
- The only workaround is to disable default ro/r services and recreate them as managed services with explicit IP family config — but the rw service cannot be disabled, making full dual-stack impossible without manual intervention

## Proposed API

Add a `defaults` field to `spec.managed.services` that provides a service template applied to all default services:

```yaml
kind: Cluster
spec:
  managed:
    services:
      defaults:
        serviceTemplate:
          spec:
            ipFamilyPolicy: RequireDualStack
            ipFamilies:
              - IPv4
              - IPv6
```

This is consistent with the existing `additional` managed services API which already supports `ServiceTemplateSpec`.

## Related

- Discussion: #7116
- Existing workaround uses `managed.services.additional` with `disabledDefaultServices` but cannot disable the rw service

cagataygurturk:
## Motivation

Currently, the default services (rw, ro, r, any) are created without `ipFamilyPolicy` or `ipFamilies`, inheriting the Kubernetes cluster default (typically SingleStack IPv4). The only workaround is to disable default ro/r services and recreate them as managed services with explicit IP family config — but the rw service cannot be disabled, making full IPv6/dual-stack impossible without manual intervention.

## Changes

Adds a `spec.managed.services.serviceTemplate` field that acts as a defaults layer for **all** services created by the operator -- both the 4 default services (rw, ro, r, any) and additional managed services. Per-service values always take precedence; the template only fills in unset fields.

Example:

```yaml
spec:
  managed:
    services:
      serviceTemplate:
        spec:
          ipFamilyPolicy: SingleStack
          ipFamilies:
            - IPv6
```

While IPv6/dual-stack is the primary driver, the field is general-purpose and can configure any `ServiceSpec` field (e.g., `externalTrafficPolicy`, `internalTrafficPolicy`, `loadBalancerClass`) across all cluster services.

Tested manually on a live dual-stack cluster — all default services received IPv6 ClusterIPs.

Closes #10424
Relates to https://github.com/cloudnative-pg/cloudnative-pg/discussions/7116


sebhoss:
would this cover services for operator-managed pgbouncers as well?

cagataygurturk:
Hi @jsilvela @NiccoloFei  @litaocdl I appreciate a review here, it seems the request has quite some demand.

PR #10428, opened 2026-04-10
Operator initiates failover without waiting for kubelet to confirm pod is not ready

### LINKED ISSUE #10423: [Bug]: Operator initiates failover without waiting for kubelet to confirm pod is not ready

### Contact Details

_No response_

### Version

trunk (main)

_No response_

Cloud: Azure AKS

_No response_

### What happened?

The operator decides to initiate failover based solely on its own HTTP health check to the pod IP (30s timeout in `rawInstanceStatusRequest`). It does not check whether the kubelet has also detected the failure by marking the pod as NotReady.

This can cause premature failovers in environments where pods can be temporarily frozen without crashing (e.g. Azure VM live migration). During a VM migration, the node is paused for a few seconds (typically 8-30s). If a reconciliation cycle happens to run during the freeze, the operator's HTTP request times out and it initiates failover, even though the kubelet is also frozen and still considers the pod Ready. When the VM migration completes, the old primary comes back with PostgreSQL still running as primary.

The operator already collects `IsPodReady` from the kubelet's `ContainersReady` condition in the `PostgresqlStatus` struct (`pkg/postgres/status.go:103`), but does not use it in the failover decision path.

#### Proposed fix

In `reconcileTargetPrimaryFromPods` (`replicas.go`), before initiating failover, check that the current primary's pod has been marked NotReady by the kubelet. If the operator's HTTP check fails but `IsPodReady` is still true, the operator should wait rather than immediately triggering failover.

This naturally provides ~30s of protection (readiness probe period=10s, default failure threshold=3), which covers most VM live migration windows.

### Cluster resource

```shell

```

### Relevant log output

```shell

```

### Code of Conduct

bramhanandlingala:
Closes #10423

Currently, the operator:

Uses HTTP health check (/pg/status)
Ignores Kubernetes Pod readiness (IsPodReady)
This causes premature failovers and split-brain risk.

The failover logic relies on AllReadyInstancesStatusUnreachable(), which treats HTTP failures as full unavailability without considering Kubernetes Pod readiness (IsPodReady).

AllReadyInstancesStatusUnreachable() to:
Consider only pods that are marked Ready by the kubelet
Ignore transient HTTP failures if the pod is still Ready
Trigger failover only when pods are both NotReady and unreachable via HTTP


bramhanandlingala:
Currently, the operator:
Uses HTTP health check (/pg/status)
Ignores Kubernetes Pod readiness (IsPodReady)

This causes Premature failovers and Split-brain risk

The failover logic relies on `AllReadyInstancesStatusUnreachable()`, which treats HTTP failures 
as full unavailability without considering Kubernetes Pod readiness (`IsPodReady`).

fix could be below changes 
Update `AllReadyInstancesStatusUnreachable()` to:
- Consider only pods that are marked Ready by the kubelet
- Ignore transient HTTP failures if the pod is still Ready
- Trigger failover only when pods are both NotReady and unreachable via HTTP

issue #10423 
closes #10423

bramhanandlingala:
Please review this PR, Any blockers please let me know

bramhanandlingala:
Hi @cloudnative-pg/maintainers, @jsilvela @NiccoloFei @litaocdl
Could you please review this pull request? Let me know if there are any blockers or feedback. Thank you!

vamsivasireddy-ops (approved):

PR #10452, opened 2026-04-15
chore: return of GOLANG_VERSION to have specific version on govulncheck

### LINKED ISSUE #10450: [Bug]: govulncheck is using the wrong Go version for checking

### Contact Details

_No response_

### Version

1.29 (latest patch)

1.35

YAML manifest

### What happened?

After we merged of https://github.com/cloudnative-pg/cloudnative-pg/pull/10134 we removed the `go-version` field on the govulncheck action, but in this action the go-version is set by default to `stable` and thus skip the `go-version-file` having the following warning:
```
Warning: Both go-version and go-version-file inputs are specified, only go-version will be used
```
When setting up Go, this lead to have the wrong version in the govulncheck making the check useless

### Cluster resource

```shell

```

### Relevant log output

```shell

```

### Code of Conduct

sxd:
The check requries to have a specific version passed instead of
using the one from go.mod, otherwise it will always pass the
`stable` as the version which is not the desire behavior, better
to be explicit.

Closes #10450

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -18,6 +18,8 @@ permissions: read-all
 
 # set up environment variables to be used across all the jobs
 env:
+  # renovate: datasource=golang-version depName=golang versioning=loose
+  GOLANG_VERSION: "1.26.1"
   # renovate: datasource=github-releases depName=golangci/golangci-lint versioning=loose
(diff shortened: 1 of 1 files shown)


gbartolini:
Why did you close this PR @sxd? Wasn't it good?

PR #10464, opened 2026-04-16
chore: add native fuzz tests and a `make fuzz` target

### LINKED ISSUE #10459: [Feature]: Add fuzz command to Makefile

The fuzz command should be used to run fuzz test by a short time that can be expanded

The fuzz command should be used to run fuzz test by a short time that can be expanded

None

### Additional context

_No response_

### Backport?

No

Yes

### Code of Conduct

sxd:
Adds a first batch of native fuzz targets and a `make fuzz` target that
discovers them automatically; new fuzz functions are picked up without
touching the Makefile. Default fuzz time is 30s, overridable via
`FUZZ_TIME`.

Not wired into CI: manual hardening runs and an OSS-Fuzz integration
seed. The current targets cover annotation/label inheritance, namespace
and plugin list parsing, and the postgresql.conf rendering pipeline
including a well-formedness check on the output.

Closes #10459

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -72,6 +72,7 @@ OPM_VERSION ?= v1.65.0
 PREFLIGHT_VERSION ?= 1.17.1
 OPENSHIFT_VERSIONS ?= v4.16-v4.21
 ARCH ?= amd64
+FUZZ_TIME ?= 30s
 
 export CONTROLLER_IMG
 export BUILD_IMAGE
@@ -130,6 +131,10 @@ test-race: generate fmt vet manifests envtest ## Run tests enabling race detecti
 	source <(${ENVTEST} use -p env --bin-dir ${ENVTEST_ASSETS_DIR} ${ENVTEST_K8S_VERSION}) ;\
 	go run github.com/onsi/ginkgo/v2/ginkgo -r -p --skip-package=e2e \
 	  --race --keep-going --fail-on-empty --randomize-all --randomize-suites
+
+fuzz: ## Run native fuzzing targets.
+	go test -run=^$$ ./pkg/postgres -fuzz=FuzzCreatePostgresqlConfigurationSanitization -fuzztime=$(FUZZ_TIME)
+	go test -run=^$$ ./internal/configuration/... -fuzz=FuzzConfigurationParsing -fuzztime=$(FUZZ_TIME)
 
 e2e-test-kind: ## Run e2e tests locally using kind.
 	CLUSTER_ENGINE=kind hack/e2e/run-e2e-local.sh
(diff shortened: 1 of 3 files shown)


internal/configuration/fuzz_test.go
```diff
@@ -0,0 +1,78 @@
...
+			IncludePlugins:       includePlugins,
+		}
+
+		expectedMatch := evaluateGlobPatterns(patterns, value)
+		if got := config.IsAnnotationInherited(value); got != expectedMatch {
+			t.Fatalf("annotation inheritance mismatch: got %t want %t", got, expectedMatch)
+		}
+		if got := config.IsLabelInherited(value); got != expectedMatch {
+			t.Fatalf("label inheritance mismatch: got %t want %t", got, expectedMatch)
+		}
```
GabriFedi97: Are these kind of assertions useful? I don't think they can ever fail actually. Isn't fuzzing more about testing potential crashes rather then inspecting the output of individual functions?

Makefile
```diff
@@ -72,6 +72,7 @@ OPM_VERSION ?= v1.65.0
 PREFLIGHT_VERSION ?= 1.17.1
 OPENSHIFT_VERSIONS ?= v4.16-v4.21
 ARCH ?= amd64
+FUZZ_TIME ?= 30s
```
GabriFedi97, line 75: Wasn't this supposed to be 10s based on the PR description?

sxd: I changed the time after XD good catch! I'll keep 30s because looked better than 10s

GabriFedi97 (approved):

litaocdl (approved):

internal/configuration/fuzz_test.go
mnencia: You were right. Replaced the wrapper-equality assertions. The fuzz now asserts independent properties: any literal pattern equal to `value` must match (exact-match semantics of `path.Match`), and the comma-list parsers are fixed points under join-then-split. Go's fuzz harness still catches panics implicitly; these are the value-add on top.

mnencia (approved):

mnencia:
/ok-to-merge no impact on the code, only manual testing

pkg/postgres/configuration_fuzz_test.go
```diff
@@ -0,0 +1,160 @@
...
+			continue
+		}
+		key, value, ok := strings.Cut(line, " = ")
+		if !ok {
+			t.Fatalf("line %d has no `key = value` separator: %q", i, line)
+		}
+		if key == "" || strings.ContainsAny(key, " \t'\"") {
+			t.Fatalf("line %d has malformed key %q", i, key)
+		}
+		if len(value) < 2 || value[0] != '\'' || value[len(value)-1] != '\'' {
```
GabriFedi97:
```suggestion
		if len(value) < 2 || !strings.HasPrefix(value, "'") || !strings.HasSuffix(value, "'") {
			t.Fatalf("line %d value not single-quoted: %q", i, value)
		}
```

PR #10515, opened 2026-04-22
fix: escape backslashes and control chars in PostgreSQL config values

### LINKED ISSUE #10506: [Bug]: Some strings aren't being escaped properly

### Contact Details

_No response_

### Version

1.29 (latest patch)

1.35

YAML manifest

### What happened?

Some strings in that arrive to `custom.conf` will arrive without being properly escaped creating corrupted configuration that may or may not be ended up in a cluster that will not start or reload properly.

Some of the places for this happening are:
https://github.com/cloudnative-pg/cloudnative-pg/blob/main/pkg/postgres/configuration.go#L904
https://github.com/cloudnative-pg/cloudnative-pg/blob/main/api/v1/cluster_funcs.go#L1474

There's probably more places for this to happen.

We should validate and/or escape the strings before arriving to `custom.conf` file.

### Cluster resource

```shell

```

### Relevant log output

```shell

```

### Code of Conduct

armru:
Values written into custom.conf were only escaped for single quotes. Backslashes and control characters (LF, CR, TAB, BS, FF) passed through unescaped, which either corrupted the configuration (e.g. a user-supplied recovery_target_name containing ' or a newline) or silently mangled the runtime value (a \ in log_line_prefix stripped by PostgreSQL's config lexer).

Introduce configfile.RenderPostgresConfiguration as the single canonical helper that emits a postgresql.conf-style fragment from a map[string]string, with values escaped according to the guc-file.l rules (\, ', \n, \r, \t, \b, \f). Migrate every config-file-emitting site to it: pkg/configfile.UpdateConfigurationContents (replacing pq.QuoteLiteral, which emits an E'...' SQL-style literal that the config lexer rejects when the value contains a backslash), pkg/postgres.CreatePostgresqlConfFile, api/v1.(*RecoveryTarget).BuildPostgresOptions (previously did no escaping at all for recovery_target_name/_xid/_lsn/_timeline/_time), and pkg/management/postgres.getRestoreWalConfig and RestoreSnapshot's restore_command construction. After this change the inner escapePostgresConfLiteral is private to pkg/configfile.

Upgrade impact: clusters whose PostgreSQL parameters contain \ or a control character (most commonly a literal tab in log_line_prefix) will see one pg_reload_conf() on the first reconcile after upgrade because the config-file SHA256 changes. No restart, no connection drops. Clusters with only default or quote-only special values are unaffected.

Related #10518
Closes #10506

Reported-by: Koda Reef <kodareef5@users.noreply.github.com>

### DIFF
--- a/api/v1/cluster_funcs.go
+++ b/api/v1/cluster_funcs.go
@@ -41,6 +41,7 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 
 	"github.com/cloudnative-pg/cloudnative-pg/internal/configuration"
+	"github.com/cloudnative-pg/cloudnative-pg/pkg/configfile"
 	"github.com/cloudnative-pg/cloudnative-pg/pkg/system"
 	"github.com/cloudnative-pg/cloudnative-pg/pkg/utils"
 	contextutils "github.com/cloudnative-pg/cloudnative-pg/pkg/utils/context"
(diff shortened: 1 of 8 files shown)


leonardoce (approved):

mnencia (approved):

NiccoloFei (approved):

sxd:
@mnencia @armru @leonardoce shouldn't we try to have some kind of validation also at OpenAPI level?

mnencia:
@sxd #10565 already does this for the recovery target fields: admission-time validation on `targetName` (rejects ASCII control chars, enforces MAXFNAMELEN) and `targetXID` (non-negative 32-bit int). It's webhook validation rather than OpenAPI annotations, but it runs at admission time too. Let's move the discussion there.

PR #10544, opened 2026-04-24
feat(e2e): add Helm deployment method for local testing

### LINKED ISSUE #10052: [Feature]: install CNPG via  OSS helm chart

Currently we only install CNPG build from source in our E2E and CD.

We want to  install CNPG using Helm chart present in [charts](https://github.com/cloudnative-pg/charts).

NA

### Additional context

_No response_

### Backport?

Yes

Yes

### Code of Conduct

gbartolini:
Introduce CNPG_DEPLOYMENT_METHOD=helm as an alternative deployment path for e2e runs on kind and k3d clusters with OPERATOR=local. When set, the operator is deployed via the official cloudnative-pg/charts Helm chart instead of a raw kustomize manifest.

Key changes:

- Add deploy_operator_with_helm() to common/20-utils-k8s.sh. When OPERATOR=local the chart's default image is overridden with the locally-built one. CRDs are applied via 'kubectl -k config/helm' so they always match the operator binary under test.
- Wire the new deploy-from-helm action into manage.sh; the existing deploy dispatcher now checks CNPG_DEPLOYMENT_METHOD before OPERATOR.
- run-e2e.sh dispatches to deploy_operator_with_helm when CNPG_DEPLOYMENT_METHOD=helm.
- run-e2e-local.sh exports CNPG_DEPLOYMENT_METHOD so it is visible inside the nested ginkgo suite.
- suite_test.go logs the active deployment method in the test configuration header.
- continuous-delivery.yml exposes cnpg_deployment_method as a workflow_dispatch input and threads it through the kind and k3d jobs.
- Documentation updated accordingly.

Closes #10052

Assisted-by: Claude


hack/e2e/run-e2e.sh
```diff
@@ -149,10 +149,14 @@ if [[ "${TEST_CLOUD_VENDOR}" != "ocp" ]]; then
   bright=${bright:-}
   reset=${reset:-}
   source "${ROOT_DIR}/hack/testing-tools/common/20-utils-k8s.sh"
-  if [[ "${OPERATOR:-local}" == "local" ]]; then
+  if [[ "${OPERATOR:-local}" == "local" ]] && [[ "${CNPG_DEPLOYMENT_METHOD}" == "manifest" ]]; then
```
danishedb:
```suggestion
  if [[ "${OPERATOR}" == "local" ]] && [[ "${CNPG_DEPLOYMENT_METHOD}" == "manifest" ]]; then
```

danishedb: Since we are already defaulting to local at the top.

hack/setup-cluster.sh
```diff
@@ -133,6 +132,14 @@ Options:
                           snapshot for that branch.
                           Env: OPERATOR
 
+    -d|--deployment-method
+      <METHOD>            Deployment method for the operator. Use 'manifest'
+                          (default) to deploy via a kustomize-generated
+                          manifest, or 'helm' to deploy via the official
+                          Helm chart. Only 'manifest' is supported when
```
danishedb:
```suggestion
                          Helm chart "https://cloudnative-pg.github.io/charts". Only 'manifest' is supported when
```

danishedb (approved):

hack/e2e/run-e2e.sh
NiccoloFei: Fixed in https://github.com/cloudnative-pg/cloudnative-pg/pull/10544/changes/5a1afb18f866dea4c607886ba8bc9b3f829d28b3

NiccoloFei:
E2E runs:
* On Kind via helm: https://github.com/cloudnative-pg/cloudnative-pg/actions/runs/25445955982
* On K3D via helm: https://github.com/cloudnative-pg/cloudnative-pg/actions/runs/25496211949

NiccoloFei:
LGTM, although there are some limitations currently:
1. **Deploy Chart Releases**: It's only possible to perform a local deploy with helm. Ideally you should also be able to deploy a published chart release, and in that case it should use the CRDs from the release itself. 
However this is a feature that doesn't really bring additional benefits at the moment, since it would only make sense locally, and we currently don't have any specific need for this.

2. **Continuos Delivery E2Es via helm**:  E2Es via helm are currently only triggerable manually. We should set up our scheduled E2Es to automatically use `helm` as method when deploying on K3D Clusters (as previously agreed upon). Opening a follow up ticket for this: https://github.com/cloudnative-pg/cloudnative-pg/issues/10653

3. **Upgrade E2Es**: Upgrade E2Es via helm are currently disabled, so when `CNPG_DEPLOYMENT_METHOD=helm` Upgrade E2E tests are just skipped. Upgrade E2Es currently perform the following 2 cases (each for both online and rolling upgrades):
      * upgrading from the most recent tag to the current operator (this would require point 1 to allow deploying the most recent Chart Release)
      * upgrading from the current operator to a `prime` operator with a new hash
   
    Both cases would require additional logic inside `tests/e2e/upgrade_test.go` to support the `helm` method. I'm not sure if this is something we eventually want to chase or if we're fine with performing upgrade E2Es just via manifest, it requires more discussion.


Opened also issue https://github.com/cloudnative-pg/cloudnative-pg/issues/10652.

NiccoloFei (approved):

mnencia (approved):

PR #10552, opened 2026-04-24
feat(cluster): retry transient wal-restore failures before promotion

armru:
Add spec.walRestoreRetryTimeout and a retry loop inside `manager wal-restore` so that transient failures during WAL replay — network blips, flaky object stores, plugin glitches — do not cause a replica to promote on an incomplete archive and lose already-archived transactions.

The command now classifies each attempt's outcome:
  * barman-cloud exit 0 or gRPC Ok    -> success (exit 0)
  * barman-cloud exit 1 or gRPC NotFound -> WAL not found (exit 1,
    unchanged; PostgreSQL can transition to streaming as before)
  * anything else -> transient, retry with exponential backoff up to
    retryBackoffCap (30s), within the user-configurable budget

If the budget is exhausted the command exits with code 255, which tells PostgreSQL to stop log-shipping replication rather than promote on a partial archive. The budget defaults to 5 minutes; the new spec.walRestoreRetryTimeout field (*metav1.Duration) lets operators tune it, and the admission webhook rejects zero or negative values so there is no silent opt-out of the protection.


leonardoce (approved):

gbartolini:
Is this related to https://github.com/EnterpriseDB/barman/commit/4f43dc59df5e598de002fea8c8aaa393e68c4dd6?

PR #10565, opened 2026-04-28
fix: tighten recovery target validation

mnencia:
TargetXID was previously unchecked and a malformed value surfaced only at PostgreSQL startup. Validate it as a non-negative 32-bit integer at admission, matching PG's TransactionId width so values that would silently truncate to a different XID are rejected up front.

TargetName accepts arbitrary text in pg_create_restore_point, but ASCII control characters (NUL, newline, DEL) are never legitimate and would either break the line-oriented config file or signal a hostile spec. Reject them. Also enforce the 64-byte MAXFNAMELEN limit at admission rather than letting PG fail at recovery.

Related #10515

Suggested-by: Koda Reef <kodareef5@users.noreply.github.com>

### DIFF
--- a/internal/webhook/v1/cluster_webhook.go
+++ b/internal/webhook/v1/cluster_webhook.go
@@ -1406,6 +1406,39 @@ func (v *ClusterCustomValidator) validateRecoveryTarget(r *apiv1.Cluster) field.
 		}
 	}
 
+	// PostgreSQL casts recovery_target_xid to TransactionId (uint32); a value
+	// above 2^32-1 would be silently truncated. Use bitSize 32 so we reject
+	// rather than admit a different XID than the user wrote.
+	if recoveryTarget.TargetXID != "" {
+		if _, err := strconv.ParseUint(recoveryTarget.TargetXID, 10, 32); err != nil {
+			result = append(result, field.Invalid(
+				field.NewPath("spec", "bootstrap", "recovery", "recoveryTarget", "targetXID"),
+				recoveryTarget.TargetXID,
+				"recovery target xid must be a non-negative 32-bit integer"))
+		}
+	}
+
+	// PostgreSQL enforces MAXFNAMELEN (64) on recovery_target_name and on
+	// pg_create_restore_point; mirror it at admission for a better error.
+	if len(recoveryTarget.TargetName) >= 64 {
+		result = append(result, field.Invalid(
+			field.NewPath("spec", "bootstrap", "recovery", "recoveryTarget", "targetName"),
+			recoveryTarget.TargetName,
+			"recovery target name must be shorter than 64 bytes"))
+	}
+
+	// pg_create_restore_point accepts arbitrary text, but a name with
+	// newlines or NUL is never legitimate and is a strong signal of a
+	// malformed or hostile spec.
+	if strings.ContainsFunc(recoveryTarget.TargetName, func(r rune) bool {
+		return r < 0x20 || r == 0x7F
+	}) {
+		result = append(result, field.Invalid(
+			field.NewPath("spec", "bootstrap", "recovery", "recoveryTarget", "targetName"),
+			recoveryTarget.TargetName,
+			"recovery target name must not contain ASCII control characters"))
+	}
+
 	// When using a backup catalog, we can identify the backup to be restored
 	// only if the PITR is time-based. If the PITR is not time-based, the user
 	// need to specify a backup ID.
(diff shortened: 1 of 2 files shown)


leonardoce (approved):

sxd:
@mnencia why not doing the verification also in the API? This because the webhook can be skipped but not the API

mnencia:
Fair point about API-level validation. The other recovery target fields are already validated in the webhook, so this PR follows the existing pattern. Moving these checks to CEL would be a good improvement, but I'd do it as a separate patch that also wires up envtest in this package so the rules get test coverage.

gbartolini (approved):

sxd:
@mnencia I'm creating the issue for it then

sxd:
Follow up issue created here https://github.com/cloudnative-pg/cloudnative-pg/issues/10621

gbartolini:
> Follow up issue created here #10621

Isn't this a broader issue, unrelated to this specific ticket? My understanding is that we have normally relied on the webhook for this kind of validation.

sxd:
@gbartolini @mnencia I don't think we should merge this before handling the situation described in #10621 since it will be a partial fix. Probably the PR for #10621 can be merged into this PR to have a full fix?

@gbartolini yes, normally we relied on it, but webhook can be skipped and there has been discussions about having these webhook removed in some configurations, so we should also start adding these verification on the API. On the other hand, having the validation on the API helps any GUI to validated objects before sinding it to the kubernetes clsuter.

I think it's not necessary to say that if webhook can be skipped, all the validations there are useless and will arrive to the PostgreSQL configuration.

leonardoce:
I believe we should merge this and we should merge #10621 too.
I don't see so much difference between merging a single PR or two PRs in a row it they both reach the release.

You can [skip the webhook](https://cloudnative-pg.io/docs/1.29/labels_annotations#predefined-labels) but, doing that:
> [you] permit unsafe or destructive operations. Use this setting with caution and at your own risk.

About webhooks being not installed... I worked on a patch a while ago to prevent the operator and the instance manager from receiving invalid CR definitions:  https://github.com/cloudnative-pg/cloudnative-pg/commit/09c0ade30d33f597b8294f06c55d983ab70a58cf

It is a separate topic, but would something like this help?

gbartolini:
Thanks for the follow-up, @sxd. I want to address the premise here, because I think it affects the conclusion.

My understanding and position is that currently webhooks are not optional in a supported CloudNativePG installation.

The operator itself treats their absence as an anomaly: in [`setDefaults`](https://github.com/cloudnative-pg/cloudnative-pg/blob/main/internal/controller/cluster_controller.go#L788), the controller detects when the mutating webhook did not run, logs `"Admission controllers (webhooks) appear to have been disabled. Please enable them for this object/namespace"`, and patches the object itself to apply the missing defaults. This is a remediation path for a broken configuration, not a supported deployment variant.

The [`cnpg.io/validation` annotation](https://cloudnative-pg.io/docs/1.29/labels_annotations#predefined-annotations) exists as a per-object escape hatch, but it is documented with an explicit warning that disabling validation may permit unsafe or destructive operations. Similarly, the [webhook disabled warning](https://github.com/cloudnative-pg/cloudnative-pg/blob/main/internal/webhook/v1/common.go#L111) in the admission path states that all changes are accepted without validation and urges extreme caution. The [installation documentation](https://cloudnative-pg.io/docs/1.29/installation_upgrade) does not describe any supported configuration without webhooks — the GKE note instructs users to open port 9443, not to remove the webhook.

So, in my opinion, the validation in this PR is not partial. It is complete relative to the architecture we support and the existing pattern across all other recovery target fields.

@mnencia's response on 4 May already addressed this correctly: CEL is a worthwhile future improvement but belongs in a separate patch with proper envtest coverage, not folded into this fix. Coupling this PR to #10621 would block a well-scoped, tested fix on work that is substantially larger in scope and not yet designed.

Unless there is a specific technical concern about the validation logic itself, I think this is ready to merge.

mnencia:
@sxd I think the framing is off. The real fix for malformed recovery target values is #10515, which is already merged. That PR makes the operator escape whatever the user provides, so it cannot corrupt `postgresql.conf` or get silently mangled by PostgreSQL's config lexer. Safety is in place regardless of any validation layer above it.

This PR is a UX improvement on top: reject obviously-bad input at admission, so users get a clear error instead of seeing PostgreSQL fail to start later. #10621 would be another UX layer (CEL on the API for clients that bypass the webhook). Neither is essential for safety; both just improve the UX at different surfaces.

So this is not a partial fix. The real fix already shipped. Let's land this UX improvement now and tackle #10621 separately when we plan CEL coverage and envtest setup for the package.

Issue #10600, opened 2026-05-03
[Bug]: Operator images not using OCI format for attestation


sxd:
### Is there an existing issue already for this bug?

- [x] I have searched for an existing issue, and could not find anything. I believe this is a new bug.

### I have read the troubleshooting guide

- [x] I have read the troubleshooting guide and I think this is a new bug.

### I am running a supported version of CloudNativePG

- [x] I have read the troubleshooting guide and I think this is a new bug.

### Version

1.29 (latest patch)

### What version of Kubernetes are you using?

1.35

### What is your Kubernetes environment?

Self-managed: kind (evaluation)

### How did you install the operator?

YAML manifest

### What happened?

The images are currently having the attestation (SBOM basically) without the right `artifactType` in the manifest making them undetectable by tools like [GUAC](https://github.com/guacsec/) 

Documentation https://docs.docker.com/build/exporters/#oci-media-types

### Cluster resource

```shell

```

### Relevant log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

mnencia:
Fixed by #10601.

BuildKit's default image exporter does not set `artifactType` or an OCI 1.1 Referrers `subject` on attestation manifests. The PR adds `oci-artifact=true`, so SBOM and provenance attestations now carry both and are discoverable through the Referrers API.

GUAC was the example mentioned here, but it has a separate limitation tracked upstream at guacsec/guac#1370: its OCI collector filters Referrers descriptors by an allowlist that does not include the BuildKit attestation manifest type, so it drops them regardless of how the manifest is shaped. That is for the GUAC project to fix; nothing on the cnpg side will help.

gbartolini:
Yes, in general it would be appreciated to have clearer issue descriptions, with ways to detect the current problem and how to determine the success of the PR. This fosters collaboration, facilitates buy-in, and reduces the effort in review and approval.

PR #10601, opened 2026-05-03
chore: enable OCI artifact type for attestation manifests

sxd:
BuildKit's default image exporter does not set `artifactType` or the OCI 1.1 Referrers `subject` backlink on attestation manifests. Setting `oci-artifact=true` adds both, so SBOM and provenance attestations become discoverable through the OCI Referrers API.

`oci-mediatypes=true` is redundant (BuildKit already defaults it to true when pushing to a registry, which is why the top-level index has been `application/vnd.oci.image.index.v1+json` for a while) but kept for clarity.

GUAC is mentioned in the linked issue, but it has a separate bug with multi-arch images that include attestations and will not pick up the change on its own. This PR makes the manifests Referrers-compliant; GUAC discovery needs a fix upstream.

Refs #10600

### DIFF
--- a/docker-bake.hcl
+++ b/docker-bake.hcl
@@ -96,7 +96,7 @@ target "default" {
   }
 
   output = [
-    "type=image,registry.insecure=${insecure}",
+    "type=image,registry.insecure=${insecure},oci-mediatypes=true,oci-artifact=true",
   ]
 
   attest = [


gbartolini:
Hi @sxd, can you please help me verify the change?

There must be something wrong I am doing, as I verified this both locally and against the image built from this branch on GHCR without success.

**Local build**, using:

```sh
./hack/setup-cluster.sh create load
```

This pushed the image to localhost:5000/cloudnative-pg-testing:latest-prime. I then inspected it with:

```
crane manifest localhost:5000/cloudnative-pg-testing:latest-prime | jq .
```

I also tried with:

```
docker buildx imagetools inspect localhost:5000/cloudnative-pg-testing:latest-prime --raw | jq .
```

**GHCR image built by this PR** (ghcr.io/cloudnative-pg/cloudnative-pg-testing:dev-10600@sha256:10a631571f03005e4231dba033de508d1894e503ce38c3a9988b32493c267be2), using:

```
crane manifest ghcr.io/cloudnative-pg/cloudnative-pg-testing:dev-10600@sha256:10a631571f03005e4231dba033de508d1894e503ce38c3a9988b32493c267be2 | jq .
```

In both cases, `artifactType` is absent from the manifest.

I also checked the current production image:

```
crane manifest ghcr.io/cloudnative-pg/cloudnative-pg:1.29.0 | jq .
```

It already uses OCI media types (`application/vnd.oci.image.index.v1+json`), even though it was built from main, where `docker-bake.hcl` does not have `oci-mediatypes=true`. This suggests BuildKit is already defaulting to OCI media types, so the problem described in the PR may not actually exist.

Could you help me understand what `oci-artifact=true` is supposed to do here, and what specific tool or scenario revealed the issue? I want to make sure I'm not missing something before drawing any conclusions.

sxd:
@gbartolini the documentation is here: 
* https://docs.docker.com/build/exporters/#oci-media-types
* https://docs.docker.com/build/exporters/image-registry/

Also in the issue and here I mentioned GUAC as the one that I used to find out https://github.com/guacsec/
You have to look into the layers because the top manifest doesn't mention anything since it's multiarch, so you have too look into these layers:
* ghcr.io/cloudnative-pg/cloudnative-pg:1.29.0@sha256:1b0fff73e12b7539bc356b8a6fa525a134b0000e75bdbb3f4bb2e1a329875706
* ghcr.io/cloudnative-pg/cloudnative-pg-testing:dev-10600@sha256:f6436386857cf1785538def37e7bfb5b8b4cf9b4a0545c5ef6af3732ca0fd507

Hopefully that helps

Also it's a bit tricky because due to a bug in GUAC it will not even show up after the change because GUAC doesn't support multiarch images with buildkit that include attestation.

gbartolini:
Got it. Found it:

```sh
docker buildx imagetools inspect --raw ghcr.io/cloudnative-pg/cloudnative-pg-testing:dev-10600@sha256:277a0a3d3a60f052a361b174a0e84c30b9de0d09e6657cf22d245209b4dc5dd1 | jq .
{
  "schemaVersion": 2,
  "mediaType": "application/vnd.oci.image.manifest.v1+json",
  "artifactType": "application/vnd.docker.attestation.manifest.v1+json",
  "config": {
    "mediaType": "application/vnd.oci.empty.v1+json",
    "digest": "sha256:44136fa355b3678a1146ad16f7e8649e94fb4fc21fe77e8310c060f61caaff8a",
    "size": 2,
    "data": "e30="
  },
  "layers": [
    {
      "mediaType": "application/vnd.in-toto+json",
      "digest": "sha256:88086847517d317f761c2ce19a58cfc571a8e25ca8bbbd94edb72e16141b8771",
      "size": 927156,
      "annotations": {
        "in-toto.io/predicate-type": "https://spdx.dev/Document"
      }
    },
    {
      "mediaType": "application/vnd.in-toto+json",
      "digest": "sha256:1be4fc2240c9f2ba7223401ce749ac083950df72a44fc742fa7e244022439f29",
      "size": 37042,
      "annotations": {
        "in-toto.io/predicate-type": "https://slsa.dev/provenance/v1"
      }
    }
  ],
  "subject": {
    "mediaType": "application/vnd.oci.image.manifest.v1+json",
    "digest": "sha256:37751f556f8d7a1ba222fe8f78bcf811a76514ec2ca21ce3ab8b31efe08c41ba",
    "size": 4414
  }
}
```

gbartolini (approved):

mnencia (approved):

PR #10620, opened 2026-05-05
feat(database): skip finalizer for resources with databaseReclaimPolicy retain

### LINKED ISSUE #10535: [Bug]: Finalizer blocks namespace deletion when operator and operand in same namespace

### Contact Details

Max.a.lengdell@ericsson.com

### Version

1.29 (latest patch)

1.35

Helm

### What happened?

CloudNativePG Operator adds finalizers to Plugin Service, Database, Subscription and Publication CR. If these are deployed in the same namespace as the operator, when a `kubectl delete namespace <namespace>` is invoked, resources could be left stuck with namespace deletion pending. The finalizer is blocking the deletion, but since the operator is also deleted it may not have time to remove the finalizers before termination.

Since the plugin must be deployed in the same namespace as the operator, this race condition is somewhat unavoidable.
The CRD based resource finalizers can be discussed as a separate PR whether that's a bug or not

### Installing operator
`
helm install cnpg https://github.com/cloudnative-pg/charts/releases/download/cloudnative-pg-v0.28.0/cloudnative-pg-0.28.0.tgz -n <TEST_NAMESPACE> --set config.clusterWide=false`

### Create cluster
```
cat <<EOF | kubectl apply -n <TEST_NAMESPACE> -f -
# Example of PostgreSQL cluster
kind: Cluster
metadata:
  name: cluster-example
spec:
  instances: 3
  storage:
    size: 1Gi
EOF

```
### Create Database CR
```
cat <<EOF | kubectl apply  -n <TEST_NAMESPACE> -f -
# Example of PostgreSQL cluster
kind: Database
metadata:
  name: app-db
spec:
  cluster:
    name: cluster-example
  name: app
  owner: app
  schemas:
    - name: data
      owner: app
EOF
```

If you want to inspect the finalizer 
`kubectl get database -n finalizer -o jsonpath='{range .items[*]}{.metadata.name}: {.metadata.finalizers}{"\n"}{end}'`

### Delete namespace
`kubectl delete namespace <TEST_NAMESPACE>
`
namespace deletion is now stuck. 

`kubectl describe namespace <TEST_NAMESPACE>`

```
Name:         <TEST_NAMESPACE>
Labels:       kubernetes.io/metadata.name=finalizer
Annotations:  <none>
Status:       Terminating
Conditions:
  Type                                         Status  LastTransitionTime               Reason                Message
  ----                                         ------  ------------------               ------                -------
  NamespaceDeletionDiscoveryFailure            False   Thu, 23 Apr 2026 14:08:14 +0200  ResourcesDiscovered   All resources successfully discovered
  NamespaceDeletionGroupVersionParsingFailure  False   Thu, 23 Apr 2026 14:08:14 +0200  ParsedGroupVersions   All legacy kube types successfully parsed
  NamespaceDeletionContentFailure              False   Thu, 23 Apr 2026 14:08:14 +0200  ContentDeleted        All content successfully deleted, may be waiting on finalization
  NamespaceContentRemaining                    True    Thu, 23 Apr 2026 14:08:22 +0200  SomeResourcesRemain   Some resources are remaining: databases.postgresql.cnpg.io has 1 resource instances
  NamespaceFinalizersRemaining                 True    Thu, 23 Apr 2026 14:08:22 +0200  SomeFinalizersRemain  Some content in the namespace has finalizers remaining: cnpg.io/deleteDatabase in 1 resource instances

```

### Cluster resource

```shell
kind: Cluster
metadata:
  name: cluster-example
spec:
  instances: 3
  storage:
    size: 1Gi
```

### Relevant log output

```shell
{"level":"info","ts":"2026-04-23T12:08:14.7511441Z","msg":"Stopping and waiting for non leader election runnables"}
{"level":"info","ts":"2026-04-23T12:08:14.752751053Z","msg":"Stopping and waiting for leader election runnables"}
{"level":"info","ts":"2026-04-23T12:08:14.752795589Z","msg":"Shutdown signal received, waiting for all workers to finish","controller":"cluster","controllerGroup":"postgresql.cnpg.io","controllerKind":"Cluster"}
{"level":"info","ts":"2026-04-23T12:08:14.752805809Z","msg":"Shutdown signal received, waiting for all workers to finish","controller":"pooler","controllerGroup":"postgresql.cnpg.io","controllerKind":"Pooler"}
{"level":"info","ts":"2026-04-23T12:08:14.752812511Z","msg":"Shutdown signal received, waiting for all workers to finish","controller":"plugin","controllerGroup":"","controllerKind":"Service"}
{"level":"info","ts":"2026-04-23T12:08:14.752817434Z","msg":"Shutdown signal received, waiting for all workers to finish","controller":"backup","controllerGroup":"postgresql.cnpg.io","controllerKind":"Backup"}
{"level":"info","ts":"2026-04-23T12:08:14.752822343Z","msg":"Shutdown signal received, waiting for all workers to finish","controller":"scheduled-backup","controllerGroup":"postgresql.cnpg.io","controllerKind":"ScheduledBackup"}
{"level":"info","ts":"2026-04-23T12:08:14.752851356Z","msg":"All workers finished","controller":"backup","controllerGroup":"postgresql.cnpg.io","controllerKind":"Backup"}
{"level":"info","ts":"2026-04-23T12:08:14.753193796Z","msg":"All workers finished","controller":"pooler","controllerGroup":"postgresql.cnpg.io","controllerKind":"Pooler"}
{"level":"info","ts":"2026-04-23T12:08:14.753316669Z","msg":"All workers finished","controller":"scheduled-backup","controllerGroup":"postgresql.cnpg.io","controllerKind":"ScheduledBackup"}
{"level":"info","ts":"2026-04-23T12:08:14.755176034Z","msg":"All workers finished","controller":"plugin","controllerGroup":"","controllerKind":"Service"}
{"level":"info","ts":"2026-04-23T12:08:14.751137579Z","msg":"Stopping and waiting for warmup runnables"}
{"level":"error","ts":"2026-04-23T12:08:14.779362398Z","msg":"Reconciler error","controller":"cluster","controllerGroup":"postgresql.cnpg.io","controllerKind":"Cluster","Cluster":{"name":"cluster-example","namespace":"finalizer"},"namespace":"finalizer","name":"cluster-example","reconcileID":"6c1c1017-a88a-4f4a-ab1c-7130956f8d86","error":"cannot reconcile required plugins: context canceled","stacktrace":"sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).reconcileHandler\n\tpkg/mod/sigs.k8s.io/controller-runtime@v0.23.3/pkg/internal/controller/controller.go:495\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).processNextWorkItem\n\tpkg/mod/sigs.k8s.io/controller-runtime@v0.23.3/pkg/internal/controller/controller.go:438\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).Start.func1.1\n\tpkg/mod/sigs.k8s.io/controller-runtime@v0.23.3/pkg/internal/controller/controller.go:313"}
{"level":"info","ts":"2026-04-23T12:08:14.800383441Z","msg":"All workers finished","controller":"cluster","controllerGroup":"postgresql.cnpg.io","controllerKind":"Cluster"}
{"level":"info","ts":"2026-04-23T12:08:14.80044489Z","msg":"Stopping and waiting for caches"}
{"level":"info","ts":"2026-04-23T12:08:14.809943355Z","msg":"Stopping and waiting for webhooks"}
{"level":"info","ts":"2026-04-23T12:08:14.810851666Z","logger":"controller-runtime.webhook","msg":"Shutting down webhook server with timeout of 1 minute"}
{"level":"info","ts":"2026-04-23T12:08:14.811019537Z","msg":"Stopping and waiting for HTTP servers"}
{"level":"info","ts":"2026-04-23T12:08:14.81104744Z","logger":"controller-runtime.metrics","msg":"Shutting down metrics server with timeout of 1 minute"}
{"level":"info","ts":"2026-04-23T12:08:14.816212465Z","msg":"Wait completed, proceeding to shutdown the manager"}
{"level":"error","ts":"2026-04-23T12:08:14.856473876Z","msg":"error received after stop sequence was engaged","error":"leader election lost","stacktrace":"sigs.k8s.io/controller-runtime/pkg/manager.(*controllerManager).engageStopProcedure.func1\n\tpkg/mod/sigs.k8s.io/controller-runtime@v0.23.3/pkg/manager/internal.go:533"}
```

### Code of Conduct

GellertBabel:
Fixes #10535

Database, Publication and Subscription resources with a retain reclaim policy no longer receive a finalizer. This prevents the resource from getting stuck with a deletionTimestamp when the operator or operand is unavailable, which would block namespace deletion.

### DIFF
--- a/internal/management/controller/database_controller.go
+++ b/internal/management/controller/database_controller.go
@@ -213,6 +213,7 @@ func NewDatabaseReconciler(
 		mgr.GetClient(),
 		utils.DatabaseFinalizerName,
 		dr.evaluateDropDatabase,
+		func(db *apiv1.Database) bool { return db.Spec.ReclaimPolicy == apiv1.DatabaseReclaimDelete },
 	)
 
 	return dr
--- a/internal/management/controller/database_controller_test.go
+++ b/internal/management/controller/database_controller_test.go
@@ -111,6 +111,7 @@ var _ = Describe("Managed Database status", func() {
 			fakeClient,
 			utils.DatabaseFinalizerName,
 			r.evaluateDropDatabase,
+			func(db *apiv1.Database) bool { return db.Spec.ReclaimPolicy == apiv1.DatabaseReclaimDelete },
 		)
 	})
 
@@ -205,7 +206,7 @@ var _ = Describe("Managed Database status", func() {
 	})
 
 	When("reclaim policy is retain", func() {
-		It("on deletion it removes finalizers and does NOT drop the DB", func(ctx SpecContext) {
+		It("on deletion it does NOT add a finalizer and does NOT drop the DB", func(ctx SpecContext) {
 			database.Spec.ReclaimPolicy = apiv1.DatabaseReclaimRetain
 			Expect(fakeClient.Update(ctx, database)).To(Succeed())
 
@@ -226,19 +227,12 @@ var _ = Describe("Managed Database status", func() {
 			err := reconcileDatabase(ctx, fakeClient, r, database)
 			Expect(err).ToNot(HaveOccurred())
 
-			// Plain successful reconciliation, finalizers have been created
-			Expect(database.GetFinalizers()).NotTo(BeEmpty())
+			// No finalizer added — retain policy means the DB is not owned by the operator lifecycle
+			Expect(database.GetFinalizers()).To(BeEmpty())
 			Expect(database.Status.Applied).Should(HaveValue(BeTrue()))
 			Expect(database.Status.Message).Should(BeEmpty())
 
-			// The next 2 lines are a hacky bit to make sure the next reconciler
-			// call doesn't skip on account of Generation == ObservedGeneration.
-			// See fake.Client known issues with `Generation`
-			// https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/client/fake@v0.19.0#NewClientBuilder
-			database.SetGeneration(database.GetGeneration() + 1)
-			Expect(fakeClient.Update(ctx, database)).To(Succeed())
-
-			// We now look at the behavior when we delete the Database object
+			// Deleting the k8s object succeeds immediately with no finalizer blocking it
 			Expect(fakeClient.Delete(ctx, database)).To(Succeed())
 
 			err = reconcileDatabase(ctx, fakeClient, r, database)
--- a/internal/management/controller/finalizers.go
+++ b/internal/management/controller/finalizers.go
@@ -27,26 +27,31 @@ import (
 )
 
 type finalizerReconciler[T client.Object] struct {
-	cli           client.Client
-	finalizerName string
-	onRemoveFunc  func(ctx context.Context, resource T) error
+	cli                  client.Client
+	finalizerName        string
+	onRemoveFunc         func(ctx context.Context, resource T) error
+	shouldAddFinalizer   func(resource T) bool
 }
 
 func newFinalizerReconciler[T client.Object](
 	cli client.Client,
 	finalizerName string,
 	onRemoveFunc func(ctx context.Context, resource T) error,
+	shouldAddFinalizer func(resource T) bool,
 ) *finalizerReconciler[T] {
 	return &finalizerReconciler[T]{
-		cli:           cli,
-		finalizerName: finalizerName,
-		onRemoveFunc:  onRemoveFunc,
+		cli:                cli,
+		finalizerName:      finalizerName,
+		onRemoveFunc:       onRemoveFunc,
+		shouldAddFinalizer: shouldAddFinalizer,
 	}
 }
 
 func (f finalizerReconciler[T]) reconcile(ctx context.Context, resource T) error {
-	// add finalizer in non-deleted publications if not present
 	if resource.GetDeletionTimestamp().IsZero() {
+		if f.shouldAddFinalizer != nil && !f.shouldAddFinalizer(resource) {
+			return nil
+		}
 		if !controllerutil.AddFinalizer(resource, f.finalizerName) {
 			return nil
 		}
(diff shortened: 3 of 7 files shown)


armru:
Thanks for working on this. Before getting into the implementation details, let me check my understanding of the scenario, because I have a concern with the approach — correct me if I'm getting something wrong.

If I understand it correctly, the hang described in #10535 only occurs when the operator runs in the same namespace as the cluster. In the standard topology, with the operator in its own namespace, deleting the operand namespace should not get stuck today: as soon as the Cluster is deleted, the operator itself removes the finalizers from every Database, Publication and Subscription belonging to it (notifyDeletionToOwnedResources, internal/controller/finalizers_delete.go), even when the instance pods are already gone. If that's right, the fix would only benefit the shared-namespace topology, while the behavior change applies to every deployment.

My worry is what the other deployments would lose: the deletion-time evaluation of the reclaim policy. Today the finalizer is present from the first reconciliation, so switching a resource from retain to delete and deleting it right afterwards still guarantees the drop — the instance manager re-reads the policy while processing the deletion. With this change, a resource created under retain (which is the default) carries no finalizer, so a flip to delete followed by a quick deletion could skip the drop entirely, and the database would silently survive in PostgreSQL despite an explicit delete request.

If my reading is correct, I'd rather scope the mitigation to the shared-namespace topology, or make it an explicit opt-in, than change the default semantics globally. But I may be missing a case where the current finalizer hurts outside that topology — happy to be corrected.

I'd also like the other maintainers to weigh in on this one before we move forward.

GellertBabel:
A user should wait for observedGeneration to equal generation before deleting. Even today I wouldn't assume it's safe to delete the CR right away, a reclaim-policy change is a spec change, and it only takes effect once reconciled.

Yes, the problem is more important for the shared-namespace use case, and **I'm fine limiting the code to that**. That said, it can still happen in a separate-namespace setup. When multiple namespaces are deleted at the same time, or the operator simply isn't running at that moment, the finalizers never get removed and the namespace hangs.

Not adding the finalizer at all when not needed. Makes for a safer but less feature-rich operator, which is the tradeoff the users hitting this actually want.

PR #10627, opened 2026-05-06
feat: introduce primary lease for safe primary election

### LINKED ISSUE #10807: [Feature]: Add a primary Lease to improve failover correctness with asynchronous replication

When users need **data durability**, they should use **synchronous replication**.

When they choose **asynchronous replication**, CloudNativePG assumes their primary goal is **self-healing** through **high availability** and that they accept some data loss on failover: the operator favours keeping the service available and promotes a replica quickly, accepting that the promoted instance may not be fully caught up with the old primary.

Even within this availability-first model, correctness can still be improved. Today, promotion is driven by the `targetPrimary/currentPrimary` fields the operator writes to the Cluster status, and there is no independent mechanism that prevents a newly promoted instance from starting to act as primary while a previous primary is still running, even if temporarily — for example, during a network partition, a slow shutdown, or a failed in-place online upgrade that leaves PostgreSQL up.

For asynchronous clusters, we want to improve resilience and correctness *without* changing the availability-first trade-off by adding an explicit lock that an instance must hold before acting as primary, narrowing the window in which two instances could both behave as primary.

Introduce a per-cluster **primary Lease** using the standard `coordination.k8s.io/v1` API as a mutex that gates primary behaviour.

Further analysis to be done during exploration, providing details and defining the scope of the intervention.

- Status-only coordination (current behaviour) — no separate lock acquired before promotion, leaving the transitional window unguarded.
- A custom CRD lock — reinvents what Lease + upstream leader election already provide.

### Additional context

N/A

### Backport?

No

Yes

### Code of Conduct

leonardoce:
Add a Kubernetes `coordination.k8s.io/v1` `Lease` object (named after the cluster) as a mutex to synchronize primary promotion. The instance manager must hold the lease before acting as PostgreSQL primary, and releases it explicitly on a clean shutdown so replicas can promote without waiting for the full TTL.

The lease timings are tunable via the new `.spec.primaryLease` stanza. Defaults match controller-runtime's stock values, except `retryPeriodSeconds` is 5s instead of 2s to reduce apiserver write load when a single operator manages many clusters.

The Lease is a promotion gate, not a fence: it serializes who is allowed to promote, while the primary isolation check remains responsible for fencing an isolated primary.

Closes #10807


leonardoce:
I'm hardcoding the lease timing parameters (leaseDuration, renewDeadline, retryPeriod) for now rather than exposing them as API fields, for two reasons:

- The values (15s / 10s / 2s) match kube-controller-manager's own leader election defaults, which are battle-tested.
- In the graceful shutdown path, Release() is called after PostgreSQL has already stopped, so replicas can acquire the lease immediately without waiting for the full TTL. The 15s floor only matters in crash scenarios.

That said, I'm open to surfacing these, leaseDuration.
Does this seem reasonable, or is there a use case I'm missing?

armru (approved):

internal/cmd/manager/instance/run/lease/runnable.go
```diff
@@ -0,0 +1,248 @@
...
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/leaderelection"
+	"k8s.io/client-go/tools/leaderelection/resourcelock"
+
+	"github.com/cloudnative-pg/cloudnative-pg/pkg/management/postgres"
+)
+
+const (
+	// leaseDuration is how long a lease is valid before it expires.
+	leaseDuration = 15 * time.Second
```
gbartolini: Can we add these parameters in a stanza `.spec.primaryLease` or similar, with these defaults, to enable customisations?

mnencia (approved):

gbartolini (approved):
Nice addition to the docs.

PR #10642, opened 2026-05-07
chore: enable ImageVolume featureGate on K3D 1.33/1.34

### LINKED ISSUE #10652: [chore]: enable ImageVolume featureGate on K3D 1.33/1.34 Clusters

While testing https://github.com/cloudnative-pg/cloudnative-pg/pull/10544 I've noticed that ImageVolume Extension E2Es are failing on K3D 1.33/1.34 Clusters because the imageVolume FeatureGate is not enabled.
Check out: https://github.com/cloudnative-pg/cloudnative-pg/actions/runs/25496211949

Also, we should avoid manually enabling the featureGate on K8S >= 1.35.0, since it's enabled by default.

Enable the ImageVolume featureGate while bootstrapping a `K3D` Cluster on 1.33/1.34 via `hack/setup-cluster.sh`, similarly to how it's already being done for `Kind`.

None

### Additional context

_No response_

### Backport?

Yes

Yes

### Code of Conduct

NiccoloFei:
* Enable the ImageVolume featureGate on K3D Clusters (only from 1.33.5 or newer, because older minor versions of 1.33 do not carry a `containerd` version that's new enough `>= 2.1.0 `)
* Added an upper-bound check so the ImageVolume featureGate block is skipped on 1.35.0+ (for both kind and k3d). Since K8S 1.35, ImageVolume is enabled by default.
* Simplified the version check into a reusable function
* Fixed an issue with grep swallowing an error message when attempting a K3D deploy with a version of K3S that doesn't exist (`set -eEuo pipefail` would kill the script before the error msg is fired because grep would exit with status 1)

Closes #10652

### DIFF
--- a/hack/testing-tools/k8s-engines/k3d/setup.sh
+++ b/hack/testing-tools/k8s-engines/k3d/setup.sh
@@ -51,7 +51,7 @@ function create_cluster_k3d() {
   local cluster_name=$2
 
   local latest_k3s_tag
-  latest_k3s_tag=$(k3d version list k3s | grep -- "^${k8s_version//./\\.}"'\+-k3s[0-9]$' | tail -n 1)
+  latest_k3s_tag=$(k3d version list k3s | { grep -- "^${k8s_version//./\\.}"'\+-k3s[0-9]$' || true; } | tail -n 1)
   if [[ -z "${latest_k3s_tag}" ]]; then
     echo "ERROR: No k3s image found for version ${k8s_version}" >&2
     exit 1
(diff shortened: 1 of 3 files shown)


NiccoloFei:
E2E run: https://github.com/cloudnative-pg/cloudnative-pg/actions/runs/25509666712

danishedb (approved):

hack/testing-tools/k8s-engines/k3d/setup.sh
```diff
@@ -51,7 +51,7 @@ function create_cluster_k3d() {
   local cluster_name=$2
 
   local latest_k3s_tag
-  latest_k3s_tag=$(k3d version list k3s | grep -- "^${k8s_version//./\\.}"'\+-k3s[0-9]$' | tail -n 1)
+  latest_k3s_tag=$(k3d version list k3s | { grep -- "^${k8s_version//./\\.}"'\+-k3s[0-9]$' || true; } | tail -n 1)
```
sxd:
```suggestion
  latest_k3s_tag=$(k3d version list k3s -l1 -s desc --include "^${k8s_version}.*")
```

litaocdl (approved):

mnencia (approved):

PR #10660, opened 2026-05-10
feat(status): add --watch / -w flag to kubectl cnpg status

### LINKED ISSUE #10659: [Feature]: add --watch / -w flag to kubectl cnpg status

When performing operations that change cluster state—rolling updates, switchovers, 
failovers, backup runs—operators must repeatedly execute `kubectl cnpg status <cluster>` 
to monitor progress.

Many users work around this with the external `watch` command:
`watch -n2 kubectl cnpg status cluster-example`

However:
- `watch` is not available on macOS without a Homebrew install
- It does not respect the plugin's `--output json/yaml` flag
- It cannot cleanly handle Ctrl-C within the plugin context
- The output header is generic and not CNPG-aware

Add a `--watch` / `-w` flag and `--watch-interval` / `-i` flag (default: 2s):

```bash
kubectl cnpg status cluster-example --watch
kubectl cnpg status cluster-example -w --watch-interval 5s
```

Behaviour:
1. Print status output, then sleep for the interval
2. Clear the terminal (ANSI escape) and reprint with a header line:
   `Every 2.0s: kubectl cnpg status cluster-example`
3. Exit cleanly on Ctrl-C (SIGINT via signal.NotifyContext)
4. Transient API errors print as warnings (not fatal exits), improving 
   resilience during failovers when pods are briefly unreachable

The `--output json/yaml` and `--verbose` flags must be fully respected in watch mode.

- External `watch` command: not cross-platform, ignores --output flag
- `kubectl get cluster -w`: shows raw resource diffs, not the rich formatted 
  status the cnpg plugin provides

### Additional context

This follows the same pattern as the recently merged `--timeout` flag (#9201) 
on the same command — a small additive flag with no breaking changes.

Implementation would touch:
- `internal/cmd/plugin/status/status.go` (~80 lines)
- `internal/cmd/plugin/status/status_test.go` (unit tests)  
- `docs/src/kubectl-plugin.md` (documentation)

I am happy to submit a PR for this.

### Backport?

No

Yes

### Code of Conduct

Ram04102007:
## What does this PR do?

Adds `--watch` / `-w` and `--watch-interval` flags to `kubectl cnpg status` so you can monitor a cluster in real time without reaching for an external tool.

## Why?

During failovers, rolling updates, or backup runs I kept doing:
  watch -n2 kubectl cnpg status cluster-example
Two problems with that: `watch` isn't on macOS without Homebrew, and it completely ignores `--output json/yaml`. Having it built into the plugin just makes sense.

## How?

- When `--watch` is set, the command loops: clear screen → print a header (`Every 2s: kubectl cnpg status <cluster>`) → run status → sleep → repeat
- Ctrl-C exits cleanly via `signal.NotifyContext`
- `--output json/yaml` skips the loop and runs once — piping a clear-screen JSON stream to `jq` would be a mess
- Transient errors during the loop are printed as warnings rather than killing the process (useful when pods are briefly unreachable mid-failover)

## Checklist

- [x] Tests pass
- [x] Commits signed (`git commit -s`)
- [x] Conventional commit prefix (`feat(status):`)

Closes #10659


Ram04102007:
Ready for review — happy to make any changes needed 
from the maintainers.

PR #10664, opened 2026-05-11
chore(test): avoid the MinIO pods deployment at the beginning of tests

### LINKED ISSUE #6572: [Bug]: Remove the fixed usage of MinIO in the E2E tests

### Contact Details

_No response_

### Version

1.25 (latest patch)

1.32

YAML manifest

### What happened?

Currently, every time we trigger the backups, we create a new deployment of MinIO even sometimes when this is not needed using resources and consuming up to 5 minutes for that, we should create it on demand.

The proposal is to create in the package minio package the proper functions to request and create all the needed resources in the requested namespace. Also, the package should query if the central minio deployment is already running, if that's the case just return with the already running deployment, if that's not the case, should wait for the deployment to be ready or create it and then wait.

All this should include a new set of functions to work with the ObjectStore and check for the content, properties, etc, stuff we already do, also it should be able to work with any object store, for this, we should create a new package in the e2e tests, this will help to remove the current minio client pod too, and improve the checks for other object stores. 

### Cluster resource

```shell

```

### Relevant log output

```shell

```

### Code of Conduct

sxd:
The deployment of MinIO and `mc` container was using most of the time at the beginning of the e2e tests making really slow to trigger a non related tests. This commit fix that by making MinIO a resource that is asked by the tests that require it and reusing the already existing one if it was requested before by another tests. It will not save much time when running backup tests, but it will save a lot of time just for other non related tests.

This is a first step to reduce the MinIO need into an encapsulated form for a future easy removal and/or replacement, from now on it will not be a strong need for the starting and a couple of functions are MinIO specific.

Another introduced benefits it's a set of function that query the store instead of trying to deal with all the ways of working of each Object Store provider. These functions must be cleaned and improved in the future for a more generic work, but for now it's enough to work with.

Closes #6572


mnencia:
Abandoned. Minio has been replaced by RustFS in #10865

PR #10666, opened 2026-05-11
fix: do not require restart for replica cluster promotion

leonardoce:
When promoting a replica cluster to a primary cluster, PostgreSQL previously needed a restart because archive_mode changed from "always" to "on", which is a fixed parameter.

This change always sets archive_mode to "always" and moves the decision of whether to actually ship WAL into the archiver:

- Primary clusters: the archiver archives from the current primary, skips silently on HA replicas (standby.signal is present), and keeps WAL in the ready queue on a former primary being demoted (standby.signal absent, CurrentPrimary no longer matches).
- Replica clusters: behaviour is unchanged — only the designated primary (CurrentPrimary or TargetPrimary) archives WAL.

Promoting a replica cluster to a primary cluster does not requires a restart.

### DIFF
--- a/docs/src/samples/dc/cluster-dc-a.yaml
+++ b/docs/src/samples/dc/cluster-dc-a.yaml
@@ -12,8 +12,11 @@ spec:
 
   backup:
     barmanObjectStore:
+      endpointCA:
+        name: minio-server-tls
+        key: tls.crt
       destinationPath: s3://backups/
-      endpointURL: http://minio:9000
+      endpointURL: https://minio:9000
       s3Credentials:
         accessKeyId:
           name: minio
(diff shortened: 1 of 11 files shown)


leonardoce:
Why the E2E tests are failing in this PR

With this change, archive_mode=always is set on every PostgreSQL instance in a primary cluster, not just the primary. This means HA replicas will now also run archive_command, exiting 0 without actually archiving (the primary is still the one that archives), consistent with the operator's existing archiving semantics.

However, this exposes a pre-existing limitation in our E2E test environment, unrelated to correctness of the fix itself:

Barman Cloud always invokes pg_backup_stop with wait_for_archive=true. When archive_mode=always is set, PostgreSQL requires a WAL switch to occur (and be archived) before pg_backup_stop can return, on any instance where archive_mode=always, not just the primary. In our E2E environment there's no write workload on the cluster during backup tests, so no natural WAL switch happens, and pg_backup_stop can hang waiting for archiving to catch up.

Traditional (non-cloud) Barman avoids this because it holds two connections - one to the backup target and one to the primary - and explicitly issues a WAL switch (pg_switch_wal) on the primary to force pg_backup_stop to complete. Barman Cloud has no such mechanism: it only connects to the instance being backed up and has no way to trigger a switch on the primary.

PR #10668, opened 2026-05-11
chore(lint): upgrade golangci-lint to v2.12.2 and fix new warnings

leonardoce:
Bump golangci-lint to v2.12.2 in CI. The new version bundles an updated goconst that also scans composite literals (jgautheron/goconst#46), surfacing repeated strings that were previously invisible to the linter.

Address the new warnings by:
- replacing "postgresql.cnpg.io" literals with apiv1.SchemeGroupVersion.Group across webhooks, specs, and RBAC helpers; this also fixes a latent bug in scheduledbackup_webhook.go where the group was incorrectly "scheduledBackup.cnpg.io"
- introducing pgdataVolumeName and walLevelMinimal constants for strings repeated across volume and WAL-level validation logic
- replacing reflect.Ptr with reflect.Pointer in configparser (govet inline)
- renaming gomodguard to gomodguard_v2 per the new linter registry name

Configure goconst with min-occurrences=10 and ignore RBAC verb strings ("get", "list", "watch", etc.) which are stable Kubernetes API values not defined as constants in k8s.io/api/rbac/v1 — extracting them would add noise without reducing risk. Extend the goconst exclusion to cover tests/ in addition to *_test.go files.


.golangci.yml
```diff
@@ -43,6 +43,18 @@ linters:
     - wastedassign
     - whitespace
   settings:
+    goconst:
+      min-occurrences: 10
+      # RBAC verb strings are stable Kubernetes API values not defined as constants
+      # in k8s.io/api/rbac/v1 — extracting them adds noise without reducing risk.
```
sxd, line 49: add noise ? or you mean reduce noise ?

sxd (approved):

sxd (approved):

gbartolini (approved):

armru (approved):

PR #10687, opened 2026-05-14
ci: align E2E cluster node sizing across cloud providers

### LINKED ISSUE #8331: [Bug]: major upgrade e2e can fail on Azure

### Contact Details

_No response_

### Version

1.26 (latest patch)

1.33

YAML manifest

### What happened?

When running the e2e on Azure, Azure choose a nodepool automatically.

Sometimes, we can get an arm64 one, as we can see from these labels:

```
`"beta.kubernetes.io/arch":"arm64","beta.kubernetes.io/instance-type":"Standard_D4plds_v5"`
```

In these cases, the PostGIS major upgrade test will fail, since we do not build the image for arm64/

Additional info:
https://learn.microsoft.com/en-us/azure/aks/quotas-skus-regions#cluster-configuration-presets-in-the-azure-portal

### Cluster resource

```shell

```

### Relevant log output

```shell

```

### Code of Conduct

mnencia:
GKE nightly E2E fails intermittently when a replica pod cannot be scheduled: the zonal PD is pinned to a node that the kubelet has tainted under disk pressure caused by the container image cache filling the 20 GiB boot disk. AKS occasionally picks Standard_D4plds_v5 (arm64) when --node-vm-size is unset, mixing architectures across the matrix and reducing test consistency.

Pin all three managed providers to x86_64 4-vCPU SKUs and raise GKE's boot disk to match EKS: e2-standard-4 with 80 GiB on GKE, Standard_D4s_v5 on AKS, m6a.xlarge on EKS.

Closes #8331

### DIFF
--- a/.github/workflows/continuous-delivery.yml
+++ b/.github/workflows/continuous-delivery.yml
@@ -1209,7 +1209,7 @@ jobs:
             az aks create --resource-group ${{ secrets.AZURE_RESOURCEGROUP }} \
               --name ${AZURE_AKS} \
               --tier standard \
-              --node-count 3 -k v${K8S_VERSION} --generate-ssh-keys --enable-addons monitoring \
+              --node-count 3 --node-vm-size Standard_D4s_v5 -k v${K8S_VERSION} --generate-ssh-keys --enable-addons monitoring \
               --workspace-resource-id ${{ secrets.AZURE_WORKSPACE_RESOURCE_ID }} \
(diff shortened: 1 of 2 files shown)


mnencia:
Ci on EDB fork https://github.com/EnterpriseDB/cloudnative-pg/actions/runs/25868748038

fcanovai (approved):

NiccoloFei (approved):

PR #10724, opened 2026-05-19
feat: SCRAM-SHA-256 encode role passwords operator-side

leonardoce:
Before emitting `CREATE`/`ALTER ROLE ... PASSWORD '...'`, the operator now SCRAM-SHA-256 encodes any cleartext password operator-side (client-side from PostgreSQL's point of view). The literal PostgreSQL parses (and that extensions such as `pg_stat_statements` or `pgaudit` may capture) is the SCRAM verifier ultimately stored in `pg_authid.rolpassword`, never the cleartext secret. This matches what `psql \password` and libpq's `PQencryptPasswordConn` do, and applies to every basic-auth Secret the operator consumes: the `postgres` superuser, the application user, and managed-role password secrets. The existing statement-log suppression for CREATE/ALTER ROLE remains in effect.

Passwords already supplied to PostgreSQL in its MD5 or SCRAM-SHA-256 shadow form are forwarded unchanged, so users who store a pre-hashed value in the Secret retain full control over the stored hash.

Opt-out: set the annotation `cnpg.io/passwordPassthrough: "enabled"` on the basic-auth Secret to have the operator forward the password literal verbatim and let PostgreSQL encode it according to its own `password_encryption` GUC. The opt-out is per-Secret, so a single cluster can mix passthrough and operator-encoded secrets freely, restoring the pre-change behaviour for clusters that explicitly run with `password_encryption = md5` or otherwise need PostgreSQL to choose the hash format.


leonardoce:
Important: this PR needs https://github.com/cloudnative-pg/machinery/pull/260 to be merged and released.

armru (approved):

internal/management/controller/roles/postgres.go
```diff
@@ -375,7 +386,19 @@ func roleHasPassword(role DatabaseRole) bool {
 	return !role.ignorePassword && role.password.Valid
 }
 
-func appendPasswordOption(role DatabaseRole, query *strings.Builder) {
+// resolveRolePassword returns the value that should be embedded in the
+// PASSWORD clause of CREATE/ALTER ROLE for the given role. Cleartext
+// passwords are SCRAM-SHA-256 encoded so the SQL literal is never
+// cleartext. Returns an empty string when the role has no password to
+// set (ignorePassword, or PASSWORD NULL).
+func resolveRolePassword(role DatabaseRole) (string, error) {
```
jsilvela: firstly, the name of the function is not great IMO. "resolve" is unspecific like "compute", but since it's non-standard, it makes one take notice. And it's just a wrapper.
Secondly, this is a shallow function (c.f. John Ousterhout's book and talk on software design).
As I will argue below, I think it's unnecessary.

internal/management/controller/roles/postgres.go
```diff
@@ -375,7 +386,19 @@ func roleHasPassword(role DatabaseRole) bool {
...
+	if !roleHasPassword(role) || !role.password.Valid {
+		return "", nil
+	}
+	return postgresutils.EnsureEncryptedPassword(role.password.String)
+}
+
+func appendPasswordOption(role DatabaseRole, encryptedPassword string, query *strings.Builder) {
```
jsilvela: Why pass the extra argument?
Instead, the encryption should be handled within this function by directly calling `EnsureEncryptedPassword`

internal/management/controller/roles/postgres.go
```diff
@@ -117,6 +117,12 @@ func Update(ctx context.Context, db *sql.DB, role DatabaseRole) error {
 	wrapErr := func(err error) error {
 		return fmt.Errorf("while updating role %s with role reconciler: %w", role.Name, err)
 	}
+
+	encryptedPassword, err := resolveRolePassword(role)
+	if err != nil {
+		return wrapErr(err)
+	}
```
jsilvela: This call and the other are in functions that end up calling `appendPasswordOption`. I think the encryption should be handled there.

jsilvela (commented):
The "resolve" wording threw me, and thinking about it, that function is unnecessary.

internal/management/controller/roles/postgres.go
armru: done, added you as coauthor given that you raised the point!

jsilvela: Thank you Armando! That's very kind.

docs/src/declarative_role_management.md
```diff
@@ -179,23 +179,54 @@ stringData:
   password: SCRAM-SHA-256$<iteration count>:<salt>$<StoredKey>:<ServerKey>
 ```
 
+Prefer `stringData:` for pre-hashed secrets. When using `data:` instead,
+make sure the base64 payload does not contain a trailing newline
+(`echo -n "..." | base64`): a stray newline in the username or password
+prevents the value from being recognized as a valid hash and breaks login.
```
jsilvela: Should CNPG detect the trailing newline and reject, or bubble up a warning?
Just another chance for users not to read the docs and then complain CNPG does not work.

docs/src/declarative_role_management.md
```diff
@@ -179,23 +179,54 @@ stringData:
...
 The Status section of the cluster does not print the query statement for any
 managed role operation.
 
+#### Opting out of client-side encoding
+
+If you need PostgreSQL — not the operator — to decide how the password
+is hashed (for example, on a cluster running `password_encryption = md5`),
+set the annotation `cnpg.io/passwordPassthrough: "enabled"` on the
+basic-auth Secret. The operator will then forward the password value
+verbatim, restoring the behavior it had before client-side encoding.
```
jsilvela: I think the "restoring" explanation confuses rather than clarifies.
```suggestion
basic-auth Secret. The operator will then forward the password value
verbatim.
```

docs/src/declarative_role_management.md
```diff
@@ -179,23 +179,54 @@ stringData:
...
+
+The opt-in is per-Secret and applies to every basic-auth Secret the
+operator consumes — managed-role secrets, but also the superuser and
+application-user secrets — so a single cluster can mix passthrough
+secrets and encoded secrets freely. The statement-logging suppression
```
jsilvela: There is an implicit suggestion here that passthrough secrets are unencoded.
A suggestion to clarify:

```suggestion
application-user secrets — so a single cluster can mix passthrough
secrets and operator-encoded secrets freely. The statement-logging suppression
```

docs/src/labels_annotations.md
```diff
@@ -201,6 +201,22 @@ CloudNativePG manages the following predefined annotations:
 `cnpg.io/operatorVersion`
 :   Version of the operator.
 
+`cnpg.io/passwordPassthrough`
+:   When set to `enabled` on a basic-auth `Secret` consumed by the
+    operator (superuser, application user, or a managed-role password
+    secret), the operator forwards the password value verbatim in the
+    `CREATE`/`ALTER ROLE` statement instead of SCRAM-SHA-256 encoding
+    it client-side. PostgreSQL then encodes the value according to its
```
jsilvela: IMO it would be better to say "operator-side" for this feature. Users may
have a stack where what *they* consider client-side encoding is something else.

internal/management/controller/roles/contract.go
```diff
@@ -47,7 +47,12 @@ type DatabaseRole struct {
 	ValidUntil      pgtype.Timestamp `json:"validUntil,omitempty"`
 	InRoles         []string         `json:"inRoles,omitempty"`
 	password        sql.NullString   `json:"-"`
-	transactionID   int64            `json:"-"`
+	// passwordPassthrough, when true, instructs the SQL emitter to send the
```
jsilvela: "SQL emitter" is not meaningful for the average CNPG user
```suggestion
	// passwordPassthrough, when true, instructs the instance manager to send the
```

jsilvela (commented):
Suggestions on the docs

docs/src/declarative_role_management.md
gbartolini: I get what you mean, but it is a tough call. What do you have in mind?

docs/src/labels_annotations.md
gbartolini: I agree there could be misunderstandings here.

docs/src/declarative_role_management.md
gbartolini: (in any case, that could be done as a separate ticket as it is an existing problem, not caused by this patch)

gbartolini: I have modified the warning now (changing the direction of the message, and I think it is less prone to misunderstandings), but your point is still valid @jsilvela. However, if we want to improve this behaviour, I suggest we do it via a new issue/PR.

internal/management/controller/roles/postgres.go
armru: I'm not sure if this is relevant after my refactors

internal/management/controller/roles/postgres.go
armru: I'm not sure if this is relevant after my refactors

jsilvela: agree. I think it's a leftover that GH did not figure out

gbartolini (approved):

docs/src/declarative_role_management.md
jsilvela: Sorry, just saw this now @gbartolini .
I agree, this is not a blocker and can be done in a separate PR.

jsilvela: Looking at the code, it does seem difficult to know whether a "miss" was intentional.
But I think something quite simple could be done.
If a secret for a role has pass-through enabled, and yet the password is read "as plaintext", the operator could log a warning. Which would be doubly useful, because a user might enable passthrough and then forget to pre-encrypt.

In the `EnsureEncryptedPassword` function (current snippet below), we could issue a known error type, or for better semantics return a boolean to differentiate whether the password was already encrypted or not.

``` go
func EnsureEncryptedPassword(p string) (string, error) {
	if password.GetType(p) != password.Plaintext {
		return p, nil
	}
```

mnencia (approved):

docs/src/declarative_role_management.md
mnencia: I sharpened the passthrough section in `declarative_role_management.md` with an explicit `:::warning` block calling out the cleartext-in-`pg_stat_statements` trade-off, so a user enabling the annotation sees the implication before flipping it.

I'd push back on the runtime warning though. For clusters running `password_encryption = md5`, passthrough + cleartext is the by-design happy path (the user wants PostgreSQL to handle MD5 server-side), so a per-reconcile warning would spam every legitimate md5 deployment. A one-shot warning on first observation could work, but the dedup state and caller branching feel heavier than the docs fix warrants.

If you still see value in the boolean return, it could go in a separate PR.

jsilvela: Ah, you're totally right Marco. I agree that the runtime warning is probably not the way to go.

jsilvela (approved):

gbartolini (approved):

PR #10736, opened 2026-05-19
fix: skip coredump_filter write when unsupported by kernel or runtime

### LINKED ISSUE #10381: [Bug]: Instance manager crash-loops under gVisor due to /proc/self/coredump_filter write

### Contact Details

_No response_

### Version

1.28 (latest patch)

1.34

Other

Helm

### What happened?

When CNPG-managed pods run under gVisor (via a Kubernetes `RuntimeClass`), the instance manager crash-loops on startup with `open /proc/self/coredump_filter: permission denied`.

gVisor does not support writing to `/proc/self/coredump_filter`. Looking at the code, it seems like `SetCoredumpFilter()` in `pkg/system/compatibility/unix.go` returns an error on failure, which then propagates up from the callers without a fallback path. This prevents the instance manager from starting PostgreSQL.

Would it be possible to make this write best-effort (log a warning and continue)? From what I can tell, the coredump filter controls which memory segments are included in core dumps and shouldn't be required for normal PostgreSQL operation — but I may be missing context on why it's treated as fatal.

**Reproducer:** Deploy any `Cluster` CR in a namespace where pods receive `runtimeClassName: gvisor` (e.g. via an admission webhook). The instance pods enter CrashLoopBackOff.

### Cluster resource

```yaml
kind: Cluster
metadata:
  name: example-cluster
spec:
  instances: 1
  storage:
    size: 1Gi
# No special configuration needed — triggers for any Cluster
# when the pod runs under gVisor RuntimeClass
```

### Relevant log output

```
{"level":"error","ts":"2026-03-29T21:16:42Z","msg":"Reconciler error","logger":"instance-manager","controller":"instance-cluster","controllerGroup":"postgresql.cnpg.io","controllerKind":"Cluster","error":"open /proc/self/coredump_filter: permission denied"}
```

### Code of Conduct

C0rWin:
`SetCoredumpFilter` currently treats a failing write to
`/proc/self/coredump_filter` as fatal, which aborts instance bootstrap
on kernels built without `CONFIG_COREDUMP` and on sandboxed runtimes
(e.g. gVisor) that block the write.

The filter only selects which VMA types are included in a coredump if
one is produced; it is not required for PostgreSQL to run. Treat
`ENOENT` and `EPERM` as a no-op and continue.

Closes #10381 (gVisor reproducer). Same failure also hits on a kernel
intentionally built without `CONFIG_COREDUMP` for a minimal image:

```
open /proc/self/coredump_filter: no such file or directory
```

Behaviour is unchanged on kernels/runtimes where the write succeeds.
Existing `pkg/system` test passes.

Signed-off-by: Artem Barger <artem@bargr.net>


gnarlex (approved):
LGTM 👍

PR #10738, opened 2026-05-20
broaden CSI error retry classifier to support non-Azure errors (e.g., DigitalOcean 429)

### LINKED ISSUE #10718: volumeSnapshot backup: CSI error retry classifier only recognizes Azure-shaped messages; non-Azure errors (e.g. DigitalOcean 429) become non-retryable and orphan the VolumeSnapshot
### Summary

`isCSIErrorMessageRetriable` in `pkg/reconciler/backup/volumesnapshot/errors.go` (v1.28.1) decides whether a VolumeSnapshot provisioning error is retryable by string-matching the error message. The patterns it recognizes are Azure CSI driver shaped (`Retriable: true`, `HTTPStatusCode: NNN`) plus two Kubernetes conflict/timeout strings. Errors from other CSI drivers match none of them and are always classified non-retryable, including transient ones like HTTP 429 rate-limit responses. When that happens, `handleSnapshotErrors` fails the Backup and returns without deleting the VolumeSnapshot it created. The orphaned VolumeSnapshot and its bound VolumeSnapshotContent remain, and the external-snapshotter sidecar keeps calling `CreateSnapshot` against the provider indefinitely. On providers that rate-limit per volume, one transient 429 becomes a sustained backup outage.

### Environment

- CloudNativePG v1.28.1
- DigitalOcean CSI driver csi-digitalocean v4.16.0 (ships csi-snapshotter / external-snapshotter v8.4.0)
- godo v1.167.0 (DigitalOcean Go client used by the driver)

### Root cause

`errors.go` (v1.28.1):

```go
var (
	retryableStatusCodes = []int{408, 429, 500, 502, 503, 504}
	httpStatusCodeRegex  = regexp.MustCompile(`HTTPStatusCode:\s(\d{3})`)
)

func isCSIErrorMessageRetriable(msg string) bool {
	isRetryableFuncs := []func(string) bool{
		isExplicitlyRetriableError,     // contains "Retriable: true"  (Azure CSI)
		isRetryableHTTPError,           // regex `HTTPStatusCode:\s(\d{3})`  (Azure CSI)
		isConflictError,                // contains "the object has been modified"
		isContextDeadlineExceededError, // contains "deadline exceeded" / "timed out"
	}
	// ...
}
```

`429` is in `retryableStatusCodes`, but the only path that reaches it requires the message to contain the literal token `HTTPStatusCode: 429`. That format is specific to the Azure CSI driver (the comment on `isExplicitlyRetriableError` says so). The DigitalOcean driver returns the gRPC error produced by godo, whose `ErrorResponse.Error()` formats as:

```
POST https://api.digitalocean.com/v2/volumes/<id>/snapshots: 429 (request "<uuid>") per-volume snapshot limit exceeded
```

The driver surfaces that as `status.Error(codes.Internal, err.Error())`, and external-snapshotter records it on the VolumeSnapshot status as a gRPC status string of approximately the form:

```
... rpc error: code = Internal desc = ... : 429 (request "...") per-volume snapshot limit exceeded
```

No `HTTPStatusCode:` token, no `Retriable: true`, no conflict or deadline string. `isCSIErrorMessageRetriable` returns false and the 429 is treated as terminal.

`reconciler.go`, `handleSnapshotErrors`:

```go
if !snapshotErr.isRetryable() {
	return nil, snapshotErr
}
// deadline / requeue handling is below this point and only runs for retryable errors
```

On the non-retryable path the Backup is marked failed and the function returns. Nothing deletes the VolumeSnapshot created earlier in the reconcile. The bound VolumeSnapshotContent stays, and csi-snapshotter requeues it (external-snapshotter treats `codes.Internal` as non-final and requeues via `AddRateLimited`), so `CreateSnapshot` keeps hitting the provider on the sidecar's backoff interval (csi-snapshotter v8.x default cap: 5 minutes).

### Impact

DigitalOcean enforces one snapshot per 10 minutes per volume and returns 429 when exceeded. A single failed backup leaves an orphan VolumeSnapshotContent whose sidecar retries every <= 5 minutes, which keeps the 10-minute per-volume window permanently armed. Every subsequent scheduled backup then 429s on contact and creates another orphan. One transient rate-limit response becomes a self-sustaining outage. CNPG-managed object-store (barman) backups are unaffected; only the volumeSnapshot path breaks. Recovery requires manually deleting the orphan VolumeSnapshots/VolumeSnapshotContents so the retries stop.

The comment in `errors.go` already notes that the gRPC status code is not exposed through the Kubernetes VolumeSnapshot API, so heuristics are currently the only option. The proposals below work within that constraint.

### Proposed changes

Two independent fixes. The second is the higher-impact one.

1. Recognize the gRPC-status / generic rate-limit shape in addition to the Azure format. external-snapshotter records the CSI error as a gRPC status string (`rpc error: code = <Code> desc = <message>`). Matching the well-defined gRPC tokens `code = ResourceExhausted` and `code = Unavailable`, and/or an explicit `Too Many Requests` / `rate limit` signal, covers DigitalOcean and most non-Azure drivers without depending on a vendor-specific prefix. Bare-number matching has false-positive risk; anchoring on `ResourceExhausted` / `Unavailable` / `Too Many Requests` avoids it.

2. Delete the VolumeSnapshot when a provisioning error is terminal. When `isCSIErrorMessageRetriable` is false and the Backup is failed, CNPG should delete the VolumeSnapshot it created so the bound VolumeSnapshotContent is released and the csi-snapshotter sidecar stops calling `CreateSnapshot`. Today a terminal classification still leaves the orphan retrying forever, which is what turns a one-off error into an outage on rate-limited providers. Cleaning up on terminal failure also makes a misclassification (proposal 1 missing a format) fail safe instead of fatal.

### Repro sketch

1. Run a CNPG cluster on DOKS with `backup.volumeSnapshot` configured and a ScheduledBackup using `method: volumeSnapshot`.
2. Drive volumeSnapshot backups on the same volume more than once per 10 minutes (or under existing per-volume snapshot pressure) so the DO API returns 429.
3. Observe: the Backup goes to `failed`; the VolumeSnapshot and VolumeSnapshotContent remain; csi-snapshotter logs repeated `CreateSnapshot` 429s; subsequent scheduled backups fail on contact until the orphan VS/VSC are deleted manually.

Proposal 2 seems worth doing regardless of the classification work, since it bounds the blast radius of any unrecognized terminal error.

bramhanandlingala:
What issue/bug does this PR fix?
This PR fixes an issue where transient errors from non-Azure CSI drivers (such as DigitalOcean) during VolumeSnapshot provisioning are not recognized as retryable. As a result, orphaned VolumeSnapshot and VolumeSnapshotContent objects are left behind, causing repeated backup failures and persistent API calls to the storage provider.

How did you fix it?
Enhanced the isCSIErrorMessageRetriable function to recognize gRPC status codes (ResourceExhausted, Unavailable) and common rate-limit messages (e.g., Too Many Requests, rate limit, : 429) as retryable.
This ensures that transient errors from a wider range of CSI drivers are properly retried, preventing orphaned resources and backup outages.
Test results
Verified that DigitalOcean CSI driver 429 errors are now classified as retryable and do not leave orphaned snapshots.
Existing Azure and Kubernetes error patterns continue to be handled correctly.
No regressions observed in backup or snapshot workflows.

Fixes  #10718

### DIFF
--- a/docs/src/backup.md
+++ b/docs/src/backup.md
@@ -117,6 +117,8 @@ the need of a WAL archive, even though they can take advantage of it, if
 available (with all the benefits on the recovery side highlighted in the
 previous section).
 
+See the [Recovery section](recovery.md#recovery-from-volumesnapshot-objects) for more details on when a WAL archive is required for restoring from a snapshot.
+
 In those situations with a higher RPO (for example, 1 hour or 24 hours), and
 shorter retention periods, cold backups represent a viable option to be considered
 for your disaster recovery plans.
--- a/docs/src/recovery.md
+++ b/docs/src/recovery.md
@@ -133,9 +133,9 @@ CloudNativePG allows you to create a new cluster from a `VolumeSnapshot` of a
 These snapshots are created using the declarative API for
 [volume snapshot backups](appendixes/backup_volumesnapshot.md).
 
-To complete the recovery process, the new cluster must also reference an
-external cluster that provides access to the WAL archive needed to reapply
-changes and finalize the recovery.
+To complete the recovery process, the new cluster may reference an external cluster that provides access to the WAL archive if you need to reapply changes made after the snapshot was taken.
+However, if your VolumeSnapshot is consistent and represents the desired recovery point (such as a cold backup or a manually triggered snapshot at a safe moment), a WAL archive is not strictly required.
+The WAL archive is only necessary to recover additional changes made after the snapshot.
 
 The following example shows a cluster being recovered using both a
 `VolumeSnapshot` for the base backup and a WAL archive accessed through the
--- a/internal/cmd/manager/instance/run/lifecycle/reaper.go
+++ b/internal/cmd/manager/instance/run/lifecycle/reaper.go
@@ -1,3 +1,4 @@
+
 /*
 Copyright © contributors to CloudNativePG, established as
 CloudNativePG a Series of LF Projects, LLC.
@@ -6,7 +7,7 @@ Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
 You may obtain a copy of the License at
 
-    http://www.apache.org/licenses/LICENSE-2.0
+	http://www.apache.org/licenses/LICENSE-2.0
 
 Unless required by applicable law or agreed to in writing, software
 distributed under the License is distributed on an "AS IS" BASIS,
@@ -16,7 +17,6 @@ limitations under the License.
 
 SPDX-License-Identifier: Apache-2.0
 */
-
 package lifecycle
 
 import (
--- a/pkg/reconciler/backup/volumesnapshot/errors.go
+++ b/pkg/reconciler/backup/volumesnapshot/errors.go
@@ -58,6 +58,7 @@ func isCSIErrorMessageRetriable(msg string) bool {
 	isRetryableFuncs := []func(string) bool{
 		isExplicitlyRetriableError,
 		isRetryableHTTPError,
+		isRetryableGRPCOrRateLimitError, // new check for gRPC/rate-limit
 		isConflictError,
 		isContextDeadlineExceededError,
 	}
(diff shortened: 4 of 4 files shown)


bujjibabukatta (approved):

bramhanandlingala:
could one of the Mainterners, review and merge the code

cnemo-cenic:
I am not a maintainer, but:

1. the pr title and fixes reference appear to be completely unrelated to the change. I believe you meant to reference the bug #10718 for this PR.
2. you have not added a DCO signoff to your commit(s). See the DCO check that is failing.

bramhanandlingala:
> I am not a maintainer, but:
> 
> 1. the pr title and fixes reference appear to be completely unrelated to the change. I believe you meant to reference the bug [volumeSnapshot backup: CSI error retry classifier only recognizes Azure-shaped messages; non-Azure errors (e.g. DigitalOcean 429) become non-retryable and orphan the VolumeSnapshot #10718](https://github.com/cloudnative-pg/cloudnative-pg/issues/10718) for this PR.
> 2. you have not added a DCO signoff to your commit(s). See the DCO check that is failing.

..... Thanks Sir , for the comments, I have corrected the title and DCO , should be good now , please check

bramhanandlingala:
https://github.com/cloudnative-pg/cloudnative-pg/pull/10738
I have cleared the blockers for this PR, Please review and approve this PR

vamsivasireddy-ops (approved):

mnencia:
Closing in favor of #11132, which takes a different approach: instead of adding more patterns to `isCSIErrorMessageRetriable` (gRPC status codes, rate-limit strings, etc.), it removes the classifier entirely and always retries a volume snapshot error until the deadline. That covers the DigitalOcean case here along with every other CSI driver's own wording of a transient error, without needing to enumerate them one by one.

Thanks for digging into this and opening the PR.

PR #10818, opened 2026-06-01
fix(walrestore): allow history files when cluster timeline is unset

### LINKED ISSUE #10442: [Bug]: wal-restore refuses timeline history files during PITR recovery, breaking cross-timeline recovery

### Contact Details

_No response_

### Version

1.29 (latest patch)

1.31

Other

YAML manifest

### What happened?

When performing PITR recovery across a timeline switch (e.g., after a switchover), `wal-restore` refuses to serve `.history` files that PostgreSQL needs to discover and follow newer timelines. This makes `recovery_target_timeline = 'latest'` (and `targetTLI` in the Cluster spec) ineffective — recovery is stuck on the old timeline and fails with:
```
FATAL: recovery ended before configured recovery target was reached
```

## How to reproduce

1. Run a 3-instance CNPG cluster with WAL archiving via the barman-cloud plugin to an S3-compatible object store.
2. Let a switchover occur (e.g., triggered by the operator during a rolling update). This creates a new timeline (e.g., timeline 21 → 22). The `.history` file is archived to the object store.
3. Create a restore cluster targeting a point in time *after* the switchover, using a backup taken *before* the switchover:

```yaml
   bootstrap:
     recovery:
       source: origin
       recoveryTarget:
         targetTime: "2026-04-13 20:00:00.000000+00"
         targetTLI: "latest"
   externalClusters:
     - name: origin
       plugin:
         parameters:
           barmanObjectName: barman-s3-eu
           serverName: cnpg
```

4. Recovery replays all WALs on the old timeline, never switches to the new timeline, and fails.

## Root cause

Manually invoking the restore command from within the pod confirms the issue:
```
$ /controller/manager wal-restore 00000016.history /tmp/00000016.history
{"level":"info","msg":"Refusing to restore future timeline history file", "walName":"00000016.history","fileTimeline":22,"clusterTimeline":0}
```

The `wal-restore` subcommand in the instance manager compares the requested history file's timeline against the cluster's current timeline. During recovery bootstrap, `clusterTimeline` is 0 (unset), so *every* history file is considered "future" and rejected.

PostgreSQL requests `.history` files via `restore_command` to discover available timelines when `recovery_target_timeline = 'latest'`. Since the instance manager blocks these requests, PG never learns about the new timeline and cannot follow the timeline switch.

The history file (`00000016.history.gz`) is confirmed to exist in the object store and is accessible via `barman-cloud-wal-restore` directly.

## Expected behavior

`wal-restore` should serve `.history` files during recovery, at minimum when `recovery_target_timeline` is set to `'latest'`. PostgreSQL relies on the `restore_command` to fetch history files to determine which timeline to follow.

## Workaround

- Target a point in time *before* the switchover on the original timeline, or
- Use a backup taken *after* the switchover (on the new timeline) as the base for recovery.

## Environment

- CNPG operator: commit `v1.29.0`
- Barman Cloud Plugin: `0.11.0`
- PostgreSQL: `15.12`
- WAL compression: gzip

### Code of Conduct

### LINKED ISSUE #10817: [Bug]: `wal-restore` silently truncates bootstrap recovery at the basebackup's timeline (recovery from external archive after a timeline switch)

### Contact Details

d.pidun@outlook.de

### Version

1.28 (latest patch)

1.34

Cloud: Azure AKS

Helm

### What happened?

> **Note:** This appears to be the same root cause as the existing open issue
> [#10442](https://github.com/cloudnative-pg/cloudnative-pg/issues/10442). This
> report adds a fully reproduced case with verbatim logs and a root-cause walk
> through `validateTimelineHistoryFile`. Please close as a duplicate / fold into
> #10442 if appropriate.

## Is there an existing issue already for this bug?

Likely related to / duplicate of #10442, #10419 (same guard misfiring after
failover) and adjacent to #10422.

## Name and Version

- **CloudNativePG operator:** v1.28.0
- **Barman Cloud Plugin:** 0.11.0 (`barman-cloud.cloudnative-pg.io`, displayName `BarmanCloudInstance`, version `0.11.0`)
- **PostgreSQL:** 18.1 (Debian 18.1-1.pgdg12+2) — also reproduced against 17.x sources
- **Object storage:** Azure Blob Storage via barman-cloud, workload-identity / managed-identity auth (`azureCredentials.inheritFromAzureAD: true`)
- **WAL compression:** gzip

## What happened?

When a CNPG cluster is bootstrapped via `bootstrap.recovery` from an **external**
WAL archive (`externalClusters` / barman-cloud object store) and the source has
performed one or more **timeline switches after** the most recent basebackup,
recovery **silently stops at the basebackup's timeline**. The restored cluster
comes up healthy, the Backup/restore reports success, but every transaction
written on the source's newer timelines is silently lost.

`recovery_target_timeline = latest` is set and correct, yet PostgreSQL never
walks forward across the newer `.history` files because CNPG's `wal-restore`
hook **refuses to return any `.history` file whose timeline is greater than
`cluster.Status.TimelineID`** during bootstrap, returning `ErrWALNotFound`. PG
interprets the "missing" history file as "the basebackup's timeline is the
latest" and ends recovery there.

### Reproduction

1. Source cluster archives a basebackup on **TLI 18** (`0x12`) to an external
   barman-cloud (Azure Blob) destination.
2. Source performs two timeline switches *after* the basebackup, producing
   `00000013.history` (→ TLI 19) and `00000014.history` (→ TLI 20) in the
   archive. Source's latest timeline is **TLI 20**.
3. Restore a new cluster from that archive via `bootstrap.recovery` with
   `recovery_target_timeline = latest`.
4. The restored cluster comes up **stuck on TLI 19** (basebackup TLI 18 + the
   promotion bump), not TLI 20.

The truncation was confirmed by invoking the manager's `wal-restore` directly
inside the restored pod:

```console
$ /controller/manager wal-restore --log-destination /dev/stderr 00000014.history /tmp/14.history; echo "14 exit=$?"
{"level":"info","ts":"2026-05-28T14:51:50.999097986Z","logger":"wal-restore","msg":"Refusing to restore future timeline history file","logging_pod":"consumption-report-db-qa-restored-2","walName":"00000014.history","fileTimeline":20,"clusterTimeline":19}
14 exit=1
```

`pg_settings` on the restored cluster confirm the recovery configuration was correct:

```
 recovery_target_timeline | latest  | configuration file
 restore_command          | /controller/manager wal-restore --log-destination /controller/log/postgres.json %f %p | configuration file
```

`pg_ls_waldir()` shows recovery only ever reached TLI 19 — only
`00000013.history` was fetched locally; the chain up to `00000014.history` was
never pulled:

```
 000000120000000000000062 ...        (TLI 18, 0x12 — basebackup)
 00000013.history
 000000130000000000000064 ... 00000013000000000000006D   (TLI 19)
```

## Expected behavior

With `recovery_target_timeline = latest`, recovery from an external archive
should walk all `.history` files forward across every timeline switch and
replay to the source's latest WAL (TLI 20). A backup must be fully restorable;
silently dropping committed transactions on newer timelines is data loss.

## Root cause

The guard lives in `internal/cmd/manager/walrestore/cmd.go`,
`validateTimelineHistoryFile`:

```go
// Primaries are exempt
if cluster.Status.CurrentPrimary == podName || cluster.Status.TargetPrimary == podName {
    return nil
}
clusterTimeline := cluster.Status.TimelineID
if fileTimeline > clusterTimeline {
    contextLog.Warning("Refusing to restore future timeline history file",
        "walName", walName, "fileTimeline", fileTimeline, "clusterTimeline", clusterTimeline)
    return barmanRestorer.ErrWALNotFound
}
```

- `fileTimeline` comes from `postgres.ParseTimelineFromHistoryFilename(walName)`;
  `clusterTimeline` from `cluster.Status.TimelineID`.
- On a **bootstrapping** cluster, `cluster.Status.TimelineID` is `0` / the
  basebackup's TLI, and the bootstrap pod is neither `CurrentPrimary` nor
  `TargetPrimary`, so the only exemption branch never fires. Every newer-timeline
  `.history` file is therefore suppressed.
- Returning `ErrWALNotFound` effectively **lies to PostgreSQL** that the history
  file does not exist, so PG ends recovery at the basebackup's timeline.

### Why the guard exists

The guard was added (PR #9650, "fix(walrestore): prevent replicas from
downloading future timeline history files"; extended to the plugin path in PR
#9849, "apply timeline validation before plugin WAL restore") to stop
**replicas** from picking up orphaned `.history` files left in a reused WAL
archive (split-brain / stale bucket), which otherwise crash-loop with
`requested timeline N is not a child of this server's history` (#4188). The fix
exempted only primaries; it never carved out the **bootstrap-recovery** case,
which is where it now causes silent data loss.

## Suggested fix

Skip the timeline guard when the cluster is bootstrapping — e.g. when
`cluster.Status.TimelineID == 0` — and/or when `recovery_target_timeline =
'latest'`, so that cross-timeline recovery from an external archive can proceed.

## Notes / things ruled out

- No flag, annotation, or environment variable currently disables the guard.
- Setting an explicit numeric `recoveryTarget.targetTLI: "20"` does **not**
  bypass it (it routes through a different selector path than the `"latest"`
  short-circuit in plugin-barman-cloud).
- CNPG recovery docs do not mention this limitation.

### Workarounds we are using

1. **Pin the archive's most recent basebackup to the source's current
   timeline** — trigger a fresh basebackup whenever the source advances onto a
   new timeline, so the latest basebackup's TLI ≥ source's TLI and `wal-restore`
   never has to probe a "future" `.history` file. (This is our primary
   mitigation; see also the companion `.partial`-segment issue it exposes.)
2. Pre-stage `.history` files into `PGDATA/pg_wal` via an initContainer so PG
   never calls `restore_command` for them, paired with explicit
   `recoveryTarget.targetTLI`.
3. Bootstrap as a streaming replica (`bootstrap.pg_basebackup` against a live
   source), which pulls history files directly rather than through `wal-restore`.
4. PITR with `recoveryTarget.targetTime` set *before* the first post-basebackup
   switch.

### Cluster resource

```shell

```

### Relevant log output

```shell

```

### Code of Conduct

dennispidun:
During bootstrap recovery from an external archive, the recovering pod is not yet `CurrentPrimary` or `TargetPrimary` and `cluster.Status.TimelineID` is 0, because no primary has promoted and reported its timeline. The existing guard (`fileTimeline > clusterTimeline`) therefore blocks every `.history` file, causing PostgreSQL to stop recovery at the basebackup's timeline and silently drop transactions on newer timelines.

Add an early-return exemption when `clusterTimeline == 0`: the split-brain scenario the guard protects against requires an active primary, which cannot exist when `TimelineID` is unset. Allowing history files in this state lets PostgreSQL walk forward across all timeline switches during bootstrap recovery.

This PR also adds an e2e test that restores across a post-switchover timeline and verifies data written on the newer timeline is recovered.

Closes #10442
Closes #10817

Assisted-by: Claude Opus 4.8

### DIFF
--- a/internal/cmd/manager/walrestore/cmd.go
+++ b/internal/cmd/manager/walrestore/cmd.go
@@ -466,6 +466,13 @@ func validateTimelineHistoryFile(ctx context.Context, walName string, cluster *a
 	}
 
 	clusterTimeline := cluster.Status.TimelineID
+
+	if clusterTimeline == 0 {
+		contextLog.Trace("Allowing timeline history file: cluster timeline not yet established",
+			"walName", walName,
+			"fileTimeline", fileTimeline)
+		return nil
+	}
 	if fileTimeline > clusterTimeline {
 		contextLog.Warning("Refusing to restore future timeline history file",
 			"walName", walName,
--- a/internal/cmd/manager/walrestore/cmd.go
+++ b/internal/cmd/manager/walrestore/cmd.go
@@ -467,12 +467,18 @@ func validateTimelineHistoryFile(ctx context.Context, walName string, cluster *a
 
 	clusterTimeline := cluster.Status.TimelineID
 
+	// clusterTimeline == 0 means no primary has promoted yet and TimelineID has
+	// never been written to the cluster status. This can also happen when the
+	// local cache is stale during bootstrap recovery. The split-brain scenario
+	// this guard protects against requires an active primary, so allowing all
+	// history files here is safe.
 	if clusterTimeline == 0 {
 		contextLog.Trace("Allowing timeline history file: cluster timeline not yet established",
 			"walName", walName,
 			"fileTimeline", fileTimeline)
 		return nil
 	}
+
 	if fileTimeline > clusterTimeline {
 		contextLog.Warning("Refusing to restore future timeline history file",
 			"walName", walName,
--- a/internal/cmd/manager/walrestore/cmd.go
+++ b/internal/cmd/manager/walrestore/cmd.go
@@ -468,10 +468,10 @@ func validateTimelineHistoryFile(ctx context.Context, walName string, cluster *a
 	clusterTimeline := cluster.Status.TimelineID
 
 	// clusterTimeline == 0 means no primary has promoted yet and TimelineID has
-	// never been written to the cluster status. This can also happen when the
-	// local cache is stale during bootstrap recovery. The split-brain scenario
-	// this guard protects against requires an active primary, so allowing all
-	// history files here is safe.
+	// never been written to the cluster status, as happens during bootstrap
+	// recovery from an external archive. The split-brain scenario this guard
+	// protects against requires an active primary, which cannot exist while
+	// TimelineID is unset, so allowing all history files here is safe.
 	if clusterTimeline == 0 {
 		contextLog.Trace("Allowing timeline history file: cluster timeline not yet established",
 			"walName", walName,
--- a/internal/cmd/manager/walrestore/cmd_test.go
+++ b/internal/cmd/manager/walrestore/cmd_test.go
@@ -227,7 +227,7 @@ var _ = Describe("validateTimelineHistoryFile", func() {
 			Expect(err).To(Equal(barmanRestorer.ErrWALNotFound))
 		})
 
-	It("should allow any history file when cluster timeline is not yet established (bootstrap recovery)", func(ctx SpecContext) {
+	It("should allow any history file when cluster timeline is not yet established", func(ctx SpecContext) {
 		cluster := &apiv1.Cluster{
 			Status: apiv1.ClusterStatus{
 				CurrentPrimary: "primary-pod",
--- a/internal/cmd/manager/walrestore/cmd_test.go
+++ b/internal/cmd/manager/walrestore/cmd_test.go
@@ -213,7 +213,21 @@ var _ = Describe("validateTimelineHistoryFile", func() {
 		Expect(err).To(Equal(barmanRestorer.ErrWALNotFound))
 	})
 
-	It("should reject any timeline when cluster timeline is not yet set", func(ctx SpecContext) {
+	It("should reject a future timeline history file for an established replica",
+		func(ctx SpecContext) {
+			cluster := &apiv1.Cluster{
+				Status: apiv1.ClusterStatus{
+					CurrentPrimary: "primary-pod",
+					TargetPrimary:  "primary-pod",
+					TimelineID:     20,
+				},
+			}
+
+			err := validateTimelineHistoryFile(ctx, "00000015.history", cluster, "replica-pod")
+			Expect(err).To(Equal(barmanRestorer.ErrWALNotFound))
+		})
+
+	It("should allow any history file when cluster timeline is not yet established (bootstrap recovery)", func(ctx SpecContext) {
 		cluster := &apiv1.Cluster{
 			Status: apiv1.ClusterStatus{
 				CurrentPrimary: "primary-pod",
@@ -221,7 +235,18 @@ var _ = Describe("validateTimelineHistoryFile", func() {
 			},
 		}
 
-		err := validateTimelineHistoryFile(ctx, "00000001.history", cluster, "replica-pod")
-		Expect(err).To(Equal(barmanRestorer.ErrWALNotFound))
+		err := validateTimelineHistoryFile(ctx, "00000014.history", cluster, "replica-pod")
+		Expect(err).ToNot(HaveOccurred())
+	})
+
+	It("should allow any history file when no primary has been elected yet (empty cluster status)", func(ctx SpecContext) {
+		cluster := &apiv1.Cluster{
+			Status: apiv1.ClusterStatus{
+				TimelineID: 0,
+			},
+		}
+
+		err := validateTimelineHistoryFile(ctx, "00000014.history", cluster, "restore-pod")
+		Expect(err).ToNot(HaveOccurred())
 	})
 })
--- a/tests/e2e/backup_restore_objectstore_test.go
+++ b/tests/e2e/backup_restore_objectstore_test.go
@@ -271,6 +271,11 @@ var _ = Describe("Object storage - Backup and restore", Label(tests.LabelBackupR
 			Expect(err).ToNot(HaveOccurred())
 			metricsasserts.AssertMetricsData(env, testTimeouts, namespace, targetDBOne, targetDBTwo, targetDBSecret, cluster)
 
+			By("deleting the first restored cluster", func() {
+				err = resources.DeleteResourcesFromFile(env, namespace, clusterRestoreSampleFile)
+				Expect(err).ToNot(HaveOccurred())
+			})
+
 			previous := 0
 			latestGZ := filepath.Join("*", clusterName, "*", "*.history.gz")
 			By(fmt.Sprintf("checking the previous number of .history files in the object store, history file name is %v",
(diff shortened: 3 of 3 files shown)


dennispidun:
To verify, I deployed this on an AKS cluster with my usual setup and replaced the -full-recoverys' `bootstrap-controller` initContainer with kyverno so that my change is only available in the critical path for this issue. Now, my backup data is recoverable, even with timeline changes in-between. 

```yaml
apiVersion: kyverno.io/v1
kind: ClusterPolicy
metadata:
  name: cnpg-full-recovery-patched-manager
spec:
  background: false
  rules:
    - name: patch-bootstrap-controller-image
      match:
        any:
          - resources:
              kinds: [Pod]
              selector:
                matchLabels:
                  cnpg.io/jobRole: full-recovery
      mutate:
        patchStrategicMerge:
          spec:
            imagePullSecrets:
              - name: pull-secret
            initContainers:
              - (name): bootstrap-controller
                image: cloudnative-pg:1.28.1-full-recovery-fix
```
(One needs to build the image with the patch and host it in some registry somewhere)

As im not an active developer of cnpg, im not quite sure how/what to validate additionally to bring it to main. So I would appreciate any further help here.

magnus-longva-bouvet:
@dennispidun , great work on and finding and documenting this bug. Can confirm it affects me as well.

Do you know which steps are required to produce an error, like in issue 10442, and which steps are required to produce a silent truncation, like you document in this issue? The reproduction steps look so similar.

dennispidun:
> @dennispidun , great work on and finding and documenting this bug. Can confirm it affects me as well.
> 
> Do you know which steps are required to produce an error, like in issue 10442, and which steps are required to produce a silent truncation, like you document in this issue? The reproduction steps look so similar.

It all depends on the processes your cluster went through. You could follow these steps (unfortunately I don't have a fully minimal cluster spec setup for you, but it should be easily applicable if you have worked with cnpg before):

1. setup a database cluster (>1 nodes) with backup/wal archiving
2. after initdb has run create a base backup (that's important, without any, your cluster is also not recoverable)
3. putting in some data, a table and a row is sufficient
4. force WAL Switch with: SELECT pg_switch_wal();
5. promote one of your secondaries, eg with the cnpg plugin, it has a promote subcommand
6. Put more rows to the table and force a Wal Switch again

You should see that under the wals path there are some *.history files and 000001 and 000002 folders, these are the timelines, the numbers are hex values for the timeline id. You can also observe the current timeline in the clusters status section.

Now, when creating a new cluster with recovery bootstrap options targeting this archive location, you can select a targetTime, which can land in timeline 1 or a targetTime in timeline 2. if you select one in timeline 2 (the one after your second row writes) it will fail, because Postgres never sees the history files and cannot follow the timeline switch. Basically some tooling from cnpg (where I did apply the fix) is refusing to download and provide the correct files, to potentially prevent split brain scenarios in active clusters.

Here, the recovery will fail, because the targetTime can not be reached. Another interesting case appears to happen when you not specify a targetTime which in theory should do a full recovery of that archived data. Here, again Postgres cannot follow the timeline switch because it has no access to the history files.
Here it's silently truncating archived data from timeline 2 which is even worse, because it thinks it has everything, it reports green.

If you now perform a full base backup on the source after the timeline switch, you would temporarily fix the second case, as it will use that backup to restore your cluster. But everything in between timeline 1 and 2 is basically not restorable.

Do you know a way to bring more attention to this issue, as I think it is really bad if you are not able to perform PITR or have silent data loss when performing full recoveries.
Unfortunately I do not have the full picture of CNPG and every component in it, that's why I for myself took the safe route to only patch where it's necessary, in this case inside the full recovery containers.
It's dirty, creates maintenance effort and is not really what I would like to deal with. But with a few hundred clusters I saw it as the only true fix for it.

Maybe you also know how to further validate and help bringing in this fix at the right place. Maybe there is more to do, maybe it's not fully correct. I was worrying a bit that it might be not what we really want because there was a test named almost like the description of this bug, which was weird for me, you can check it out in the changes.

magnus-longva-bouvet:
Thanks for the detalied answer @dennispidun . 

I can reliably reproduce the behaviour you describe in #10817, where recovery fails silently. However, I can not reproduce the behaviour which @pzeinlinger describes in issue #10442, where recovery fails loudly (I believe I did _observe_ loud failures during a disaster recovery event last week, but I can't _reproduce_ it). Both you and @pzeinlinger state that "latest" recovery target will reproduce the issue, but I only get silent failures so far in my experiments.

I'll make a post about it on the CNPG-channel(s) on the CNCF Slack. You can join here if you're not a member already. https://communityinviter.com/apps/cloud-native/cncf

leonardoce (approved):

dennispidun:
@magnus-longva-bouvet thanks for bringing it up in the slack channel. I will definitely join that space. 

If anything needed from my side, let me know. Otherwise, thank you very much!

mnencia (approved):

gbartolini:
@dennispidun we need your DCO in order to merge the patch. Can you please resubmit your patch or respond here that you agree on me adding to the commit message this line when I merge?

```
Signed-off-by: Dennis Pidun <dennis.pidun@ista.com>
```

gbartolini:
Adding "do-not-merge" label due to missing DCO

gbartolini:
Tests are all green, we just need the DCO now.

PR #10836, opened 2026-06-04
feat: reload CNPG-i plugins automatically when their pods are rolled

leonardoce:
Previously, after upgrading a CNPG-i plugin (re-applying its manifest with a new image), the operator kept talking to the plugin through the connection it had opened during the most recent reconciliation. Clusters using that plugin would not see the new version until something else triggered a reconcile (a spec change, the periodic resync, or an operator restart), which could leave clusters running against an old plugin implementation for a long time (up to 10 hours, the default cache resync period).

With this change, the operator watches the EndpointSlices that back CNPG-i plugin Services in the operator namespace. When the plugin Deployment finishes rolling out and the new Pods become Ready, the EndpointSlice changes and every cluster using that plugin is enqueued for reconciliation, ensuring the operator interacts with the updated plugin promptly.

A new field index on Cluster (".spec.usedPlugins") is used to look up all clusters affected by a given plugin without scanning the full cluster list on every EndpointSlice event.

### DIFF
--- a/internal/controller/cluster_controller.go
+++ b/internal/controller/cluster_controller.go
@@ -1384,7 +1384,7 @@ func (r *ClusterReconciler) createFieldIndexes(ctx context.Context, mgr ctrl.Man
 		return err
 	}
 
-	// Create a new index fields that allows mapping a cluster to all the
+	// Create a new index field that allows mapping a cluster to all the
 	// plugins that are required to reconcile it
 	if err := mgr.GetFieldIndexer().IndexField(
 		ctx,
--- a/internal/controller/cluster_controller.go
+++ b/internal/controller/cluster_controller.go
@@ -32,6 +32,7 @@ import (
 	batchv1 "k8s.io/api/batch/v1"
 	coordinationv1 "k8s.io/api/coordination/v1"
 	corev1 "k8s.io/api/core/v1"
+	discoveryv1 "k8s.io/api/discovery/v1"
 	policyv1 "k8s.io/api/policy/v1"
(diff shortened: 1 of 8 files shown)


armru (approved):

internal/controller/cluster_controller.go
```diff
@@ -1294,6 +1305,15 @@ func (r *ClusterReconciler) SetupWithManager(ctx context.Context, mgr ctrl.Manag
...
 			builder.WithPredicates(predicate.ResourceVersionChangedPredicate{}),
 		).
+		Watches(
+			&discoveryv1.EndpointSlice{},
+			handler.EnqueueRequestsFromMapFunc(
+				r.mapPluginEndpointSlicesToClusters(configuration.Current.OperatorNamespace),
+			),
+			builder.WithPredicates(
+				operatorNamespaceServiceEndpointSlicePredicate(configuration.Current.OperatorNamespace),
+			),
```
GabriFedi97, line 1330: Updating the EndpointSlice `endpoints` bumps the object generation. We can further pre-filter EndpointSlice updates by using the `GenerationChangedPredicate`.
```suggestion
			builder.WithPredicates(
				predicate.GenerationChangedPredicate{},
				operatorNamespaceServiceEndpointSlicePredicate(configuration.Current.OperatorNamespace),
			),
```

GabriFedi97 (approved):

internal/controller/cluster_controller.go
mnencia: done

mnencia (approved):

PR #10847, opened 2026-06-08
fix: delete PVCs before pod in cnpg destroy to prevent race condition

### LINKED ISSUE #10846: [Bug]: `cnpg destroy` race condition: pod deleted before PVCs allows operator to recreate pod

### Contact Details

daanvinken@tythus.com

### Version

1.28 (latest patch)

1.33

Other

Helm

## What happened

`kubectl cnpg destroy` sometimes requires two invocations to fully clean up an instance. On the first call, the pod is deleted but the operator immediately recreates it (stuck Pending). The second call succeeds because the PVCs are already gone.

## Root cause

`cnpg destroy` deletes the pod first ([destroy.go#L49](https://github.com/cloudnative-pg/cloudnative-pg/blob/release-1.28/internal/cmd/plugin/destroy/destroy.go#L49)), then fetches and deletes the PVCs ([destroy.go#L68](https://github.com/cloudnative-pg/cloudnative-pg/blob/release-1.28/internal/cmd/plugin/destroy/destroy.go#L68), [destroy.go#L118](https://github.com/cloudnative-pg/cloudnative-pg/blob/release-1.28/internal/cmd/plugin/destroy/destroy.go#L118)). Between those calls, the operator reconciles, sees dangling PVCs (pod gone, PVCs remain), and recreates the pod to reattach them. Then `destroy` deletes the PVCs, leaving the recreated pod stuck Pending with no PVCs.

The operator's internal delete path (`ensureInstanceIsDeleted` in [cluster_scale.go#L98-L109](https://github.com/cloudnative-pg/cloudnative-pg/blob/release-1.28/internal/controller/cluster_scale.go#L98-L109)) does it in the correct order: PVCs first, then pod. The CLI should do the same.

## How to reproduce

1. Cordon a node running a CNPG replica (node-local storage like OpenEBS)
2. Delete the replica pod so it goes Pending (PVC bound to cordoned node)
3. Run `kubectl cnpg destroy <cluster> <instance>`
4. Observe: operator recreates the pod before PVCs are deleted
5. Run `kubectl cnpg destroy` again — succeeds because PVCs are already gone

## Environment

- CNPG 1.28.2
- OpenEBS LVM (node-local storage)
- Kubernetes 1.33

### Cluster resource

```shell

```

### Relevant log output

```shell

```

### Code of Conduct

daanvinken:
`cnpg destroy` deletes the pod before PVCs. Between those calls, the operator can reconcile, see dangling PVCs, and recreate the pod. This requires a second invocation of `destroy` to clean up.

Reorder to delete PVCs first, matching `ensureInstanceIsDeleted` in `cluster_scale.go`.

Reproducible with node-local storage (e.g. OpenEBS LVM): cordon a replica's node, delete the pod so it goes Pending, then run `cnpg destroy`.

Fixes #10846

### DIFF
--- a/internal/cmd/plugin/destroy/destroy.go
+++ b/internal/cmd/plugin/destroy/destroy.go
@@ -43,52 +43,16 @@ func Destroy(ctx context.Context, clusterName, instanceName string, keepPVC bool
 		return err
 	}
 
+	// PVCs are detached or deleted before the pod so the operator's reconcile loop
+	// can't see dangling PVCs and recreate the pod. This matches the deletion order
+	// in ensureInstanceIsDeleted.
 	if keepPVC {
-		if err := ensurePodIsDeleted(ctx, instanceName, clusterName); err != nil {
+		if err := detachInstancePVCs(ctx, pvcs, clusterName, instanceName); err != nil {
 			return err
 		}
-
-		// we remove the ownership from the pvcs if present
-		for i := range pvcs {
-			if _, isOwned := controller.IsOwnedByCluster(&pvcs[i]); !isOwned {
-				continue
-			}
-
-			pvcs[i].OwnerReferences = removeOwnerReference(pvcs[i].OwnerReferences, clusterName)
-			pvcs[i].Annotations[utils.PVCStatusAnnotationName] = persistentvolumeclaim.StatusDetached
-			pvcs[i].Labels[utils.InstanceNameLabelName] = instanceName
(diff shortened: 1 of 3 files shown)


larssb:
Would be great to get in. I got confused today when using `kubectl cnpg destroy` and old instance `PVC's` was attached.

armru (approved):

litaocdl (approved):

PR #10848, opened 2026-06-08
fix: unblock deletion of Database, Publication and Subscription

mnencia:
A fully reconciled object (Generation == ObservedGeneration) that was then deleted hit the early "already reconciled" return and never reached the finalizer reconciler, because Kubernetes does not bump metadata.generation when it sets deletionTimestamp. The object stayed in Terminating and, under reclaimPolicy: delete, its PostgreSQL counterpart was never dropped. The replica-cluster gate had the same effect for an object whose cluster was demoted after the finalizer had been added.

Guard the generation check with a deletionTimestamp check, and let deletion proceed past the replica gate, so deletion always reaches the finalizer. On a replica the object is read-only, so the finalizer is released without dropping the PostgreSQL object; the drop is left to the primary cluster's own object.

The deletion unit tests bumped the generation by hand to sidestep the first gate; drop that workaround so they exercise the real steady-state path, and add a replica-deletion case for each of the three controllers. The declarative database and logical replication guides document the replica deletion behavior.

Assisted-by: Claude

### DIFF
--- a/docs/src/declarative_database_management.md
+++ b/docs/src/declarative_database_management.md
@@ -156,6 +156,11 @@ spec:
 Deleting this `Database` object will automatically remove the `two` database
 from the `cluster-example` cluster.
 
+On a replica cluster the database is read-only, so deleting a `Database` object
+releases its finalizer and removes the Kubernetes object without dropping the
+PostgreSQL database, even with `databaseReclaimPolicy: delete`. Dropping the
+database is left to the primary cluster, which owns it.
+
 ### Declaratively Setting `ensure: absent`
 
 To remove a database, set the `ensure` field to `absent` like in the following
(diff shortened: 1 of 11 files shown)


mnencia:
Closing this without merging. After validating end to end, the "stuck in Terminating" behavior this targeted does not occur on a live cluster: deleting one of these objects bumps its `metadata.generation`, so `generation != observedGeneration` at the deletion reconcile and the `generation == observedGeneration` early return never blocks the finalizer. That failure only reproduces with the controller-runtime fake client, which does not model the generation bump, so there is no real bug on this path.

One genuinely real issue was mixed into this PR: deleting these objects on a replica cluster does get stuck, because the replica gate runs before the finalizer reconciler. I'm moving that, with proper tests, into a focused follow-up PR rather than having it reviewed here.

mnencia:
Follow-up with the replica-deletion fix and tests: #10853

PR #10850, opened 2026-06-08
test(e2e): cover backups through plugin-barman-cloud

armru:
In-core Barman Cloud support is deprecated in favour of the Barman Cloud plugin, so the e2e suite needs to cover backups taken through the plugin. The plugin follows its own release cycle, so the build to test is chosen at invocation time: the latest release, the current main, or a pinned version when reproducing an issue.

On kind and k3d, once the operator is up, the suite installs cert-manager (a plugin prerequisite) and the Barman Cloud plugin into cnpg-system, defaulting to the latest plugin release; the resolved operator and plugin versions are printed at the start of the run.

The suite then covers, through the plugin:

- a smoke round-trip: a cluster archives through the plugin, the operator reports it as loaded, and a backup is taken, restored into a new cluster via the externalClusters plugin source, and verified
- a replica cluster bootstrapped from a plugin backup, then streaming from the source
- tablespaces surviving a backup and restore round-trip
- a plugin-backed cluster across an in-place operator upgrade, with a backup taken through the plugin afterwards
- a ScheduledBackup with the immediate option, a prefer-standby backup target verified to run on a replica, and a point-in-time recovery from an external cluster definition

The in-core variants stay until the in-core support is removed; their clusters are shared with volume-snapshot scenarios that have no plugin counterpart yet. The new specs use a dedicated plugin-barman-cloud label: "plugin" already selects the kubectl-cnpg tests, whose certificate spec restarts the operator and would fail anything running alongside it.

cert-manager is now installed on every kind/k3d e2e run; its version is Renovate-tracked. The operator under test stays the local build by default.

To pick the plugin build a run validates, pass barman_plugin (alias bp) to the /test command on a pull request — /test barman_plugin=release (the default), /test bp=main, or /test bp=v0.12.0 for a pinned version — or set the barman_plugin workflow_dispatch input to the same value. Locally, the equivalent is BARMAN_PLUGIN_VERSION=main ./hack/e2e/run-e2e-local.sh.


NiccoloFei (approved):

hack/testing-tools/common/20-utils-k8s.sh
```diff
@@ -20,6 +20,14 @@
 
 # This file contains functions for interacting with the Kubernetes API using $K8S_CLI.
 
+# _github_auth_header: returns curl auth flags when GITHUB_TOKEN is set.
+# Avoids 403 rate-limiting on unauthenticated downloads from GitHub Actions.
+function _github_auth_header() {
+    if [[ -n "${GITHUB_TOKEN:-}" ]]; then
```
GabriFedi97: If this is something specific for Github Actions, should we check for the `CI` or `GITHUB_ACTIONS` env variable instead? 
```suggestion
    if [[ -n "${CI:-}" ]]; then
```

hack/testing-tools/common/20-utils-k8s.sh
```diff
@@ -229,7 +244,8 @@ function deploy_operator_from_manifest() {
     fi
 
     local manifest_file="${TEMP_DIR}/cnpg-operator-manifest.yaml"
-    if ! curl -fsSL --retry 5 --retry-delay 2 -o "${manifest_file}" "${manifest_url}"; then
+    # shellcheck disable=SC2046
+    if ! curl -fsSL --retry 5 --retry-delay 2 $(_github_auth_header) -o "${manifest_file}" "${manifest_url}"; then
```
GabriFedi97: This doesn't seem to work properly, by executing it I get the following error:

`curl: (6) Could not resolve host: Bearer`

I'd suggest to move here the "-H" option and wrap `$(_github_auth_header)` in double quotes:

```suggestion
    if ! curl -fsSL --retry 5 --retry-delay 2 -H "$(_github_auth_header)" -o "${manifest_file}" "${manifest_url}"; then
```

GabriFedi97 (approved):

mnencia (approved):

PR #10869, opened 2026-06-10
feat(security): enable FIPS 140-3 by default (#10564)

### LINKED ISSUE #10564: [Feature]: Enable FIPS 140-3 by default

Now that https://github.com/golang/go/issues/69536 has been closed and Go has the FIPS-140-3 certification, we should start building only using FIPS as the default, this will increase the security in the operator and the binaries that we produce by default and help to pass any certification process that we will require in the future for the Incubation process

Compile using [FIPS](https://go.dev/doc/security/fips140) during development and release process.

None

### Additional context

_No response_

### Backport?

Yes

No

### Code of Conduct

GellertBabel:
Set GOFIPS140=latest in goreleaser env and Makefile so all binaries are compiled with Go's native FIPS 140-3 certified crypto module.

Closes #10564


mnencia:
Thanks for the work here, @GellertBabel, and sorry to close it after you put in the effort.

After the discussion on #10564 the maintainers decided not to enable FIPS 140-3 by default in the community builds. The short version is that the operator flag does not stand alone: our community PostgreSQL images link against Debian's OpenSSL, which is not FIPS-140-3 validated, so setting `GOFIPS140` on the Go binaries alone would signal a property the full stack does not have. The reasoning is laid out in full on the issue.

Closing this in line with that decision. The effort is appreciated regardless.

PR #10874, opened 2026-06-11
feat: keep defaulting and validation working when admission webhooks are unavailable

### LINKED ISSUE #10942: [Feature]: Keep defaulting and validation working when admission webhooks are unavailable

CloudNativePG relies on admission webhooks to default and validate its resources (Cluster, Backup, ScheduledBackup, Pooler, Database). When the webhooks are not installed, or are temporarily unable to respond, resources are admitted without ever being defaulted or validated. The operator then reconciles specs that are incomplete or invalid, and the user gets no clear signal about what went wrong. The webhook is effectively a single point at which correctness is enforced, and when it is absent that enforcement disappears silently.

Defaulting and validation should keep working even when the webhook did not run. During reconciliation the operator should notice a resource that is missing its defaults and either apply them itself or wait for the webhook to do so, and it should validate the spec as a fallback. When validation fails, the reason should be surfaced on the resource status so the user can see exactly why the resource was rejected, instead of having to infer it from downstream reconcile failures.

Making the webhooks mandatory with a failing failure policy would block all writes whenever the webhook is down, including unrelated ones, and it does not help installation methods that intentionally omit the webhooks. Documenting that the webhooks are required does not actually protect a cluster whose webhook is missing or briefly unavailable. Handling defaulting and validation as a reconciliation fallback covers both cases without making the operator brittle.

### Additional context

Implemented in #10874. Related to #9278, which discusses making the webhook dependency optional.

### Backport?

Yes

Yes

### Code of Conduct

leonardoce:
When admission webhooks are not installed (or configured to ignore failures), resources previously skipped defaulting and validation entirely, leaving the operator to reconcile invalid or incomplete specs.

Resources are now defaulted and validated during reconciliation as a fallback. When a resource is missing defaults, the operator either applies them directly or waits for the webhook to do so. When validation fails, the offending fields are surfaced in the resource status, and reconciliation stops, with full details in the operator logs, so users can see what was rejected even if the webhook never ran.

Webhook serving-certificate management is also decoupled from CA-bundle injection: the new `MANAGE_WEBHOOK_CONFIGURATIONS` option (enabled by default) lets webhook configurations be managed externally, e.g., by cert-manager's CA injector or by GitOps, while the operator continues rotating the serving cert.

Closes #10942


api/v1/backup_funcs.go
```diff
@@ -322,3 +322,11 @@ func (b BackupMethod) IsManagedByInstance() bool {
 func (b BackupMethod) IsManagedByOperator() bool {
 	return b == BackupMethodVolumeSnapshot
 }
+
+// SetAdmissionError sets the admission error status on the Backup resource
+func (backup *Backup) SetAdmissionError(msg string) {
+	if len(msg) > 0 {
+		backup.Status.Phase = BackupPhaseDefinitionInvalid
+		backup.Status.Error = msg
+	}
```
GellertBabel, line 342:
```suggestion
	}  else {
		backup.Status.Error = ""
	}
```

api/v1/cluster_funcs.go
```diff
@@ -1599,3 +1599,11 @@ func (cluster *Cluster) GetServiceAccountName() string {
 	}
 	return cluster.Name
 }
+
+// SetAdmissionError sets the admission error status on the Cluster resource
+func (cluster *Cluster) SetAdmissionError(msg string) {
+	if len(msg) > 0 {
+		cluster.Status.Phase = PhaseDefinitionInvalid
+		cluster.Status.PhaseReason = msg
+	}
```
GellertBabel, line 1668:
```suggestion
	} else {
		cluster.Status.PhaseReason = ""
	}
```

GellertBabel (dismissed):
I think without the else path the system might have short-lived state where the status is not correct on the CR.

api/v1/backup_types.go
```diff
@@ -53,6 +53,9 @@ const (
 
 	// BackupPhaseWalArchivingFailing means wal archiving isn't properly working
 	BackupPhaseWalArchivingFailing = "walArchivingFailing"
+
+	// BackupPhaseDefinitionInvalid means the backup definition is invalid
+	BackupPhaseDefinitionInvalid = "invalid cluster definition"
```
maxlengdell: Should perhaps be `invalid backup definition`?

maxlengdell (approved):
Looks good, tried it locally and it works as expected. Added optional comment

mnencia (approved):

gbartolini:
With the help of Claude, I pushed three fixes for the clear correctness issues; the rest are discussion points.

Fixes pushed:

1. Completed/Failed Backups could be clobbered to invalid backup definition. The guard ran before the terminal-phase short-circuit in backup_controller.go. The Backup validator's Method == VolumeSnapshot && !HaveVolumeSnapshot() check depends on the runtime environment, not the immutable spec — so if the VolumeSnapshot CRD is later absent, a completed backup would be re-validated and flipped to BackupPhaseDefinitionInvalid, destroying the record (and since the original phase is overwritten in place, clearing the error reset it to empty rather than restoring it, potentially re-running the backup). → moved the guard after the terminal-phase switch.

2. Instance manager could stall on an invalid Cluster. With ApplyChanges: false (the operand reconciling a Cluster it doesn't own), a validation failure returned reconcile.TerminalError, which stops controller-runtime from requeueing — local PostgreSQL reconciliation could stall until an unrelated event woke it up. → requeue after a short delay instead, matching the defaulting path.

3. Invalid Cluster kept reporting Ready=True. Cluster.SetAdmissionError set the phase directly, bypassing RegisterPhase, so the readiness condition was never updated. → set Ready=False in SetAdmissionError, mirroring SetClusterReadyCondition.

Discussion points (not addressed):

4. The guard only ever calls ValidateCreate. When webhooks are down — exactly the scenario this targets — update-only invariants (validateClusterChanges: storage shrink, image major downgrade, replication-slot / replica-cluster immutability) are never enforced, and create-only checks gated on len(ResourceVersion)==0 misfire on existing objects. A real fix needs the old object for ValidateUpdate, which the loop doesn't have. Can we branch create-vs-update, or should we document the limitation explicitly?

5. Cluster is defaulted twice with conflicting semantics. The guard uses cluster.Default() (preserve user fixed params = true); the pre-existing fallback in cluster_controller.go uses SetDefaults() (= false). They're now redundant and disagree on fixed parameters. Suggest removing the old fallback and standardizing on the guard's Default() (which matches the webhook).

6. Cluster/Database never persist the cleared admission error. GetAdmissionError() returns "" for these types, so the guard relies on the phase machinery to clear the status — which can be skipped if reconciliation returns early before the first RegisterPhase, leaving a stale "Invalid definition". Coupled to 7 below.

7. The GetAdmissionError sentinel. Three of five types hardcode return "" to opt out of guard-driven clearing. Modeling this as an explicit PersistClearedError bool on AdmissionParams would remove the magic-empty-string contract and resolve 6 cleanly.

8. Cost on the hot path. The guard runs full defaulting + validation + two go-spew hashes on every reconcile, including when webhooks are installed and already admitted the object (and on every operand instance loop). Worth gating on a cheap webhook-availability signal.

9. Boilerplate. The EnsureResourceIsAdmitted block is copy-pasted across 6 reconcilers, and each NewXxxAdmissionGuard() restates the defaulter/validator pairing already in Setup*WebhookWithManager. A small helper/registry would prevent drift.

gbartolini (approved):

PR #10876, opened 2026-06-11
fix: use WebSocket for exec/attach streaming

### LINKED ISSUE #10873: [Bug]: cnpg status command to display database size is unable to upgrade connection

### Contact Details

_No response_

### Version

1.28 (latest patch)

1.33

Other

YAML manifest

### What happened?

The `kubectl cnpg status -n <my-ns> <my-cluster>` displays many characteristics of the cluster. However, the value of field `Size` reports `cmd: [du -sLh /var/lib/postgresql/data/pgdata]`. The latest patch version number (v1.28.3) prints `error: unable to upgrade connection: empty server response` as well, earlier patch version only print `error: `.

I would have expected to just see the database size, e.g, 320G.

Apparently, the exec still uses the deprecated SPDY protocol for streaming connections. The websocket protocol is the default since Kubernetes 1.31.

Note that in local setups using kind or minikube this will not pop up.

### Cluster resource

```shell

```

### Relevant log output

```shell

```

### Code of Conduct

bartscheers:
Fixes #10873

### DIFF
--- a/pkg/utils/exec.go
+++ b/pkg/utils/exec.go
@@ -202,7 +202,7 @@ func execCommandOnce(
 		Stderr:    true,
 	}, scheme.ParameterCodec)
 
-	executor, err := remotecommand.NewSPDYExecutor(&newConfig, "POST", req.URL())
+	executor, err := remotecommand.NewWebSocketExecutor(&newConfig, "POST", req.URL().String())
 	if err != nil {
 		return "", "", err
 	}


gbartolini (approved):

gbartolini:
This can only be merged once we remove support for Kubernetes 1.29 (#10881) and can only be backported to CNPG 1.29.

leonardoce (approved):

Issue #10826, opened 2026-06-03
[Feature]: Support for database-level `GRANT`/`REVOKE` privileges in the `Database` CRD


gbartolini:
### Is there an existing issue already for this feature request/idea?

- [x] I have searched for an existing issue, and could not find anything. I believe this is a new feature request to be evaluated.

### What problem is this feature going to solve? Why should it be added?

The `Database` CRD currently has no declarative way to manage privileges on the database object itself. Users who need to grant or revoke `CONNECT`, `CREATE` or `TEMPORARY` fall back to post-initialisation SQL scripts, which are imperative, one-off and fragile.

The most common real-world operation here is hardening: PostgreSQL grants `CONNECT` and `TEMPORARY` to `PUBLIC` by default, so almost every production setup needs to run `REVOKE CONNECT ON DATABASE x FROM PUBLIC` and then grant `CONNECT` back to specific roles. Today there is no first-class way to express that.

This is a natural companion to #7872 (schema-level `CREATE`/`USAGE` grants). Both address the same gap — declarative privilege management on objects already owned by the `Database` CRD — and should share design and reconciliation machinery.

### Describe the solution you'd like

PostgreSQL's database-level `GRANT` syntax:

```sql
GRANT { { CREATE | CONNECT | TEMPORARY | TEMP } [, ...] | ALL [ PRIVILEGES ] }
    ON DATABASE database_name [, ...]
    TO role_specification [, ...] [ WITH GRANT OPTION ]
    [ GRANTED BY role_specification ]
```

As with #7872, `WITH GRANT OPTION` and `GRANTED BY` are unsuitable for a declarative environment and are out of scope. We focus on `CREATE`, `CONNECT` and `TEMPORARY` (`TEMP` is an alias).

A `permissions` stanza on the `Database` spec, keyed by privilege, with each entry carrying a role `name` and a `type` of `grant` (default) or `revoke`:

```yaml
spec:
  permissions:
    connect:
    - name: public
      type: revoke
    - name: app
      type: grant
    create:
    - name: app
      type: grant
    temporary:
    - name: app
      type: revoke
```

[...]

PR #10910, opened 2026-06-15
feat(database): add `USAGE` and `CREATE` schema privilege grants

### LINKED ISSUE #7872: [Feature]: Support for `CREATE` and `USAGE` grants on database schemas

Right now, the `Database` CRD only manages the `owner` of the schema.

PostgreSQL provides two privileges as part of the `GRANT` command on schemas: `CREATE` and `USAGE`. See ["SQL Grant"](https://www.postgresql.org/docs/current/sql-grant.html).

Users are now using the post initialization scripts for this, making it imperative, one-off and thus fragile.

Providing usage permissions on managed schema objects as part of the `Database` CRD.

The `GRANT` SQL command in Postgres has the following syntax:

```sql
GRANT { { CREATE | USAGE } [, ...] | ALL [ PRIVILEGES ] }
    ON SCHEMA schema_name [, ...]
    TO role_specification [, ...] [ WITH GRANT OPTION ]
    [ GRANTED BY role_specification ]
```

Both `GRANTED BY` and `WITH GRANT OPTION` are not suitable for a declarative environment typical of CloudNativePG, so we will focus only on `CREATE` and `USAGE`.

A possible implementation could be:

```yaml
# ...
spec:
  schemas:
  - name: app
    owner: app
    create:
    - name: angus
      type: grant
    - name: brian
      type: revoke
    usage:
    - name: angus
      type: grant
    - name: brian
      type: grant
# ...
```

The default `type` is `grant`.

For example, the above will check that `angus` hasn't got yet the `create` privilege for the `app` database and run the required `GRANT CREATE ON app TO angus`. Similarly, if `brian` has `create` usage on the same `app` schema, it will run `REVOKE CREATE ON app FROM angus`.

See: [`REVOKE`](https://www.postgresql.org/docs/current/sql-revoke.html).

Please feel free to change the above proposal if you have a better idea.

We could do nothing, and delegate this operation to another tool.

However, I personally think that this is a good responsibility for the `schema` management feature in the `Database` CRD.

### Additional context

N/A

### Backport?

No

Yes

### Code of Conduct

BRGOVIND:
Adds a `permissions` field to schema entries in the `Database` CRD, letting users declaratively grant or revoke the `USAGE` and `CREATE` privileges on a schema, alongside the existing `owner`/`ensure` management:

    schemas:
      - name: app
        owner: app
        permissions:
          usage:
            - name: public
              type: revoke

Each entry reuses the existing grant/revoke model used for FDW and foreign-server usage (`name` + `type: grant|revoke`, default: `grant`), so public targets the PostgreSQL PUBLIC pseudo-role as it already does elsewhere.

Closes #7872


gbartolini:
Please see my comment here: https://github.com/cloudnative-pg/cloudnative-pg/issues/7872#issuecomment-4712780917

BRGOVIND:
Thanks @gbartolini — I've reworked the API to follow the proposed design.

Schema grants now use a `permissions` stanza keyed by privilege, reusing the existing `UsageSpec` `{name, type: grant|revoke}` model rather than the previous flat `usagePrivileges`/`createPrivileges` lists:

```yaml
schemas:
  - name: myschema
    owner: app
    permissions:
      usage:
        - name: readonly
          type: grant
      create:
        - name: app
          type: grant
```

This keeps the surface consistent with the FDW/server usage permissions and mirrors the shape proposed for database-level grants in #10826, building on the shared `applyObjectPrivilege` primitive from #10829. I've regenerated the deepcopy, CRD manifests and API reference docs, and added unit tests covering grant/revoke on schema create and update (`go build`/`go vet`/`go test` all green locally).

On the idempotent ACL-diffing point from #10826: I kept this PR on the existing fire-and-forget `applyObjectPrivilege` path to stay consistent with the currently-merged FDW/server usage behaviour. Happy to follow up with a shared read-current-ACL-and-diff helper (usable by both #7872 and #10826) as a separate change if you'd prefer that to land here instead. Let me know which you'd like.

BRGOVIND:
Gentle ping on this one 🙂 — it adds explicit `USAGE` and `CREATE` schema privilege grants for the `Database` resource. Branch is rebased and green. Is there anything you'd like me to adjust to make it easier to review?

Utkarsh-sharma47:
Hi @BRGOVIND, I was looking through the validation path for the new permissions stanza and noticed a possible consistency gap.

FDW/server usage entries currently go through validateNameOptionsUsages, which prevents the same role from appearing more than once. Unless I'm missing another validation path, the new schema permissions.usage / permissions.create lists don't seem to get an equivalent check in validateSchemas.

For example, would this currently be accepted?

permissions:
  usage:
    - name: reader
      type: grant
    - name: reader
      type: revoke

If so, both operations would be applied in list order. Would it make sense to reuse the existing duplicate-name validation independently for each privilege list?

Happy to help with the validation/tests if useful.

BRGOVIND:
> Hi @BRGOVIND, I was looking through the validation path for the new permissions stanza and noticed a possible consistency gap.
> 
> FDW/server usage entries currently go through validateNameOptionsUsages, which prevents the same role from appearing more than once. Unless I'm missing another validation path, the new schema permissions.usage / permissions.create lists don't seem to get an equivalent check in validateSchemas.
> 
> For example, would this currently be accepted?
> 
> permissions: usage: - name: reader type: grant - name: reader type: revoke
> 
> If so, both operations would be applied in list order. Would it make sense to reuse the existing duplicate-name validation independently for each privilege list?
> 
> Happy to help with the validation/tests if useful.

Thanks for taking a look!

Good catch. I don't believe the new schema permissions currently go through the same duplicate entry validation as the existing FDW server usage lists. I'll take a look at reusing the existing validation so duplicate role entry isrejected consistently for each privilege list, and I'll add/update the relevant tests.

armru:
Hello, thanks for the contribution, I pushed some changes if you could take a look

BRGOVIND:
> Hello, thanks for the contribution, I pushed some changes if you could take a look

I did; Thanks !
i there anything more you'd like me to add ?

gbartolini:
I have tested the patch using the following YAML manifest, which creates a cluster (`acdc`), a new user called `acdc` that owns a new database called `acdc`, with a schema called `acdc`. I then create another user called `angus`, which should be used to then use the schema.

```yaml
kind: Cluster
metadata:
  name: acdc
spec:
  instances: 1
  storage:
    size: 1Gi
  postgresql:
    pg_hba:
      - hostssl acdc acdc all cert
      - hostnossl acdc acdc all reject
---
kind: DatabaseRole
metadata:
  name: acdc
spec:
  cluster:
    name: acdc
  name: acdc
  login: true
  clientCertificate:
    enabled: true
  databaseRoleReclaimPolicy: retain
---
kind: DatabaseRole
metadata:
  name: angus
spec:
  cluster:
    name: acdc
  name: angus
  login: true
  clientCertificate:
    enabled: true
  databaseRoleReclaimPolicy: retain
---
kind: Database
metadata:
  name: acdc
spec:
  name: acdc
  owner: acdc
  cluster:
    name: acdc
  schemas:
  - name: public
    owner: acdc
    permissions:
      usage:
      - name: public
        type: revoke
  - name: acdc
    owner: acdc
    permissions:
      create:
      - name: acdc
        type: grant
      usage:
      - name: angus
        type: grant
      - name: public
        type: revoke
```

As you can see, a couple of interesting points:

1. I remove `USAGE` permissions from `PUBLIC` schema to `PUBLIC` (all users)
2. I grant `CREATE` privileges to the `acdc` user on the `acdc` schema
3. I grant `USAGE` privileges to the `angus` user on the `acdc` schema
4. I revoke `USAGE` privileges from `acdc` schema to `PUBLIC` (all users)

I have then created this job:

```yaml
apiVersion: batch/v1
kind: Job
metadata:
  name: acdc-schema-init
spec:
  template:
    metadata:
      name: acdc-schema-init
    spec:
      restartPolicy: OnFailure
      securityContext:
        runAsNonRoot: true
        runAsUser: 26
        fsGroup: 26
        seccompProfile:
          type: RuntimeDefault
      containers:
        - name: psql-init
          image: ghcr.io/cloudnative-pg/postgresql:18-minimal-trixie
          securityContext:
            allowPrivilegeEscalation: false
            capabilities:
              drop:
                - ALL
          resources:
            requests:
              cpu: "100m"
              memory: "64Mi"
            limits:
              cpu: "500m"
              memory: "128Mi"
          command:
            - psql
            - "postgres://acdc@acdc-rw:5432/acdc?sslmode=verify-full&sslcert=/etc/certs/tls.crt&sslkey=/etc/certs/tls.key&sslrootcert=/etc/ca/ca.crt"
            - -c
            - |
              CREATE TABLE IF NOT EXISTS acdc.i (
                n           INTEGER NOT NULL
              );
              SELECT
                  username,
                  has_schema_privilege(username, 'acdc', 'USAGE') AS can_usage_acdc,
                  has_schema_privilege(username, 'acdc', 'CREATE') AS can_create_acdc,
                  has_schema_privilege(username, 'public', 'USAGE') AS can_usage_public,
                  has_schema_privilege(username, 'public', 'CREATE') AS can_create_public
              FROM (
                  VALUES
                      ('app'),
                      ('acdc'),
                      ('angus')
              ) AS users(username);
          volumeMounts:
            - name: certs
              mountPath: /etc/certs
              readOnly: true
            - name: ca
              mountPath: /etc/ca
              readOnly: true
      volumes:
        - name: certs
          secret:
            secretName: acdc-client-cert
            defaultMode: 0640
        - name: ca
          secret:
            secretName: acdc-ca
            defaultMode: 0640
```

This is what happens:

```
$ kubectl logs jobs/acdc-schema-init
CREATE TABLE
 username | can_usage_acdc | can_create_acdc | can_usage_public | can_create_public
----------+----------------+-----------------+------------------+-------------------
 app      | f              | f               | f                | f
 acdc     | t              | t               | t                | t
 angus    | t              | f               | f                | f
```

As expected, the `app` user (automatically created by CNPG) is not allowed to enter either schema (`public` or `acdc`). The `acdc` user basically "owns" both, while the `angus` user can only use the `acdc` schema (without creating objects inside).

gbartolini:
@Utkarsh-sharma47 @BRGOVIND I have made some small changes, including documentation and testing. Let me know what you think. Thanks.

tests/e2e/fixtures/declarative_databases/database.yaml.template
```diff
@@ -26,10 +26,17 @@ spec:
...
+      # privilege held by PUBLIC satisfies has_schema_privilege() for every
+      # role, so granting both to "public" would make the assertion on the
+      # named role pass even if that grant never ran.
       permissions:
         usage:
           - name: public
             type: grant
+        create:
+          - name: app
+            type: grant
```
gbartolini, line 39: This test is useless as that permission is already present (hence why I removed it)

armru: On test_schema the CREATE grant isn't already present. The instance manager connects to the target database as postgres, and   test_schema has no owner: in the fixture, so CREATE SCHEMA runs without AUTHORIZATION and the schema is owned by postgres.                                                                      
                                                                                                                              
That would be true on public, though: it's owned by pg_database_owner and app owns the database, so app has CREATE there      
regardless. Happy to drop it if you'd rather keep the fixture minimal

BRGOVIND:
> @Utkarsh-sharma47 @BRGOVIND I have made some small changes, including documentation and testing. Let me know what you think. Thanks.

i did 
sorry for the delay
anything more i need to add ?

armru (approved):

PR #10972, opened 2026-06-20
fix(psql): support Windows execution

### LINKED ISSUE #10553: [Bug]: psql command of kubectl cnpg plugin does not work on Windows

### Contact Details

github@jockie.de

### Version

1.29 (latest patch)

1.34

Self-managed: RKE

Helm

### What happened?

When I run the psql sub-command on Windows (either CMD.EXE or PowerShell prompt) I immediately get the following error output:

`PS > kubectl cnpg psql my-cnpg-cluster`
`Error: not supported by windows`
`Error: exit status 1`

Namespace is set via kubens, so that's not the issue.
Other cnpg sub-commands work just fine.

My gut says there might be some Go command in the connection creation that is specific to Unices.
Maybe something like a file (stat, chmod) or pipe operation.

### Cluster resource

```shell
not applicable
```

### Relevant log output

```shell

```

### Code of Conduct

Utkarsh-sharma47:
The psql plugin replaced the current process with a `kubectl exec` invocation through `syscall.Exec`, which is not available on Windows. As a result `kubectl cnpg psql` failed there with "not supported by windows".

Execution is now platform-specific: Unix-like systems keep using `syscall.Exec`, while Windows launches `kubectl exec` as a child process via `exec.Command`.

Closes #10553

### DIFF
--- a/internal/cmd/plugin/psql/exec_unix.go
+++ b/internal/cmd/plugin/psql/exec_unix.go
@@ -1,0 +1,33 @@
+//go:build !windows
+
+/*
+Copyright © contributors to CloudNativePG, established as
+CloudNativePG a Series of LF Projects, LLC.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+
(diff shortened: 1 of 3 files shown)


Utkarsh-sharma47:
@gbartolini I'd appreciate a review when you have time. Thanks!

jockie:
Did compile and `kubectl cnpg psql <cluster>` now also works on Windows. Thanks for the effort @Utkarsh-sharma47 !

Utkarsh-sharma47:
Thanks for taking the time to test it on Windows and for confirming the fix works in your environment, @jockie .

I appreciate the verification. I'll wait for maintainer review and let me know if you notice any other issues related to the command behavior.

leonardoce (approved):

leonardoce:
I tested it on Windows, it works fine to me.

mnencia (approved):

PR #10975, opened 2026-06-21
fix(external): prevent stale certificate data and partial reads after secret rotation

### LINKED ISSUE #10892: [Bug]: External cluster TLS files are rewritten without O_TRUNC, leaving stale bytes after secret rotation

### Contact Details

### Version

trunk (main)

1.35

YAML manifest

## Summary

`dumpSecretKeyRefToFile` writes the TLS material (`sslcert`, `sslkey`, `sslrootcert`) for `externalClusters` entries to a stable, reused path, but opens the file with `os.O_WRONLY | os.O_CREATE` and **without `os.O_TRUNC`**. When a referenced secret is rotated to **shorter** content, the new bytes overwrite only the prefix and stale trailing bytes from the previous content remain on disk.

This was found by code inspection on the `main` branch; it is environment-independent.

## Affected code

`pkg/management/external/utils.go:113`

```go
filePath := filepath.Join(directory, fmt.Sprintf("%v_%v", selector.Name, selector.Key))
f, err := os.OpenFile(filepath.Clean(filePath), os.O_WRONLY|os.O_CREATE, 0o600)
...
_, err = f.Write(value)
```

- The path is a stable constant: `/controller/external/<serverName>/<secretName>_<secretKey>` (via `getExternalSecretsPath()`).
- It is rewritten on every reconcile of the external-servers reconciler (`internal/management/controller/externalservers/reconciler.go:79`, which watches `Cluster`).
- The function's own doc comment states its purpose is to update the file when the secret changes **without restarting the pod**, so this rewrite path is normal operation.

## Why it happens

Opening with `O_WRONLY|O_CREATE` (no `O_TRUNC`) starts writing at offset 0 but does **not** clear existing content. If the new value is shorter than what was previously written, the old tail is left in place.

## Impact

The three files flow directly into libpq as `sslcert` / `sslkey` / `sslrootcert` (`pkg/management/external/external.go:86-104`), used by replica clusters, `pg_basebackup` bootstrap, recovery-from-external-source, and logical import.

The defect is most visible with **multi-object PEM files**: a `sslrootcert` CA bundle (or a `sslcert` chain) rotated to fewer/shorter objects leaves a complete stale PEM block on disk — so a **removed CA can remain trusted**, or a partial leftover block causes a TLS load error. (For a single-object PEM shrunk shorter, the leftover usually lacks a `-----BEGIN-----` marker, so parsers often skip it — which is why this has gone unnoticed.)

## Steps to reproduce

1. Configure an `externalCluster` with `sslRootCert` pointing to a Secret holding a **2-certificate** CA bundle.
2. Let the instance reconcile (the file is written under `/controller/external/...`).
3. Rotate the Secret to a **single, shorter** certificate.
4. After the next reconcile, inspect the file inside the instance pod — the removed second certificate is still present and the file length is unchanged.

Isolated reproduction (same flags as the code path): writing a 98-byte 2-cert value, then rewriting a 49-byte 1-cert value to the same path, leaves a 98-byte file containing the removed cert.

## Expected behaviour

After a Secret rotation, the on-disk file should contain **exactly** the new secret value, with no leftover bytes.

## Suggested fix

Add `os.O_TRUNC` to the open flags:

```go
f, err := os.OpenFile(filepath.Clean(filePath), os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0o600)
```

or replace the open/write/close block with `os.WriteFile(filePath, value, 0o600)` (which truncates), matching how the sibling pgpass path is written.

## Additional context

- The buggy line was introduced together with the physical replica clusters feature (commit `f6cf8d9d`, "feat: add support for physical replica clusters") and has only been touched by a `golangci-lint` cleanup (#10008) since — it has never been fixed.
- The `pkg/management/external/` package currently has no unit tests covering this path.

## Are you willing to contribute?

Yes — I'd like to be assigned to this. My plan: add `os.O_TRUNC` (or switch to `os.WriteFile`) and add a unit test in a new `pkg/management/external/utils_test.go` that writes a longer value, rewrites a shorter value to the same path, and asserts the file content exactly equals the new value (fails on `main`, passes with the fix).

### Cluster resource

```shell
N/A — found by source-code inspection, not from a running cluster.
```

### Relevant log output

```shell
N/A — code-level defect; no runtime logs apply.
```

### Code of Conduct

Anand-240:
When an external server's secret was rotated to shorter content - for
example, a CA bundle shrinking from two certificates to one - the
previous certificate bytes could remain visible on disk, causing libpq
to reject the connection or trust an unintended CA.

The same in-place write also left a window where libpq could read a
partially written file during a streaming reconnect, leading to
transient connection failures at rotation time.
  
Both problems are fixed by writing the file atomically: the new content
is written to a temporary file and renamed into place, so readers always
see either the old or the new value, never a mix.

Closes #10892

### DIFF
--- a/pkg/management/external/suite_test.go
+++ b/pkg/management/external/suite_test.go
@@ -1,0 +1,32 @@
+/*
+Copyright © contributors to CloudNativePG, established as
+CloudNativePG a Series of LF Projects, LLC.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package external
+
+import (
+	"testing"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+func TestExternal(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "External Suite")
+}
--- a/pkg/management/external/utils.go
+++ b/pkg/management/external/utils.go
@@ -111,14 +111,9 @@ func dumpSecretKeyRefToFile(
 	}
 
 	filePath := filepath.Join(directory, fmt.Sprintf("%v_%v", selector.Name, selector.Key))
-	// Write the file atomically (temp file + rename), as the sibling pgpass
-	// writer in this package already does. This file lives on a stable path and
-	// is reused across reconciliations: an in-place write would leave stale
-	// trailing bytes (e.g. a removed certificate from a CA bundle) when the
-	// secret is rotated to shorter content, and would expose a window where a
-	// concurrent libpq read of sslcert/sslkey/sslrootcert observes a truncated
-	// or partially written file. WriteFileAtomic also fsyncs and skips the write
-	// when the content is unchanged.
+	// Write atomically: this file is reused across reconciliations and read
+	// concurrently by libpq, so an in-place rewrite could expose a partial file
+	// or leave stale bytes behind when the secret rotates to shorter content.
 	if _, err := fileutils.WriteFileAtomic(filePath, value, 0o600); err != nil {
 		return "", err
 	}
--- a/pkg/management/external/utils.go
+++ b/pkg/management/external/utils.go
@@ -25,6 +25,7 @@ import (
 	"os"
 	"path/filepath"
 
+	"github.com/cloudnative-pg/machinery/pkg/fileutils"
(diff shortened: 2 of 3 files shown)


mnencia:
Thanks for tracking this down, the bug and the rotation test are spot on.

Rather than adding `O_TRUNC`, I'd suggest fixing it the way the sibling `createPgPassFile` in this package already does: route the write through `fileutils.WriteFileAtomic`. It writes to a temp file and renames into place, so the replacement is atomic. That also closes a window the current open/write/close has (and that `O_TRUNC` keeps): between truncating and finishing the write, a concurrent libpq read of `sslcert`/`sslkey`/`sslrootcert` on a streaming reconnect can see an empty or partial file. As a bonus it fsyncs and skips the write when the content is unchanged.

Anand-240:
Thanks @mnencia, great suggestion — the atomic write is clearly better. I've switched `dumpSecretKeyRefToFile` to route through `fileutils.WriteFileAtomic`, matching the sibling pgpass writer in this package. This makes the replacement atomic, closes the truncate/write window where a concurrent libpq read could see an empty or partial file, fsyncs, and skips the write when the content is unchanged. The rotation test still passes. Updated in fffce8dc.

mnencia (approved):

leonardoce (approved):

PR #11031, opened 2026-06-24
feat(roles): introduce `roleGrants` field with role membership options

### LINKED ISSUE #10007: [Feature]: Support role membership options `ADMIN`, `INHERIT`, `SET`

I don't like when changes are made in the database with the personal role, where the applications role should have been used to have the owner set correctly, only because the user was not prompted to "become" another role, because they already had the required permissions inherited.

But I also want to have some roles where the permissions should be granted to every personal role directly.
`inherit` would have to be true for only these roles, but with CNPG I can only set it as a default for the user to be applied to all role memberships.

I'd like to have a new field in the `RoleConfiguration` that holds information about the parent roles the configured role should be added to with additional fields for each membership option as described by the official PostgreSQL documentation.

```yaml
# example taken from documentation https://cloudnative-pg.io/docs/1.28/declarative_role_management
kind: Cluster
spec:
  managed:
    roles:
    - name: dante
      ensure: present
      comment: Dante Alighieri
      login: true
      superuser: false
      inRoles:
        - pg_monitor
        - pg_signal_backend
      # new field
      roleGrants:
        - role: my_application_role
          inherit: false
          admin: false
          set: true
```
I am aware that these are the default options for `admin` and `set` - they could be omitted, I only put them here to clarify the schema I have in mind.
For `inherit`, the default is whatever is specified for the user, which would be `true` if not specified otherwise. So with my proposal, you'd be able to override this on a per-role level.

I considered to manage the grants externally and only use CNPG to set up the cluster, but I think this is a feature that could be useful to others as well.

For the schema I also considered changing the `inRoles` definition to allow for passing the options, but this would be a breaking change to the Cluster CRD. And the new field could not 

### Additional context

Looking at the implementation, the role memberships defined in `roleGrants` would not be managed when creating the user, but during `UpdateMembership`. This means if I want my user creation to fail if the role does not exist, I specify the membership in `inRoles`.
But it also means the order of `.spec.managed.roles` does not matter, because they would be created first and only after all of them are present the memberships would be reconciled.

When the same role is specified in both `inRoles` and `roleGrants`, the options specified in `roleGrants` take precedence.

Updating the options is done by simply `GRANT`-ing again with the updated options set. They will be overridden.

I have to adjust the `GetParentRoles` command to return the currently set options as well.

Please let me know your feedback to this.
I'd like to attempt implementing it and we could discuss further implementation details on the PR I'm going to open.

### Backport?

N/A

Yes

### Code of Conduct

jonasbadstuebner:
Complements the limited `inRoles` string array with a structured `roleGrants` object array in both the DatabaseRole CRD and managed roles definition to support advanced PostgreSQL membership options, `ADMIN`, `INHERIT` and `SET`.

**Why this change?**
The `inRoles` option only allows basic role membership. `roleGrants` provides precise control for environments requiring specific PostgreSQL role behaviors, allowing explicit toggling of `ADMIN`, `INHERIT`, and `SET` membership options per granted role.

Behavior & Compatibility:
- `inRoles` stays supported.
- If both are provided, `roleGrants` options takes precedence, but new roles are still created with the `IN ROLE` option
- If only `roleGrants` is provided, the role is created first without any memberships and the `GRANT` statements with granular options are run right after creation.
- Default behavior for `roleGrants` matches PostgreSQL defaults if flags are omitted/only the name is given.

**Why not a Grant CRD?**
Choosing to embed `roleGrants` into existing definitions rather than building a broad Grant CRD enforces a strict architectural boundary: the CNPG operator manages cluster-level role membership down to the database, while fine-grained permissions of objects inside databases remain the responsibility of application migrations.

Closes #10007


internal/management/controller/roles/contract_test.go
```diff
@@ -44,18 +45,35 @@ var _ = Describe("DatabaseRole implementation test", func() {
...
 			Name:    "abc",
 			Inherit: true,
 			InRoles: []string{
-				"role1", "Userrole2", "Testxxx",
+				"Userrole2", "Testxxx",
+			},
+			RoleGrants: []DatabaseRoleGrant{
+				{
+					Name:    "role1",
+					Admin:   ptr.To(true),
```
marvin-roesch, line 58: Go 1.26 allows `new(true)`, so using the `ptr` package no longer is required.

internal/management/controller/roles/postgres.go
```diff
@@ -44,13 +44,24 @@ func List(ctx context.Context, db *sql.DB) ([]DatabaseRole, error) {
...
-			FROM pg_catalog.pg_auth_members GROUP BY member
+			SELECT
+				pg_catalog.array_agg(
+						pg_catalog.pg_get_userbyid(am.roleid) ||
+						'|' ||
+						am.admin_option ||
+						'|' ||
+						am.inherit_option ||
+						'|' ||
+						am.set_option
```
marvin-roesch: Rather than converting to strings and then parsing them afterwards, this can use pgx's built-in capabilities for scanning nested records as structs:
```suggestion
						ROW(
							pg_catalog.pg_get_userbyid(am.roleid),
							am.admin_option,
							am.inherit_option,
							am.set_option
						)
```
Then this can be directly scanned to `&role.RoleGrants`. Thus, `PopulateRoleGrantsFromStringArray` becomes obsolete.

internal/management/controller/roles/contract.go
```diff
@@ -73,17 +143,44 @@ func (d *DatabaseRole) hasSameCommentAs(inSpec apiv1.RoleConfiguration) bool {
 }
 
 func (d *DatabaseRole) isInSameRolesAs(inSpec apiv1.RoleConfiguration) bool {
```
marvin-roesch: This should probably renamed to something like `hasMatchingRoleGrants`.

api/v1/cluster_types.go
```diff
@@ -2657,6 +2657,13 @@ type PluginStatus struct {
 	Status string `json:"status,omitempty"`
 }
 
+type RoleGrant struct {
+	Name    string `json:"name"`
+	Admin   *bool  `json:"admin,omitempty"`
+	Inherit *bool  `json:"inherit,omitempty"`
+	Set     *bool  `json:"set,omitempty"`
```
marvin-roesch, line 2731: The semantics of these values being optional should be documented. The way things currently are implemented, AFAICT, omitting them leads to reconciliation not touching them at all. I think that is a sensible behavior, but definitely needs to be documented. This would mean that setting e.g. `admin` to `true` and then simply removing it from the spec (instead of changing to `false`) would keep the role grant with the admin option.

Alternatively, possible semantics could be that omitting the value in the spec forces a value to have the Postgres default value. That might be more in-line with what users of the CRD could expect, but it convolutes the implementation some.

marvin-roesch (commented):
Some minor code improvements to point out and one larger question about optional field semantics.

internal/management/controller/roles/contract_test.go
jonasbadstuebner: A quick grep says there are over 400 occurrences of `ptr.To(...)` in the code base and only a handful of `new(...)`.
I wanted to keep the general style of dealing with bool pointers, but if the maintainers agree with you, @marvin-roesch, I can easily replace `ptr.To` with `new` for this newly added code.

api/v1/cluster_types.go
jonasbadstuebner: Addressed in the latest commit, please let me know if the description is roughly what you imagined and if all important information is present.

internal/management/controller/roles/contract.go
jonasbadstuebner: You're right, I renamed it - the function is not only comparing the roles a user is in but also the membership options.

internal/management/controller/roles/postgres.go
jonasbadstuebner: I did not succeed with scanning it directly into the `RoleGrants` field due to a different driver interface being used.
But I got rid of the `PopulateRoleGrantsFromStringArray` function by using JSON.

Unfortunately I cannot do the same for the E2E tests, where the query is executed remotely on a node of the cluster and the result is only returned as a string.

api/v1/cluster_types.go
marvin-roesch: Looks good to me :+1:

PR #11077, opened 2026-06-29
fix(webhook): reject slotPrefix that overflows the replication slot n…

### LINKED ISSUE #11072: [Bug]: Unvalidated slotPrefix length can push HA replication slot names past PostgreSQL's 63-char limit

### Contact Details

### Version

1.30.0-rc1

1.36

YAML manifest

## Summary

The cluster name is validated to a maximum of 50 characters so that derived object names stay within PostgreSQL limits, but the user-configurable `spec.replicationSlots.highAvailability.slotPrefix` is validated **only for allowed characters, not for length**. HA replication slot names are built as `<slotPrefix><instanceName>` with no truncation, so a long prefix combined with a long cluster name produces a slot name longer than PostgreSQL's 63-character limit. The webhook accepts the spec, but PostgreSQL then rejects the slot at runtime, so the HA replication slots are never created.

Found by source-code inspection on `main`; environment-independent.

## Root cause

`slotPrefix` validates characters but not length — `api/v1/cluster_types.go:1365-1367`:

```go
// +kubebuilder:validation:Pattern=^[0-9a-z_]*$
// +optional
SlotPrefix string `json:"slotPrefix,omitempty"`
```

The slot name is built by concatenation with no length cap — `api/v1/cluster_funcs.go:307`:

```go
slotName := fmt.Sprintf("%s%s", r.GetSlotPrefix(), instanceName)
sanitizedName := slotNameNegativeRegex.ReplaceAllString(strings.ToLower(slotName), "_")
return sanitizedName // never truncated to PostgreSQL's NAMEDATALEN-1 (63)
```

The webhook's `validateReplicationSlots` (`internal/webhook/v1/cluster_webhook.go:2311-2338`) checks only the SynchronizeReplicas regex, never the resulting slot-name length. By contrast the cluster name *is* capped at 50 (`cluster_webhook.go:1849`) and recovery target names *are* checked against 64 (`cluster_webhook.go:1460`) — but the slot prefix is not. The slot is created at runtime in `internal/management/controller/slots/infrastructure/postgresmanager.go:98` via `pg_create_physical_replication_slot`, where PostgreSQL enforces `NAMEDATALEN` (64) and rejects names of 64+ bytes.

## Expected behaviour

The webhook should reject a `slotPrefix` whose combination with the longest possible instance name would exceed 63 characters, failing fast at admission with a clear message, instead of accepting a spec that silently breaks HA replication slots at runtime.

## Steps to reproduce

1. Create a Cluster with a long-ish name (~45 chars) and a long `slotPrefix` (see Cluster resource below).
2. The Cluster is admitted — the webhook passes.
3. In the instance-manager logs, HA replication slot creation fails because the generated slot name `<slotPrefix><cluster>-<serial>` exceeds 63 chars (`replication slot name "..." is too long`). The HA slots are never created.

## Impact

A spec accepted by the operator silently fails to deliver HA replication slots — the feature meant to protect replicas from losing required WAL. There is no admission-time error; the failure only shows up as runtime log noise, and replicas run without the HA slots they were configured to use.

## Suggested fix

In `validateReplicationSlots` (`internal/webhook/v1/cluster_webhook.go:2311`), validate that `len(slotPrefix) + maxInstanceNameLength <= 63` (the cluster name is already capped at 50, plus the `-` and the serial suffix) and return a `field.Invalid` on the `slotPrefix` path otherwise, mirroring the existing length checks used for the cluster name (line 1849) and recovery target name (line 1460).

### Cluster resource

```shell
kind: Cluster
metadata:
  name: a-fairly-long-cluster-name-used-for-this-repro
spec:
  instances: 2
  storage:
    size: 1Gi
  replicationSlots:
    highAvailability:
      enabled: true
      slotPrefix: "my_very_long_custom_slot_prefix_here_"
```

### Relevant log output

```shell
N/A — found by source-code inspection. At runtime PostgreSQL rejects the slot with: replication slot name "<slotPrefix><cluster>-<serial>" is too long
```

### Code of Conduct

Anand-240:
…ame length

The Cluster name is capped at 50 characters so derived object names stay within PostgreSQL limits, but spec.replicationSlots.highAvailability.slotPrefix was only validated for allowed characters, not for length. HA replication slot names are built as "<slotPrefix><instanceName>" with no truncation, so a long prefix produced a slot name longer than PostgreSQL's 63-character limit (NAMEDATALEN-1). The webhook accepted the spec, but pg_create_physical_replication_slot rejected the name at runtime, so the HA slots were never created.

Validate in validateReplicationSlots that the longest slot name the cluster can generate (using the highest assignable node serial, which equals the instance count) stays within the limit, and reject the slotPrefix otherwise. Export the existing PostgresIdentifierMaxLen constant as the single source for the 63-char bound. Add unit tests covering overflow, the default prefix at the maximum cluster-name length, and the HA-disabled no-op case.

Closes #11072

### DIFF
--- a/internal/webhook/v1/cluster_webhook.go
+++ b/internal/webhook/v1/cluster_webhook.go
@@ -2325,16 +2325,39 @@ func (v *ClusterCustomValidator) validateReplicationSlots(r *apiv1.Cluster) fiel
 		return nil
 	}
 
-	if err := r.Spec.ReplicationSlots.SynchronizeReplicas.ValidateRegex(); err != nil {
-		return field.ErrorList{
-			field.Invalid(
-				field.NewPath("spec", "replicationSlots", "synchronizeReplicas", "excludePatterns"),
-				err,
-				"Cannot configure synchronizeReplicas. Invalid regexes were found"),
-		}
+	var result field.ErrorList
+
+	if err := replicationSlots.SynchronizeReplicas.ValidateRegex(); err != nil {
+		result = append(result, field.Invalid(
+			field.NewPath("spec", "replicationSlots", "synchronizeReplicas", "excludePatterns"),
+			err,
+			"Cannot configure synchronizeReplicas. Invalid regexes were found"))
 	}
 
-	return nil
+	// When HA replication slots are enabled the operator derives a slot name for
+	// every instance as "<slotPrefix><instanceName>". PostgreSQL rejects
+	// replication slot names longer than NAMEDATALEN-1 (63) characters, so reject
+	// at admission a slotPrefix that would push the longest possible slot name
+	// past that limit. The highest node serial a cluster can assign equals its
+	// number of instances: serials are kept contiguous because new instances
+	// fill the lowest free serial (generateNodeSerial) and scale-down always
+	// removes the highest serial (findDeletableInstance). The instance with the
+	// highest serial therefore has the longest instance name, and so the longest
+	// slot name, the cluster can produce. GetSlotNameFromInstanceName returns an
+	// empty string when HA slots are disabled, so this is a no-op in that case.
+	longestInstanceName := specs.GetInstanceName(r.Name, r.Spec.Instances)
+	if slotName := r.GetSlotNameFromInstanceName(longestInstanceName); len(slotName) > postgres.PostgresIdentifierMaxLen {
+		result = append(result, field.Invalid(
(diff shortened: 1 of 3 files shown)


Anand-240:
Just a gentle follow-up on this one. It's been ready for about a week now with all the automated checks passing and the fix covered by tests. Would really appreciate a review whenever someone gets a chance, and I'm happy to adjust anything if needed. Thanks!

PR #11089, opened 2026-06-29
feat(logs): propagate OTel log endpoint to instance pods

### LINKED ISSUE #11085: [Feature]: Support Otel Log streaming w/o daemonset

Currently, collecting logs from CloudNativePG (operator, instance manager, and PostgreSQL) requires an external log collection infrastructure — either a DaemonSet (e.g. Fluent
Bit, Fluentd) reading container stdout from the node filesystem, or a sidecar tailing log files.

This is problematic because:

1. CNPG's architecture makes external log capture difficult. The operator generates operand pod specs and injects its own instance manager binary via an init container. The
instance manager acts as a custom process manager — PostgreSQL runs as a child process and writes to a CSV FIFO (not stdout). The instance manager reads, parses, and re-emits
these logs through Go's logr interface. There is no shell entrypoint to hook into for stdout redirection.

2. DaemonSets violate Kubernetes best practices for workload portability. The [Red Hat Best Practices for Kubernetes](https://redhat-best-practices-for-k8s.github.io/guide/#k8s-best-practices-no-naked-pods) recommend that applications should not depend on node-level infrastructure or assume DaemonSets are present. Log collection that relies on a
DaemonSet tailing node-local files creates a hard dependency on cluster-wide infrastructure outside the application's control. This makes CNPG non-portable to environments where
DaemonSets are restricted (multi-tenant clusters, managed Kubernetes offerings with locked-down node access, edge deployments, or clusters with strict pod security policies).

3. Sidecar log tailers can't work. CNPG doesn't write to a shared file volume — logs flow through an in-process Go logger to stdout. There is no file for a sidecar to tail. The
LogShipper pattern (Fluent Bit sidecar reading from /logs/) is incompatible because CNPG manages its own pod specs and uses a different log format.

Adding native OpenTelemetry log export solves all of these: logs push directly from the process to an OTel Collector via OTLP gRPC, with full structure preserved, no DaemonSet
dependency, no sidecar, and no modification to the container image or entrypoint. The application becomes self-contained for observability — portable across any Kubernetes
environment regardless of what node-level infrastructure is available.

When configured with an OpenTelemetry Collector endpoint, CNPG natively pushes all its logs — from the operator, the instance manager, and PostgreSQL — directly to the collector via OTLP gRPC. Logs
continue to appear on stdout as they do today; the OTel export is additive.

This covers all three log sources without requiring any external infrastructure (no DaemonSet, no sidecar, no file volume). PostgreSQL log fields (severity, SQL state, query text, session ID, etc.) are
preserved as structured attributes in the exported log records rather than being flattened into a JSON string.

The feature is opt-in via a single environment variable or CLI flag, and has no effect on existing behavior when not configured.

1. DaemonSet log collector (Fluent Bit / Fluentd): Captures stdout from node-local files. Works but requires cluster-wide infrastructure, adds per-node resource overhead, and is not viable in environments that restrict DaemonSets.

2. LogShipper sidecar (Fluent Bit reading from shared file volume): Incompatible with CNPG. The operator controls pod specs and uses a custom process manager — there is no shared log file volume to tail.

3. Stdout redirect / wrapper script: Not possible. The operand container entrypoint is the instance manager binary injected from the operator image via init container. There is no shell, no
docker-entrypoint.sh, and no interception point between the Go process and stdout.

4. OTel Collector sidecar with filelog receiver: Would require shareProcessNamespace: true to read the main container's stdout from /proc/1/fd/1, or writing logs to a shared volume — neither of which CNPG
supports. Still involves re-parsing JSON that was already structured in-process.

5. Standalone Deployment streaming logs via Kubernetes API: A separate deployment that calls the K8s log API (kubectl logs --follow) for each CNPG pod. Functional but adds API server load and introduces
latency.

### Additional context

The extension would be to the cloudnative-pg/machinery repo where the [OTel Logs API](https://opentelemetry.io/docs/specs/otel/logs/api/) would be extended. 

The functionality could be provided as a experimental feature like: [traefik](https://doc.traefik.io/traefik/v3.4/observability/logs/#opentelemetry)

### Backport?

No

Yes

### Code of Conduct

maxlengdell:
Pass CNPG_LOG_OTEL_ENDPOINT from the operator environment to operand pods so that instance managers also export logs directly to the configured OpenTelemetry Collector.

Depends on: cloudnative-pg/machinery#290
Closes: #11085

## How to test manually

1. Deploy an OTel Collector with debug exporter:

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: otel-collector-config
  namespace: observability
data:
  config.yaml: |
    receivers:
      otlp:
        protocols:
          grpc:
            endpoint: 0.0.0.0:4317
    exporters:
      debug:
        verbosity: detailed
    service:
      pipelines:
        logs:
          receivers: [otlp]
          exporters: [debug]
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: otel-collector
  namespace: observability
spec:
  replicas: 1
  selector:
    matchLabels:
      app: otel-collector
  template:
    metadata:
      labels:
        app: otel-collector
    spec:
      containers:
        - name: collector
          image: otel/opentelemetry-collector-contrib:0.127.0
          args: ["--config=/etc/otelcol/config.yaml"]
          ports:
            - containerPort: 4317
          volumeMounts:
            - name: config
              mountPath: /etc/otelcol
      volumes:
        - name: config
          configMap:
            name: otel-collector-config
---
apiVersion: v1
kind: Service
metadata:
  name: otel-collector
  namespace: observability
spec:
  selector:
    app: otel-collector
  ports:
    - port: 4317
      targetPort: 4317
```

2. Set the endpoint on the operator:

```bash
kubectl -n cnpg-system set env deployment/cnpg-controller-manager \
  CNPG_LOG_OTEL_ENDPOINT=otel-collector.observability.svc:4317
```

3. Create a cluster and generate a PG error:

```bash
kubectl exec <cluster-pod> -- psql -U postgres -c "SELECT 1/0;"
```

4. Verify logs in the collector:

```bash
# PostgreSQL logs
kubectl -n observability logs deployment/otel-collector | grep "division by zero"

# Instance manager logs
kubectl -n observability logs deployment/otel-collector | grep "instance-manager"

# Operator logs
kubectl -n observability logs deployment/otel-collector | grep "Reconcil"
```


go.mod
```diff
@@ -110,15 +130,18 @@ require (
...
-	k8s.io/kube-openapi v0.0.0-20260603220949-865597e52e25 // indirect
+	k8s.io/kube-openapi v0.0.0-20260502001324-b7f5293f4787 // indirect
 	sigs.k8s.io/json v0.0.0-20250730193827-2d320260d730 // indirect
 	sigs.k8s.io/kustomize/api v0.21.1 // indirect
 	sigs.k8s.io/kustomize/kyaml v0.21.1 // indirect
 	sigs.k8s.io/randfill v1.0.0 // indirect
 	sigs.k8s.io/structured-merge-diff/v6 v6.4.0 // indirect
 )
+
+replace github.com/cloudnative-pg/machinery => ../machinery
```
maxlengdell, line 147: This is temporary to make it testable in case you checkout the corresponding machinery PR in neighbouring folder. Will be removed in final version

smiyc:
@mnencia can you put this on the Roadmap for 1.31?

PR #11109, opened 2026-07-01
fix(logpipe): prevent pgaudit records being silently dropped by zap sampler

### LINKED ISSUE #10815: [Bug]: PG audit logs are dropped

### Contact Details

will@murnane.family

### Version

1.29 (latest patch)

1.35

Cloud: Amazon EKS

Helm

### What happened?

I configured a Cluster with postgreSQL auditing enabled with pgaudit. Then I ran a pair of queries: one that produces a lot of auditing output quickly, and one that's got a nice distinctive error message.
```bash
kubectl exec -it mycluster-0 -- bash -c 'psql "\df *.*"; psql "CREAT TABLE foo(id int)"'
```
Meanwhile, I watched the logs from the cluster in another terminal:
```bash
kubectl logs -f mycluster-0 | grep 'CREAT '
```
and the log message never came through.

I believe this is due to a few factors:
* [CNPG logs `"msg": "record"` on every record it outputs](https://github.com/cloudnative-pg/cloudnative-pg/blob/7d4c9a3f253084837108306bece943840459e9c5/pkg/management/postgres/logpipe/writer.go#L27-L41) [via this function](https://github.com/cloudnative-pg/machinery/blob/main/pkg/log/log.go#L132).
* [controller-runtime defaults to adding a sampler](https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/log/zap/zap.go#L204) to its logger
* [A sampler discards messages which contain the same message](https://github.com/uber-go/zap/blob/master/zapcore/sampler.go#L127) if they occur too often within a specified timeframe
* CNPG uses the controller-runtime logger for all of its output.

This is not an ideal outcome, audit logs should not be dropped. However, I'm not sure of the right way to fix it; it might be best for the controller to determine "this cluster's logs should not be sampled" or maybe the instance manager should do so, maybe it should be communicated from the controller to the instance manager with a flag or an environment variable, maybe audit logs should always be sent through an un-sampled logger or maybe all Postgres logs should or maybe it should always be up to an administrator, and so on.

I'm happy to implement the code to make the changes if desired, I came up with a dirty hack that makes the audit logs work properly for my particular circumstances and could adapt it if that would be helpful, but the questions above (and probably many more) made me think twice about just starting a pull request that makes a bunch of decisions up front.

### Cluster resource

```shell

```

### Relevant log output

```shell

```

### Code of Conduct

akashbhujbalwebsite:
Fixes a silent audit-log loss in `LogRecordWriter`. Every PostgreSQL audit record it emits shares the same log message key (`"record"`), so the zap sampler controller-runtime installs in production mode (passing the first 100 identical messages per second, then roughly 1 in 100 after that) treats a burst of audit activity as duplicates and drops most of it, including security-relevant events such as DDL errors or privilege escalations, with no indication of the gap.

`LogRecordWriter` now writes through a dedicated logger with sampling disabled, built once when the log pipe starts and reused across every reconnect. That logger otherwise mirrors the global instance-manager logger for Info-level output: the same JSON/RFC3339Nano encoding, the same effective `--log-level`, and the same `--log-field-level`/`--log-field-timestamp` remapping, so audit records only differ from the rest of the pod's log lines in that they're never sampler-dropped.

Fixes #10815

----

**Root cause**

```go
// writer.go (before)
const logRecordKey = "record" // constant — never varies

func (writer *LogRecordWriter) Write(record NamedRecord) {
    log.WithName(record.GetName()).Info(logRecordKey, logRecordKey, record)
    //                                  ^^^^^^^^^^^^
    //             every audit record uses msg="record" → sampler treats all as duplicates
}
```

controller-runtime v0.24.1 adds this sampler automatically when `Development=false`:

```go
zap.WrapCore(func(core zapcore.Core) zapcore.Core {
    return zapcore.NewSamplerWithOptions(core, time.Second, 100, 100)
})
```

That's fine for operator reconcile-loop noise. It's not fine for audit records, where every entry represents a distinct real-world database event.

**How to verify**

1. Deploy a cluster with `pgaudit.log: all` and `shared_preload_libraries: pgaudit`
2. Fire a burst of 500+ SQL statements within a single second:
   ```sql
   DO $$ DECLARE i INT; BEGIN
     FOR i IN 1..500 LOOP EXECUTE 'SELECT ' || i; END LOOP;
   END $$;
   CREAT TABLE marker_should_appear(id int); -- intentional typo
   ```
3. Check pod logs: `kubectl logs <pod> -c postgres | grep marker_should_appear`
   - **Before fix**: 0 results (marker event dropped by sampler)
   - **After fix**: 1 result (all records forwarded)

**Tests**

`writer_test.go` calls `buildUnsampledLogger` directly: one spec floods it with a burst and confirms every record is forwarded, another confirms it honors the effective level of the global instance-manager logger, and two more cover `applyFieldRemap`'s key-renaming logic. A separate spec demonstrates the underlying zap sampler dropping records in isolation (not through CNPG code) to document why the bug is real.

Add to the release notes:

Fixed an issue where a burst of PostgreSQL audit activity (`pgaudit`) could cause most audit records to be silently dropped from the pod's log stream due to log sampling.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -14,6 +14,7 @@ require (
 	github.com/davecgh/go-spew v1.1.2-0.20180830191138-d8f796af33cc
 	github.com/evanphx/json-patch/v5 v5.9.11
 	github.com/go-logr/logr v1.4.3
+	github.com/go-logr/zapr v1.3.0
 	github.com/grpc-ecosystem/go-grpc-middleware/v2 v2.3.3
 	github.com/jackc/pgx/v5 v5.10.0
 	github.com/jackc/puddle/v2 v2.2.2
@@ -57,7 +58,6 @@ require (
 	github.com/fsnotify/fsnotify v1.9.0 // indirect
 	github.com/fxamacker/cbor/v2 v2.9.2 // indirect
 	github.com/go-errors/errors v1.5.1 // indirect
-	github.com/go-logr/zapr v1.3.0 // indirect
 	github.com/go-openapi/jsonpointer v0.23.1 // indirect
 	github.com/go-openapi/jsonreference v0.21.5 // indirect
 	github.com/go-openapi/swag v0.26.0 // indirect
--- a/pkg/management/postgres/logpipe/logpipe.go
+++ b/pkg/management/postgres/logpipe/logpipe.go
@@ -130,8 +130,8 @@ func (p *LogPipe) Start(ctx context.Context) error {
 }
 
 // collectLogsFromFile opens (blocking) the FIFO file, then starts reading the csv file line by line
-// until the end of the file or an error. recordLogger is the unsampled logger
-// built once by Start, reused across every retry.
+// until the end of the file or an error. recordLogger is built once by Start
+// and passed through.
 func (p *LogPipe) collectLogsFromFile(ctx context.Context, recordLogger logr.Logger) error {
 	filenameLog := log.FromContext(ctx).WithValues("fileName", p.fileName)
 
--- a/pkg/management/postgres/logpipe/logpipe.go
+++ b/pkg/management/postgres/logpipe/logpipe.go
@@ -35,6 +35,8 @@ import (
 	"github.com/cloudnative-pg/machinery/pkg/fileutils"
 	"github.com/cloudnative-pg/machinery/pkg/fileutils/compatibility"
 	"github.com/cloudnative-pg/machinery/pkg/log"
+	"github.com/go-logr/logr"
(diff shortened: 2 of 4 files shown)


mnencia:
Closing this in favor of #11131, which fixes the same issue at the process level rather than inside the log pipe. Credit to you and to @willmurnane, whose diagnosis in #10815 identified the sampler mechanism from the start, it's cited directly in #11131's description and is what made that fix possible. Thanks for digging into this.

PR #11115, opened 2026-07-02
feat: disable automountServiceAccountToken on instance pods and jobs

### LINKED ISSUE #7896: [Feature]: Configurable 'automountServiceAccountToken', extra 'volume', extra 'volumeMount' for Cluster Pods and Jobs.

In some installations with high security standards, the requirement to disable automatic service account token mounting is starting to appear.
_A pod that does not have automountServiceAccountToken: false set is not created/executed at all_.
_automountServiceAccountToken: false for the ServiceAccount is not sufficient solution._
In the case of Cloudnative PG this can only be done for the Deployment with operator, using Helm Chart. For dynamically created Pods with instances and Jobs, no solution has been found other than a modification in the operator code.

Configure these things in CRD kind Cluster, via the new podTemplate section.

```
kind: Cluster
spec:
  podTemplate:
    spec:
      automountServiceAccountToken: false
      volumes:
        - name: kube-api-access
          projected:
            defaultMode: 0o444
            sources:
              - serviceAccountToken:
                  expirationSeconds: 3607
                  path: token
              - configMap:
                  items:
                    - key: ca.crt
                      path: ca.crt
                  name: kube-root-ca.crt
              - downwardAPI:
                  items:
                    - fieldRef:
                        apiVersion: v1
                        fieldPath: metadata.namespace
                      path: namespace
      volumeMounts:
        - mountPath: /var/run/secrets/kubernetes.io/serviceaccount
          name: kube-api-access
          readOnly: true
          recursiveReadOnly: Disabled
```
Content of volumes and volumeMounts should be merged into resulting Pod specification.
The change must be reflected in **all containers in both Pod and Job.**

I've tried setting it up through Helm Charter, but I couldn't find a workable way.

### Additional context

_No response_

### Backport?

No

No

### Code of Conduct

armru:
The operator now unconditionally sets automountServiceAccountToken:
false on instance Pods, instance creation Jobs (including major
upgrade), and the ServiceAccount it generates, so clusters comply out
of the box with admission policies that forbid automatic mounting of
the ServiceAccount token. ServiceAccounts supplied by the user via
serviceAccountName are left untouched.

Since the instance manager still needs access to the Kubernetes API,
the operator projects an explicit kube-api-access volume at the
canonical path on every container that needs it, mirroring what the
ServiceAccount admission controller would have injected. The
bootstrap init container, which only copies the manager binary, does
not receive this mount.

Clusters upgrading from a previous release will see a one-time
rolling update, since their existing Pods don't have this field set;
this is expected. ServiceAccount drift is corrected on the next
Cluster reconcile rather than immediately.

Closes #7896


leonardoce:
The last commit removes `.spec.automountServiceAccountToken` and modifies the operator to mount a service account token only where needed within containers.
This means that updating the operator will trigger a pod rollout, even when in-place updates are enabled.

If we wish to avoid this behavior, we can simply disable the rollout for this change and defer it until the next time it’s actually required.
If we’re okay with adding this new field to the API, then dropping the latest commit is sufficient.

gabriele-wolfox (approved):

leonardoce (approved):

PR #11130, opened 2026-07-04
fix(rollout): compare the running configuration hash, not the on-disk one

### LINKED ISSUE #11129: [Bug]: restart-requiring parameter change with primaryUpdateMethod: switchover can trigger a double switchover that leaves the primary unchanged

### Contact Details

_No response_

### Version

trunk (main)

1.33

Cloud: Amazon EKS

YAML manifest

### What happened?

This is a recurrence of #6956, which was fixed by #6991. The fix does not fully close the window, and the double switchover reappeared on the `main` nightly e2e (`configuration_update`, "performing a rolling update when a GUC requiring restart is modified").

On a 3-instance cluster with `primaryUpdateMethod: switchover`, changing a parameter that requires a restart (`shared_buffers`) triggers two switchovers in quick succession that cancel out, leaving `status.currentPrimary` on the original pod. Same second:

- `03:39:04.100` primary `-1` reloads, reports the pending restart
- `03:39:04.194` / `.200` replicas `-2` / `-3` reload, report the pending restart
- `03:39:04.215` operator status snapshot: `-1 pendingRestart=true`, `-2`/`-3 pendingRestart=false` (all `Ready`, no errors)
- `03:39:04` switchover `-1 -> -2`; `-2` is promoted in place without a restart, still running the old `shared_buffers`
- `03:39:35` the new primary `-2` now reports the pending restart, switchover back `-2 -> -1`
- `currentPrimary` ends on `-1`, the original primary; no further switchovers

**Expected:** a single switchover to a replica that has already applied the change, so the primary role moves once and stays.

Why the #6991 gate does not catch it: the "delay rollout until every Pod received the same configuration" check keys on `LoadedConfigurationHash`, read from `pg_show_all_file_settings()` (`pkg/management/postgres/probes.go`), i.e. the hash of the `postgresql.conf` on disk. That goes uniform as soon as the operator distributes the config files, before the instances reload them. `pending_restart` (from `pg_settings`) only flips after each instance actually reloads. In the window between "files distributed" and "replicas reloaded", the gate sees a uniform config and proceeds while the replicas still report `pendingRestart: false`, and `updatePrimaryPod` promotes a replica that needs the same restart. Same race as #6956, at ~90ms under cloud latency instead of the artificial one-minute sleep from the original repro. The relevant code is unchanged across the current release branches (1.28, 1.29, 1.30) and `main`.

Impact is unchanged from #6956: self-healing (the parameter is applied, no data loss), but an extra failover and a brief window with a primary running stale configuration.

### Cluster resource

```yaml
kind: Cluster
metadata:
  name: pg-configuration-update
spec:
  instances: 3
  primaryUpdateMethod: switchover
  storage:
    size: 1Gi
  postgresql:
    parameters:
      shared_buffers: 128MB   # then changed to 256MB to reproduce
```

### Relevant log output

```
03:39:04.100  pg-configuration-update-1  parameter "shared_buffers" cannot be changed without restarting the server
03:39:04.194  pg-configuration-update-2  parameter "shared_buffers" cannot be changed without restarting the server
03:39:04.200  pg-configuration-update-3  parameter "shared_buffers" cannot be changed without restarting the server

03:39:04.215  pod status  name=pg-configuration-update-1 isPrimary=true  isPodReady=true pendingRestart=true
03:39:04.215  pod status  name=pg-configuration-update-2 isPrimary=false isPodReady=true pendingRestart=false
03:39:04.215  pod status  name=pg-configuration-update-3 isPrimary=false isPodReady=true pendingRestart=false

03:39:04  The primary needs to be restarted, we'll trigger a switchover to do that  currentPrimary=pg-configuration-update-1 targetPrimary=pg-configuration-update-2
03:39:06  The PostgreSQL instance has been promoted successfully  (pg-configuration-update-2, same postmaster session, no restart)

03:39:35  The primary needs to be restarted, we'll trigger a switchover to do that  currentPrimary=pg-configuration-update-2 targetPrimary=pg-configuration-update-1
03:39:41  Setting primary label  pod=pg-configuration-update-1
```

### Code of Conduct

mnencia:
The gate that delays a rolling update until every instance reports the same configuration compared the hash from `pg_show_all_file_settings()`, which reflects the configuration file on disk and updates as soon as the operator writes it, before the instance reloads it. `PendingRestart`, in contrast, only changes after the reload. In the window between writing the configuration and reloading it, the gate saw a uniform configuration while the replicas had not reloaded yet, so changing a parameter that requires a restart could switch over to a replica that still needed the same restart, causing a second switchover that left the primary unchanged.

Read the hash with `current_setting()` so it reflects the configuration PostgreSQL has actually loaded. That value and `PendingRestart` are then produced by the same reload and stay consistent within a single probe.

Closes #11129


armru (approved):

tests/e2e/configuration_update_test.go
```diff
@@ -347,6 +350,17 @@ var _ = Describe("Configuration update", Label(tests.LabelClusterMetadata), func
...
+			By("verifying that a single switchover happened", func() {
+				// Every promotion advances the PostgreSQL timeline by one.
+				// Promoting a replica whose restart was still pending would
+				// force a second switchover and leave the timeline at
+				// oldTimeline+2 (issue #11129).
+				Eventually(func() (int, error) {
+					cluster, err := clusterutils.Get(env.Ctx, env.Client, namespace, clusterName)
+					return cluster.Status.TimelineID, err
+				}, RetryTimeout).Should(Equal(oldTimeline + 1))
+			})
```
armru, line 363: this could be excluded from the backport if needed

armru (approved):

leonardoce (approved):

PR #11131, opened 2026-07-06
fix(logging): stop sampling logs of the pods' processes

### LINKED ISSUE #10815: [Bug]: PG audit logs are dropped

armru:
Every process the manager binary runs inside a Cluster's pods now logs without sampling. The controller keeps the duplicate-message sampler that controller-runtime applies to every operator.

The manager's logger passed only the first 100 messages sharing the same (level, message) pair each second, and roughly 1 in 100 afterwards. Every PostgreSQL record forwarded by the instance manager's log pipe is logged with the same message (`record`), so a burst of database activity, such as pgaudit output under load, was silently collapsed and most records were lost from the pod's log stream with no indication of the gap.

The sampler is installed by the controller-runtime zap builder at construction time and cannot be stripped afterwards, so the fix lives in machinery: `ConfigureLogging` gained a `WithDisabledSampling()` option, and the manager now passes it for every subcommand except `controller`. The operator keeps sampling, since a reconciliation storm can repeat one message well beyond the threshold, which is the case the sampler exists for.

Credit to @willmurnane for the initial diagnosis in #10815, and to @akashbhujbalwebsite for narrowing it down to the sampler mechanism in #11109.

Closes #10815


ardentperf:
would be nice to see this land. i suspect many people dont realize sampling can be applied to pgaudit records under extreme load

akashbhujbalwebsite:
Thanks for the broader architectural fix, @armru — I agree disabling sampling at the process level is the right long-term approach.

One question: for backports to release-1.28, 1.29, and 1.30, where machinery is pinned at v0.5.0 and WithDisabledSampling() isn't available yet, would #11109 still be useful as a temporary fix until machinery#292 is released? Happy to defer to the maintainers' judgment.

NiccoloFei (approved):

mnencia:
> Thanks for the broader architectural fix, @armru — I agree disabling sampling at the process level is the right long-term approach.
> 
> One question: for backports to release-1.28, 1.29, and 1.30, where machinery is pinned at v0.5.0 and WithDisabledSampling() isn't available yet, would #11109 still be useful as a temporary fix until machinery#292 is released? Happy to defer to the maintainers' judgment.

@akashbhujbalwebsite Good question, and thanks again for the root-cause diagnosis, it's what made this fix possible. Credit also to @willmurnane, whose original diagnosis in #10815 identified the sampler mechanism from the start. The plan is to merge this PR now on the pseudo-version pin, then cut and tag a machinery release before the next cnpg release and bump the pin to that tag. Backports to 1.28/1.29/1.30 will ride the tagged version once it's out, so there's no need to keep #11109 alive as a stopgap.

mnencia (approved):

PR #11132, opened 2026-07-06
fix(backup): always retry VolumeSnapshot errors until the deadline

### LINKED ISSUE #8608: [Bug] VolumeSnapshot on OKE/OCI fails with HTTP 409 “backup in progress”

mnencia:
VolumeSnapshot.Status.Error is free text from the CSI driver, with no structured code, and its shape is provider-specific. CNPG classified some of these messages as permanent and failed the Backup instead of letting the existing volumeSnapshotDeadline-bounded retry loop run, but the pattern list only recognized Azure's wording: OCI's 409 "backup in progress" (#8608) and GCP's transient 502 both fell through and killed backups the CSI driver was about to complete on its own.

There's no reliable way to tell a permanent condition from a transient one in this text, and what looks permanent today, like a missing VolumeSnapshotClass or bad credentials, can resolve itself while CNPG is retrying. Always retry until the deadline (10 minutes by default) elapses instead of trying to classify the message, and drop the pattern matching entirely.

Closes #8608

Assisted-by: Claude

----

Add to the release notes:

```
Fixed a bug where a VolumeSnapshot backup could fail immediately on a transient error from some CSI drivers (for example OCI's "backup in progress" conflict, or a transient error from the GCP Persistent Disk driver) instead of retrying until the configured deadline.
```

### DIFF
--- a/docs/src/appendixes/backup_volumesnapshot.md
+++ b/docs/src/appendixes/backup_volumesnapshot.md
@@ -283,22 +283,20 @@ If not specified, the default retry deadline is **10 minutes**.
 
 ### Error Handling
 
-When a retryable error occurs during a volume snapshot operation:
+When an error occurs during a volume snapshot operation:
 
 1. CloudNativePG records the time of the first error.
 2. The system retries the operation every **10 seconds**.
 3. If the error persists beyond the specified deadline (or the default 10
    minutes), the backup is marked as **failed**.
 
-### Retryable Errors
-
-CloudNativePG treats the following types of errors as retryable:
-
-- **Server timeout errors** (HTTP 408, 429, 500, 502, 503, 504)
-- **Conflicts** (optimistic locking errors)
-- **Internal errors**
-- **Context deadline exceeded errors**
-- **Timeout errors from the CSI snapshot controller**
+The CSI driver reports snapshot errors as free text, with no reliable way to
+tell a permanent condition (for example, a missing `VolumeSnapshotClass` or
+invalid credentials) from a transient one, and a condition that looks
+permanent can become resolvable at any point, for example if someone creates
+the missing class or fixes the credentials while CloudNativePG is retrying.
+For this reason, CloudNativePG retries every volume snapshot error the same
+way, until the deadline is reached.
 
 ### Examples
 
--- a/docs/src/appendixes/backup_volumesnapshot.md
+++ b/docs/src/appendixes/backup_volumesnapshot.md
@@ -266,12 +266,12 @@ for details on this standard behavior.
 
 ## Backup Volume Snapshot Deadlines
 
-CloudNativePG supports backups using the volume snapshot method. In some
-environments, volume snapshots may encounter temporary issues that can be
-retried.
+CloudNativePG supports backups using the volume snapshot method. Volume
+snapshot operations may encounter errors, which CloudNativePG retries for a
+configurable amount of time before giving up.
 
 The `backup.cnpg.io/volumeSnapshotDeadline` annotation defines how long
-CloudNativePG should continue retrying recoverable errors before marking the
+CloudNativePG should continue retrying these errors before marking the
 backup as failed.
 
 You can add the `backup.cnpg.io/volumeSnapshotDeadline` annotation to both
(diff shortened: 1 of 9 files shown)


leonardoce (approved):

mnencia:
/ok-to-merge E2E tests failed for infra glitch (too many requests to github api)

armru (approved):

PR #11135, opened 2026-07-06
fix: Mark replicas unhealthy on stalled WAL replay error loops

### LINKED ISSUE #11134: [Bug]: Replica can remain healthy after unrecoverable WAL replay loop

### Contact Details

schneider@ardentperf.com

### Version

1.30.0

1.36

Cloud: Azure AKS

YAML manifest

### What happened?

A replica can enter an unrecoverable WAL replay failure loop while CNPG continues to treat the instance as healthy enough for normal operation. In this state the instance repeatedly logs WAL replay errors at the same LSN, but the operator does not get a hard status signal from the instance manager.

Reproducer: https://github.com/ardentperf/bug-repro-wal-restore

Observed impact:

- the broken replica becomes Ready even though WAL replay is stuck
- the pod remains eligible for read traffic through normal service endpoints
- the replica remains a possible failover/promotion target because `/pg/status` does not report an error
- diagnosis currently depends on inspecting instance logs directly

Expected behavior:

When the instance manager detects repeated unrecoverable WAL replay errors with no replay progress, it should mark the instance unhealthy through existing readiness and status paths. CNPG should stop routing traffic to the replica and existing target-selection logic should avoid it.

Known log patterns include:

- `record with incorrect prev-link ... at ...`
- `contrecord is requested by ...`

A future enhancement may be useful to surface "user action needed" conditions directly on the Cluster object, but operator and instance logs are sufficient for the initial fix.

### Cluster resource

```shell

```

### Relevant log output

```shell

```

### Code of Conduct

ardentperf:
This change detects replicas stuck in repeated unrecoverable WAL replay errors and marks the instance unhealthy using existing CNPG health paths.

Reproducer: https://github.com/ardentperf/bug-repro-wal-restore

**important: if someone went into a pod and manually executed pg_rewind to recover, current PR design would still require a pod restart to clear the flag allowing CNPG to mark it healthy**

The patch intentionally avoids automatic remediation. It does not delete pods, recreate PVCs, fence instances, or mark the Cluster unrecoverable. It only makes the broken replica stop looking healthy.

Main changes:

- Adds an instance `statusError` field.
  - A non-empty value causes `GetStatus()` and therefore `/pg/status` to fail.
  - The operator records the failed `/pg/status` call as a status collection error.
  - Existing target selection already avoids instances with status collection errors.
  - This is process-local state and is cleared by restarting the affected Pod.
- Adds a log monitor path for CSV log records.
  - `LogPipe` now accepts a `RecordWriter`, and a new `MultiRecordWriter` fans each parsed log record out to every configured `RecordWriter`.
  - The instance manager builds a `MultiRecordWriter` holding the WAL replay detector alongside the normal log writer, so multiple independent detectors can observe records without chaining.
  - The WAL replay detector is a pure observer — it inspects records in memory and no longer forwards them itself (delivery is handled by the `MultiRecordWriter`).
- Adds WAL replay loop detection.
  - Tracks repeated matching WAL replay errors at the same LSN.
  - Requires 3 repeated matches within the configured window.
  - Skips primaries.
  - On detection, sets `canCheckReadiness=false` and sets the instance status error.

Behavior after detection:

- readiness fails, so Kubernetes removes the replica from service endpoints
- `/pg/status` fails with a clear message
- normal cluster failover target selection avoids the instance through existing status-error handling
- logs remain the primary visibility mechanism for the detailed condition

This PR deliberately does not add Cluster phase reporting for this condition. A follow-up feature could add a generic mechanism for instance-manager detected faults that require user action to be surfaced in Cluster status, but that is kept out of this minimal fix.

Tests run:

```bash
go test ./internal/cmd/manager/instance/run ./pkg/management/postgres ./pkg/management/postgres/logpipe ./pkg/postgres ./internal/controller
```

Fixes #11134

### DIFF
--- a/internal/cmd/manager/instance/run/cmd.go
+++ b/internal/cmd/manager/instance/run/cmd.go
@@ -306,8 +306,10 @@ func runSubCommand( //nolint: gocyclo,gocognit
 	}
 
 	// postgres CSV logs handler (PGAudit too)
-	postgresLogPipe := logpipe.NewLogPipe(
-		newWALReplayErrorDetector(instance, &logpipe.LogRecordWriter{}))
+	postgresLogPipe := logpipe.NewLogPipe(logpipe.MultiRecordWriter{
+		newWALReplayErrorDetector(instance),
+		&logpipe.LogRecordWriter{},
+	})
 	if err := mgr.Add(postgresLogPipe); err != nil {
 		contextLogger.Error(err, "unable to add CSV logs handler")
 		return err
--- a/internal/cmd/manager/instance/run/cmd.go
+++ b/internal/cmd/manager/instance/run/cmd.go
@@ -306,7 +306,8 @@ func runSubCommand( //nolint: gocyclo,gocognit
 	}
 
 	// postgres CSV logs handler (PGAudit too)
-	postgresLogPipe := logpipe.NewLogPipe()
+	postgresLogPipe := logpipe.NewLogPipe(
+		newWALReplayErrorDetector(instance, &logpipe.LogRecordWriter{}))
 	if err := mgr.Add(postgresLogPipe); err != nil {
 		contextLogger.Error(err, "unable to add CSV logs handler")
 		return err
--- a/internal/cmd/manager/instance/run/wal_replay_detector.go
+++ b/internal/cmd/manager/instance/run/wal_replay_detector.go
@@ -1,0 +1,107 @@
+/*
+Copyright © contributors to CloudNativePG, established as
+CloudNativePG a Series of LF Projects, LLC.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package run
+
+import (
+	"fmt"
+	"regexp"
+	"time"
+
+	"github.com/cloudnative-pg/cloudnative-pg/pkg/management/postgres"
+	"github.com/cloudnative-pg/cloudnative-pg/pkg/management/postgres/logpipe"
+)
+
+const (
+	walReplayErrorThreshold = 3
+	walReplayErrorWindow    = 10 * time.Minute
+	walReplayStalledError   = "unrecoverable WAL replay error loop detected"
+)
+
+var (
+	incorrectPrevLinkRegex = regexp.MustCompile(`record with incorrect prev-link [0-9A-F]+/[0-9A-F]+ at ([0-9A-F]+/[0-9A-F]+)`)
+	contrecordRegex        = regexp.MustCompile(`contrecord is requested by ([0-9A-F]+/[0-9A-F]+)`)
+)
+
+type walReplayErrorDetector struct {
+	instance *postgres.Instance
+	next     logpipe.RecordWriter
+	now      func() time.Time
+
+	lsn       string
+	count     int
+	firstSeen time.Time
+}
+
+func newWALReplayErrorDetector(instance *postgres.Instance, next logpipe.RecordWriter) logpipe.RecordWriter {
+	return &walReplayErrorDetector{
+		instance: instance,
+		next:     next,
+		now:      time.Now,
+	}
+}
+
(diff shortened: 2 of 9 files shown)


armru:
The PR looks good to me, and I like the same-LSN repetition idea: from
what I can tell from the PostgreSQL sources, these messages only show up
when a WAL source keeps serving actual bad bytes, and the benign
end-of-WAL repeats seem to be suppressed to DEBUG1, so I don't think
they can reach the threshold. There are a few points I'd like to
discuss though.

**1. I think the detector should also check where the message comes
from.** If I'm reading it right, any CSV record whose message matches
would count, regardless of which backend emitted it, so for example
three `RAISE EXCEPTION 'contrecord is requested by 0/5A000028'` from a
connected user would mark the replica unhealthy (pg_walinspect seems
able to produce the same text inside its error messages too). Since the
genuine loop messages should only come from the startup process, and
`LoggingRecord` already parses `backend_type` (PG 13+), wouldn't a
`BackendType == "startup"` check close this?

**2. Have you considered clearing `statusError` when the postmaster
restarts in place?** The instance manager re-enables `canCheckReadiness`
at every postmaster session start (lifecycle/run.go), so after an
in-place restart (config change, unfencing) the two signals seem to
diverge: the pod becomes ready again and rejoins the `-ro` service while
`/pg/status` keeps failing. Clearing the error at that same point feels
more consistent to me, and I think it would also address the pg_rewind
caveat from your description: a still-broken replica should re-trip
within seconds, while a fixed one recovers without deleting the pod.

**3. I'm a bit worried about the fencing interaction.** The early
return in `GetStatus` is above the `MightBeUnavailable` masking defer,
so if someone fences the broken replica (which I'd expect to be a common
reaction) I believe `/pg/status` would keep hard-failing instead of
reporting as expected-down, leaving cluster reconciliation blocked.
Would moving the check below the defer, or clearing the error on
fencing, make sense?

Two smaller things:

- If the replica stays permanently not-ready, I believe `reconcilePods`
  stops at its "waiting for pods to be ready" gates, so scale changes
  and rolling updates would freeze until someone intervenes (failover
  and switchover should still work). Maybe that's the right trade-off
  for a replica that needs human attention, but I'd call it out in the
  PR body.
- The messages are translated under a non-English `lc_messages`, so
  detection would silently turn off there; probably fine, but worth a
  comment (`backend_type` is never translated, so the check from point 1
  should be safe). Also, having the actual LSN in the status error
  message would help debugging.

The MultiRecordWriter change looks good to me.

ardentperf:
thanks for the review and feedback

> (pg_walinspect seems able to produce the same text inside its error messages too). Since the genuine loop messages should only come from the startup process, and `LoggingRecord` already parses `backend_type` (PG 13+), wouldn't a `BackendType == "startup"` check close this?

good catch on pg_walinspect. i think responding to an explicit `RAISE EXCEPTION` might be useful for testing (unlikely someone does that accidentally) - but while unlikely, i think pg_walinspect could happen while someone was debugging. i'll work on an update to filter on BackendType

> **2. Have you considered clearing `statusError` when the postmaster restarts in place?** ... Clearing the error at that same point feels more consistent to me, and I think it would also address the pg_rewind caveat from your description

i agree this makes sense. i think that resetting the state with in-place postmaster restarts will be intuitive for users.

> early return in `GetStatus` is above the `MightBeUnavailable` masking defer, so if someone fences the broken replica (which I'd expect to be a common reaction) I believe `/pg/status` would keep hard-failing instead of reporting as expected-down, leaving cluster reconciliation blocked. Would moving the check below the defer, or clearing the error on fencing, make sense?

random comment: `MightBeUnavailable` seems like an odd name to me for fencing :)

i hadn't fully thought through the interaction with fencing *after* the instance gets into an unrecoverable state. this is helpful.

i think the error should be left in place when an instance is fenced; i haven't traced all the code yet but i'm thinking masked errors might still be visible? but i do think it would make sense to clear the error on un-fencing, if that won't already happen. if instances can be un-fenced without any in-place postmaster restart the it might be worth an extra line of code to do this explicitly.

> * If the replica stays permanently not-ready, I believe `reconcilePods`
>   stops at its "waiting for pods to be ready" gates, so scale changes
>   and rolling updates would freeze until someone intervenes (failover
>   and switchover should still work). Maybe that's the right trade-off
>   for a replica that needs human attention, but I'd call it out in the
>   PR body.

actually i dont think this is the right tradeoff - but i think this should be a separate PR. IMO we should think more broadly about "how should reconciliation work with pods that aren't ready"

> * The messages are translated under a non-English `lc_messages`, so
>   detection would silently turn off there; probably fine, but worth a
>   comment (`backend_type` is never translated, so the check from point 1
>   should be safe). Also, having the actual LSN in the status error
>   message would help debugging.

i originally had LSN in the message, during an early draft, and removed it from the message bc that seemed a little more consistent with other messages to me :) happy to add it back in, that works for me

i'll work on a new revision of this patch. thanks again for reviewing!!

ardentperf:
nb. i will also address linter feedback in the PR

PR #11137, opened 2026-07-06
feat: Capture shutdown diagnostics before immediate PostgreSQL shutdown

### LINKED ISSUE #11136: [Feature]: Capture shutdown diagnostics before immediate PostgreSQL shutdown

We recently had a fairly painful incident on CNPG. The root cause was a hung postgres shutdown which has been reported to hackers - but there was not enough diagnostics to further root cause.

https://www.postgresql.org/message-id/flat/CA%2BfnDAa8Kbbx3KkqpfxutTx41RchJsT3Lykqwcedm0CVr5Tw%3DA%40mail.gmail.com

We would like to preserve diagnostic evidence immediately before CNPG requests abortive PostgreSQL shutdown. Currently, when shutdown hangs until CloudNativePG escalates to immediate shutdown, the remaining PostgreSQL process state is the most useful evidence but is often lost as the pod exits or is killed.

Want to add a minimal instance-manager diagnostics collector that runs only when a fast PostgreSQL shutdown fails and CloudNativePG is about to issue `pg_ctl -m immediate stop`.

An operator manually collects troubleshooting info when a hung shutdown happens. Requires oncall/paging, immediate response, and requires delaying resolution for collecting data.

### Additional context

_No response_

### Backport?

Yes

Yes

### Code of Conduct

ardentperf:
## Summary

This change adds a minimal instance-manager diagnostics collector that runs only when a fast PostgreSQL shutdown fails and CloudNativePG is about to issue `pg_ctl -m immediate stop`.

The collector logs three structured diagnostics records, one immediately and two more at three-second intervals, under the existing short diagnostics timeout. Each record contains a `sample` number and a `processes` payload collected from `/proc/[0-9]*`.

A related PostgreSQL hackers thread documents a real-world shutdown stall involving autovacuum which occurred on CNPG: https://www.postgresql.org/message-id/flat/CA%2BfnDAa8Kbbx3KkqpfxutTx41RchJsT3Lykqwcedm0CVr5Tw%3DA%40mail.gmail.com

The goal here is not to change shutdown semantics. It is to preserve diagnostic evidence immediately before CNPG requests abortive PostgreSQL shutdown. Currently, when shutdown hangs until CloudNativePG escalates to immediate shutdown, the remaining PostgreSQL process state is the most useful evidence but is often lost as the pod exits or is killed.

- Adds `pkg/management/postgres/shutdown_diagnostics.go`.
- Calls `logShutdownDiagnostics(ctx)` only from `TryShuttingDownFastImmediate`, after fast shutdown returns an `exec.ExitError` and before issuing immediate shutdown.
- Reads procfs directly instead of depending on `ps` or a process-listing library.
- Emits three structured log records:
  - `message`: `PostgreSQL shutdown diagnostics`
  - `sample`: `1`, `2`, or `3`
  - `processes`: array of `{ pid, files }`
- Collects `cmdline`, `comm`, `status`, `wchan`, `io`, `syscall`, `stack`, and `sched`.
- Trims leading/trailing whitespace from collected procfs lines to reduce payload size while preserving internal formatting.
- Stores proc read failures directly in the relevant field value, for example:
  - `"stack": ["open /proc/123/stack: permission denied"]`
- Keeps `stack` and `syscall` reads best-effort because they are often ptrace/capability gated.
- Keeps `io` and `sched` because three samples make per-process IO and scheduling counters useful for identifying whether a stuck backend is making progress.

Example shape:

```json
{
  "msg": "PostgreSQL shutdown diagnostics",
  "sample": 1,
  "processes": [
    {
      "pid": "123",
      "files": {
        "cmdline": ["postgres autovacuum worker"],
        "comm": ["postgres"],
        "status": ["Name:\tpostgres", "State:\tT (stopped)"],
        "io": ["rchar: 208321", "write_bytes: 0"],
        "sched": ["postgres (123, #threads: 1)", "nr_switches : 8"],
        "wchan": ["do_signal_stop"]
      }
    }
  ]
}
```

## Testing

Ran:

```console
go test ./pkg/management/postgres/...
```

The test coverage verifies that:

- procfs process data is collected into the structured payload;
- null-separated `cmdline` content is normalized;
- leading/trailing whitespace is removed from collected lines.

Also validated manually in AKS with a custom CNPG image:

- Built the modified manager and overlaid it on `ghcr.io/cloudnative-pg/cloudnative-pg:1.29.2`.
- Deployed the custom image as both the operator image and `OPERATOR_IMAGE_NAME`.
- Created a one-instance cluster with `stopDelay: 30` and `switchoverDelay: 30`.
- Started a user `psql` session, found its PostgreSQL backend, and sent `SIGSTOP`.
- Confirmed `/proc/<pid>/status` showed `State: T (stopped)` and `/proc/<pid>/wchan` showed `do_signal_stop`.
- Fenced the instance to trigger the fast-then-immediate shutdown path.
- Verified fast shutdown timed out, three diagnostics records were emitted, the stopped backend appeared in all samples, and immediate shutdown followed.

Closes #11136

### DIFF
--- a/pkg/management/postgres/instance.go
+++ b/pkg/management/postgres/instance.go
@@ -682,6 +682,7 @@ func (instance *Instance) TryShuttingDownFastImmediate(ctx context.Context) erro
 	)
 	var exitError *exec.ExitError
 	if errors.As(err, &exitError) {
+		logShutdownDiagnostics(ctx)
 		contextLogger.Info("Graceful shutdown failed. Issuing immediate shutdown",
 			"exitCode", exitError.ExitCode())
 		err = instance.Shutdown(ctx,
--- a/pkg/management/postgres/shutdown_diagnostics.go
+++ b/pkg/management/postgres/shutdown_diagnostics.go
@@ -29,22 +29,64 @@ import (
 	"github.com/cloudnative-pg/machinery/pkg/log"
 )
 
-func logShutdownDiagnostics(ctx context.Context) {
-	diagCtx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
-	defer cancel()
+const (
+	shutdownDiagnosticsSamples  = 3
+	shutdownDiagnosticsInterval = 3 * time.Second
+	// shutdownDiagnosticsFirstSampleWait caps how long the shutdown escalation
+	// waits for the first sample: reads of /proc/[pid]/{cmdline,status} can
+	// block in the kernel and cannot be cancelled from Go.
+	shutdownDiagnosticsFirstSampleWait = 3 * time.Second
+)
 
+// logShutdownDiagnostics logs a few samples of the state of the processes
+// still running in the container. It waits for the first sample, so that the
+// processes that resisted the fast shutdown are captured before the immediate
+// shutdown starts killing them, while the remaining samples are collected in
+// the background: the processes that survive them are the ones worth
+// investigating.
+func logShutdownDiagnostics(ctx context.Context) {
 	contextLogger := log.FromContext(ctx)
-	contextLogger.Info("PostgreSQL shutdown diagnostics",
-		"sample", 1,
-		"processes", collectProcDiagnostics(diagCtx, "/proc"))
-	time.Sleep(3 * time.Second)
-	contextLogger.Info("PostgreSQL shutdown diagnostics",
-		"sample", 2,
-		"processes", collectProcDiagnostics(diagCtx, "/proc"))
-	time.Sleep(3 * time.Second)
-	contextLogger.Info("PostgreSQL shutdown diagnostics",
-		"sample", 3,
-		"processes", collectProcDiagnostics(diagCtx, "/proc"))
+
+	firstSampleLogged := make(chan struct{})
+	go func() {
+		// The collection is detached from the caller's cancellation on
+		// purpose: the samples must survive the manager teardown that may
+		// be in progress while this code runs.
+		diagCtx, cancel := context.WithTimeout(context.WithoutCancel(ctx), 10*time.Second)
+		defer cancel()
+
+		for sample := 1; sample <= shutdownDiagnosticsSamples; sample++ {
+			processes := collectProcDiagnostics(diagCtx, "/proc")
+			for i := range processes {
+				// One record per process: a single record with every process
+				// would exceed the 16KiB line limit enforced by the container
+				// runtimes, and would be split into fragments of invalid JSON.
+				contextLogger.Info("PostgreSQL shutdown diagnostics",
+					"sample", sample,
+					"pid", processes[i].PID,
+					"files", processes[i].Files)
+			}
+			if sample == 1 {
+				close(firstSampleLogged)
+			}
+			if sample == shutdownDiagnosticsSamples {
+				return
+			}
+
+			select {
+			case <-time.After(shutdownDiagnosticsInterval):
+			case <-diagCtx.Done():
+				return
+			}
+		}
+	}()
+
+	select {
+	case <-firstSampleLogged:
+	case <-time.After(shutdownDiagnosticsFirstSampleWait):
+		contextLogger.Info(
+			"Shutdown diagnostics collection is still running, proceeding with the immediate shutdown")
+	}
 }
 
 type procDiagnostics struct {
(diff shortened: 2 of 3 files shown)


ardentperf:
@armru @mnencia  not sure i agree with the idea of backgrounding this.  getting multiple samples is actually really important and useful for debugging complex issues, because a single sample doesnt tell you if processes are stuck or slow.  its the multiple samples that tell you if a process did nothing for 3 seconds, or if it did IO, or consumed cpu, etc.

original patch had a max 10 second pause in the shutdown (context deadline enforced) - i had considered this part carefully; i think a 10 second delay is acceptable any time we're triggering timeouts and escalating to SIGKILL on postgres

realistically, I think it will rarely exceed six seconds because the only thing it’s doing is reading from /proc

part of the underlying idea behind this is that our desired shutdown path is to avoid SIGKILL, and any time we're going the SIGKILL route we want to collect quality diagnostics data to root cause it

from past experience, i can say that it's a common debugging procedure to grab a few samples of system state, separated by a few seconds - not just a single sample

——

Switching to one log record per process seems ok - I don’t see any downside there

ardentperf:
I'm reviewing logs from the original incident I saw, which is described in the hackers email linked above.

Looking closer, I actually don't think `TryShuttingDownFastImmediate()` was the shutdown path after all.

The exact operations during the incident were a `kubectl cnpg destroy` issued on the primary instance, while a failover was still in progress before it completed (the new primary was delayed with becoming ready). ISTM that the destroy command's `ensurePodIsDeleted()` does a straight pod delete. https://github.com/cloudnative-pg/cloudnative-pg/blob/main/internal/cmd/plugin/destroy/destroy.go#L118

I think instance manager intercepts sigint/sigterm and calls `TryShuttingDownSmartFast()` instead of immediate. https://github.com/cloudnative-pg/cloudnative-pg/blob/main/internal/cmd/manager/instance/run/lifecycle/lifecycle.go#L142

And I think that the SIGKILL i saw was actually coming from kubelet, i suspect based on CNPG setting `terminationGracePeriodSeconds` to the value of `stopDelay` with default of 1800

So this PR is actually injecting the diagnostics collection in the wrong spot. i don't think a k8s `preStop` hook is going to work either, from initial searching it seems like that fires right away instead of waiting for `terminationGracePeriodSeconds`

What I actually want is a way to grab these diagnostics before `terminationGracePeriodSeconds` triggers a SIGKILL on the pod

ardentperf:
by the way - IMO we should remove the comment in `lifecycle.go` suggesting that sigkill can result in data corruption. sigkill is not ideal and should be avoided if possible - but what is the actual risk of corruption? especially with data checksums in place (which should always be enabled), even a torn write would never result in silent corruption. in fact postgres is writing complete pages into the page cache and the OS flushes to disk on fsync independently of whether a process gets killed. transactions are always written and fsync'd in WAL before commits are acknowledged to clients. it's good to attempt a clean shutdown first but sigkill likely is not as risk as this comment makes it seem. and that discussion is better left for the hackers list (the level of risk is far from being clear & agreed)

ardentperf:
this seems connected to #2835 and subsequent commits where we extended stopDelay to 1800 (30 minutes)

we will likely need to configure a more aggressive shutdown -- especially now that we have primary lease... the result here will be a 30-minute outage during a routine failover if primary isn't going down.  i dont want stuck shutdowns killing SLO budgets when sigkill is safe (which it is). I think we will want to run with stopDelay well under a minute, probably 30 seconds or less.  (cc @gbartolini )  it's better for SLO's to let postgres do crash recovery on startup, versus hoping for a clean checkpoint+shutdown and then still having to do crash recovery in addition to that whenever it doesn't complete successfully.

*(also we really need to remove the comment suggesting that sigkill can result in data corruption)*

@armru @mnencia what do you think about changing to hard coded 2 second sleeps (instead of 3 seconds) and then doing something like this:

```go
func (instance *Instance) TryShuttingDownSmartFast(ctx context.Context) error {
	contextLogger := log.FromContext(ctx)

	var err error

	cluster := instance.GetClusterOrDefault()
	smartTimeout := cluster.GetSmartShutdownTimeout()
	maxStopDelay := cluster.GetMaxStopDelay()

	// Capture diagnostics 5 seconds before the stop-delay budget runs out,
	// even if we're still blocked in pg_ctl below.
	if maxStopDelay > 10 {
		diagTimer := time.AfterFunc(
			time.Duration(maxStopDelay-5)*time.Second,
			func() { logShutdownDiagnostics(ctx) },
		)
		defer diagTimer.Stop()
	}

	// ... existing smart/fast logic unchanged ...
}
```

and it could be ok to keep the call in `TryShuttingDownFastImmediate()` - just wasn't my primary focus since that wasn't the call path in the incident i saw.

PR #11138, opened 2026-07-07
fix(instance): skip startup probe during active WAL replay

### LINKED ISSUE #11024: [Bug]: Replica enters crash/restart loop after restart with overloaded primary checkpointer

### Contact Details

schneider@ardentperf.com

### Version

older minor (unsupported)

1.34

Cloud: Azure AKS

Helm

### What happened?

>
> ```c
> {
>    ereport(DEBUG2,
>    ...
>    return false;
>}
>```
>

### Cluster resource

```shell

```

### Relevant log output

```shell

```

### Code of Conduct

ardentperf:
The startup probe currently depends on PostgreSQL accepting connections. That can make a replica look failed while PostgreSQL is still making forward progress through crash recovery or WAL replay, causing Kubernetes to restart the pod and repeat the same work. This builds on the logpipe infrastructure from https://github.com/cloudnative-pg/cloudnative-pg/pull/11135 and records the timestamp of PostgreSQL `redo in progress` log messages that include the current LSN. While one of those messages was seen in the last five minutes, `/startupz` reports `Skipped` instead of failing the startup probe. If the signal is stale or absent, the code stops replay-progress detection for this process and falls back to the existing startup checker. The detector also stops as soon as PostgreSQL logs that the database system is ready to accept connections. At that point replay-progress detection is no longer useful, and stopping it avoids running the regex against later log messages.

Prerequisite: #11135
Fixes: #11024

### DIFF
--- a/internal/cmd/manager/instance/run/cmd.go
+++ b/internal/cmd/manager/instance/run/cmd.go
@@ -308,6 +308,7 @@ func runSubCommand( //nolint: gocyclo,gocognit
 	// postgres CSV logs handler (PGAudit too)
 	postgresLogPipe := logpipe.NewLogPipe(logpipe.MultiRecordWriter{
 		newWALReplayErrorDetector(instance),
+		newWALReplayProgressDetector(instance),
 		&logpipe.LogRecordWriter{},
 	})
 	if err := mgr.Add(postgresLogPipe); err != nil {
--- a/internal/cmd/manager/instance/run/cmd.go
+++ b/internal/cmd/manager/instance/run/cmd.go
@@ -306,8 +306,10 @@ func runSubCommand( //nolint: gocyclo,gocognit
 	}
 
 	// postgres CSV logs handler (PGAudit too)
-	postgresLogPipe := logpipe.NewLogPipe(
-		newWALReplayErrorDetector(instance, &logpipe.LogRecordWriter{}))
+	postgresLogPipe := logpipe.NewLogPipe(logpipe.MultiRecordWriter{
+		newWALReplayErrorDetector(instance),
+		&logpipe.LogRecordWriter{},
+	})
 	if err := mgr.Add(postgresLogPipe); err != nil {
 		contextLogger.Error(err, "unable to add CSV logs handler")
 		return err
--- a/internal/cmd/manager/instance/run/cmd.go
+++ b/internal/cmd/manager/instance/run/cmd.go
@@ -306,7 +306,8 @@ func runSubCommand( //nolint: gocyclo,gocognit
 	}
 
 	// postgres CSV logs handler (PGAudit too)
-	postgresLogPipe := logpipe.NewLogPipe()
+	postgresLogPipe := logpipe.NewLogPipe(
+		newWALReplayErrorDetector(instance, &logpipe.LogRecordWriter{}))
 	if err := mgr.Add(postgresLogPipe); err != nil {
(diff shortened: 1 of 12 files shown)


armru:
Thanks for this work, and especially for the diagnosis in #11024 — the
restartpoint analysis there is spot on, and I fully agree with the
direction: while replay is making progress, failing the startup probe and
losing the work is the worst outcome.

While reviewing I ran into a few things that, as far as I can tell,
prevent the detection from engaging in practice. I may well be missing
something, so please double check my reasoning — I'm sharing what I
found in the PostgreSQL sources and what I observed running this branch
on a kind cluster.

**1. I couldn't find the `redo in progress` message on standbys.** It
seems to be guarded by `if (!StandbyMode)` in `PerformWalRecovery`
(xlogrecovery.c, REL_15 through REL_18), and if I read `InitWalRecovery`
correctly, a replica enables standby mode *before* redo starts (both the
backup_label and the minRecoveryPoint paths). That would mean the pod
class from #11024, a restarted replica, doesn't produce matching log
lines. In my tests (full standby replays on PostgreSQL 18.4 with this
branch deployed) I indeed saw zero matches and the detector never fired —
but if you observed it firing on a replica in your environment I'd really
like to see those logs, since that would mean I'm misreading the guard.
Also worth noting: the message seems to arrive with PG 15
(`log_startup_progress_interval`), so PG 14 wouldn't have it either way.

**2. On primaries, the first probe may disable detection before the
first message can appear.** The first `redo in progress` line can only
show up `log_startup_progress_interval` (10s) after redo starts, while
the first startup probe tends to reach the handler earlier (we set no
`initialDelaySeconds`, 10s period). When it finds no signal it calls
`StopWALReplayProgressDetection()`, which if I understand the intent
correctly is one-way. In my crash recovery test (22.9s recovery) the
first message appeared at T+10s but a probe had already come at T+3s, so
the pod went through recovery on the standard path. Maybe I got unlucky
with timing, but I couldn't construct a run where detection survived the
first probe.

**3. I'm not sure the staleness fallback can ever run.** Once the probe
returns 200, the kubelet marks the container Started and stops calling
the startup probe (kubelet `prober/worker.go`), and I couldn't find
another caller of `/startupz`. If that's right, the 5-minute window is
never re-evaluated after the first skip, and since our replica liveness
always returns OK, a replay that stalls after that point would leave the
pod unready indefinitely, which quietly changes the meaning of
`startDelay`. It would at least deserve a docs note.

To be clear on severity: even if I'm right about 1 and 2, the effect is
that the change becomes a no-op rather than something harmful — nothing
breaks, it just wouldn't help #11024.

Since I believe the probe-side idea itself is the right one, I also
explored an alternative signal on top of it: sampling the startup process
title (`postgres: ... startup recovering <segment>`, set in
`XLogFileRead` for all WAL sources), which appears to advance per segment
in every recovery form, on every supported version, independently of the
log locale. I'd like to open it as an alternative PR crediting you as
co-author for the diagnosis and the design idea, and iterate on it
together — but first I'd love your take on the findings above, in case
there's context I'm missing. #11135 looks unaffected by all of this and
I'll review it separately.

ardentperf:
thanks for taking a look. sampling the startup process title actually seems like a better approach to me than looking for the `redo in progress` message - i agree with a pivot to that approach

ardentperf:
i'm going to close this in favor of #11141

PR #11141, opened 2026-07-07
wip(instance): keep the startup probe from failing during WAL replay

### LINKED ISSUE #11024: [Bug]: Replica enters crash/restart loop after restart with overloaded primary checkpointer

armru:
Closes #11024


armru:
closing in favor of #11152

PR #11144, opened 2026-07-07
feat(synchronous): failure domain-aware synchronous replication

leonardoce:
Let CloudNativePG restrict synchronous standby selection to instances in a failure domain different from the primary's, such as another availability zone, so a single-zone outage cannot take out the primary and all its synchronous replicas at once.

- Add mutually exclusive `podFailureDomainKeys` and `nodeFailureDomainKeys` to `.spec.postgresql.synchronous`: pod labels are resolved without ever consulting the node, while node labels remain available for clusters where pod-level topology labels aren't populated (Kubernetes 1.34 and older, or where the `PodTopologyLabelsAdmission` admission plugin isn't enabled - see KEP4742).
- Apply the constraint only when it is fully satisfiable by the configured `number` of synchronous standbys (or, with `preferred` data durability, by a single cross-domain standby); otherwise, fall back to selecting standbys as if the keys were not configured, so write transactions never wait on acknowledgments that cannot arrive.
- Add a `SyncReplicationTopologySatisfied` cluster condition reporting whether the placement preference is currently honored, evaluated only against cluster instances (not `standbyNamesPre`/`standbyNamesPost`).
- Show a Failure Domain Topology section in `kubectl cnpg status`, grouping instances by domain, marking not-ready instances, and skipping the table for hibernated clusters.
- Update the kind/k3d e2e setup scripts to wait for every node to register and label them with distinct `topology.kubernetes.io/region`/`zone` values, so the feature can be exercised locally out of the box.


leonardoce:
This work can actually be split in 3 different PR. The last one can't be backported.

ardentperf:
what is the fallback model, if we need N sync targets but we can't satisfy that number from available replicas based on the current topology? assuming it's just around ordering, that it prioritizes the failure domain compliant replicas above others, but still adds others if needed?

leonardoce:
> what is the fallback model, if we need N sync targets but we can't satisfy that number from available replicas based on the current topology? assuming it's just around ordering, that it prioritizes the failure domain compliant replicas above others, but still adds others if needed?

It is just the same as what the "old" API was doing. Only the replicas that are not in the same failure domain of the primary are listed in synchronous_standby_names -  besides of their readiness status.

docs/src/replication.md
```diff
@@ -478,6 +478,73 @@ spec:
...
 
+### Failure Domain-Aware Synchronous Replication
+
+CloudNativePG can restrict the choice of synchronous standbys to instances
+located in a failure domain different from the primary's, such as another
+availability zone. The labels that define the failure domains are declared
+through one of the following options of the `.spec.postgresql.synchronous`
+stanza:
+
+- `nodeFailureDomainKeys`: a list of node label keys, resolved from the labels
```
gbartolini, line 491: I would flip the order, given that from K8s 1.35 some key labels will be present on the pod itself.

gbartolini (approved):

armru (approved):

armru:
A few commits on top of this branch change the original design, so, briefly:

The cross-domain list is now used only when it can satisfy the requested
durability, otherwise we fall back to the unfiltered one. With the hard filter, a
cluster with fewer cross-domain candidates than `number` gets
`ANY 2 ("cluster-2")` and every commit hangs, and one node failure is enough to
get there: better to lose the domain distribution than the writes. Does that
trade-off sound right, or should it be configurable?

`failureDomainKey` is split into `podFailureDomainKeys` and
`nodeFailureDomainKeys`, mutually exclusive: with the pod-then-node fallback the
same key meant different things on different instances, and a typo on a pod label
silently resolved to the node one. Pod labels are strict now, a missing key fails
the extraction.

The condition uses the same helper as the election, so it can't say "satisfied"
while the constraint isn't applied.

Last commit restores the "no synchronous replication configured" warning: it had
been dropped here while it duplicated #11148, which is now merged, so the removal
was deleting a feature that lives on main.

Feedback welcome, especially on the first point.

ardentperf:
How does this interact with failover quorum?

leonardoce:
> How does this interact with failover quorum?

This restrict the replicas listed in synchronous_standby_names, and is reflected on the FailoverQuorum custom resource.
From there the code path is the same, influencing the quorum check directly via the cardinality of the synchronous_standby_names and of the read set.

PR #11145, opened 2026-07-08
feat: read topology labels from pods before falling back to nodes

### LINKED ISSUE #11146: Read topology labels from pods before falling back to nodes

### Contact Details

leonardo.cecchi@gmail.com

### Version

1.30.0

1.36

YAML manifest

### What happened?

Prefer topology labels available directly on the `Pod` object over fetching the `Node` object when computing pod topology for synchronous replica election.

## Motivation

Kubernetes 1.35 promotes [KEP-4742](https://github.com/kubernetes/enhancements/tree/master/keps/sig-scheduling/4742-node-topology-labels-on-pods) (`PodTopologyLabelsAdmission`) to beta, enabled by default. This feature propagates node topology labels (e.g. `topology.kubernetes.io/zone`) onto the `Pod` object at scheduling time.

Today, `getPodsTopology` in `internal/controller/cluster_status.go` always fetches the `Node` object for each pod to read its topology labels. This requires an extra API call/cache lookup and fails outright (returning an empty topology) if the node can no longer be found — for example, when a node has been drained or removed after the pod was scheduled.

By reading topology labels from the pod first, we avoid an unnecessary dependency on the `Node` object being present, and make synchronous replica election more resilient on clusters where nodes may be scaled down or drained after pod scheduling.

## References

- KEP-4742: Node topology labels on Pods (`PodTopologyLabelsAdmission`, beta in Kubernetes 1.35)
- Implementation: `internal/controller/cluster_status.go` (`getPodsTopology`)

### Cluster resource

```shell

```

### Relevant log output

```shell

```

### Code of Conduct

leonardoce:
When extracting topology information for synchronous replica election, prefer labels available directly on the Pod object (propagated by Kubernetes via KEP-4742 / PodTopologyLabelsAdmission, beta and enabled by default in Kubernetes 1.35) over fetching the Node object.

The Node fetch is retained as a fallback for clusters running Kubernetes 1.34 or with the feature gate disabled, preserving full backwards compatibility. No behaviour change for existing users.

Closes: #11146

### DIFF
--- a/internal/controller/cluster_status.go
+++ b/internal/controller/cluster_status.go
@@ -832,18 +832,26 @@ func getPodsTopology(
 	for _, pod := range pods {
 		podName := apiv1.PodName(pod.Name)
 		data[podName] = make(map[string]string, 0)
-		node, ok := nodes[pod.Spec.NodeName]
-		if !ok {
-			// node not found, it means that:
-			// - the node could have been drained
-			// - others
-			contextLogger.Debug("node not found, skipping pod topology matching")
-			return apiv1.Topology{}
-		}
-
 		nodesMap[pod.Spec.NodeName] = append(nodesMap[pod.Spec.NodeName], podName)
 
 		for _, labelName := range topology.NodeLabelsAntiAffinity {
+			// Prefer pod labels: KEP-4742 propagates node topology labels onto pods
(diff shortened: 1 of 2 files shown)


leonardoce:
We abandoned this PR to focus on supporting Pod labels explicitly in #11144, without improving the deprecated synchronous replica API.

PR #11150, opened 2026-07-08
ci: authenticate raw.githubusercontent.com fetches in CD/CI

mnencia:
Anonymous requests to raw.githubusercontent.com share a per-IP rate limit across all GitHub Actions runners and can 429 during a nightly burst, as seen on the EKS external-snapshotter CRD fetch. An authenticated request gets its own, much higher quota, so route every unauthenticated fetch of that kind through the ambient GITHUB_TOKEN: the kind/k3d hostpath CSI driver setup (deploy_csi_host_path, the widest blast radius since it runs on every kind and k3d job), the EKS snapshotter setup, the AKS Rook setup, and the OpenShift bundle test.

Retry is kept only as a fallback for genuine transient network errors; it doesn't help with sustained rate-limiting the way authentication does.

### DIFF
--- a/.github/workflows/continuous-delivery.yml
+++ b/.github/workflows/continuous-delivery.yml
@@ -1246,6 +1246,12 @@ jobs:
         # It can still take a while to deploy rook.
         name: Set up Rook
         uses: nick-fields/retry@ad984534de44a9489a53aefd81eb77f87c70dc60 # v4
+        env:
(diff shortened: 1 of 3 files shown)


mnencia:
E2E tests running on EDB fork https://github.com/EnterpriseDB/cloudnative-pg/actions/runs/29035948401

NiccoloFei (approved):

litaocdl (approved):

PR #11152, opened 2026-07-09
fix(probes): treat a PostgreSQL rejecting connections as started up

### LINKED ISSUE #11024: [Bug]: Replica enters crash/restart loop after restart with overloaded primary checkpointer

armru:
When a PostgreSQL instance has to replay a large amount of WAL before reaching consistency (for example a standby restarting after a crash, as reported in #11024), `pg_isready` reports that the server is alive but rejecting connections until recovery completes. The startup probe treated that state as a failure, so when recovery outlasted `startDelay` the kubelet killed the pod. Every restart threw away the replay progress and the instance could never catch up, looping forever.

With the default `pg_isready` strategy, the startup probe now considers a server that is alive but rejecting connections as already started up:

- The liveness probe will not restart it, because it returns OK unconditionally for a non-primary instance.
- The readiness probe keeps failing on a server that rejects connections, so a recovering instance stays out of the ready endpoints of the services and is not eligible for promotion until it accepts connections.

The `query` and `streaming` startup strategies are unchanged: they require PostgreSQL to accept connections by design, and the documentation now points out that `startDelay` must account for recovery time when using them.

One trade-off worth noting: the startup probe was previously the only mechanism that restarted a pod whose PostgreSQL was permanently stuck while rejecting connections, even though in the #11024 scenario those restarts were exactly what prevented recovery from completing. With this change such an instance is reported as started and left to continue attempting recovery, with no automatic restart if it never completes. A specific form of this, repeated WAL-replay error loops, is proposed in #11135, which does not cover a replica silently stalled with no matching errors.

This was verified on a kind cluster by reproducing the scenario from the issue: a two-instance cluster with `startDelay: 10`, one-second probe periods, and `checkpoint_timeout: 5h` so that the standby takes no restartpoints. After writing about 4.7GiB of WAL and deleting the standby pod, the instance came back in `shut down in recovery`, the first startup probe reported it as started, the pod stayed not ready while replaying (readiness failing with HTTP 500), and it became ready about 40 seconds later with zero container restarts. The same configuration on the previous build restarted the pod every 10 seconds.

Unit tests cover the new startup checker and the probe runner selection, which previously had no coverage.

Thanks to Jeremy Schneider for reporting this issue and providing a detailed analysis of the failure.

Closes #11024

### DIFF
--- a/docs/src/instance_manager.md
+++ b/docs/src/instance_manager.md
@@ -50,8 +50,11 @@ The `.spec.startDelay` parameter specifies the maximum time, in seconds,
 allowed for the startup probe to succeed.
 
 By default, the `startDelay` is set to `3600` seconds. It is recommended to
-adjust this setting based on the time PostgreSQL needs to fully initialize in
-your specific environment.
+adjust this setting based on the time PostgreSQL needs to start in your
+specific environment. With the default `pg_isready` strategy, this only
+needs to cover the time required for the postmaster to come alive, while
+with the `query` and `streaming` strategies it must also account for the
+time needed to complete recovery and accept connections.
 
 :::warning
     Setting `.spec.startDelay` too low can cause the liveness probe to activate
--- a/docs/src/instance_manager.md
+++ b/docs/src/instance_manager.md
@@ -25,7 +25,7 @@ of the Pod, the instance manager acts as a backend to handle the
 ## Startup Probe
 
 The startup probe ensures that a PostgreSQL instance, whether a primary or
-standby, has fully started.
+standby, has started up.
 
 :::info
     By default, the startup probe uses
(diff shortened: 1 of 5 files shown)


leonardoce:
I can't reproduce the problem anymore, and I can see these lines in the log of my replica:

```
{
  "level": "info",
  "ts": "2026-07-17T09:09:09.250604551Z",
  "msg": "PostgreSQL is alive but rejecting connections, considering the instance as started up",
  "logging_pod": "cluster-example-2"
}
```

leonardoce (approved):

mnencia (approved):

PR #11175, opened 2026-07-13
Remove GH account for ADOPTERS.md

FloorD:
Former employee of Rewe has asked us to remove their GitHub account from the ADOPTERS file, as per the GDPR regulations. I've replaced their handle by a "-". We might want to seek out a new contact at this company.


gbartolini:
Mhh ... Are we sure we can keep the adopter's name without a Github handle? In any case, approving it for now.

gbartolini (approved):

mnencia (approved):

PR #11182, opened 2026-07-14
feat: read topology labels from pods before falling back to nodes

### LINKED ISSUE #11179: [Feature]: Read failure-domain topology labels from Pods before falling back to Nodes

To evaluate the syncReplicaElectionConstraint / failure-domain-aware synchronous replication topology, the operator currently fetches the Node object hosting every instance Pod and reads the anti-affinity labels (e.g. topology.kubernetes.io/zone) from it. This means:

- An extra Node GET (or a full Node list watch/cache) is required purely to read a couple of well-known topology labels.
- If a Node has been drained or deleted after the Pod was scheduled (e.g. during a rolling node replacement), topology extraction fails outright for the whole cluster, even though the information needed is often still available.
- Kubernetes 1.35 promotes KEP-4742 (PodTopologyLabelsAdmission) to beta and enables it by default: the kube-apiserver now copies topology.kubernetes.io/zone and topology.kubernetes.io/region from the Node onto the Pod at scheduling time. On clusters running this feature, the operator is doing unnecessary Node lookups for information already present on the Pod it already has in hand.

When extracting topology labels for synchronous replica election, read each configured label key from the Pod's own labels first. Only fall back to fetching the Node object when the label is not present on the Pod (i.e. clusters on Kubernetes 1.34 or earlier, or with PodTopologyLabelsAdmission disabled). This:

- Removes the Node dependency entirely for clusters on Kubernetes 1.35+ with the feature enabled.
- Preserves full backward compatibility for older clusters or clusters with the gate disabled, via the existing Node-fetch path.
- Causes no behavior change for existing users — the resolved label values are identical, only the source read changes.

- Keep reading exclusively from Node objects and require KEP-4742 as a hard dependency — rejected, as it would break clusters on Kubernetes < 1.35 or with the gate off.
- Make the pod-vs-node label source explicitly configurable per key — considered as a separate, more advanced follow-up (see the later podFailureDomainKeys/nodeFailureDomainKeys split), but out of scope for this specific change, which is meant to be a transparent optimization/fallback.

### Additional context

Reference: [KEP-4742 — PodTopologyLabelsAdmission](https://github.com/kubernetes/enhancements/issues/4742) (beta, enabled by default in Kubernetes 1.35).

### Backport?

Yes

Yes

### Code of Conduct

gbartolini:
When extracting topology information for synchronous replica election, prefer labels available directly on the Pod object (propagated by Kubernetes via KEP-4742 / PodTopologyLabelsAdmission, beta and enabled by default in Kubernetes 1.35) over fetching the Node object.

The Node fetch is retained as a fallback for clusters running Kubernetes 1.34 or with the feature gate disabled, preserving full backwards compatibility. No behaviour change for existing users.

Closes #11179

### DIFF
--- a/internal/controller/cluster_status.go
+++ b/internal/controller/cluster_status.go
@@ -832,18 +832,26 @@ func getPodsTopology(
 	for _, pod := range pods {
 		podName := apiv1.PodName(pod.Name)
 		data[podName] = make(map[string]string, 0)
-		node, ok := nodes[pod.Spec.NodeName]
-		if !ok {
-			// node not found, it means that:
-			// - the node could have been drained
-			// - others
-			contextLogger.Debug("node not found, skipping pod topology matching")
-			return apiv1.Topology{}
-		}
-
 		nodesMap[pod.Spec.NodeName] = append(nodesMap[pod.Spec.NodeName], podName)
 
 		for _, labelName := range topology.NodeLabelsAntiAffinity {
+			// Prefer pod labels: KEP-4742 propagates node topology labels onto pods
+			// during scheduling, making a node fetch unnecessary on Kubernetes 1.35+.
+			if value, ok := pod.Labels[labelName]; ok {
+				data[podName][labelName] = value
+				continue
+			}
+
+			// Fall back to reading from the node object for older clusters or when
+			// the PodTopologyLabelsAdmission feature gate is disabled.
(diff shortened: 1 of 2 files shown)


gbartolini:
I just realised that the new version of #11144 completely invalidates this. Abandoning.

PR #11207, opened 2026-07-15
fix(metrics): handle NULL stats_reset in pg_stat_wal

### LINKED ISSUE #9106: [Bug]: 1.27.1 stats query bug

HoniSanders:
Since PostgreSQL 18, pg_stat_wal.stats_reset can be NULL. The metrics exporter scanned it into a non-nullable Go string, so metric collection failed with "converting NULL to string is unsupported", breaking Prometheus scrapes on otherwise-healthy clusters.

Coalesce stats_reset to '-infinity' in both pg_stat_wal queries in TryGetPgStatWAL, matching how fillArchiverStatus already handles the nullable last_archived_time and last_failed_time columns in this same file. Keeping the column as timestamptz rather than casting to ::text leaves the metric label format unchanged for non-NULL values.

Closes #9106

----

## Summary

Since PostgreSQL 18, `pg_stat_wal.stats_reset` can be `NULL`. The metrics
exporter scanned it into a non-nullable Go `string`, so metric collection
failed with:

```
sql: Scan error on column index 4, name "stats_reset": converting NULL to string is unsupported
```

This broke Prometheus scrapes on otherwise-healthy clusters running PostgreSQL 18.

## Fix

Coalesce `stats_reset` to `'-infinity'` in both `pg_stat_wal` queries in
`TryGetPgStatWAL`. This matches how `fillArchiverStatus` already handles the
nullable `last_archived_time` / `last_failed_time` columns in this same file
(`COALESCE(col, '-infinity')` into a plain string), so `TryGetPgStatWAL` is no
longer the odd one out scanning a nullable timestamptz directly.

Keeping the column as a `timestamptz` (rather than casting to `::text`) leaves
the metric label format unchanged for non-`NULL` values and is consistent with
the sibling archiver fields.

Thanks to @mnencia for pointing to the existing pattern.

## Testing

- Unit tests cover the coalesced `stats_reset` value scanning into the string
  field for both the pre-18 and 18+ query shapes.
- `go build`, `go vet`, and `go test` pass for `pkg/management/postgres` and
  `pkg/management/postgres/webserver/metricserver`.

### DIFF
--- a/pkg/management/postgres/probes.go
+++ b/pkg/management/postgres/probes.go
@@ -619,7 +619,7 @@ func (instance *Instance) TryGetPgStatWAL() (*PgStatWal, error) {
 			wal_sync,
 			wal_write_time,
 			wal_sync_time,
-			stats_reset
+			COALESCE(stats_reset, '-infinity')
 			FROM pg_catalog.pg_stat_wal`)
 		if err := row.Scan(
 			&pgWalStat.WalRecords,
@@ -643,7 +643,7 @@ func (instance *Instance) TryGetPgStatWAL() (*PgStatWal, error) {
 		wal_fpi,
 		wal_bytes,
 		wal_buffers_full,
-		stats_reset
+		COALESCE(stats_reset, '-infinity')
 	    FROM pg_catalog.pg_stat_wal`)
 		if err := row.Scan(
 			&pgWalStat.WalRecords,
--- a/pkg/management/postgres/probes_test.go
+++ b/pkg/management/postgres/probes_test.go
@@ -156,6 +156,81 @@ var _ = Describe("updateResultForDecrease", func() {
 })
 
 var _ = Describe("probes", func() {
+	Context("pg_stat_wal", func() {
+		// A never-reset instance (or PostgreSQL 18+, where stats_reset starts as
+		// NULL) is coalesced to '-infinity' in the query, so the value always
+		// scans into the plain string field instead of failing on a NULL.
+		It("scans the coalesced stats_reset before PostgreSQL 18", func() {
+			db, mock, err := sqlmock.New(sqlmock.QueryMatcherOption(sqlmock.QueryMatcherEqual))
+			Expect(err).ToNot(HaveOccurred())
+			DeferCleanup(func() {
+				_ = db.Close()
+			})
+
+			const query = "SELECT pg_stat_wal"
+			rows := sqlmock.NewRows([]string{
+				"wal_records",
+				"wal_fpi",
+				"wal_bytes",
+				"wal_buffers_full",
+				"wal_write",
+				"wal_sync",
+				"wal_write_time",
+				"wal_sync_time",
+				"stats_reset",
+			}).AddRow(int64(1), int64(2), int64(3), int64(4), int64(5), int64(6), 7.5, 8.5, "-infinity")
+
+			mock.ExpectQuery(query).WillReturnRows(rows)
+
+			var walStat PgStatWal
+			err = db.QueryRow(query).Scan(
+				&walStat.WalRecords,
+				&walStat.WalFpi,
+				&walStat.WalBytes,
+				&walStat.WALBuffersFull,
+				&walStat.WalWrite,
+				&walStat.WalSync,
+				&walStat.WalWriteTime,
+				&walStat.WalSyncTime,
+				&walStat.StatsReset,
+			)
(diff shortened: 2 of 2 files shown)


mnencia:
One thought on the approach here: `pkg/postgres/status.go`'s `LastArchivedWALTime`/`LastFailedWALTime` handle the same situation (nullable `last_archived_time`/`last_failed_time` in `pg_stat_archiver`) by coalescing at the SQL level (`COALESCE(last_archived_time, '-infinity')`) and keeping the Go field a plain `string`. That's the pattern used everywhere else in this file for nullable Postgres columns; `TryGetPgStatWAL` is the only place scanning directly instead.

`COALESCE(stats_reset::text, '')` in both queries here would get the same result without touching `PgStatWal.StatsReset`'s type or `wal.go` at all, and it sidesteps the infinity/-infinity concern entirely since you'd be coalescing text, not a timestamptz that has to round-trip through a `time.Time`-shaped driver value.

Not a blocker, `sql.NullString` fixes the bug correctly either way. Just flagging the smaller, more consistent option in case you'd rather go that route.

HoniSanders:
Thanks for the suggestion!  I see that `fillArchiverStatus` in this same file is the precedent, and `TryGetPgStatWAL` was the outlier. Switched to `COALESCE(stats_reset, '-infinity')` in both queries and reverted `PgStatWal.StatsReset` back to a plain string (`wal.go` is now untouched).

I went without the `::text` cast on purpose: keeping it a `timestamptz` preserves the exact label format the exporter emits today and matches how `last_archived_time` renders, whereas `stats_reset::text` would switch non-NULL values to Postgres's native text format. 

- Current behavior (non-NULL values today): the exporter scans stats_reset — a `timestamptz` — directly into a Go string. `pgx` hands it back as a `time.Time`, which database/sql renders as RFC3339Nano (`2024-01-01T12:00:00.123456Z`). That's the label format clusters emit right now, and it's also how the sibling last_archived_time field renders, since `COALESCE(col, '-infinity')` keeps the value a `timestamptz`. `COALESCE(stats_reset, '-infinity')` preserves that format exactly.
- `COALESCE(stats_reset::text, '')`: the ::text cast runs server-side, so Postgres emits its native text format (`2024-01-01 12:00:00.123456+00`). That changes the label string for every non-NULL value relative to what's emitted today — a one-time series-identity change for anyone already scraping the metric — and is inconsistent with how `last_archived_time` renders.

Happy to use `COALESCE(stats_reset::text, '')` instead if you'd prefer the empty-string sentinel over -infinity — let me know which you'd like.

HoniSanders:
@mnencia What are the next steps?

HoniSanders:
@jsilvela Can you help push this version of the PR forward?  It is more closely tailored to @mnencia 's suggestions.

mnencia (approved):

NiccoloFei (approved):

PR #11210, opened 2026-07-15
fix(webhook): reject extension paths that escape via absolute traversal

### LINKED ISSUE #11208: [Bug]: Extension path validation accepts absolute paths that escape the extension directory

### Contact Details

### Version

trunk (main)

1.36

YAML manifest

## Summary

The webhook validation for per-extension paths (`extension_control_path`, `dynamic_library_path`, `ld_library_path`, `bin_path`) rejects relative traversal (`..`) but not absolute paths. A crafted absolute path with embedded `..` bypasses the check and, at runtime, resolves to a location outside the extension image volume, contradicting the check's stated intent ("path must not escape the extension directory").

Found by source-code inspection on `main`; environment-independent.

## Root cause

`internal/webhook/v1/cluster_webhook.go:2976`

```go
if strings.HasPrefix(filepath.Clean(path), "..") {
    result = append(result, field.Invalid(
        fieldPath.Index(j),
        path,
        "path must not escape the extension directory",
    ))
    continue
}
```

This only catches relative traversal. `filepath.Clean` on an absolute path stays absolute (starts with `/`), so it never starts with `..`. At runtime, `absolutizePaths` (`pkg/postgres/configuration.go:915`) joins each path with the extension mount:

```go
for _, path := range paths {
    if !yield(filepath.Join(ext.MountPath, path)) {
```

`filepath.Join` strips the leading slash of an absolute path but still resolves embedded `..`, so a path like `/a/../../../../etc` joined with `/extensions/myext` resolves to `/etc`.

## What you expected to happen

The check should reject any path that escapes the extension directory, including absolute paths, matching its stated intent.

## How to reproduce

Running the exact webhook check and the exact runtime join:

```
path = "/a/../../../../etc"
webhook rejects?                              -> false   (Clean is "/etc", not prefixed with "..")
filepath.Join("/extensions/myext", path)      -> "/etc"  (escapes the mount)
```

So a Cluster like the following is accepted by the webhook even though the resulting GUC points outside the extension image volume:

```yaml
spec:
  postgresql:
    extensions:
    - name: myext
      dynamic_library_path:
      - "/a/../../../../lib"
```

## Impact

`extension_control_path` / `dynamic_library_path` control where PostgreSQL loads extension control files and shared libraries. Bypassing this check lets a crafted spec point them outside the read-only extension image volume the feature is meant to confine them to, defeating the isolation the check promises.

This is admin-scoped (whoever sets the Cluster spec is normally trusted), so it is validation hardening / defense-in-depth rather than a critical vulnerability. It still silently breaks a deliberate safety boundary that policy-restricted or GitOps setups may rely on, and the image-volume extensions feature is recent, so this path is lightly audited.

## Suggested fix

Reject absolute paths in addition to `..` traversal, since paths here are always relative to the extension mount (they are joined with `MountPath`):

```go
if filepath.IsAbs(path) || strings.HasPrefix(filepath.Clean(path), "..") {
```

A relative non-`..` path cannot climb out of `filepath.Join`, so rejecting both absolute and `..` fully prevents the escape. Add a test with `/a/../../../../etc` alongside the existing `../../etc` case.

### Cluster resource

```shell
kind: Cluster
metadata:
  name: extension-path-repro
spec:
  instances: 1
  storage:
    size: 1Gi
  postgresql:
    extensions:
    - name: myext
      image:
        reference: example/myext:1.0
      dynamic_library_path:
      - "/a/../../../../lib"
```

### Relevant log output

```shell
N/A — found by source-code inspection. Webhook accepts the spec; at runtime the GUC resolves outside the extension mount via filepath.Join.
```

### Code of Conduct

Anand-240:
`validateExtensions` accepted absolute paths with embedded traversal (e.g. "/a/../../../../etc") because it only checked the cleaned raw path for a ".." prefix. `CollectBinPaths` and `absolutizePaths` join these paths under the extension's mount point at runtime, and that join resolves the embedded ".." against the real mount point, so the accepted value still escaped it.

Resolve each path against "." with `filepath.Join`, then check the result with `filepath.IsLocal`. This is independent of the extension mount point's actual depth, since `IsLocal` guarantees containment against any base rather than requiring the check to mirror the real (multi-segment) mount point. Plain absolute-looking paths (e.g. /opt/custom/lib) keep working, since joining under "." first strips a leading separator the same way the runtime join does.

`CollectLibraryPaths`, `CollectBinPaths`, and `absolutizePaths` apply the same check directly, skipping any entry that fails it. This covers callers that read the Cluster spec without going through admission, such as the major-upgrade Job, and Clusters whose spec was already persisted before this fix.

This check is lexical, on the path string only; it doesn't cover symlinks inside the extension's own image content, which is a separate trust boundary tied to the image reference itself.

Closes #11208

### DIFF
--- a/internal/webhook/v1/cluster_webhook.go
+++ b/internal/webhook/v1/cluster_webhook.go
@@ -2964,8 +2964,6 @@ func (v *ClusterCustomValidator) validateExtensions(r *apiv1.Cluster) field.Erro
 		var result field.ErrorList
 		pathSet := stringset.New()
 
-		const extensionPathBasePlaceholder = "/mount"
-
 		for j, path := range paths {
 			if validateErr := ensureNotEmptyOrDuplicate(
 				fieldPath.Index(j),
@@ -2976,14 +2974,15 @@ func (v *ClusterCustomValidator) validateExtensions(r *apiv1.Cluster) field.Erro
 				continue
 			}
 
-			// Join the path under a placeholder mount point, mirroring what the
-			// instance manager does at runtime, and verify the result stays
-			// within it. This catches embedded traversal such as
-			// "/a/../../etc". The separator boundary is required
-			// to reject paths like "/mount-evil".
-			resolvedPath := filepath.Join(extensionPathBasePlaceholder, path)
-			if resolvedPath != extensionPathBasePlaceholder &&
-				!strings.HasPrefix(resolvedPath, extensionPathBasePlaceholder+string(filepath.Separator)) {
+			// filepath.IsLocal reports whether the path, after joining it under
+			// "." and resolving "..", would stay within that base. Joining
+			// under "." first (rather than a placeholder mount point) makes
+			// this independent of the extension mount point's actual depth:
+			// CollectBinPaths and absolutizePaths join these paths under a
+			// real, multi-segment mount point at runtime, and a placeholder
+			// base shallower than that real path would wrongly accept some
+			// escaping paths (e.g. "../mount/lib").
+			if !filepath.IsLocal(filepath.Join(".", path)) {
 				result = append(result, field.Invalid(
 					fieldPath.Index(j),
 					path,
--- a/internal/webhook/v1/cluster_webhook.go
+++ b/internal/webhook/v1/cluster_webhook.go
@@ -2939,19 +2939,6 @@ func (v *ClusterCustomValidator) validateLivenessPingerProbe(r *apiv1.Cluster) f
 }
 
 func (v *ClusterCustomValidator) validateExtensions(r *apiv1.Cluster) field.ErrorList {
-	// extensionPathContainmentBase is a synthetic placeholder standing in for
-	// an extension's mount point, used only to check that a user-supplied
-	// path stays contained once joined the same way the instance manager
-	// joins it at runtime (see CollectBinPaths and absolutizePaths). The
-	// webhook validates the Cluster spec at admission time, before any Pod
-	// exists, so the real mount point isn't available yet, and it doesn't
-	// need to be: filepath.Join's ".." resolution is purely syntactic, so
-	// whether the result stays under the base depends only on the shape of
-	// the joined path, not on what the base actually is. Any fixed
-	// placeholder therefore gives the same containment answer the real mount
-	// point would.
-	const extensionPathContainmentBase = "/mount"
-
 	ensureNotEmptyOrDuplicate := func(path *field.Path, list *stringset.Data, value string) *field.Error {
 		if value == "" {
 			return field.Invalid(
@@ -2977,6 +2964,8 @@ func (v *ClusterCustomValidator) validateExtensions(r *apiv1.Cluster) field.Erro
 		var result field.ErrorList
 		pathSet := stringset.New()
 
+		const extensionPathBasePlaceholder = "/mount"
+
 		for j, path := range paths {
 			if validateErr := ensureNotEmptyOrDuplicate(
 				fieldPath.Index(j),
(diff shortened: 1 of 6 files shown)


gbartolini:
Thanks for tracking down the bypass. Nice catch on the leading separator interaction with `filepath.Join`.

Here is a basic review, assisted by Claude.

## Suggested approach

One suggestion on the approach. Rather than inferring escape from the raw path string (strip a leading separator, `filepath.Clean`, check for a `..` prefix), it would be more direct to resolve the path the same way the runtime does and check containment on the result:

```go
resolvedPath := filepath.Join(extensionMountPointPlaceholder, path)
if resolvedPath != extensionMountPointPlaceholder &&
    !strings.HasPrefix(resolvedPath, extensionMountPointPlaceholder+string(filepath.Separator)) {
    // reject: escapes the extension's mount point
}
```

where `extensionMountPointPlaceholder` is a fixed synthetic base (e.g. `/mount`) standing in for whatever `CollectBinPaths` / `CollectLibraryPaths` / `absolutizePaths` actually join the path onto at runtime.

A placeholder is used instead of the real value on purpose. The webhook validates the `Cluster` spec at admission time, before any pod exists, and the real base directory (`postgres.ExtensionsBaseDirectory` or `postgres.UpgradeTargetExtensionsBaseDirectory`, joined with the extension name) is only computed later by the instance manager, and even then it differs between the two phases of a major version upgrade. The containment check does not need that real value to be correct: `filepath.Join`'s `..` resolution is purely syntactic, so whether the result stays under the base depends only on the shape of `path`, not on what the base actually is. Any fixed placeholder therefore gives the same answer as the real value would, without tying the webhook to instance manager runtime constants.

## Why this is preferable

- It mirrors the actual runtime join (`filepath.Join(baseDir, ...)`) instead of relying on the indirect (if valid) property that "a cleaned relative path starting with `..` implies escape." That property is correct, but it takes a paragraph of comment to justify. "Join against a sentinel base and check it is still inside" is self evidently correct by inspection. It is the same containment check used for zip slip prevention.
- It drops the manual `strings.TrimPrefix(path, string(filepath.Separator))` step. That step is exactly modeling what `filepath.Join` already does, so we may as well call `filepath.Join` itself rather than reimplementing its leading separator handling by hand.
- One thing to watch: the containment check must compare with a separator boundary (`resolvedPath == base || strings.HasPrefix(resolvedPath, base+"/")`), not a bare `strings.HasPrefix(resolvedPath, base)`. Otherwise `/mount` would wrongly accept something that resolves to `/mount-evil`.

Since all four fields (`extension_control_path`, `dynamic_library_path`, `ld_library_path`, `bin_path`) are joined onto a single fixed base at runtime (just with different base paths depending on the caller), one synthetic sentinel base works uniformly for the webhook check.

## Worth noting on the docs side

Per `docs/src/imagevolume_extensions.md`, all four of these fields are documented as taking **relative** paths within the container image:

> - `extension_control_path`: A list of relative paths within the container image to be appended to PostgreSQL's `extension_control_path`...
> - `dynamic_library_path`: A list of relative paths within the container image to be appended to PostgreSQL's `dynamic_library_path`...
> - `ld_library_path`: A list of relative paths within the container image to be appended to the `LD_LIBRARY_PATH` environment variable...
> - `bin_path`: A list of relative paths within the container image to be appended to the `PATH` environment variable...

So an absolute-looking value such as `/opt/custom/lib` is already outside the documented contract. It is only accepted today because `CollectBinPaths` and friends treat a leading `/` as equivalent to no leading separator when joining. That equivalence is what both the current fix and the containment based approach need to preserve for backward compatibility, but it is worth keeping in mind that the "real" contract is relative paths only, and the validation exists to make sure that contract cannot be subverted regardless of how the value is spelled.

## Tests

The two new test cases you added (embedded traversal rejection, leading `/` without escape acceptance) are exactly the right ones to keep. They would just need to double as regression tests for the containment based implementation too.

Anand-240:
Updated to the containment check with the placeholder base, and fixed the separator boundary issue you pointed out. Kept both test cases since you're right that they work as regression tests either way. Pushed.

gbartolini (approved):

gbartolini:
Thanks a lot @Anand-240, looks good to me!

internal/webhook/v1/cluster_webhook.go
```diff
@@ -61,6 +61,18 @@ const sharedBuffersParameter = "shared_buffers"
...
+// extension's mount point, used only to check that a user-supplied path
+// stays contained once joined the same way the instance manager joins it at
+// runtime (see CollectBinPaths and absolutizePaths). The webhook validates
+// the Cluster spec at admission time, before any Pod exists, so the real
+// mount point isn't available yet, and it doesn't need to be: filepath.Join's
+// ".." resolution is purely syntactic, so whether the result stays under the
+// base depends only on the shape of the joined path, not on what the base
+// actually is. Any fixed placeholder therefore gives the same containment
+// answer the real mount point would.
+const extensionPathContainmentBase = "/mount"
```
GabriFedi97: I think we can move the definition of this const directly to the `validateExtensions` function as that's the only place where we are using it.

internal/webhook/v1/cluster_webhook_test.go
```diff
@@ -6125,6 +6125,91 @@ var _ = Describe("validateExtensions", func() {
...
+				PostgresConfiguration: apiv1.PostgresConfiguration{
+					Extensions: []apiv1.ExtensionConfiguration{
+						{
+							Name: "extOne",
+							ImageVolumeSource: corev1.ImageVolumeSource{
+								Reference: "extOne",
+							},
+							ExtensionControlPath: []string{
+								"/opt/custom/share",
+							},
```
GabriFedi97, line 6164: We could add a test-case showing that a path containing `../` can still be valid as long as it doesn't escape the extension directory. 

```suggestion
							ExtensionControlPath: []string{
								"/opt/custom/share",
							},
							DynamicLibraryPath: []string{
								"/a/b/../../share",
							},
```

internal/webhook/v1/cluster_webhook.go
Anand-240: Moved the const local to validateExtensions, and added the a/b/../../share test case showing internal .. is fine as long as it doesn't escape. Pushed in 26606e9.

internal/webhook/v1/cluster_webhook_test.go
Anand-240: Added that DynamicLibraryPath case with a/b/../../share, same as your suggestion. Pushed in 26606e9.

NiccoloFei (approved):

GabriFedi97 (approved):

mnencia (approved):

Issue #10924, opened 2026-06-16
[Bug]: High memory usage/OOMKill of operator if there are many large secrets in the namespace


bergdavi:
### Is there an existing issue already for this bug?

- [x] I have searched for an existing issue, and could not find anything. I believe this is a new bug.

### I have read the troubleshooting guide

- [x] I have read the troubleshooting guide and I think this is a new bug.

### I am running a supported version of CloudNativePG

- [x] I have read the troubleshooting guide and I think this is a new bug.

### Contact Details

david.berghammer@ericsson.com

### Version

1.29 (latest patch)

### What version of Kubernetes are you using?

1.35

### What is your Kubernetes environment?

Self-managed: kind (evaluation)

### How did you install the operator?

YAML manifest

### Issue

When there are many large secrets in a namespace that the CNPG operator watches the memory usage of the operator becomes high, and eventually reaches the configured limit and gets OOMKilled.

This can realistically happen if there is a long helm release history of large charts in the namespace (e.g. 400 chart versions stored). It could probably happen with other kind of resources that the operator watches, but helm secrets is a typical scenario of many large resources.

### Steps to reproduce

1. Install CNPG operator based on the quick start guide (operand not needed)
```
kubectl apply --server-side -f \
  https://raw.githubusercontent.com/cloudnative-pg/cloudnative-pg/release-1.29/releases/cnpg-1.29.1.yaml
```
2. Check operator memory usage
```
kubectl top -n cnpg-system pod
NAME                                       CPU(cores)   MEMORY(bytes)
cnpg-controller-manager-7d9fd94fb7-xtjfc   2m           15Mi
```
3. Create many large secrets
```
#!/bin/bash
# Create 200 large fake helm release secrets (~100KB each)
# Helm release secrets have label "owner=helm" and type "helm.sh/release.v1"

NAMESPACE="cnpg-system"
LARGE_DATA=$(head -c 100000 /dev/urandom | base64 | tr -d '\n')

for i in $(seq 1 200); do
  kubectl -n "$NAMESPACE" create secret generic "sh.helm.release.v1.fake-release.v${i}" \
    --from-literal=release="$LARGE_DATA" \
    --dry-run=client -o yaml | \
    kubectl label --local -f - owner=helm app.kubernetes.io/managed-by=Helm --dry-run=client -o yaml | \
    kubectl apply -f - > /dev/null 2>&1
  if [ $((i % 50)) -eq 0 ]; then
[...]

PR #11211, opened 2026-07-16
fix: stop caching Helm release secrets in the operator

leonardoce:
Clusters with large Helm releases installed alongside CloudNativePG were seeing the operator's memory usage spike and, in some cases, crashloop. The operator's controller-runtime cache was keeping a full copy of every Secret in the watched namespace(s), including the large Helm release manifests Helm stores as Secrets, even though the operator never reads them.

Helm release Secrets are now excluded from the cache, so they are no longer listed, watched, or held in memory. This covers the Helm release scenario specifically; the general case of large Secrets of other types filling the cache is not addressed here.

Refs #10924


GellertBabel:
It works, code looks clean.

gbartolini (approved):

NiccoloFei (approved):

mnencia (approved):

PR #11213, opened 2026-07-16
test(e2e): stop losing transient-error retries to a conflict-only predicate

mnencia:
client-go's retry.RetryOnConflict retries only on HTTP 409 Conflict, so a
transient webhook-proxy 500 (the AKS Konnectivity flake that tests/utils/objects
already retries around for Patch/PatchStatus/Update) passes straight through
instead of being retried, and these read-modify-write loops bail on the
first hit. RetryOnConflict was chosen assuming the only failure mode was a
concurrent-write race; it silently gives zero protection against the other one.

Export the same transient-write predicate objects.Patch/Update rely on,
combined with a Conflict check, and switch these loops from
retry.RetryOnConflict to retry.OnError with it. Each loop already re-fetches
and reapplies its mutation on every attempt, so retrying on either condition
is safe.

### DIFF
--- a/tests/e2e/backup_restore_objectstore_test.go
+++ b/tests/e2e/backup_restore_objectstore_test.go
@@ -739,7 +739,7 @@ var _ = Describe("Object storage - Backup and restore", Label(tests.LabelBackupR
 			})
 
 			By("scaling first cluster to 2 instances", func() {
-				err := retry.RetryOnConflict(retry.DefaultBackoff, func() error {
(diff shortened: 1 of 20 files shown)


NiccoloFei (approved):

NiccoloFei:
Full E2E run (with cloud-providers) here: https://github.com/EnterpriseDB/cloudnative-pg/actions/runs/29507416609

fcanovai (approved):

PR #11215, opened 2026-07-16
fix(controller): re-run owned resource finalizer cleanup after a restart

armru:
The Database, Publication and Subscription finalizers are removed either by their reconcilers in the instance manager when the object itself is deleted, or by the cluster controller through `notifyDeletionToOwnedResources` when the owning Cluster is deleted. The latter is the only path left once the instance pods are gone, and it runs exclusively while the cluster controller is reconciling the deleted Cluster.

Nothing adds a finalizer to the Cluster object, so when a namespace holding the Cluster and these resources is deleted the Cluster disappears at once. If the operator is not running during that window, on restart it never observes the Cluster again: the object is gone, the cluster controller does not watch these resources, and every other object it maps has already been removed. The cleanup is never triggered, the finalizers remain, and the namespace hangs in Terminating until someone removes them by hand.

This change makes the cluster controller watch the owned Database, Publication and Subscription resources, restricted to the objects that are being deleted. On the initial cache sync after a restart the lingering resources are delivered as create events, which enqueue a reconcile for their referenced Cluster and let `notifyDeletionToOwnedResources` strip the finalizers. The reference is resolved by name rather than through an owner reference, so it still points at the Cluster after the object itself is gone.

Refs #10535

### DIFF
--- a/internal/controller/cluster_controller.go
+++ b/internal/controller/cluster_controller.go
@@ -1346,6 +1346,11 @@ func (r *ClusterReconciler) SetupWithManager(ctx context.Context, mgr ctrl.Manag
 			&apiv1.Subscription{},
 			handler.EnqueueRequestsFromMapFunc(mapClusterOwnedResourceToCluster),
 			builder.WithPredicates(isBeingDeletedPredicate),
+		).
+		Watches(
+			&apiv1.DatabaseRole{},
+			handler.EnqueueRequestsFromMapFunc(mapClusterOwnedResourceToCluster),
+			builder.WithPredicates(isBeingDeletedPredicate),
 		)
 
 	if configuration.Current.OperatorNamespace != "" {
@@ -1628,11 +1633,9 @@ func (r *ClusterReconciler) mapDatabaseRolesToClusters() handler.MapFunc {
 }
 
 // mapClusterOwnedResourceToCluster maps a namespaced resource that references its
-// Cluster through the `cluster` field (Database, Publication, Subscription) to a
-// reconcile request for that Cluster. It resolves the reference by name instead of
-// through an owner reference, so it keeps working after the Cluster object is gone:
-// that is what allows notifyDeletionToOwnedResources to strip a lingering finalizer
-// when the operator sees the resource again after a restart.
+// Cluster through the `cluster` field to a reconcile request for that Cluster.
+// It resolves the cluster reference by the spec instead of through an owner reference,
+// so it keeps working after the Cluster object is gone.
 func mapClusterOwnedResourceToCluster(_ context.Context, obj client.Object) []reconcile.Request {
 	owned, ok := obj.(interface {
 		GetClusterRef() corev1.LocalObjectReference
--- a/internal/controller/cluster_controller.go
+++ b/internal/controller/cluster_controller.go
@@ -1323,6 +1323,29 @@ func (r *ClusterReconciler) SetupWithManager(ctx context.Context, mgr ctrl.Manag
 			// The cluster only consumes spec.passwordSecret (to maintain the
 			// instance RBAC), so status-only changes are irrelevant here.
 			builder.WithPredicates(predicate.GenerationChangedPredicate{}),
+		).
+		// Watch the owned Database, Publication and Subscription resources only
+		// while they are being deleted. Their reconcilers run in the instance
+		// manager, so when the cluster is torn down together with its pods the
+		// finalizer can only be removed by notifyDeletionToOwnedResources during
+		// the cluster's own deletion. If the operator is down while the namespace
+		// is deleted, the Cluster object is gone and nothing would ever re-trigger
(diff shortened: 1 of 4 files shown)


internal/controller/cluster_predicates.go
```diff
@@ -74,6 +74,15 @@ var (
 			return isUsefulClusterSecret(e.ObjectNew)
 		},
 	}
+
+	// isBeingDeletedPredicate admits only objects that already carry a
+	// deletionTimestamp. The cluster controller watches the owned Database,
+	// Publication and Subscription resources solely to re-run the deletion
+	// cleanup after a restart, so resources that are not being deleted must not
+	// enqueue a Cluster reconcile.
```
GabriFedi97: > The cluster controller watches the owned Database,
	Publication and Subscription resources solely to re-run the deletion
	cleanup after a restart, so resources that are not being deleted must not
	enqueue a Cluster reconcile.

I think we shouldn't limit the comment here for this function to why it was added in the first place.


```suggestion
	// isBeingDeletedPredicate admits only objects that already carry a
	// deletionTimestamp.
```

internal/controller/cluster_controller.go
```diff
@@ -1604,6 +1627,34 @@ func (r *ClusterReconciler) mapDatabaseRolesToClusters() handler.MapFunc {
 	}
 }
 
+// mapClusterOwnedResourceToCluster maps a namespaced resource that references its
+// Cluster through the `cluster` field (Database, Publication, Subscription) to a
+// reconcile request for that Cluster. It resolves the reference by name instead of
+// through an owner reference, so it keeps working after the Cluster object is gone:
+// that is what allows notifyDeletionToOwnedResources to strip a lingering finalizer
+// when the operator sees the resource again after a restart.
```
GabriFedi97:
```suggestion
// mapClusterOwnedResourceToCluster maps a namespaced resource that references its
// Cluster through the `cluster` field to a reconcile request for that Cluster.
// It resolves the cluster reference by the spec instead of through an owner reference,
// so it keeps working after the Cluster object is gone.
```

internal/controller/cluster_controller.go
```diff
@@ -1323,6 +1323,29 @@ func (r *ClusterReconciler) SetupWithManager(ctx context.Context, mgr ctrl.Manag
...
+		// manager, so when the cluster is torn down together with its pods the
+		// finalizer can only be removed by notifyDeletionToOwnedResources during
+		// the cluster's own deletion. If the operator is down while the namespace
+		// is deleted, the Cluster object is gone and nothing would ever re-trigger
+		// that cleanup after a restart; these watches deliver the lingering
+		// resources on the initial cache sync so the cleanup runs.
+		Watches(
+			&apiv1.Database{},
+			handler.EnqueueRequestsFromMapFunc(mapClusterOwnedResourceToCluster),
+			builder.WithPredicates(isBeingDeletedPredicate),
```
GabriFedi97, line 1343: While I understand this is here as an optimization, I believe it has an important implication. I noticed it can make the operator behaving differently in case it is running during the cluster deletion and in case it's not running.
In the first case, a reconciliation loop, enqueued for the cluster deletion, takes care of removing the finalizer and updating the status of the watched resources immediately , in the second case, it is deferred to the event of the watched resource deletion.

internal/controller/cluster_controller.go
```diff
@@ -1323,6 +1323,29 @@ func (r *ClusterReconciler) SetupWithManager(ctx context.Context, mgr ctrl.Manag
...
+		Watches(
+			&apiv1.Publication{},
+			handler.EnqueueRequestsFromMapFunc(mapClusterOwnedResourceToCluster),
+			builder.WithPredicates(isBeingDeletedPredicate),
+		).
+		Watches(
+			&apiv1.Subscription{},
+			handler.EnqueueRequestsFromMapFunc(mapClusterOwnedResourceToCluster),
+			builder.WithPredicates(isBeingDeletedPredicate),
 		)
```
GabriFedi97, line 1354: Any reason why the `DatabaseRole` has been left out of this path?

GabriFedi97 (approved):

internal/controller/cluster_controller.go
GabriFedi97: Resolved by https://github.com/cloudnative-pg/cloudnative-pg/pull/11215/changes/8aa21e5c3f8b43fe35dfbedb171366f69cf34b1f

internal/controller/cluster_controller.go
GabriFedi97: Resolved by https://github.com/cloudnative-pg/cloudnative-pg/pull/11215/changes/8aa21e5c3f8b43fe35dfbedb171366f69cf34b1f

litaocdl (approved):

mnencia (approved):

PR #11231, opened 2026-07-20
feat: bootstrap new instances in-process instead of with Jobs

### LINKED ISSUE #11228: [Feature]: Bootstrap PostgreSQL instances without a separate Job or init container

Today, creating a new PostgreSQL instance (a primary via initdb/recovery/pgbasebackup, or a replica via join/snapshot-restore) goes through a `batchv1.Job` that runs `instance <init|join|restore|restoresnapshot|pgbasebackup>` against the PVC, and only once that Job completes does the operator create the actual instance Pod that runs postgres.

This extra Job stage carries real cost. While any bootstrap Job for a cluster is running, the operator's `reconcileResources` loop treats it as blocking and requeues the Pod/PVC part of reconciliation until it's done; that blocking guard, plus the separate Job-creation step, keeps inviting a class of PVC/Job split-brain deadlocks, e.g. #10985 (a join Job's Pod could never schedule, so the guard never cleared) and #11036 (a status update conflicting before the initdb Job existed wedged first-primary bootstrap). Both are fixed now, but each needed its own targeted self-heal; a single-container design removes the guard and the Job-creation window those fixes work around. Completed Jobs aren't given a TTL and have to be swept up by a separate cleanup pass. PVC readiness has to be tracked out-of-band by matching PVCs to Job completion, which is its own piece of bookkeeping.

I looked at an alternative where the bootstrap step is moved into an init container on the instance Pod itself, so there's no separate Job object. That removes the Job, but the extra machinery mostly moves rather than disappears: PVC-readiness now has to be derived from the init container's termination status on the Pod instead of Job completion, the operator still has to patch that PVC annotation from outside the Pod (the instance's own ServiceAccount has no RBAC on PVCs or Pods), and the Pod spec annotation used to detect "does this Pod need updating" has to be computed with the bootstrap init container stripped out again, otherwise a completed bootstrap looks like a spec drift and triggers a rolling update.

Bootstrap the instance directly from the main container, with no separate Job and no separate init container. The instance manager already has `init`/`join`/restore logic factored as reusable library code (`InitInfo.Bootstrap`, `InitInfo.Join`, and the restore equivalents) rather than shell-outs to a subcommand, so `instance run` could detect an uninitialized PGDATA on startup and run the appropriate bootstrap step in-process before starting postgres, instead of requiring a prior process/container to have done it.

This would remove the Job entirely, remove the PVC-readiness tracking that only exists to compensate for the two-stage creation, and collapse instance creation down to "create one Pod, the Pod initializes itself." I'd like to use this issue to discuss the design before writing any code: what bootstrap-mode information the Pod needs at creation time instead of being told a Job's command afterwards, how failures should surface (an instance stuck failing to bootstrap forever vs. today's Job backoff-driven `PhaseUnrecoverable`), and whether there's anything the two-stage approach is quietly relying on that a single-container design would need to replace.

Moving bootstrap into an init container on the instance Pod instead of a Job, described above: it drops the Job object but relocates rather than removes the complexity, so it isn't a clear enough win to be worth the churn on its own.

### Additional context

_No response_

### Backport?

No

Yes

### Code of Conduct

armru:
Bootstrapping an instance today goes through a dedicated `batch/v1` Job per mode (`<instance>-initdb`, `-join`, `-pgbasebackup`, `-full-recovery`, `-snapshot-recovery`): the operator creates the PVCs, waits for the Job to initialize them, marks them ready, and only then creates the instance Pod. Every bootstrap therefore pays two pod schedules and image pulls, reconciliation stays blocked while the Job runs, and the operator has to coordinate two consumers of the same PVCs.

With this change the operator creates the instance Pod directly. When the pgdata PVC is not marked ready yet, the Pod carries a bootstrap overlay: `instance run` receives `--bootstrap-mode=<mode>` plus exactly the flags the Job builders used to emit, and the Pod is annotated with `cnpg.io/bootstrapInstance` for observability. Before starting the instance manager, `instance run` initializes PGDATA in-process with the same library code the `instance <mode>` subcommands are built on, while a minimal webserver keeps the kubelet probes and `wal-restore` working: the startup probe is answered during the whole data copy instead of timing out, readiness stays negative, and `/pg/status` answers 503 so the operator sees the instance as not reporting yet. The overlay is deliberately left out of the pod-template annotation, so a freshly bootstrapped Pod is not considered drifted and no rollout is triggered.

Completion is tracked at two levels. A marker file inside PGDATA records the mode and the instance identity (namespace, cluster name and UID, pod name): a container restarting mid-bootstrap finds no marker and retries the mode from scratch (except snapshot recovery, which resumes in place, since a snapshot-seeded PGDATA cannot be re-obtained), while a marker that traveled inside a volume snapshot belongs to a different identity and is ignored, so a snapshot-cloned PVC is bootstrapped properly instead of being mistaken for an already-initialized one. The `cnpg.io/pvcStatus` annotation remains the operator-side memory, and now flips to `ready` only when the owning Pod becomes Ready.

User-visible effects: no bootstrap Jobs are created anymore (the `pg_upgrade` Job is unchanged); bootstrap progress and failures surface on the instance Pod itself, in its logs and restart count; and since the Pod's hostname is the instance name from the start (the Jobs used the Job name), `application_name` and everything else derived from `POD_NAME` is correct during the initial data copy too. Recovery sources protected by a private CA no longer rely on a Secret mounted over `/controller/certificates`: the bootstrap writes the CA file itself, which also removes a mount that would have shadowed the directory where the instance manager maintains the server certificates.

Operator upgrades need no migration: the Job read/guard/cleanup paths are all kept, so an in-flight legacy bootstrap Job still blocks reconciliation, still marks its PVC ready on completion, and still surfaces its failure; from then on those paths only ever see the `pg_upgrade` Job.

Recovery through a CNPG-I plugin now runs inside the instance Pod, so plugins must serve the restore protocol from their instance sidecar (there is no restore Job carrying a dedicated sidecar anymore): cloudnative-pg/plugin-barman-cloud#1025 is the companion change. Removing the now-unused Job builders and the five bootstrap subcommands, and reworking how a failed bootstrap reaches `PhaseUnrecoverable` without a Job `backoffLimit`, are left to follow-ups; the failure policy is the open question tracked on the issue.

The e2e bootstrap assertions are retargeted from Jobs to Pods and now also verify that no bootstrap Job is ever created; a `walStorage` variant covers the separate WAL volume layout.

Closes #11228


mnencia (approved):

mnencia:
Abandoned as superseded by #11319

PR #11241, opened 2026-07-22
fix: honor unrecoverable annotation on pending and terminating instances

### LINKED ISSUE #10926: [Bug]: `alpha.cnpg.io/unrecoverable` annotation has no effect on Terminating or Pending pods

### Contact Details

daanvinken@tythus.com

### Version

1.29 (latest patch)

1.36

Other

YAML manifest

### What happened

The `alpha.cnpg.io/unrecoverable` annotation is ineffective for pods in Terminating or Pending state. These are the two most common states during node maintenance with local storage (e.g., OpenEBS LVM):

1. **Unreachable node:** pods get a `DeletionTimestamp` (Terminating) that the kubelet can't confirm, so they stay Terminating indefinitely.
2. **Drained node with local storage:** pods get evicted and recreated, but the new pod can't be scheduled because the PVC is bound to the cordoned node. The pod sits in Pending.

In both cases, applying `alpha.cnpg.io/unrecoverable=true` has no effect because the reconciliation loop exits before reaching `reconcileUnrecoverableInstances`:

1. `getManagedInstances` includes Terminating and Pending pods (no filter)
2. [`allInstancesAreActive()`](https://github.com/cloudnative-pg/cloudnative-pg/blob/main/internal/controller/cluster_controller.go#L863) returns false because [`IsPodActive`](https://github.com/cloudnative-pg/cloudnative-pg/blob/main/pkg/utils/pod_conditions.go#L72-L77) rejects both `DeletionTimestamp != nil` and `PodPending` phase
3. The reconciler returns early with a 1-second requeue, never reaching [`reconcilePods`](https://github.com/cloudnative-pg/cloudnative-pg/blob/main/internal/controller/cluster_controller.go#L1046) → [`reconcileUnrecoverableInstances`](https://github.com/cloudnative-pg/cloudnative-pg/blob/main/internal/controller/cluster_controller.go#L1089)

### Relationship to #9947 / #9968

PR #9968 solved a related but different problem: pods that are not ready but still active (e.g., postgres not running, startup probe failing). That fix moved `reconcileUnrecoverableInstances` before the non-ready gate inside `reconcilePods`. However, the [`allInstancesAreActive()`](https://github.com/cloudnative-pg/cloudnative-pg/blob/main/internal/controller/cluster_controller.go#L863) gate runs before `reconcilePods` is called, so Terminating and Pending pods are still blocked.

### Expected behavior

Annotating a Terminating or Pending pod with `alpha.cnpg.io/unrecoverable=true` should cause the operator to delete the pod and its PVCs, allowing a new instance to be created on a healthy node. The PVCs will go Terminating but are already [ignored by the instance counter](https://github.com/cloudnative-pg/cloudnative-pg/blob/main/pkg/reconciler/persistentvolumeclaim/status.go#L201) due to `DeletionTimestamp` filtering.

For the Terminating variant, the `pvc-protection` finalizer holds PVC deletion while the pod object exists, so the implementation would also need to force-remove the pod (with appropriate safety checks if it was a primary).

### Possible approaches

1. Move `reconcileUnrecoverableInstances` before the `allInstancesAreActive()` gate (similar to what #9968 did for the non-ready gate)
2. Exclude pods annotated with `alpha.cnpg.io/unrecoverable=true` from the `allInstancesAreActive()` check

**I'm happy to create/contribute myself once we settle on an approach.**

### Environment

- CNPG 1.28.x, verified against latest `main` (`946085b0b`)
- Bare-metal Kubernetes with OpenEBS LVM local storage (~200 clusters)

### Related

- #9947
- #9968
- [My comment on #9968](https://github.com/cloudnative-pg/cloudnative-pg/pull/9968#issuecomment-4354608774)

### Cluster resource

```shell
NA
```

### Relevant log output

```shell
None, not triggered.
```

### Code of Conduct

armru:
The alpha.cnpg.io/unrecoverable annotation deletes an instance's Pod and PVCs so the instance is recreated, but it did nothing while the Pod was Pending or Terminating. reconcileResources returns early at the active-instances gate whenever an instance is not active, and the unrecoverable handling lived downstream of that gate (in reconcilePods), so it was never reached for exactly the states where a stuck instance most needs it. A Terminating Pod needs more than reaching the delete: a graceful delete is a no-op on it, so its PVCs stay blocked by the pvc-protection finalizer.

This moves the unrecoverable handling ahead of the active-instances gate, alongside the unschedulable-instances handling that already runs there for the same structural reason. A Pod stuck Terminating past its deletion deadline is force-removed first so its PVCs can be released, and deleting an unrecoverable instance now emits a Kubernetes event. The current primary and the designated target primary stay excluded in every state.

Closes #10926

### DIFF
--- a/docs/src/labels_annotations.md
+++ b/docs/src/labels_annotations.md
@@ -303,8 +303,19 @@ CloudNativePG manages the following predefined annotations:
 :   Experimental annotation applied to a `Pod` running a PostgreSQL instance.
     It instructs the operator to delete the `Pod` and all its associated PVCs.
     The instance will then be recreated according to the configured join
-    strategy. This annotation can only be used on instances that are neither the
-    current primary nor the designated target primary.
+    strategy. The annotation is honored regardless of the instance's state, so
+    it also takes effect on a `Pod` that is Pending, Running but not ready, or
+    Terminating. When a Terminating `Pod` has passed its deletion deadline (the
+    deletion request time plus its termination grace period), the operator force
+    removes it so that its PVCs can be deleted and the instance recreated. This
+    annotation can only be used on instances that are neither the current primary
+    nor the designated target primary, which stay excluded even when marked
+    unrecoverable.
+
+    **⚠️ WARNING:** Force removing a `Pod` on an unreachable node does not stop
+    any process that might still be running there, and whether the underlying
+    volumes are detached depends on the storage driver. This is acceptable for
+    this annotation, whose contract already forfeits the instance's data.
 
 ## Prerequisites
 
--- a/internal/controller/cluster_controller.go
+++ b/internal/controller/cluster_controller.go
@@ -886,6 +886,16 @@ func (r *ClusterReconciler) reconcileResources(
 		return *result, err
 	}
 
+	// Handle instances explicitly marked as unrecoverable before the active-instances
+	// gate below. A Pod that is Pending or Terminating is never "active", so the gate
+	// would otherwise return early and the alpha.cnpg.io/unrecoverable annotation would
+	// have no effect on exactly those states. This mirrors the unschedulable-instances
+	// handling above, which sits before the same gate for the same structural reason
+	// (its target Pods are Pending too).
+	if res, err := r.reconcileUnrecoverableInstances(ctx, cluster, resources); !res.IsZero() || err != nil {
+		return res, err
+	}
+
 	if !resources.allInstancesAreActive() {
 		contextLogger = contextLogger.WithValues(
 			"inactiveInstances", resources.inactiveInstanceNames())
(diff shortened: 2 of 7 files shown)


daanvinken (approved):
(creater of issue)

Thanks for taking the time on this, I don't see anything worth reconsidering here. Glad we're calling out the accepted risks that something may still be running there, it's indeed up to the (human) operator to work with this annotation.

litaocdl (approved):

GabriFedi97 (approved):

leonardoce (approved):

docs/src/labels_annotations.md
```diff
@@ -303,8 +303,19 @@ CloudNativePG manages the following predefined annotations:
...
+    strategy. The annotation is honored regardless of the instance's state, so
+    it also takes effect on a `Pod` that is Pending, Running but not ready, or
+    Terminating. When a Terminating `Pod` has passed its deletion deadline (the
+    deletion request time plus its termination grace period), the operator force
+    removes it so that its PVCs can be deleted and the instance recreated. This
+    annotation can only be used on instances that are neither the current primary
+    nor the designated target primary, which stay excluded even when marked
+    unrecoverable.
+
+    **⚠️ WARNING:** Force removing a `Pod` on an unreachable node does not stop
```
gbartolini, line 315: We should have used the docusaurus admonitions here

PR #11243, opened 2026-07-23
fix #963 accept null postgres UID GID

### LINKED ISSUE #963: cluster does not allow null PostgresUID/PostgresGID
# Issue

cluster forces a default `PostgresUID/PostgresGID = 26` (see https://github.com/cloudnative-pg/cloudnative-pg/blob/main/pkg/specs/pods.go#L456). It is impossible to remove this default. One can set a value >= 0, but not null.

# Expected behavior

There should be a way to set the values such that any containers should be created without SCC.

# Why this matter

In non-root environment like OpenShift/OKD, with restricted-v2/v3, it is not possible to set UIDs. So a pod without SCC will inherit one with a forced `UID/GID/fsGroup` in a dynamic range depending on the namespace, so it cannot be hard-coded in `values.yaml`. Right now, it is not possible to deploy a cluster of cnpg inside such a cluster.

# Suggestion of fix
1. in go code
    1. Check uid/git at https://github.com/cloudnative-pg/cloudnative-pg/blob/main/pkg/specs/pods.go#L464C18-L464C37
If null, we ignore these values, so `RunAsUser` `RunAsGroup` `FSGroup` are not set in the `defaultContext`
    2. Also check uid/git not null at https://github.com/cloudnative-pg/cloudnative-pg/blob/v1.30.0/pkg/specs/pods.go#L471

1. in helm cluster chart
    1. Make crd `postgresUID` `postgresGID` optional (so they can be null) and remove the default value of `26`
    2. To avoid regression with other environments, the default `values.yaml` should contains the default `postgresUID`, `postgresGID` set to `26`

There is a case of regression for people setting `cluster.useTimescaleDBDefaults` to `timescaledb`. Somehow, this requires UID/GID=1000. These people should write the UID/GID in their own `values.yaml`.

Please tell me if you have another suggestion. One alternative would be to create a new field like `openshiftmode = true` like in some other chart, but it is a bit uglier because it is in contradiction with non null UID/GID.

### LINKED ISSUE #11245: [Bug]: accept null postgres UID GID

### Contact Details

_No response_

### Version

1.30.0

1.36

Other

Helm

### What happened?

Impossible to deploy in OpenShift/OKD with restricted-v3 SCC. It requires podSecurityContext.runAsUser, runAsGroup, fsGroup to not be defined, so that the Kubernetes will apply automatically some values to it.

### Cluster resource

```shell

```

### Relevant log output

```shell

```

### Code of Conduct

# Additional info

As dosu mentioned, the current code normally automatically detect if we are in OpenShift mode. However this detection is based on apiresource `security.openshift.io/v1`, which is not exposed in my case, of a vcluster inside an OpenShift cluster. The vcluster brings a virtual Kubernetes which allows when installed in a shared Kubernetes cluster the installation of cluster scope like CRDs, which is not normally allowed at host level.

antoinetran:
Fixes cloudnative-pg/charts#963 https://github.com/cloudnative-pg/charts/issues/963
Fixes #11245

Description: accept `-1` as value for UID/GID, it has a special meaning: `runAsUser` `runAsGroup` `fsGroup` will not be defined.

### DIFF
--- a/pkg/specs/pods.go
+++ b/pkg/specs/pods.go
@@ -461,6 +461,17 @@ func GetPodSecurityContext(cluster *apiv1.Cluster) *corev1.PodSecurityContext {
 		FSGroup:        &gid,
 	}
 
+	// Special case meaning we want to nil the nil (for OpenShift-like cluster who assign a value >0 automatically)
+	if uid = -1 {
+		defaultContext.RunAsUser = nil
+	}
+	
+	// Special case meaning we want to nil the group (for OpenShift-like cluster who assign a value >0 automatically)
+	if gid = -1 {
+		defaultContext.RunAsGroup = nil
+		defaultContext.FSGroup = nil
+	}
+
 	if cluster.Spec.PodSecurityContext == nil {
 		return defaultContext
 	}


antoinetran:
Fix not needed with the workaround https://github.com/cloudnative-pg/cloudnative-pg/issues/11245#issuecomment-5118254672

PR #11246, opened 2026-07-23
fix: recover log streaming when a stale non-FIFO blocks the log pipe

### LINKED ISSUE #11201: [Bug]: wal-archive/wal-restore log destination can be silently and permanently downgraded from a FIFO to a dead regular file

### Contact Details

_No response_

### Version

trunk (main)

1.34

YAML manifest

### What happened?

`wal-archive` and `wal-restore` (invoked by PostgreSQL as `archive_command`/`restore_command`, and by `pg_rewind --restore-target-wal`) are given `--log-destination <path>` pointing at a path the instance manager expects to be a FIFO, tailed by a `logpipe.LineLogPipe` reader and converted into the structured JSON stream visible in `kubectl logs`. Creating that FIFO is conditional: it only happens if the path doesn't already exist (that check lives in `cloudnative-pg/machinery`; see cloudnative-pg/machinery#298). Nothing in `cloudnative-pg/cloudnative-pg` ever confirms the FIFO actually exists before code that could write to that path runs.

Two places start these readers, and neither synchronizes on their readiness. `pkg/management/postgres/instance.go`'s `WithActiveInstance` (bootstrap only, used by `initdb.go`/`restore.go`) launches the readers as bare goroutines and calls `instance.Startup()` right after, with no wait in between; its own comment describes the hazard ("Without these readers nothing calls Mkfifo on the log paths, the writers fall back to creating regular files, and their log output is silently lost"), but starting the goroutines a few statements earlier only narrows the window, it doesn't close it. `LineLogPipe` exposes a condition (`GetExecutedCondition()`) that's broadcast right after the FIFO is created, which is exactly what a caller would wait on to close this window, but no caller anywhere in the codebase does. The steady-state path, `internal/cmd/manager/instance/run/cmd.go`, has an even wider opportunity: it registers the same readers and the `InstanceReconciler` as independent controller-runtime Runnables via separate `mgr.Add(...)` calls, with substantially more setup code in between and no ordering at all.

Losing that race even once is enough to break log capture for the path's entire remaining lifetime, not just for the one invocation that raced. `postgres.LogPath` lives under `ScratchDataDirectory` (`/controller`), backed by the pod's `scratch-data` `EmptyDir` volume, which survives container restarts and is only wiped when the pod itself is deleted and recreated. Nothing in the codebase cleans that directory on a container restart. So once any writer creates a plain file at the log-destination path instead of a FIFO (whether during this pod's original bootstrap or any subsequent restart), it stays a plain file across every later restart of that same pod, because the FIFO-creation check never revisits a path that already exists.

This matches a real incident: a demoted former primary running `pg_rewind --restore-target-wal` failed 22 times in a row with `could not restore file ... from archive` while walking backward through WAL, and every one of those `wal-restore` invocations produced zero log lines in the captured JSON stream for the entire ~45-minute window, which is more consistent with a log destination that was already a dead regular file than with a race being freshly lost partway through. I haven't been able to independently confirm exactly when in that pod's history the path was first corrupted, only that the mechanism described above would produce exactly this symptom and that nothing in the code would ever self-heal it once it happened.

I'd expect the log-destination path to be verified as a real FIFO, not just as "some file exists here," before anything is allowed to write to it, and for that verification to happen on every restart rather than being skippable once a stale file is present from an earlier one.

Related: cloudnative-pg/machinery#298 (CreateFifo doesn't validate an existing path is actually a FIFO) and cloudnative-pg/machinery#299 (log-destination open flags overwrite instead of appending), both surfaced by this same investigation.

### Cluster resource

_(code-analysis finding, not tied to a specific cluster spec)_

### Relevant log output

_(representative symptom: zero `"logger":"wal-restore"` JSON lines in the captured pod logs despite repeated invocations)_

### Code of Conduct

armru:
`wal-archive` and `wal-restore` run as separate `/controller/manager` processes and are told to write their logs to `--log-destination <LogPath>/postgres.json`, the same path the instance manager's JSON log pipe uses for its FIFO. If one of those commands opens the destination before the pipe has created the FIFO, `customDestination` creates a plain regular file there (it opens with `O_CREATE`). The pipe then finds a non-FIFO at the path; until now it silently accepted it, so the reader opened a static file, reached EOF, and no `wal-archive`/`wal-restore` output was ever captured. That is the silence reported in #11201.

Now that `CreateFifo` reports `ErrExistsNotFifo` for a non-FIFO, the pipe reacts to it: it removes the stale entry and lets the next iteration create the FIFO, so streaming recovers without operator action. The path lives on an `emptyDir` that outlives container restarts, so without this the only recovery is recreating the pod. The removal is deliberately narrow: it fires only on `ErrExistsNotFifo`, logs a warning first, and drops just the name at the path (a symlink is removed, never its target). An in-flight writer holding the old inode loses its buffered lines, but the next invocation opens the fresh FIFO; that is convergent and acceptable for a transient log-staging path. The retry loops also gain a short, cancellable backoff so a persistent failure no longer busy-spins.

The machinery dependency is pinned (operator and e2e modules) to machinery's `main` tip, which carries the `ErrExistsNotFifo` fix (cloudnative-pg/machinery#301, merged) ahead of a tagged release; re-pin to a tagged release once machinery cuts one.

Closes #11201

### DIFF
--- a/internal/cmd/manager/instance/run/cmd.go
+++ b/internal/cmd/manager/instance/run/cmd.go
@@ -333,6 +333,14 @@ func runSubCommand( //nolint: gocyclo,gocognit
 	postgresStartConditions = append(postgresStartConditions, jsonPipe.GetExecutedCondition())
 	exitedConditions = append(exitedConditions, jsonPipe.GetExitedCondition())
 
+	// Give Rewind the same log-pipe readiness conditions, so pg_rewind
+	// never races these readers for the log-destination FIFOs.
+	instance.SetLogPipesReadyCondition(concurrency.MultipleExecuted{
+		postgresLogPipe.GetInitializedCondition(),
+		rawPipe.GetExecutedCondition(),
+		jsonPipe.GetExecutedCondition(),
+	})
+
 	if err := instancestorage.ReconcileWalDirectory(ctx); err != nil {
 		contextLogger.Error(err, "unable to move `pg_wal` directory to the attached volume")
 		return err
--- a/internal/cmd/manager/instance/run/cmd.go
+++ b/internal/cmd/manager/instance/run/cmd.go
@@ -271,7 +271,7 @@ func runSubCommand( //nolint: gocyclo,gocognit
 		contextLogger.Error(err, "unable to create instance controller")
 		return err
 	}
-	postgresStartConditions = append(postgresStartConditions, reconciler.GetExecutedCondition())
+	postgresStartConditions = append(postgresStartConditions, reconciler.GetInitializedCondition())
 
 	// database reconciler
 	dbReconciler := controller.NewDatabaseReconciler(
@@ -321,7 +321,7 @@ func runSubCommand( //nolint: gocyclo,gocognit
 		contextLogger.Error(err, "unable to add raw logs handler")
 		return err
 	}
-	postgresStartConditions = append(postgresStartConditions, rawPipe.GetExecutedCondition())
+	postgresStartConditions = append(postgresStartConditions, rawPipe.GetInitializedCondition())
 	exitedConditions = append(exitedConditions, rawPipe.GetExitedCondition())
 
 	// json logs handler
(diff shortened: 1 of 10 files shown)


bujjibabukatta (approved):
Changes looks good.

GabriFedi97 (approved):

litaocdl (approved):

mnencia (approved):

Issue #10419, opened 2026-04-06
[Bug]: Replica permanently stuck in wal-restore loop after failover: walreceiver never reconnects, operator reports cluster healthy, burns CPU indefinitely


jmealo:
## What happened?

After a failover, a demoted primary became permanently stuck in archive-restore-only mode. The walreceiver was killed ~5 minutes after starting and never reconnected. The replica has been stuck for **40+ hours**, burning **~9.7 CPU cores** continuously while the operator reports the cluster as fully healthy (`readyInstances: 3`, `phase: "Cluster in healthy state"`).

### Context

We performed a node migration on 2026-04-04: scaled the cluster from 3→6 instances (to provision fresh replicas on new nodes), promoted a new-node instance to primary, cordoned the old nodes, destroyed the old instances one at a time, then patched back to 3 instances. This completed successfully by ~02:45 UTC.

Approximately **28 hours later** (2026-04-05 ~06:01 UTC), a separate failover occurred. The logs show a **contested switchover** -- the operator set pod-18 as the target primary, but pod-17 also briefly attempted to promote itself:

1. **06:01:02Z** -- Pod-17 logs `"Setting myself as primary"` and `"I'm the target primary, wait for the wal_receiver to be terminated"`
2. **06:01:02Z** -- Walreceiver FATAL: `"replication terminated by primary server"` -- `"End of WAL reached on timeline 48 at 412/D80000A0"`, `"could not send end-of-streaming message to primary: SSL connection has been closed unexpectedly"`
3. **06:01:03Z** -- `"I'm the target primary, applying WALs and promoting my instance"` -- pod-17 tried to promote
4. **06:04-06:05Z** -- Smart shutdown of pod-17 **failed twice** (`"pg_ctl: server does not shut down"`)
5. **06:11:28Z** -- Pod-17 finally realizes pod-18 won the promotion: `"This is an old primary instance, waiting for the switchover to finish"`
6. **06:11:29Z** -- Runs `pg_rewind` against pod-18, restarts as replica

The cause of the switchover at 06:01 is unclear from available logs. It may have been triggered by a health check failure exacerbated by our aggressive `wal_receiver_timeout: 5s` / `wal_sender_timeout: 5s` settings.

### Sequence of events (2026-04-05)

**06:11:28Z** -- Instance manager on pod-17 detects switchover completed, runs `pg_rewind` against pod-18, demotes to replica, starts postmaster.

[...]

PR #11251, opened 2026-07-24
feat: detect and contain a replica diverged onto a discarded timeline

armru:
When a primary is lost abruptly and a new primary is promoted from
whichever replica had progressed furthest, any other replica that had
received more WAL than the promoted one is left on a timeline the
cluster no longer follows. That replica cannot resume streaming from
the new primary, but until now the operator kept reporting it healthy,
kept it in the read services, and kept its replication slot open on
the primary, letting it pin WAL indefinitely.

This adds detection for that condition. The operator tracks how long a
replica's WAL receiver has stalled behind the current primary's
timeline, and once it has been stalled long enough, checks the
primary's timeline history to confirm whether the replica's received
WAL actually goes past the fork point, that is, whether it is looking
at data the new timeline never produced. If the fork-point check
confirms the divergence, the replica is marked unhealthy, fenced so it
stops retrying a broken WAL stream and drops out of the read services,
and its slot on the primary is removed so it stops holding WAL back.
If the check cannot confirm a divergence, for example because the
primary was unreachable, the instance is instead reported as stuck:
surfaced, but never fenced, since it may still catch up once the
underlying cause resolves.

Recovery from a confirmed divergence is a manual rebuild for now:
`kubectl cnpg destroy` on the affected instance drops it so the
operator re-creates it as a fresh replica of the current primary.
Automatic rewind and reducing how often this situation occurs in the
first place are tracked separately in #10419 and #4990.

The new `alpha.cnpg.io/divergedReplicaHandling: detectOnly` annotation
disables the automatic fencing and slot removal, leaving the
divergence only detected and surfaced, for cases where a divergence
should be reviewed before the operator acts on it. This introduces new
`Cluster` status fields, `replicaDivergenceWatermarks` and
`replicaWalIssues`, to report the watermark tracking and the confirmed
or stuck instances.

Depends on #11274


jcouturest:
Any progress on this? This keeps biting us, and it's very painful

PR #11261, opened 2026-07-28
fix: don't treat an unreachable standby as WAL-receiver-down

armru:
Stacked on #11260 and targets that branch, so the first commit in the diff
belongs there and is not part of this review. Refs #11251, which fixes the
opposite failure of the same gate.

**The problem.** When the operator cannot reach a standby, it concludes that
the standby's WAL receiver is down. It cannot tell "this standby told me its
receiver is down" apart from "this standby told me nothing", because a failed
`/pg/status` probe leaves `IsWalReceiverActive` at its Go zero value, `false`,
which is the same value a genuinely stopped receiver reports.

**Why it matters.** That conclusion is what the operator uses to decide a
failover is safe to finalize. So a standby that is streaming happily from the
old primary, but that the operator momentarily could not reach, is counted as
proof that nothing is still advancing. The operator promotes a less advanced
candidate, the standby it could not see ends up ahead of the new primary, and
we get exactly the divergence this gate exists to prevent. It takes nothing
exotic: probe timeouts are not retried, so one slow or dropped probe is enough.

**What this PR does.** It makes the gate treat unreachable as unknown rather
than as down, deciding by kubelet readiness when our own probe fails:

| Standby's status | Before | After |
|---|---|---|
| Reports receiver up | blocks | blocks |
| Reports receiver down | proceeds | proceeds |
| Probe failed, pod still Ready per kubelet | **proceeds** | **blocks, warning event** |
| Probe failed, pod not Ready per kubelet | proceeds | proceeds |

Only the third row changes. The fourth deliberately still proceeds: if kubelet
says the pod is gone, a dead standby is not streaming, and blocking there would
stall every ordinary node-failure failover.

This is the same signal we already trust for a primary that is Ready but not
reporting. That guard has the right instinct and was simply never generalized
to standbys, which is the hole being closed.

**Does blocking on unknown introduce a new hang?** No unbounded one. A wedged
standby fails its liveness probe, kubelet restarts the container, the pod stops
being Ready, and the fourth row reclassifies it, bounded by the existing
`.spec.livenessProbeTimeout`. What remains is the narrow case where kubelet can
reach a pod that the operator cannot, and there blocking is the correct answer:
the operator provably does not know that standby's LSN, so it has no sound
basis on which to finalize an election. The warning event makes the wait
visible in `kubectl describe cluster` instead of looking like a hang.

**The rest of the diff.** The same conflation was live at three other call
sites, each fixed on top of the `InstanceSet` partition from the base branch:
the switchover target could be chosen from an instance whose status was never
observed, a backup could be assigned to one, and `kubectl cnpg status` printed
confident blanks for an unreachable primary rather than saying the status is
unknown.

### DIFF
--- a/docs/src/failover.md
+++ b/docs/src/failover.md
@@ -37,6 +37,17 @@ controller will initiate the failover process, in two steps:
     and the replicas.
 :::
 
+To decide whether a standby's WAL receiver has actually stopped, the operator
+needs to hear back from that standby. If the operator's status probe to a
+standby fails while Kubernetes still reports its pod as `Ready`, the operator
+cannot tell whether the WAL receiver is really down or the pod is simply
+unreachable, so it conservatively assumes the WAL receiver might still be
+running and defers the election, raising a `WalReceiverStatusUnknown` event
+naming the affected pod(s). Once Kubernetes stops reporting the pod as
+`Ready` (bounded by `.spec.livenessProbeTimeout`), the standby is treated as
+down, since a genuinely dead pod cannot still be streaming and blocking on it
+would stall failover after an actual node failure.
+
 During the time the failing primary is being shut down:
 
 1. It will first try a PostgreSQL's *fast shutdown* with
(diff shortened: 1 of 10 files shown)


armru:
Squashed into #11260. Its first commit duplicated that PR's fix, and keeping the rest on a separate branch on top of it made a dependency chain that is not worth reviewing across three PRs: the whole "not reporting is not the same as down" change now lives in #11260 as a single commit.

PR #11268, opened 2026-07-29
fix: demote a primary that stopped reporting its status

armru:
Based on #11260, whose commits are included here: the fallback below reads
`.status.instancesReportedState`, which only becomes trustworthy once a failed
probe stops overwriting an instance's entry with zeros.

When a cluster is switched to replica cluster mode, `replicaclusterswitch`
looks for a primary among the instances that reported their status and, when it
finds one, fences the cluster to demote it. An instance whose `/pg/status`
probe failed reports no role at all, `IsPrimary` included, so a single failing
probe made `containsPrimaryInstance` return false and the reconciler take its
"no primary instance present means that we have no work to do" branch, skipping
the transition while that primary kept accepting writes.

Nothing else in that pass fences it either. `AllReadyInstancesStatusUnreachable`
only trips when every ready instance is silent, and the `IsComplete` gate that
does stop the reconcile sits further down, past the point where the demotion
would have started. The pass is requeued about ten seconds later by
`evaluatePodReadinessGuards`, so the demotion does eventually happen, but only
once the probe answers again: for as long as the operator could not reach an
instance that clients could, the cluster was a replica cluster on paper with a
writable primary behind it.

The decision now falls back on the last state observed for the instance
recorded as the current primary. Fencing is a metadata write that goes through
the API server, so stopping that primary never required reaching it in the
first place, and the transition completes on the instance manager's own watch of
the Cluster object rather than on an operator probe.

The fallback is deliberately narrow, so that it fires only where a primary
really is at stake:

- it asks for an instance that was actually observed as a primary, which a
  designated primary never is since it runs with a `standby.signal` file, so a
  replica cluster created from scratch is left alone when its designated primary
  goes quiet;
- it requires the instance to still be among the probed ones, because one that
  has left the status list has no PostgreSQL left to stop;
- it steps aside once a demotion token exists, so a probe failing right after a
  completed transition cannot start a second one and fence a healthy replica
  cluster.

The package had no coverage of `Reconcile` at all, so the tests come with it:
the silent-primary case, the reporting-primary case that must keep working, and
one case per carve-out above.

There is no issue tracking this. It came out of reviewing #11260, which fixes
the fabricated data this fallback relies on not being fabricated.

### DIFF
--- a/internal/controller/cluster_status.go
+++ b/internal/controller/cluster_status.go
@@ -757,17 +757,22 @@ func (r *ClusterReconciler) updateClusterStatusThatRequiresInstancesState(
 	// be left untouched rather than overwritten with zeros, because "we do
 	// not know" must not erase the last thing we did know.
 	//
-	// Membership in statuses.Items this pass is what tells apart "silent but
-	// still part of the cluster" from "no longer part of the cluster"
-	// (scaled down, or the pod was replaced under the same name). AddPod
-	// (pkg/postgres/status.go) runs unconditionally regardless of the probe
-	// outcome, and GetStatusFromInstances keeps one item per pod that
-	// survives filtering rather than dropping the ones whose probe failed,
-	// so every instance that still belongs to the cluster appears in
-	// statuses.Items this pass, silent or not. Only names absent from
-	// statuses.Items entirely are pruned below, which keeps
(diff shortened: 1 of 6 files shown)


armru:
merged on 11260

PR #11274, opened 2026-07-29
fix(controller): let a cluster recover while an instance is fenced

armru:
Fencing an instance is supposed to take that one instance out of service. Instead, while any instance is fenced the cluster loses its ability to recover from an unrelated failure: it will not elect a new primary, and it will not recreate a pod that has been deleted. Un-fencing by hand is the only way out. Three separate places in the reconciler contribute, each of them reached because a fenced instance looks healthy to the operator while its pod is permanently not Ready.

The shared cause is that a fenced instance keeps answering `/pg/status`. PostgreSQL is shut down, but the instance manager is still running and reports `MightBeUnavailable`, so the status carries no error while the pod never becomes Ready. Two further details matter. A fenced replica is not necessarily last in the sorted status list: entries whose status call failed are pushed to the tail before the primary-first ordering is applied, so when the primary's pod is gone, or present but unreachable, the fenced replica ends up first. And those are precisely the situations in which the cluster needs to elect a new primary.

The first problem is the readiness guard. It defers reconciliation when the first instance in the list serves a healthy status endpoint while the kubelet still reports its pod as not Ready, which normally lasts a few seconds after un-fencing and avoids electing a primary Kubernetes would refuse to route traffic to. A fenced instance sits in that state permanently, and the guard cannot tell the two apart, so it requeues every ten seconds forever. Because it runs ahead of both the election and the reconciliation of the cluster's resources, nothing downstream of it happens at all. The guard now ignores fenced instances, whose not-Ready state is expected rather than a probe yet to catch up.

The second problem surfaces once the guard stops blocking. Nothing excluded a fenced instance from target primary candidacy, so with a fenced replica and no healthy primary the operator elected the fenced instance and then retried a promotion that can never complete, spinning on the same failover attempt. Both election paths now skip fenced instances, and when no promotable candidate is left the operator elects nobody and leaves the target primary alone rather than pointing it at an instance that will never come up. A fenced current primary was already handled earlier, by `handleSwitchover` declining the switchover before this code is reached.

The third problem is the reattachment gate in `ensureInstancesAreCreated`, which defers recreating a missing pod until `InstancesReportingStatus()` matches `cluster.Status.ReadyInstances`. The first counter treats a fenced instance as reporting; the second is pure kubelet readiness and knows nothing about fencing. With anything fenced the two can never agree, so the gate deferred forever and a pod deleted elsewhere in the cluster was never recreated. The gate now excludes fenced instances from the comparison, through a predicate-taking sibling of that counter; `InstancesReportingStatus` keeps its behaviour for its other call sites, which compare against the spec or status instance count and do converge with a fenced instance counted.

All three use the existing `Cluster.IsInstanceFenced` helper, so the fence-all wildcard is covered as well.

### DIFF
--- a/internal/controller/replicas.go
+++ b/internal/controller/replicas.go
@@ -96,18 +96,18 @@ func (r *ClusterReconciler) reconcileTargetPrimaryForNonReplicaCluster(
 ) (string, error) {
 	contextLogger := log.FromContext(ctx)
 
-	mostAdvancedInstance, hasPromotableCandidate := firstPromotableCandidate(cluster, status)
+	promotionCandidate, hasPromotableCandidate := firstPromotableCandidate(cluster, status)
 	if !hasPromotableCandidate {
 		contextLogger.Info("No promotable candidate found, every instance is fenced, "+
 			"skipping the election", "targetPrimary", cluster.Status.TargetPrimary)
 		return "", nil
 	}
-	if cluster.Status.TargetPrimary == mostAdvancedInstance.Pod.Name {
+	if cluster.Status.TargetPrimary == promotionCandidate.Pod.Name {
 		return "", nil
 	}
 
 	// If the first pod of the list has no reported status we can't evaluate the failover logic.
-	if !mostAdvancedInstance.HasHTTPStatus() {
+	if !promotionCandidate.HasHTTPStatus() {
 		return "", nil
 	}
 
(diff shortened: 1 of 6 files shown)


internal/controller/replicas.go
GabriFedi97: Should we also check for the pod readiness here in case we are not dealing with the first item of the list? The `evaluatePodReadinessGuards` acts on the first pod only and we have changed the guard to skip if the first pod in the list is fenced, regardless its readiness: https://github.com/cloudnative-pg/cloudnative-pg/blob/8c27e39e808830d4dcd47c711a267ff1a4069ae8/internal/controller/cluster_controller.go#L656

```suggestion
if !mostAdvancedInstance.HasHTTPStatus() || !mostAdvancedInstance.IsPodReady {
		return "", nil
	}
```

GabriFedi97 (approved):

litaocdl (approved):

leonardoce (approved):

Issue #11114, opened 2026-07-02
[Bug]: Potential race condition in primary update switchover


chunngaiau:
### Is there an existing issue already for this bug?

- [x] I have searched for an existing issue, and could not find anything. I believe this is a new bug.

### I have read the troubleshooting guide

- [x] I have read the troubleshooting guide and I think this is a new bug.

### I am running a supported version of CloudNativePG

- [x] I have read the troubleshooting guide and I think this is a new bug.

### Contact Details

_No response_

### Version

1.29 (latest patch)

### What version of Kubernetes are you using?

1.35

### What is your Kubernetes environment?

Cloud: Google GKE

### How did you install the operator?

YAML manifest

### What happened?

A synchronous replica was not promoted as the new primary during an unsupervised primary update switchover. Instead a less up-to-date replica was promoted. There were WAL records in the synchronous replica that were never replicated to the new primary. This is because the new primary is chosen before the current primary is shutdown.

### Timeline

1. We have a 3-instance cluster. `-2` was the primary and `-1` and `-3` were the replicas. Our relevant settings are:
```yaml
spec:
  primaryUpdateStrategy: unsupervised
  primaryUpdateMethod: switchover
  postgresql:
    synchronous:
      dataDurability: required
      failoverQuorum: true
      method: any
      number: 1
```
2. We added an entry to the `spec.postgresql.shared_preload_libraries` config which requires a restart.
3. The operator correctly restarted `-1` and `-3` for the config change.
4. The operator then selected `-1` as the new replica before restarting `-2`. In this reconciliation iteration, all instances reported the same LSN.
5. Fast shutdown was requested on `-2`. Since 4, it had written new WAL records. These new WAL records were replicated to `-3` but not `-1`.
6. After `-2` was shutdown, `-1` was promoted to the new primary.
7. `-3` fails to resume streaming from `-1` because it is ahead of the fork point of `-1`'s new timeline and crash-loops. `-2` ran `pg_rewind` and successfully resumed streaming from the new replica. The logs we saw on `-3` after the switchover are:
```
could not start WAL streaming: ERROR:  requested starting point 0/D000000 on timeline 2 is not in this server's history
DETAIL:  This server's history forked from timeline 2 at 0/C000130.
new timeline 3 forked off current database system timeline 2 before current recovery point 0/D0000A0
waiting for WAL to become available at 0/D0000B8
```

Our workaround for this is to now use `primaryUpdateMethod: restart`.

[...]

Issue #10591, opened 2026-05-01
[Bug]: Failover blocked indefinitely when walsender survives on zombie node (storage failure with kernel alive)


daanvinken:
### Is there an existing issue already for this bug?

- [x] I have searched for an existing issue, and could not find anything. I believe this is a new bug.

### I have read the troubleshooting guide

- [x] I have read the troubleshooting guide and I think this is a new bug.

### I am running a supported version of CloudNativePG

- [x] I have read the troubleshooting guide and I think this is a new bug.

### Contact Details

daanvinken@tythus.com

### Version

1.28 (latest patch)

### What version of Kubernetes are you using?

1.33

### What is your Kubernetes environment?

Self-managed: kind (evaluation)

### How did you install the operator?

YAML manifest

## Description

We hit a production incident where failover was blocked for ~45 minutes because the walsender on the old primary kept running after a storage failure. The node's NVMe died but the kernel and network stayed alive (on-prem hypervisor). The walsender doesn't need disk I/O in its steady-state loop, it reads WAL from shared memory and sends keepalives over TCP. These keepalives kept the replicas' WAL receivers connected, preventing the `AreWalReceiversDown` gate from clearing.

## What's happening

After a storage failure where the kernel stays alive:

1. PostgreSQL backends freeze on disk I/O
2. The walsender keeps running and sending keepalives every `wal_sender_timeout / 2` (~30s)
3. WAL receivers receive keepalives, resetting `wal_receiver_timeout` each time, so the timeout never fires
4. `wal_receiver_timeout` is unlikely to help. The WAL receiver solicits keepalives via the `requestReply` flag at half-timeout ([walreceiver.c:581](https://github.com/postgres/postgres/blob/REL_13_STABLE/src/backend/replication/walreceiver.c#L581)), the walsender responds immediately, resetting the timer before the full timeout fires
5. Operator detects unhealthy primary, sets `TargetPrimary = PendingFailoverMarker`
6. Old primary's instance manager receives the signal but can't shut down PostgreSQL because the checkpoint blocks on disk I/O, so the walsender never gets SIGTERM
7. `AreWalReceiversDown` returns false on every reconcile loop, failover blocked indefinitely

## Reproduction

Reproduced in a local kind cluster with CNPG 1.28.2.

3-instance cluster across 3 kind worker nodes. To simulate the zombie node:

1. SIGSTOP the kubelet on the primary's node (prevents pod restart)
[...]

PR #11294, opened 2026-08-04
fix(controller): ensure old primary is shutdown before picking new primary in unschedulable node switchovers

### LINKED ISSUE #11293: [Bug]: Race condition in unschedulable node auto-switchover

### Contact Details

_No response_

### Version

1.30.0

1.35

Cloud: Amazon EKS

YAML manifest

### What happened?

The failure mode here is the same as #11114 where a race condition in the switchover logic causes timeline divergence and lost writes on the new primary. I am opening a new issue because although the mechanism of the bug is the same, the trigger is different and is in another part of the code.

Here, the bug is caused by the automatic switchover logic in `setPrimaryToSchedulableNode` method in `replicas.go`. Similar to #11114, the new primary is chosen before the old primary is shutdown. Which means transactions can still be committed on the old primary and replicated to other replicas. Those transactions may not be replicated to the selected new primary and are lost when it is promoted. Manual intervention is then required to reconcile the diverged timelines.

This happened to two of our CNPG clusters during a node pool upgrade and analysis on one of them revealed 45 commits was lost. We run our clusters with

```yaml
spec:
  instances: 3
  postgresql:
    synchronous:
      dataDurability: required
      failoverQuorum: true
      method: any
      number: 1
```

Given this configuration, we expect the CNPG operator to ensure ~no data loss or split-brain happens~ all client-committed transactions are replicated during an automatic failover/switchover, or to delay any failover/switchover if it can't be guaranteed.

### Timeline

1. We have a 3-instance cluster. `-1` is the primary, `-2` and `-3` are the replicas.
2. Karpenter taints the `-1` node with `karpenter.sh/disrupted:NoSchedule`.
3. The operator recognises the taint and initiates a switchover to `-2`.
```
Current primary is running on unschedulable node, triggering a switchover
```
4. `-1` shuts down and `-2` is promoted to primary. `-2`'s logs:
```
Acquired primary lease
received promote request
redo done at 3DC/5222C3D8 system usage: CPU: user: 8.78 s, system: 1.64 s, elapsed: 3364.24 s
selected new timeline ID: 4
```
5. `-3` fails to resume streaming from `-2` because it was ahead of `-2`'s forkpoint for the new timeline. `-1` ran `pg_rewind` and successfully resumed streaming from `-2`, but the rewind silently dropped commits. The follow error logs was repeated on `-3`:
```
could not start WAL streaming: ERROR:  requested starting point 3DC/53000000 on timeline 3 is not in this server's history
DETAIL:  This server's history forked from timeline 3 at 3DC/5222C420.
```

### Analysis

Running `pg_waldump` on `-3`'s disk shows that it was ahead of `-2` by 45 commits when `-2` was promoted. These commits might be "lost"<sup>1</sup> on the new primary:
```console
# pg_waldump -p . --start=3DC/5222C420 -t 3 --stats=record -r Transaction 00000003000003DC00000052
WAL statistics between 3DC/5222C420 and 3DC/522A36F0:
Type                                           N      (%)          Record size      (%)             FPI size      (%)        Combined size      (%)
----                                           -      ---          -----------      ---             --------      ---        -------------      ---
Transaction/COMMIT                            45 (100.00)                 2925 (100.00)                    0 (  0.00)                 2925 (100.00)
                                        --------                      --------                      --------                      --------
Total                                         45                          2925 [100.00%]                   0 [0.00%]                  2925 [100%]
```

<sup>1</sup> It's only a problem if the old primary had acknowledged the successful commit of these transactions back to the client. If the old primary had already terminated the client connections (e.g. fast shutdown), then these commits were never successful from the perspective of the client. However, in smart shutdown mode, the old primary is still able to return success to the clients. Regardless of shutdown mode, there is still a gap between when the new target primary is selected and when the old primary's shutdown is initiated (the former triggers the latter). Unfortunately, I'm unable to confirm if any of these commits were acknowledged back to the client although it is likely. The primary is configured with `ANY 1` in its `synchronous_standby_names`, so if the commits were visible in one of the replicas, it is very likely that the primary had acknowledged the commit back to the client. Refer to the [comment](https://github.com/cloudnative-pg/cloudnative-pg/issues/11293#issuecomment-5179778941) for an explanation of Postgres replication details.

### Cluster resource

```shell

```

### Relevant log output

```shell

```

### Code of Conduct

chunngaiau:
Fixes #11293.

There is a race condition in the current switch over logic when the operator detects the primary is running on an unschedulable node. The operator picks the new primary before waiting for the old primary to fully shutdown. The old primary can still commit transactions and replicate them to any replicas. These transactions may not be replicated to the chosen new primary, especially when quorum-based synchronous replication is configured (i.e. `method: any`), and are "lost" when it is promoted. The more advanced replica enters a crash-loop because it is ahead of the new primary's fork point of the new timeline.

This PR fixes it by reusing the code path for failovers. The failover code path sets the target primary to a pending marker, which triggers the shutdown of the old primary, but waits until the primary is fully shutdown before picking the new primary. Reusing the failover code path also ensures that we perform the failover quorum check. This is only required for non-replica clusters. The designated primary of a replica cluster does not need to be shut down before a switchover.

This fix was inspired by #11185, but I've chosen not to introduce a separate pending switchover marker. I don't think we need to distinguish a failover from a switchover as their mechanics should be the same. The only difference is in how they are triggered, a failover is unintentional and a switchover is intentional. In fact, I think one of the causes of this bug and #11114 is that there are separate code paths for each. Happy to change it if this is wrong.

As mentioned in this [comment](https://github.com/cloudnative-pg/cloudnative-pg/issues/11114#issuecomment-4870867123), the failover primary selection can be further improved using the primary lease. The operator can try to acquire the primary lease before selecting the new primary, which should give better guarantees that the old primary is fully shutdown before picking the new primary. I've decided not to implement that improvement in this PR.

Appreciate any feedback as this is my first contribution to the project.

### DIFF
--- a/internal/controller/cluster_controller_test.go
+++ b/internal/controller/cluster_controller_test.go
@@ -558,6 +558,83 @@ var _ = Describe("Updating target primary", func() {
 			Expect(instanceToCreate).ToNot(BeNil())
 			Expect(instanceToCreate.Name).To(Equal(instance2Name))
 		})
+
+	It("does not select a new primary until the old one on the unschedulable node has shut down", func(ctx SpecContext) {
+		namespace := newFakeNamespace(env.client)
+		cluster := newFakeCNPGCluster(env.client, namespace, func(c *apiv1.Cluster) {
+			c.Status.Instances = 3
+			c.Status.ReadyInstances = 3
+			primaryName := specs.GetInstanceName(c.Name, 1)
+			c.Status.CurrentPrimary = primaryName
+			c.Status.TargetPrimary = primaryName
+		})
+
+		instances := generateFakeClusterPods(env.client, cluster, true)
+		managedResources := &managedResources{
+			instances: corev1.PodList{Items: instances},
+		}
+
+		By("placing the primary on an unschedulable node and the replicas on schedulable ones", func() {
+			primaryNode := corev1.Node{
+				ObjectMeta: metav1.ObjectMeta{Name: "primary-node"},
+				Spec:       corev1.NodeSpec{Unschedulable: true},
+			}
+			replicaNode1 := corev1.Node{ObjectMeta: metav1.ObjectMeta{Name: "replica-node-1"}}
+			replicaNode2 := corev1.Node{ObjectMeta: metav1.ObjectMeta{Name: "replica-node-2"}}
+			Expect(env.client.Create(ctx, &primaryNode)).To(Succeed())
+			Expect(env.client.Create(ctx, &replicaNode1)).To(Succeed())
+			Expect(env.client.Create(ctx, &replicaNode2)).To(Succeed())
+		})
+
+		primaryStatus := postgres.PostgresqlStatus{
+			IsPrimary: true,
+			Node:      "primary-node",
+			Pod:       &instances[0],
+		}
+		replica1Status := postgres.PostgresqlStatus{
+			Node:                "replica-node-1",
+			Pod:                 &instances[1],
+			IsWalReceiverActive: true,
+		}
+		replica2Status := postgres.PostgresqlStatus{
+			Node:                "replica-node-2",
+			Pod:                 &instances[2],
+			IsWalReceiverActive: true,
+		}
+
+		By("reconciling while the old primary is still up and its replicas are connected to it", func() {
+			statusList := postgres.PostgresqlStatusList{
+				Items: []postgres.PostgresqlStatus{primaryStatus, replica1Status, replica2Status},
+			}
+
+			selectedPrimary, err := env.clusterReconciler.reconcileTargetPrimaryForNonReplicaCluster(
+				ctx, cluster, statusList, managedResources,
+			)
+
+			Expect(err).To(MatchError(ErrWalReceiversRunning))
+			Expect(selectedPrimary).To(BeEmpty())
+			Expect(cluster.Status.TargetPrimary).To(Equal(apiv1.PendingFailoverMarker))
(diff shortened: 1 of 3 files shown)


ardentperf:
> The operator picks the new primary before waiting for the old primary to fully shutdown.

there is not technically any requirement to wait for cleanup of the old/dead primary to complete. Also note there is not any race condition leading to data loss or split brain here. https://github.com/cloudnative-pg/cloudnative-pg/issues/11293#issuecomment-5179778941

This PR is interesting, but it needs a different framework for evaluation. It may extend RTO/unavailability - need to reason through this carefully.

gbartolini:
Thanks for the analysis and the draft, @chunngaiau. The diagnosis in #11293 is right, and unifying the switchover and failover code paths is the direction we want to go.

We're mid-way through a broader rework of failover/switchover coordination built on the primary lease. See the status update on a related report for the failover side: #10591 (comment). That covers `reconcileTargetPrimaryForNonReplicaCluster`; it doesn't yet include the switchover paths this PR touches, but rather than land a third, separate implementation of the pending-marker pattern, we'd like to extend the same lease-gated mechanism to `updatePrimaryPod` and `setPrimaryOnSchedulableNode` once #11367 is in.

Given that, I'd suggest keeping this as a draft for now rather than pushing it toward review. We'll probably fold the unschedulable-node case into the lease-based rework and loop you in when that follow-up starts. Also worth weighing @ardentperf's RTO comment above, it applies to any version of this fix, not just this one.

PR #11308, opened 2026-08-10
fix: remove `ssl_ecdh_curve` from fixed parameters

### LINKED ISSUE #11310: [bug]: Allow ssl_ecdh_curve parameter

I'd like to use X25519 with postgres 16, but the webhook doesn't allow me to change the `ssl_ecdh_curve` parameter. I couldn't see a reason in the code to restrict setting this parameter (its not set as a result of other CR settings like many of the other ssl parameters), so raising this PR.

If I was using postgres 18 I'd be able to use the `ssl_groups` paramter instead, which isn't blocked by the webhook.

Remove the webhook block for `ssl_ecdh_curve`

Disabling the webhook for the CR

### Additional context

Would also be useful for migrating to postgres 18 later because postgres 18 will also respect the `ssl_ecdh_curve` setting, so I can upgrade then change the config to `ssl_groups` afterwards.

### Backport?

Yes

Yes (pr: https://github.com/cloudnative-pg/cloudnative-pg/pull/11308)

### Code of Conduct

Jamstah:
Enable users to configure `ssl_ecdh_curve` (since it should never have been blocked).

Closes: #11310

### DIFF
--- a/docs/src/postgresql_conf.md
+++ b/docs/src/postgresql_conf.md
@@ -767,7 +767,6 @@ Users are not allowed to set the following configuration parameters in the
 - `ssl_cert_file`
 - `ssl_crl_file`
 - `ssl_dh_params_file`
-- `ssl_ecdh_curve`
 - `ssl_key_file`
 - `ssl_passphrase_command`
 - `ssl_passphrase_command_supports_reload`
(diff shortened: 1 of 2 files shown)


gbartolini:
Thanks @Jamstah. Could you please open an issue before the PR so we can discuss it? For example, the field in PostgreSQL 18 has been renamed `ssl_groups`, which means we need to consider this from a broader perspective and potentially enable related GUCs as well.

Jamstah:
Sure, I figured the change was so small the discussion might benefit from seeing test runs too :)

Opened here: https://github.com/cloudnative-pg/cloudnative-pg/issues/11310

gbartolini (approved):

armru (approved):

PR #11319, opened 2026-08-13
feat(bootstrap): use an init container instead of a separate Job for new instances

### LINKED ISSUE #11228: [Feature]: Bootstrap PostgreSQL instances without a separate Job or init container

leonardoce:
New primary and replica instances are now bootstrapped (initdb, recovery, pg_basebackup, join, and snapshot restore) directly on the instance Pod via an init container, instead of running a separate Kubernetes Job beforehand. This removes the extra Job resource previously created for every new instance, along with its own Pod, and the wait for that Job to complete before the actual instance Pod could start.

A stuck bootstrap surfaces as PhaseUnrecoverable on the Cluster status, and a replica resuming a stalled bootstrap reports PhaseCreatingReplica with an Event.

Closes: #11228

----

## Original PR description

New primary and replica instances are now bootstrapped (initdb, recovery,
pg_basebackup, join, and snapshot restore) directly on the instance Pod via
an init container, instead of running a separate Kubernetes Job beforehand.
This removes the extra Job resource previously created for every new
instance, along with its own Pod, and the wait for that Job to complete
before the actual instance Pod could start.

# Important

This is an alternative approach for #11231 that is less invasive on the codebase.
It requires #11315 to be merged too.


armru (approved):

mnencia (approved):

mnencia:
Full E2E run on EDB fork https://github.com/EnterpriseDB/cloudnative-pg/actions/runs/33049566677

PR #11336, opened 2026-08-18
fix(controller): avoid stalling or reverting an in-progress failover

### LINKED ISSUE #11346: [Bug]: failover can be silently reverted

### Contact Details

### Version

1.30.0

1.36

YAML manifest

### What happened?

When a failover was in progress and the cluster's target primary was set to the internal "pending" marker, two separate reconciliation bugs could prevent it from ever completing:

1. The reconcile loop treated the pending marker like any other in-progress switchover and short-circuited every pass with a flat 1-second requeue, never actually checking whether the failover was ready to complete. This could cause the failover to stall indefinitely.
2. Independently, the status-update logic didn't recognize the pending marker as a special value — it looked for a running instance matching that marker, found none, and concluded the target primary was invalid. It then reset TargetPrimary back to the (unhealthy) current primary, silently undoing the failover that was already underway.

### Cluster resource

```shell

```

### Relevant log output

```shell

```

### Code of Conduct

leonardoce:
While a failover is pending, the reconciler used to short-circuit every tick: it just requeued after one second instead of checking whether the failover could complete, and a separate check mistook the pending marker for a missing pod and reset the target primary back to the unhealthy one. Either way the failover could stay stuck forever or get silently undone.

The reconciler now keeps working through a pending failover and completes it once the WAL receivers are down. Right before completing it, it also rechecks quorum safety, since the set of instances safe to promote can change while waiting for the WAL receivers to stop. If it is no longer safe, the operator keeps waiting instead of promoting a replica that might be missing acknowledged transactions.

Closes #11346


chunngaiau:
I think this change might affect when the failover quorum is checked. Currently, the failover quorum is only checked when `currentPrimary == targetPrimary` [here](https://github.com/cloudnative-pg/cloudnative-pg/blob/a95a534dbe0854495b863bbc1b7651f74d310916/internal/controller/replicas.go#L113-L122), it doesn't recheck the failover quorum when the `targetPrimary` is `pending`.

This doesn't matter currently due to the bug referenced in this PR. The failover quorum check is always performed before the `targetPrimary` is set to `pending` and always after the `updateResourceStatus` method resets it back to the `currentPrimary`. So it never sees the `pending` marker and is always performed until a new replica is chosen as the target primary, but this PR would change that.

I think maybe we need to add a `cluster.Status.TargetPrimary == apiv1.PendingFailoverMarker` condition [here](https://github.com/cloudnative-pg/cloudnative-pg/blob/a95a534dbe0854495b863bbc1b7651f74d310916/internal/controller/replicas.go#L114) to the failover quorum check too? Or simplify it to only check if the failover quorum is enabled so it is checked anytime the target primary needs updating? I'm still new to the code so I might have missed something.

We noticed this when we were reviewing the logs of a failover. These lines were repeated until all the WAL receivers are down. Crucially, we see that the quorum check is performed every time, which I think is what we want.

```
Wrong target primary, the chosen one is not active or not present
Quorum check algorithm results
Current primary isn't healthy, initiating a failover
pod status (1 of 2)
pod status (2 of 2)
Waiting for all WAL receivers to be down to elect a new primary
```

leonardoce:
> I think this change might affect when the failover quorum is checked. Currently, the failover quorum is only checked when `currentPrimary == targetPrimary` [here](https://github.com/cloudnative-pg/cloudnative-pg/blob/a95a534dbe0854495b863bbc1b7651f74d310916/internal/controller/replicas.go#L113-L122), it doesn't recheck the failover quorum when the `targetPrimary` is `pending`.

Thank you @chunngaiau for taking the time of having a look at this PR.
When `cluster.Status.TargetPrimary == apiv1.PendingFailoverMarker` the operator already committed to do a failover and the old primary may already have received the change and be demoting.

In my opinion the failover quorum rule need to be evaluated before we committed to any failover.
Did I get the scenario you were thinking of?

chunngaiau:
Thanks for the reply @leonardoce. Yes, I think we have the same scenario in mind. Basically, I think it is possible for the failover quorum check to pass when we commit to a failover but fail when we eventually pick the new primary. This might happen if more replicas become unhealthy as the old primary is shutting down.

Say we have a 3-instance cluster with `ANY 1` synchronous replication mode. First, the primary `-1` becomes unhealthy, triggering a failover. The failover quorum check passes here because replica `-2` and `-3` are healthy, so we commit to a failover by setting target primary to `pending`. The operator then waits for all WAL receivers to shut down. In the meantime, `-2` becomes unhealthy too and goes offline. Once, `-3`'s WAL receiver has shutdown, the operator proceeds to pick the new primary, but it can only retrieve the status of `-3`. Here, the failover quorum check would fail as we can't guarantee that `-3` had received all the WAL records from `-1`. Have I got this right? I think this situation is rare, but it can occur, especially if the cause of failures are correlated.

So it depends what we think is the correct action to take here? As you say, if we are already committed to a failover, then the operator should proceed with the failover and a failover quorum check is not required. Or, the operator could stall until enough replica instances become healthy again to satisfy the failover quorum.

In my company, we would prefer the latter (at the cost of availability). Any automatic failover or switchover by the operator should only proceed if it has the correct guarantees. If it can't proceed, then we can always set an alarm to fire and manually intervene, for example by using `kubectl cnpg promote`. What do you think?

leonardoce:
@chunngaiau Yes, I got it know and I believe what you're proposing is correct.

The failover quorum criteria is evaluated after the promotable replicas set, and that may change in time occurring between setting `targetPrimary` to the `pendingFailover` marker and the time in which we're able to elect a replica as the new primary.
 
@NiccoloFei added a new commit in this PR implementing this.

NiccoloFei (approved):

mnencia (approved):

PR #11337, opened 2026-08-18
feat: generate the password of a declarative role

### LINKED ISSUE #3788: [Feature]: add automatic generation of password secrets for managed roles

The current implementation of managed roles allows giving the name
of a secret where a role's password will be kept, but that secret is
assumed to exist, and its creation falls o the user.

For users migrating from other solutions, or users wanting to get on
CloudNativePG and have password-protected roles from the get-go,
it would be convenient if the secrets were generated by CloudNativePG
as needed.

See thread https://cloudnativepg.slack.com/archives/C03AX0J5P29/p1706780445388369

Allow the managed roles stanza to specify either an existing secret
to hold a role's password, or an auto-generated secret.

at the moment, automatic creation of password secrets would
fall outside of the CNPG envelope

### Additional context

_No response_

### Backport?

Yes

Yes

### Code of Conduct

armru:
A `DatabaseRole` could only take its password from a Secret created beforehand,
leaving that Secret, and the password inside it, to the user. This adds a
`password` stanza that has the operator generate the password instead and store
it in a `kubernetes.io/basic-auth` Secret shaped the way the instance manager
expects, named `<databaserole-name>-password` unless `password.secret` says
otherwise:

```yaml
kind: DatabaseRole
metadata:
  name: cluster-example-dante
spec:
  cluster:
    name: cluster-example
  name: dante
  ensure: present
  login: true
  password:
    enabled: true
    duration: 2160h
    renewBefore: 168h
    criteria:
      length: 32
      symbolCharacters: "!#$%&*+-?"
```

The `criteria` block is modelled on the external-secrets password generator:
`length`, `digits`, `symbols`, `symbolCharacters`, `noUpper` and
`allowRepeat`. Validation rules reject a combination no password could satisfy,
and a combination that only the generator can rule out is reported in
`status.password` rather than retried forever.

The password is generated once and then left alone, unless `duration` gives it a
lifetime: the operator then generates a new one `renewBefore` ahead of its
expiration, which is recorded on the Secret and mirrored in
`status.password.expiration`. Rotation is deliberately off by default, and not
derived from the operator's certificate settings as the issue suggested, because
PostgreSQL keeps a single password per role: a rotation invalidates every
credential a consumer is still holding the moment `ALTER ROLE` runs, which
should not happen ninety days after somebody asked only for a generated
password.

A Secret the operator does not own is never overwritten nor deleted, and no
password is generated on a replica cluster, where the role and its password
belong to the primary cluster; generation starts after promotion. The stanza is
mutually exclusive with `passwordSecret`, which stays as it is for the
bring-your-own-password case, and with `disablePassword`.

Along the way the Secret holding the password of a role is now resolved through
the `DatabaseRole` rather than through `passwordSecret` alone, so the condition
the instance manager watches, the password it applies, and the per-cluster Role
that lets it read that Secret all follow a generated Secret too. Two failures
this uncovered on a live cluster are fixed here: without the generated name in
that RBAC rule the instance manager cannot read the password, and since it
cannot create a role without one, the role was never created at all; and a role
built from a `RoleConfiguration` with no `passwordSecret` defaults to ignoring
its password, which is exactly how a generated password looks, so the role
reached PostgreSQL with a NULL password while the reconciliation reported
success.

Closes #3788


gbartolini:
@armru, I completely forgot that we have a field called `disablePassword` in the API. As a result, I've been exploring an alternative approach to a boolean "enabled" field, aiming to eliminate any ambiguity.

For example, I propose using a `mode` that accepts one of the following four values:

- `generate`: Generate the password and assume ownership of the secret (whether provided or not).
- `secret`: Use the provided secret.
- `external`: Leave the current password in the database unchanged, as it could be managed either manually or through a third-party solution.
- `clear`: Explicitly set the password to `NULL` in the database.

With this approach, `disablePassword` would be mutually exclusive with the `password` stanza, such as `passwordSecret`, and would become a deprecated interface.

I'll work on providing a prototype for your pull request.

armru:
disablePassword lives on RoleConfiguration, shared with Cluster.spec.managed.roles, while password is only
on DatabaseRoleSpec. So deprecating it leaves inline managed roles with no way to
say "NULL password" (password.mode: clear isn't reachable from there). Move
password down to RoleConfiguration, or keep disablePassword for the inline path
and call it superseded only for DatabaseRole? Also: does password.mode keep the
generate default or become required?

gbartolini:
> disablePassword lives on RoleConfiguration, shared with Cluster.spec.managed.roles, while password is only on DatabaseRoleSpec. So deprecating it leaves inline managed roles with no way to say "NULL password" (password.mode: clear isn't reachable from there). Move password down to RoleConfiguration, or keep disablePassword for the inline path and call it superseded only for DatabaseRole? Also: does password.mode keep the generate default or become required?

On `disablePassword` and the inline path, I would go with option 2: keep it for inline-managed roles, and call it superseded only for DatabaseRole. I don't want to move `password` down to `RoleConfiguration`. So inline managed roles keep `disablePassword` and their current behaviour, unchanged. And to be explicit: we're not removing `disablePassword` or `passwordSecret` from `DatabaseRole` any time soon either. They keep working, we just recommend the password stanza over them now that `DatabaseRole` is the recommended way to manage roles.

On the default, you've talked me out of having one. `mode` should be required whenever the `password` stanza is present, with no default, so `password: {}` is rejected at admission. I'd considered defaulting it to clear instead of generate, but rejected that primarily because it would make an empty block destructive on a role that already has a password.

armru:
Done in `037e41c3c`, option 2 as you said. `disablePassword` and `passwordSecret` stay on `RoleConfiguration`, inline roles unchanged, both still working on `DatabaseRole` and mutually exclusive with the stanza. `mode` is required, no default, so `password: {}` is rejected.

Two changes to your sketch. The fourth value is `setNull`, which says what it does to the role. And `external` sets a password the operator generated itself to NULL once before letting go: its Secret is deleted, so a credential nobody can read and nothing rotates would be left behind otherwise.

Also, the stanza cannot be removed once set: dropping it would quietly hand the role back to the legacy fields.

Covered by the CEL rules on a real API server, unit tests, and the e2e checking `rolpassword IS NULL` after `generate` to `external`.

gbartolini:
In the meantime, I have submitted this patch to the PostgreSQL project for consideration in version 20: https://www.postgresql.org/message-id/CA%2BVUV5pyUB-sXUCxnOATJdE_XDdos6rAgW5wPVTz-0AT8bLuGw%40mail.gmail.com

gbartolini:
I tested this PR thoroughly, and I would like to start hearing more feedback. I created this cluster:

```yaml
kind: Cluster
metadata:
  name: acdc
spec:
  instances: 1

  storage:
    size: 1Gi
```

I then started testing the new `mode: generate` option:

```yaml
kind: DatabaseRole
metadata:
  name: angus
spec:
  cluster:
    name: acdc
  name: angus
  login: true
  password:
    mode: generate
    duration: 1m
    renewBefore: 20s
  databaseRoleReclaimPolicy: delete
```

I requested that a password be regenerated every 40 seconds (1m - 20s of renewBefore), expiring after a minute. A cool thing about the `generate` mode is that it automatically sets the `VALID UNTIL` field in the Postgres system. Check the `.status` part below, with `appliedExpiration` stating when the database change was applied:

```apiVersion: postgresql.cnpg.io/v1
status:
  applied: true
  conditions:
  - lastTransitionTime: "2026-08-26T02:41:39Z"
    message: "55189"
    reason: ChangeDetected
    status: "True"
    type: PasswordSecretChange
  observedGeneration: 2
  password:
    appliedExpiration: "2026-08-26T02:42:39Z"
    expiration: "2026-08-26T02:42:39Z"
    issuedAt: "2026-08-26T02:41:39Z"
    secretName: angus-password
  secretResourceVersion: "55189"
```

By default, the secret is the name of the `DatabaseRole` + '-password', which guarantees uniqueness in the same namespace.

This is the secret (which you can appreciate also embeds a simple `pgpass` entry):

```yaml
apiVersion: v1
data:
  password: RGg3Zno2bzRqbnJ1NTBlWktjSXRwMkxt
  pgpass: Kjo1NDMyOio6YW5ndXM6RGg3Zno2bzRqbnJ1NTBlWktjSXRwMkxtCg==
  username: YW5ndXM=
kind: Secret
metadata:
  creationTimestamp: "2026-08-26T02:41:39Z"
  labels:
    app.kubernetes.io/managed-by: cloudnative-pg
  name: angus-password
  namespace: default
  ownerReferences:
  - apiVersion: postgresql.cnpg.io/v1
    blockOwnerDeletion: true
    controller: true
    kind: DatabaseRole
    name: angus
    uid: 169b2f0f-5266-4ca2-9ff5-90765c394f36
  resourceVersion: "55798"
  uid: 04f44ccb-98fb-4597-accb-a86e6b77a5a5
type: kubernetes.io/basic-auth
```

Here is the `pgpass` content:

```
$ echo -n "Kjo1NDMyOio6YW5ndXM6RGg3Zno2bzRqbnJ1NTBlWktjSXRwMkxtCg==" | base64 -d
*:5432:*:angus:Dh7fz6o4jnru50eZKcItp2Lm
```


This is the situation on the database side:

```console
postgres=# SELECT * FROM pg_authid WHERE rolname = 'angus';
-[ RECORD 1 ]--+--------------------------------------------------------------------------------------------------------------------------------------
oid            | 16393
rolname        | angus
rolsuper       | f
rolinherit     | t
rolcreaterole  | f
rolcreatedb    | f
rolcanlogin    | t
rolreplication | f
rolbypassrls   | f
rolconnlimit   | -1
rolpassword    | SCRAM-SHA-256$4096:svMZdo1fMxB+WV/ic/DBJw==$C4yxULI9kFvzdt5ZU0550/cbpsVZxmWVYK2X3FFDtvU=:qPuihKNawSC2dIpjmbc6vk/7BKEqljKXJ4Q88siHEFk=
rolvaliduntil  | 2026-08-26 02:45:59+00
```

The `get` command output has been improved too:

```
$ kubectl get databaseroles.postgresql.cnpg.io
NAME    AGE    CLUSTER   PG NAME   APPLIED   MESSAGE   PASSWORD EXPIRATION    CERTIFICATE EXPIRATION
angus   5m2s   acdc      angus     true                2026-08-26T02:47:19Z
```

and so the `describe` command:

```console
Name:         angus
Namespace:    default
Labels:       <none>
Annotations:  <none>
API Version:  postgresql.cnpg.io/v1
Kind:         DatabaseRole
Metadata:
  Creation Timestamp:  2026-08-26T02:41:39Z
  Finalizers:
    cnpg.io/deleteRole
  Generation:        2
  Resource Version:  56552
  UID:               169b2f0f-5266-4ca2-9ff5-90765c394f36
Spec:
  Cluster:
    Name:                        acdc
  Connection Limit:              -1
  Database Role Reclaim Policy:  delete
  Ensure:                        present
  Inherit:                       true
  Login:                         true
  Name:                          angus
  Password:
    Duration:      1m0s
    Mode:          generate
    Renew Before:  20s
Status:
  Applied:  true
  Conditions:
    Last Transition Time:  2026-08-26T02:41:39Z
    Message:               56545
    Reason:                ChangeDetected
    Status:                True
    Type:                  PasswordSecretChange
  Observed Generation:     2
  Password:
    Applied Expiration:     2026-08-26T02:48:39Z
    Expiration:             2026-08-26T02:48:39Z
    Issued At:              2026-08-26T02:47:39Z
    Secret Name:            angus-password
  Secret Resource Version:  56545
Events:
  Type    Reason             Age                  From                         Message
  ----    ------             ----                 ----                         -------
  Normal  PasswordGenerated  6m6s                 cloudnative-pg-databaserole  Generated the password of role "angus" into Secret "angus-password"
  Normal  PasswordRotated    6s (x16 over 5m26s)  cloudnative-pg-databaserole  Generated a new password for role "angus" into Secret "angus-password"
```

I also tested the manual rotation:

```sh
kubectl annotate databaserole angus cnpg.io/rotatePassword=whatever
```

I tried changing the values for duration, renewBefore, and even the criteria (to change the length and composition of the password). I tried removing the duration as well and changing the secret name (including using an existing secret name, which is overwritten only if the cluster and database role names are the same).

I also tested all transitions, with no issue.

Finally, I tested the integration with external secrets operator and updated the documentation accordingly.

gbartolini:
Regarding the `.pgpass` content, in the future, we could provide a template for the secret, allowing users to customise it and add connection strings for specific databases.

gbartolini (approved):

xgerman:
This is directly applicable to DocumentDB and we would like to use it.

The dilemma is the password write path. Though DocumentDB  uses Postgres roles, it must pre-process the plaintext password before storing it. Its  `PrehashPassword()`  path applies SASLprep and generates a SCRAM secret with a default 28-byte salt, rather than relying on PostgreSQL’s normal password encoding. A generic  `CREATE ROLE ... PASSWORD '<plaintext>'`  or  `ALTER ROLE ... PASSWORD '<plaintext>'`  can therefore create the role successfully but leave a password that is incompatible with DocumentDB authentication.

Would CNPG consider introducing a password/role provider hook around the database mutation, something along these lines?

```
type DatabaseRoleProvider interface {
    Create(ctx context.Context, spec RoleSpec, password string) error
    UpdatePassword(ctx context.Context, spec RoleSpec, password string) error
    UpdateMembership(ctx context.Context, spec RoleSpec) error
    Delete(ctx context.Context, spec RoleSpec) error
    Inspect(ctx context.Context, spec RoleSpec) (RoleState, error)
}
```

The default provider would preserve the current PostgreSQL behavior. A DocumentDB provider could invoke the extension API instead:
```
SELECT documentdb_api.create_user(
  '{"createUser":"name","pwd":"...","roles":[...],"$db":"admin"}'::documentdb_core.bson
);
SELECT documentdb_api.update_user(
  '{"updateUser":"name","pwd":"...","$db":"admin"}'::documentdb_core.bson
);
```
This is preferable to adding a DocumentDB-specific condition throughout the  DatabaseRole  reconciler, and it avoids duplicating DocumentDB’s password hashing implementation in CNPG. The provider would receive the plaintext password only in memory and let the DocumentDB extension perform the canonical pre-hashing. The generated Secret could remain unchanged.

This might also apply to other extensions which need that flexibiity.

A provider hook would also need to cover role membership, because DocumentDB’s  create_user()  resolves the requested Mongo-compatible roles and grants the corresponding PostgreSQL parent role. It should also make clear whether  Inspect  is based on PostgreSQL roles or an extension-specific query.

I am wondering if this could be part of this PR or if we should propose an additional PR on top of this.

PR #11350, opened 2026-08-20
fix(instance-manager): bound the primary lease take-over Get call

### LINKED ISSUE #11349: [Bug]: Primary lease take-over polling can stall past RetryPeriod

### Contact Details

### Version

1.30.0

1.36

YAML manifest

### What happened?

tryTakeOver reads the primary lease via r.lock.Get(ctx), using the caller's context as-is with no per-call deadline. preAcquire calls tryTakeOver in a loop and expects each attempt to return within roughly RetryPeriod (2s by default), whether it succeeds, finds the lease unavailable, or errors out.

Because the Get call carried no timeout of its own, if the Kubernetes API server was unreachable or unresponsive, the call could block for however long the underlying client-go transport takes to give up on its own — not necessarily bounded by RetryPeriod. During that time the take-over loop made no progress: it couldn't observe the lease, couldn't detect a stale holder, and couldn't retry on schedule.

The practical effect is that a replica trying to take over a stale primary's lease (e.g. during a failover) could see its take-over attempts stall for an unpredictable, potentially long time whenever the API server was slow or unreachable, instead of retrying at the intended cadence.

### Cluster resource

```shell

```

### Relevant log output

```shell

```

### Code of Conduct

leonardoce:
`tryTakeOver` reads the lease with the caller's un-deadlined context, so a `Get` against an unreachable API server could block indefinitely instead of retrying at `RetryPeriod cadence. A stalled `Get` here also keeps a primary's own lease renewal from resuming after a transient blip, risking a legitimate takeover by another node once the lease expires.

Wrap the call in a context bounded by `RenewDeadline`: tight enough to fail fast against an unreachable server, wide enough that a single still-answering-but-slow poll doesn't get discarded before the observation window can complete.

Closes #11349

### DIFF
--- a/internal/cmd/manager/instance/run/lease/runnable.go
+++ b/internal/cmd/manager/instance/run/lease/runnable.go
@@ -90,7 +90,7 @@ func defaultConfig() Config {
 // It starts idle and enters the acquisition/renewal loop only after Acquire is called.
 type Runnable struct {
 	instance *postgres.Instance
-	lock     *resourcelock.LeaseLock
+	lock     resourcelock.Interface
 
 	// config holds the lease timings. It is initialised to the defaults by New
 	// and overwritten by the first Acquire call before the runnable activates.
@@ -201,7 +201,7 @@ func (r *Runnable) Release(ctx context.Context) error {
 	if err != nil {
 		return err
 	}
-	if record.HolderIdentity != r.lock.LockConfig.Identity {
+	if record.HolderIdentity != r.lock.Identity() {
 		contextLogger.Debug("Primary lease held by another identity, nothing to release",
 			"holder", record.HolderIdentity)
 		return nil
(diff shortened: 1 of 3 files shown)


leonardoce:
This looks like CI-node/scheduling slowness, not a regression.

- 16:40:14 — test deletes the primary (cluster-fast-failover-1)
- 16:40:46 — a new primary (cluster-fast-failover-2) is confirmed via currentPrimary/currentPrimaryTimestamp — the actual fast-failover/lease-takeover mechanism worked correctly, in ~32s
- 16:40:52 — the old primary's pod (-1) is finally rescheduled as a replica (38s after deletion — pod termination + reschedule overhead on the kind node)
- 16:41:04 — its postgres container finally starts
- 16:41:13 — first readiness probe still returns 500 (only 9s after the container started — it just hadn't caught up/switched timeline yet)
- 16:41:14 — the test's 60s Eventually for "pod should be on timeline 2" times out

So of the 60s budget, ~50s was consumed just getting pod -1 rescheduled and its containers started.

Claude helped with wording and organizing the timeline.
The underlying investigation and assessment were done manually.

NiccoloFei (approved):

litaocdl (approved):

mnencia (approved):

armru (approved):

gbartolini (approved):

fcanovai (approved):

PR #11356, opened 2026-08-21
fix(instance-manager): gate PostgreSQL start on holding the primary lease

### LINKED ISSUE #11354: bug(1.30.0): align PostgreSQL startup ordering with lease acquisition on primary restart
When the instance manager reconciles a pod whose `PGDATA` is already flagged as primary and which Kubernetes already designates as `targetPrimary`, it starts PostgreSQL as soon as system initialisation completes.

Lease acquisition happens afterward, inside the `Cluster` reconcile loop, only once `IsReady()` has already confirmed PostgreSQL is up. `targetPrimary == podName` reflects Kubernetes' intent, not proof that a previous holder of the role has actually released it (e.g. a prior primary process still running after a node partition or slow termination), so starting PostgreSQL ahead of the lease means this code path offers no fencing against that scenario.

The promoted-replica path is unaffected: the lease is already acquired there before `pg_promote` is invoked, so this gap is specific to the restart-on-already-primary-PGDATA path. Reaching the race in practice requires a node partition or delayed termination combined with asynchronous replication. With synchronous replication, the recommended production configuration for avoiding data loss, the client-visible commit is gated on standby acknowledgment, which the affected instance can't obtain while partitioned, so this is a gap specific to async-only setups rather than a general regression of the lease feature.

*Note:* affects 1.30.0 only, introduced by #10627, not present in 1.29.x or earlier since the primary lease mechanism doesn't exist there.

leonardoce:
An already-primary PGDATA started PostgreSQL right after system initialization, before ever acquiring the primary lease. Lease acquisition happened later, inside the Cluster reconcile loop, only after IsReady() confirmed PostgreSQL was already up. This left a window where a restarting or rescheduled primary could serve writes without holding the lease.

Acquire the lease from verifyPgDataCoherenceForPrimary instead, which runs once during initialize(), strictly before PostgreSQL is allowed to start. reconcilePrimary already acquires the lease unconditionally at the top of every reconcile for the target primary (this incidentally gates pg_promote too). This change brings the startup path in line with that.

Closes #11354


NiccoloFei:
I've verified this manually by:
1. Creating a 1 instance cluster `cluster-example`
2. Scaling the operator to 0 (to avoid reverting RBAC changes about to be done)
3. Edit cluster-example's `Role` to remove all rules on `leases` objects
4. Run `kubectl exec cluster-example-1 -c postgres -- kill 1` to kill the container and let it restart on its own, this way the new instance-manager that comes up will run `r.initialize()` again

* On main: Postgres will be able to start and serve writes even though the lease is not getting renewed (stale `renewTime`)
* With the patch: Postgres start is correctly gated by the primary lease. The pod won't be able to start and it's continuously logging: 
```
"msg":"startup probe failing"
"msg": "Primary lease not yet acquired, retrying",
```

NiccoloFei (approved):

litaocdl (approved):

mnencia (approved):

PR #11357, opened 2026-08-21
feat(instance-manager): move primary isolation check to the primary lease

leonardoce:
Primary self-fencing on network isolation used to live in the liveness
probe: on API-server unreachability it pinged every peer's failsafe
endpoint, and a failed probe caused Kubelet to restart the pod roughly
three probe periods (~30s) after a partition started.

The primary lease now gives a faster, more precise signal for the same
condition. When the primary fails to renew its lease, it runs the same
peer-reachability check itself and, if it should step down, requests a
fast PostgreSQL shutdown directly rather than waiting on Kubelet.

The check runs as soon as renewal fails (not after waiting out the
remaining lease TTL): the RenewDeadline-to-LeaseDuration margin is the
time budget for the check-and-shutdown sequence to complete before a
replica becomes eligible to promote, not idle slack to wait through.

The step-down condition is no longer just peer unreachability: the
/failsafe entrypoint now reports the responding instance's view of the
cluster's target primary (via a new internal/management/failsafe wire
package shared by both sides), so a primary that can still reach every
peer but learns from one of them that someone else is now the target
primary steps down too. This closes a gap the reachability check alone
couldn't see: a peer being reachable doesn't mean it hasn't already
been promoted.

The liveness probe's isolation logic is removed now that the lease
handles it; IsHealthy always reports OK.


leonardoce:
**IMPORTANT**

This requires #11347 and #11350 to be merged.
The relative commits are manually cherry-picked in this PR but should be removed before merging it.

mnencia (approved):

NiccoloFei (approved):

armru (approved):

PR #11358, opened 2026-08-21
feat(instance-manager): turn the liveness probe into a watchdog

leonardoce:
The liveness probe used to always report healthy, so a Pod could never
be restarted when it truly got stuck. It now fails, so the kubelet can
kill and recreate the Pod, in two cases the primary lease and fencing
logic can't recover from on their own:

- the primary-lease loop stops attempting to renew or take over the
  lease altogether (e.g. a deadlock), as opposed to merely failing to
  renew, which is an expected condition handled separately by the
  lease/failsafe step-down check;

- PostgreSQL was asked to shut down immediately (e.g. because this
  primary had to step down) but didn't honor it within a new
  configurable `.spec.immediateShutdownTimeout` (default 30s).


leonardoce:
**IMPORTANT**

This PR depends on #11347 #11350 and #11357 that are cherry-picked here out of convenience but should be removed before merging it.

PR #11366, opened 2026-08-24
fix(pgbouncer): bind `SHOW STATS` columns by name

### LINKED ISSUE #11365: [Bug]: pooler stats metrics disappear when PgBouncer adds a SHOW STATS column
The pooler stops exporting every `cnpg_pgbouncer_stats_*` metric and logs this on each scrape:

```
Error while executing SHOW STATS ... sql: expected 25 destination arguments in Scan, not 23
```

`collectShowStats` picks the scan layout from the number of columns (`< 16`, `== 17`, default 23). PgBouncer keeps adding columns — `total_client_login_count` and `avg_client_login_count` are the newest pair — so a build past the known counts falls into the default branch and the scan fails. It already happened once with 1.24 (#6566) and will happen again on the next column.

Pool metrics keep working, so the pooler looks healthy while its throughput and wait-time series are gone.

To reproduce: run a Pooler on a PgBouncer that reports 25 SHOW STATS columns and read `/metrics` on port 9127 — only `cnpg_pgbouncer_pools_*` are there.

Fix in #11366: bind the scan destinations to the column names from `rows.Columns()` and skip names the collector does not know.

Separately, `ShowStatsMetrics.Reset` resets `TotalClientParseCount` and `TotalServerParseCount` twice and never resets `AvgClientParseCount` and `AvgServerParseCount`, so a database that disappears keeps its last averages.

AdzerKI:
The collector picked its scan layout from the number of columns returned by `SHOW STATS`, and the gate that wrote the server assignment gauges (`statColsCount == 16`) matched none of the scan branches. On the 17-column path, those two values were scanned and discarded, while the parse-count gauges (never scanned there) were exported as zeros.

Bind each scan destination to its column name instead, ignoring columns the collector does not know, so a new PgBouncer column can no longer shift the values of the others. `total_server_assignment_count` and `avg_server_assignment_count` start being exported as a result, so add them to the sample output in the documentation, and correct the help text of the total, which described a duration rather than a count.

Also reset the two avg parse-count gauges, which were shadowed by their total counterparts and kept stale values across scrapes.

Closes #11365

### DIFF
--- a/docs/src/connection_pooling.md
+++ b/docs/src/connection_pooling.md
@@ -782,6 +782,9 @@ cnpg_pgbouncer_stats_avg_recv{database="pgbouncer"} 0
 # HELP cnpg_pgbouncer_stats_avg_sent Average sent (to clients) bytes per second.
 # TYPE cnpg_pgbouncer_stats_avg_sent gauge
 cnpg_pgbouncer_stats_avg_sent{database="pgbouncer"} 0
+# HELP cnpg_pgbouncer_stats_avg_server_assignment_count Average number of times a server was assigned to a client per second in the last stat period.
+# TYPE cnpg_pgbouncer_stats_avg_server_assignment_count gauge
+cnpg_pgbouncer_stats_avg_server_assignment_count{database="pgbouncer"} 0
 # HELP cnpg_pgbouncer_stats_avg_server_parse_count Average number of prepared statements created by pgbouncer on a server.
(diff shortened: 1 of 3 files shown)


AdzerKI:
Rebased on main, tests in `pkg/management/pgbouncer/metricsserver` pass. Anything else needed from my side before this can get a review?

gbartolini (approved):

armru (approved):

PR #11367, opened 2026-08-24
fix(instance-manager): promote a replica without waiting for WAL receivers to stop

leonardoce:
The primary lease now guarantees only one instance can act as primary
at a time, so failovers and switchovers no longer need to stall until
every WAL receiver in the cluster reports itself down before a new
primary is chosen and promoted. This removes a source of delay and
stuck-waiting states during failover.


leonardoce:
**IMPORTANT**: this requires #11347 #11350 #11357 and #11358 to be merged before a real review. They are cherry-picked here out of convenience.

leonardoce:
This is what happened:

- 14:44:50.109 — Test corrupts the primary (-1) by deleting its PGDATA while postgres keeps running
- 14:44:5x – 14:45:17 — -1's postmaster starts returning FATAL: "base/5" is not a valid data directory for new connections/backends, but its pre-existing walsender connections to -2 and -3 (established before the corruption) keep running normally — they only read WAL segment files and don't need to touch the corrupted catalog
- 14:45:17.847 — Operator (decides the new primary: it snapshots replica status and finds -2 and -3 tied at receivedLsn = replayLsn = 0/6055A20; it picks -2 and sets cluster.Status.TargetPrimary = -2
- 14:45:18.276 – 18.406 — -1's own instance-manager notices it's no longer TargetPrimary, requests a fast shutdown; postgres on -1 (and both its walsenders) dies
- 14:45:18.397 / .401 — -2's and -3's walreceivers both disconnect essentially simultaneously (SSL EOF) — i.e. both were still actively streaming, at slightly different effective positions, right up to the moment -1 actually died
- 14:45:18.563 — -1 releases the primary lease
- 14:45:20.495 – 21.562 — -2 acquires the lease (fast path: empty holder, no 15s TTL wait needed) and promotes — ~2.6s after the original election snapshot, and ~1.9s after the lease was actually released
- (at promotion) — -2's fork point is 0/6055A20 (its replay had only reached there) — but -3 had already received WAL up to 0/7000000 in the meantime
- 14:45:5x → 14:51:09 (timeout) — -3 endlessly retries streaming from 0/7000000, failing forever with requested starting point 0/7000000 on timeline 1 is not in this server's history — it diverged onto WAL the new primary never had, and nothing detects/repairs this (no automatic re-clone/rewind)

So the primary lease worked correctly, but unfortunately the operator decided which replica to promote too early, without re-checking it before the promotion actually happens.

Claude was used to help wording the timeline.

leonardoce:
- 15:52:08.88 → 15:52:23.17: 8 consecutive "Waiting for the primary lease to be released or expire" reconciles, 1 every ~2s (the RequeueAfter: 1 * time.Second from handleSwitchover's ErrPrimaryLeaseHeld branch, though logs land ~2s apart)
- 15:52:25.26: FailoverTarget event finally fires — ~16.4s after the failover was first marked pending

It looks like the tests time budget is not enough to observe a failover. That's even more important now given that, in these conditions, we need to wait for the lease to expire.

gbartolini:
@leonardoce could this be related to #11114?

PR #11370, opened 2026-08-25
chore: remove `status-port-tls` flag and HTTP fallback

GellertBabel:
All instances have used TLS for operator communication by default since v1.24 (#4442). Remove the now-dead `--status-port-tls` flag, the `StatusPortTLS` instance field, and the HTTPScheme detection logic (`GetStatusSchemeFromPod`), assuming TLS unconditionally in the status webserver, probes, and instance manager client.

An in-place instance manager upgrade re-execs this binary with the `argv` chosen by the operator that created the Pod, so Pods created before 1.31 still pass `--status-port-tls` and would otherwise crash-loop. Keep the flag parseable as deprecated and ignore it when comparing container commands until 1.30 is discontinued.

Important changes (for release notes): in the unlikely event of a problem after upgrading from an operator older than v1.24, perform a manual rolling update of the affected Pods to recreate them with a TLS-enabled status port.

### DIFF
--- a/docs/src/installation_upgrade.md
+++ b/docs/src/installation_upgrade.md
@@ -408,11 +408,13 @@ for details.
 :::warning
 This protection has a hard requirement: the status port **must** be served over
 TLS, which has been the default since v1.24. Instances created by an operator
-older than v1.24 serve the status port over plain HTTP; once their instance
-manager is upgraded to 1.30.0 the operator can no longer authenticate to them
-and every call to the protected endpoints is **permanently** rejected with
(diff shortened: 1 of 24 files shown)


armru (approved):

gbartolini:
Tests are green.

leonardoce (approved):

gbartolini:
I don't think we should warn anymore. 1.24.0 was released 2 years ago. We should just mention this in the release notes.

gbartolini (approved):

PR #11437, opened 2026-09-05
fix(logpipe): do not block threads on log FIFOs

### LINKED ISSUE #11436: [Bug]: Log pipes open even when no writer will attach, leading to deadlock during initdb

### Contact Details

will.murnane@gmail.com

### Version

1.30.0

1.36

Helm

### What happened?

A bug in PR #10043 causes a leaked OS thread during initdb. This does not have a major impact on Linux systems where CNPG generally runs, but it's still a bug. 

In that PR, `jsonPipe := logpipe.NewJSONLineLogPipe(...)` was added to `WithActiveInstance()` in `pkg/management/postgres/instance.go`.
   
    1. `jsonPipe.Start()` creates `/controller/log/postgres.json` as a named pipe (FIFO) and calls `fileutils.OpenFileAsync(ctx, p.fileName, os.O_RDONLY, 0o600)`.
    2. In Linux/POSIX, calling `open(fifo, O_RDONLY)` blocks in kernel `wait_for_partner` indefinitely until an external writer connects.
    3. During `initdb` bootstrap, Postgres only opens `/controller/log/postgres.csv` (`log_destination: csvlog`). No Barman or external tools are running yet, so nothing ever opens `/controller/log/postgres.json` for write.
    4. When `WithActiveInstance` finishes and calls `ctxCancel()`, `OpenFileAsync` returns on `ctx.Done()`, but the background goroutine running `os.OpenFile` **remains permanently stuck in the blocking kernel syscall on the OS thread**.
    5. When `WithActiveInstance` runs `defer jsonPipe.GetExitedCondition().Wait()`, or when the Go runtime attempts process exit/thread synchronization, the blocked thread causes a deadlock.
    6. Writing any byte to `/controller/log/postgres.json` from the host instantly unblocks the kernel `open()` and allows `initdb` to complete immediately.

As a consequence, under amd64 emulation using Rosetta on macOS, the `initdb` processes occasionally (about 5% of the time, in my testing) deadlock. Actually running the database is fine, because during that state the `postgres.json` named pipe does get open, unblocking the thread.

I have a suggested fix, which I think is unambiguous, and will submit a PR shortly.

### Cluster resource

```shell
Any Cluster object
```

### Relevant log output

```shell
{"level":"info","ts":"2026-08-22T14:14:08.290455636Z","msg":"Configuring new PostgreSQL instance","logging_pod":"grafana-db-1-initdb"}
    {"level":"info","ts":"2026-08-22T14:14:08.729123974Z","msg":"Executing post-init SQL instructions","logging_pod":"grafana-db-1-initdb"}
    {"level":"info","ts":"2026-08-22T14:14:08.76115752Z","msg":"Executing post-init template SQL instructions","logging_pod":"grafana-db-1-initdb"}
    {"level":"info","ts":"2026-08-22T14:14:09.271973743Z","msg":"executing Application instructions","logging_pod":"grafana-db-1-initdb"}
    <process hangs here indefinitely>

  #### Kernel process call stacks (/proc/<pid>/task/*/stack):

    === TID 8362 (goroutine opening postgres.json) ===
    [<0>] wait_for_partner+0x84/0x100
    [<0>] fifo_open+0x1b8/0x2a8
    [<0>] vfs_open+0x118/0x4a8
    [<0>] path_openat+0x41c/0x1db0
    [<0>] do_file_open+0x9c/0x160
    [<0>] __arm64_sys_openat+0x120/0x1f8
```

### Code of Conduct

willmurnane:
Since #10043, WithActiveInstance starts the raw and JSON log pipes
during bootstrap (initdb, join, restore). Their readers open the
postgres.csv and postgres.json FIFOs with a blocking open(2), which the
kernel parks until a writer connects.

During an initdb bootstrap postgres logs to csvlog only, so nothing
ever opens the raw and JSON FIFOs for write. OpenFileAsync returns on
context cancellation, but the goroutine it spawned inside os.OpenFile
stays blocked in the kernel forever, pinning an OS thread that the Go
runtime cannot reclaim.

Open the FIFOs non-blockingly, so a cancelled context can no longer
strand a thread in the kernel.

Fixes #11436.

### DIFF
--- a/pkg/management/postgres/logpipe/linelogpipe.go
+++ b/pkg/management/postgres/logpipe/linelogpipe.go
@@ -149,7 +149,7 @@ func (p *LineLogPipe) collectLogsFromFile(ctx context.Context) error {
 		}
 	}()
 
-	f, err := fileutils.OpenFileAsync(ctx, p.fileName, os.O_RDONLY, 0o600)
+	f, err := fileutils.OpenFileAsync(ctx, p.fileName, os.O_RDWR, 0o600)
 	if err != nil {
 		return err
 	}
--- a/pkg/management/postgres/logpipe/logpipe.go
+++ b/pkg/management/postgres/logpipe/logpipe.go
@@ -141,7 +141,7 @@ func (p *LogPipe) collectLogsFromFile(ctx context.Context) error {
 		}
 	}()
 
-	f, err := fileutils.OpenFileAsync(ctx, p.fileName, os.O_RDONLY, 0o600)
+	f, err := fileutils.OpenFileAsync(ctx, p.fileName, os.O_RDWR, 0o600)
 	if err != nil {
 		return err
 	}


GellertBabel:
Using O_RDWR instead of O_RDONLY looks weird. Cant os.O_RDONLY|os.O_NONBLOCK be used? Also this impacts both bootstrap and instance manager, while you need it only on bootstrap?

willmurnane:
`os.O_RDONLY|os.O_NONBLOCK` sort of works, but AIUI because `bufio.Scanner` and `io.Reader` in Go expect blocking stream semantics, they would busy loop rather than waiting for bytes to show up (so every instance of `streamLogFromCSVFile` would consume a whole CPU core all the time). 

Another option I considered was opening with `O_NONBLOCK` and then using fcntl to switch back to blocking mode, but unfortunately that also leads to the busy loop behavior.

The point about changing both bootstrap and instance manager is another fair objection. I think that introduces a lot more questions about how best to implement than I'd prefer to dive into: should `NewLogPipe` take an argument for "this is always expected to have a writer, use `O_RDONLY`" versus "there may not be a writer, use `O_RDWR`"?  Or maybe a process-wide flag is available somewhere that should be used? Setting this up correctly would take more understanding of the architecture than I have at the moment, so I would prefer to let someone more familiar with the codebase implement that.

I think the current changeset is the largest possible "bang for your buck". It fixes the known problem of leaking an goroutine and OS thread during initdb, by changing two constants. That said, I'm extremely sympathetic to the possibility that this leads to another kind of problem that I haven't encountered in my relatively limited testing of this patch; after all, #10043 was a bug fix that didn't appear to have any negative ramifications at all, and it took some effort to find the problem that it caused. If there's some testing I could run, or analysis I could perform, to demonstrate that using `O_RDWR` all the time doesn't lead to new problems, I'd be happy to do so.

gbartolini:
> Using O_RDWR instead of O_RDONLY looks weird. Cant os.O_RDONLY|os.O_NONBLOCK be used? Also this impacts both bootstrap and instance manager, while you need it only on bootstrap?

I agree with @GellertBabel.

PR #11461, opened 2026-09-08
ci: replace label-checker action with github-script

mnencia:
agilepathway/label-checker builds its own Docker image from source on every run. Docker Hub base-image pulls fail intermittently during that build, breaking the check with no code change on our side (agilepathway/label-checker#492).

Reimplement the same any-of/none-of label check with actions/github-script, already used elsewhere in this repo, which runs directly on the runner with no image build.

Upstream shipped a fix for this specific failure in v1.6.66, but dropping the Docker build removes the whole class of issue rather than just this instance.

### DIFF
--- a/.github/workflows/require-labels.yml
+++ b/.github/workflows/require-labels.yml
@@ -10,7 +10,7 @@ on:
       - labeled
       - unlabeled
 
-permissions: read-all
+permissions: {}
 
 jobs:
   require-labels:
@@ -25,6 +25,6 @@ jobs:
             if (!labels.includes('ok to merge :ok_hand:')) {
               core.setFailed('Missing required label: "ok to merge :ok_hand:"');
             }
-            if (labels.includes('do not merge')) {
-              core.setFailed('Forbidden label present: "do not merge"');
+            if (labels.includes('do not merge :no_good:')) {
+              core.setFailed('Forbidden label present: "do not merge :no_good:"');
             }
--- a/.github/workflows/require-labels.yml
+++ b/.github/workflows/require-labels.yml
@@ -18,8 +18,13 @@ jobs:
     runs-on: ubuntu-24.04
     steps:
       - name: Require labels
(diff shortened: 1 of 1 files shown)


fcanovai (approved):

johnboyes:
> agilepathway/label-checker builds its own Docker image from source on every run. Docker Hub base-image pulls fail intermittently during that build, breaking the check with no code change on our side ([agilepathway/label-checker#492](https://github.com/agilepathway/label-checker/issues/492)).
> 
> Reimplement the same any-of/none-of label check with actions/github-script, already used elsewhere in this repo, which runs directly on the runner with no image build.

@mnencia apologies for the bug.  This should now be fixed in the latest [`v1.6.66`](https://github.com/agilepathway/label-checker/releases/tag/v1.6.66) version that I've just released, just an FYI should you need it.

litaocdl (approved):

litaocdl:
https://github.com/cloudnative-pg/cloudnative-pg/pull/11463

mnencia:
> @mnencia apologies for the bug. This should now be fixed in the latest [`v1.6.66`](https://github.com/agilepathway/label-checker/releases/tag/v1.6.66) version that I've just released, just an FYI should you need it.

@johnboyes Appreciate the quick fix. We'll likely stick with dropping the Docker build here rather than bumping to v1.6.66, as it avoids the whole class of Docker image build failures going forward, nothing against the fix itself.

johnboyes:
> @johnboyes Appreciate the quick fix. We'll likely stick with dropping the Docker build here rather than bumping to v1.6.66, as it avoids the whole class of Docker image build failures going forward, nothing against the fix itself.

@mnencia yes, makes sense.  I've not given the label checker any love for a while now - when I do have time to look at it properly again there will be some simplifications like that for me to look into for sure, the fewer moving parts, the better.

PR #11472, opened 2026-09-11
ci(preflight): create image-pull secret in preflight's randomized namespace

litaocdl:
Since preflight v1.20.0, the DeployableByOLM check installs the operator into a namespace with a random suffix (see redhat-openshift-ecosystem/openshift-preflight#1500). The image pull secret can no longer be pre-created, so pulling the operator image from a protected registry fails.

This watches for the namespace as it appears and creates the pull secret in it once it exists. Upstream issue requesting a way to pass pull secrets via the preflight CLI: redhat-openshift-ecosystem/openshift-preflight#1535.


mnencia (approved):

NiccoloFei:
E2E run: https://github.com/EnterpriseDB/cloudnative-pg/actions/runs/34869773163

NiccoloFei (approved):

PR #11479, opened 2026-09-14
ci(e2e): fix Rook CRD race on AKS

mnencia:
The Set up Rook step applies a CR right after its CRD, in two spots: crds.yaml before cluster-on-pvc.yaml, and csi-operator.yaml before operator.yaml. Sometimes the apiserver has not registered the new CRD yet, so the apply fails with "no matches for kind". The step has no error check, so this failure passes silently. The loop right after it then waits forever for OSD deployments that will never exist, until the 27 minute step timeout kills it.

Wait for each CRD to become Established before applying what depends on it, so the race cannot happen. Add set -euo pipefail, so any other silent failure stops the step right away instead of hanging. Give the OSD-count loop a sleep and a max try count, so it cannot spin forever again.


mnencia:
E2E AKS run on the EDB fork, combined with #11481: https://github.com/EnterpriseDB/cloudnative-pg/actions/runs/34859578872

litaocdl (approved):

NiccoloFei (approved):

PR #11481, opened 2026-09-14
ci(e2e): pin AKS Rook CSI CephX key type to aes

mnencia:
cluster-on-pvc.yaml never sets security.cephx.csi.keyType, so a fresh cluster gets whatever Rook's default is once that field exists (Rook v1.20.5+, part of the CVE-2025-30156 fix). That default needs Linux kernel 7.0+ for the kernel RBD client to map a volume. AKS nodes don't have that kernel yet, so every object-store PVC mount fails with "rbd: failed to add secret ... to kernel". The E2E suite then hangs waiting for a pod that will never become ready.

Rook's own cluster.yaml example already pins this to "aes" for this exact reason; cluster-on-pvc.yaml just never got the same line. Add it to the same yq patch that already sets storage size and class.


mnencia:
E2E AKS run on the EDB fork, combined with #11479: https://github.com/EnterpriseDB/cloudnative-pg/actions/runs/34859578872

litaocdl (approved):

NiccoloFei (approved):

PR #11488, opened 2026-09-15
fix: set --no-data-checksums when dataChecksums is false on PostgreSQL 18+

### LINKED ISSUE #11497: [Bug]:  --no-data-checksums is not when dataChecksums is false on PostgreSQL 18+

### Contact Details

_No response_

### Version

1.30.0

1.37

Other

Helm

### What happened?

The defaults for data checksums were swapped on PostgreSQL 18+ so that data checksums are enabled by default.  On Postgres 18+, when dataChecksums is false initdb --no-data-checksums is not set, so data checksums are always enabled.  It is desirable to disable data checksums on filesystems that already perform data checksums, e.g. ZFS.

### Cluster resource

```shell

```

### Relevant log output

```shell

```

### Code of Conduct

evan314159:
PostgreSQL 18 enables data checksums by default in initdb, so disabling them now requires an explicit flag. This is desirable for underlying filesystems that provide their own data checksums, like ZFS.

Fixes: #11497


gbartolini:
Hi @evan314159, thanks for your submission. It makes sense. However, if possible, can you please flip the check? I mean that, by default, consider PostgreSQL 18 onwards, and implement the < 18 case only as an exception. Could you please provide an updated version? Thanks

evan314159:
Hi @gbartolini -- Thank you for reviewing.  I made the updates to consider pgsql >= 18 as the default.  I also made the same change for upgrades, which I had missed previously.  Have a good day.

evan314159:
rebased on main to remove merge block

gbartolini:
Hi @evan314159, thanks again. A few things worth addressing before merge:

1. **`pkg/specs/jobs.go:142`**:`buildInitDBFlags` discards the error from `GetPostgresqlMajorVersion()`. Make sure you handle it.
2. **`api/v1/cluster_types.go:1938`** and **`docs/src/bootstrap.md:263`**: the `dataChecksums` doc comment/docs still say "default: false" and reference the removed `-k` flag. Both should be updated to reflect the PG18+ default flip, since `kubectl explain` and the docs page are what users will trust. I should have mentioned it.
3. **Minor/non-blocking**: the "only emit an explicit flag when it differs from the PG18-dependent default" logic is duplicated independently between `tryAddDataChecksums` (pg_upgrade path) and `buildInitDBFlags` (bootstrap path). Worth a shared helper at some point so a future version-threshold change doesn't need to be applied twice. What do you think?

evan314159:
Thank you @gbartolini -- I made those updates.  I added the helper to build the initdb flags into its own file as it didn't seem to fit elsewhere.  Have a good day.

gbartolini:
I have tested it with:

```yaml
kind: Cluster
metadata:
  name: checksums-disabled
spec:
  instances: 1

  bootstrap:
    initdb:
      dataChecksums: false
  storage:
    size: 1Gi
```

gbartolini (approved):

armru (approved):

leonardoce (approved):

PR #11490, opened 2026-09-16
test(e2e): stop postmaster for deterministic disk-space detection

### LINKED ISSUE #11489: [Bug]: disk_space_test.go times out on cloud E2E, race decides whether Postgres actually crashes

### Contact Details

_No response_

### Version

trunk (main)

1.35

Cloud: Other

Other

### What happened?

This is about our own E2E suite, not a user cluster: `tests/e2e/disk_space_test.go` times out on cloud jobs from time to time (GKE, AKS, OpenShift, not tied to one provider or one Kubernetes/PostgreSQL version). It's shown up on and off since August 2026.

The test fills the WAL volume, then expects a CHECKPOINT to fail and the operator to notice PostgreSQL has no disk space left. What actually decides the result is a race inside PostgreSQL: filling the disk can make the checkpoint hit a real WAL write and crash the backend (test passes), or it can hit a small per-backend temp file first, which just makes the query error out without crashing anything. In the second case PostgreSQL keeps running, the pod never turns NotReady, and the test times out waiting for that.

Expected: the test result should not depend on which of the two writes wins the race.

### Cluster resource

_No response_

### Relevant log output

_No response_

### Code of Conduct

mnencia:
Whether the disk-space test's primary crashes depends on which write
hits ENOSPC first: a real WAL extend crashes it, a harmless
per-backend temp file doesn't. When the second one wins, the pod
stays Ready and the test times out.

Stop the postmaster with pg_ctl -m immediate once the disk is full:
it skips the shutdown checkpoint, which would need free space, and
any postmaster exit makes the instance manager re-check disk space
before restarting.

Closes #11489

Assisted-by: Claude


mnencia:
E2E on EDB fork https://github.com/EnterpriseDB/cloudnative-pg/actions/runs/35070396346

NiccoloFei (approved):

leonardoce (approved):

PR #11518, opened 2026-09-21
docs: describe what the primary isolation check actually does

mnencia:
The isolation check only fails the liveness probe. The kubelet then restarts the container through its normal termination path, a smart shutdown that lets already-open sessions keep committing until smartShutdownTimeout elapses (180s default). The docs said the check "stops", "fences", or "shuts down" the primary, none of which is what the code does.

Point out `spec.smartShutdownTimeout: 0` as the workaround for anyone who needs the write window closed sooner than the fix landing in CloudNativePG 1.31.

Related to #11280


gbartolini (approved):

leonardoce (approved):

mnencia:
Verified the `smartShutdownTimeout: 0` workaround with a live test: local kind cluster, two 2-instance clusters (defaults vs `smartShutdownTimeout: 0`), one already-open session per primary predating the partition, isolated each primary's node from the API server and its peer only (surgical `iptables`, not a full node disconnect, so the session itself wasn't the thing broken).

Same partition, same detection timing. `smartShutdownTimeout: 0`: the open session stopped committing ~6s after the partition (detection only). Default (180s): the same open session kept committing until ~185s after the partition, matching the 180s budget plus detection. Confirms the doc change against actual current behavior, not just against the code.

Issue #8370, opened 2025-08-15
[Bug]: Setting Cluster.spec.imagePullSecrets after the Cluster has been created paralyzes CNPG


niels-hb:
### Is there an existing issue already for this bug?

- [x] I have searched for an existing issue, and could not find anything. I believe this is a new bug.

### I have read the troubleshooting guide

- [x] I have read the troubleshooting guide and I think this is a new bug.

### I am running a supported version of CloudNativePG

- [x] I have read the troubleshooting guide and I think this is a new bug.

### Version

1.26 (latest patch)

### What version of Kubernetes are you using?

1.33

### What is your Kubernetes environment?

Self-managed: k3s

### How did you install the operator?

Helm

### What happened?

Steps to reproduce:
1. Create a CNPG Cluster without image pull secrets (`imagePullSecrets: []`).
2. Wait for the Cluster to be healthy.
3. Update the Cluster CR to include at least one image pull secret.
4. The CNPG Operator will be unable to update the cluster as it tries to update the Pod in-place, which Kubernetes won't allow as `Pod.spec.imagePullSecrets` is immutable.
5. Manually deleting all (Replica + Primary) CNPG Cluster Pods fixes the issue (as this results in the recreation of Pods).
6. Manually deleting the Replica Pods does not work as the CNPG Operator will try to update the Primary first regardless.
7. Fixing the issue therefore results in reduced availability/downtime as the CNPG Operator does not launch new replicas for the Cluster until all Pods have been deleted manually.

The correct strategy needs to be a recreation of the Pods, starting with the replicas, then promoting a newly created replica Pod to the primary and finally recreating the Pod of the previous primary instance.

### Cluster resource

```shell
apiVersion: postgresql.cnpg.io/v1
kind: Cluster
metadata:
  annotations:
    cnpg.io/hibernation: "off"
  name: cnpg-cluster
  namespace: default
spec:
  [...]
  imageName: mirror.example.com/ghcr.io/cloudnative-pg/postgresql:15
  imagePullPolicy: IfNotPresent
  imagePullSecrets:
  - name: regcred
  instances: 2
[...]
```

### Relevant log output

```shell
[...]

PR #11521, opened 2026-09-21
feat(rollout): recreate Pods when a pull secret is added to the Cluster

Atoms:
The operator applies spec.imagePullSecrets to the instance ServiceAccount only; Kubernetes copies them into a Pod at creation time. A Pod already running when the secret is added never receives it.

This adds a rollout checker that recreates a Pod missing a configured pull secret, through the normal replica-then-primary update. Skipped when spec.serviceAccountName is set, since the operator doesn't manage that ServiceAccount's secrets.

Related to #8370, which this doesn't fix: the reported stall isn't caused by the operator patching pods.

----
### Original PR message
 
Summary

- The operator only applies spec.imagePullSecrets to the instance ServiceAccount and relies on Kubernetes to inject it into new Pods at creation time.
- Since Pod.spec.imagePullSecrets is immutable, Pods created before a secret was added to the cluster never picked it up, and reconciliation would stall trying (and failing) to patch it in place.
- Adds a checkPodImagePullSecretsOutdated rollout checker so affected Pods are instead recreated through the normal replica-then-primary rolling update.

~~Fixes 8370~~

Test plan

- [x] Added unit tests covering: no configured secrets, pod already has the configured secrets, pod missing a configured secret
- [x] go test ./internal/controller/... passes
- [x] go vet ./internal/controller/... passes


GellertBabel:
Looking at this I have two questions:
1. Should the check skip Clusters that set `spec.serviceAccountName`? As far as I can tell the operator never adds pull secrets to a user-supplied ServiceAccount - `createOrPatchServiceAccount` returns early via `validateExistingServiceAccount` - so would a recreated Pod not come back without them and get rolled again on every pass? Or is that combination meant to be rejected by the webhook instead, since today it only excludes `serviceAccountName` with `serviceAccountTemplate`?


```go
if cluster.Spec.ServiceAccountName != "" {
    return rollout{}, nil
}

```

2. Is it intentional that the checker runs before the cnpg.io/reconcilePodSpec: disabled early return? It's in the first checkers map, so that opt-out no longer applies to a PodSpec field.

armru:
Thanks for the PR. I pushed a commit on top and I propose a reword of the title and body:

The stall in #8370 is not caused by the operator. The failing request is the
labels/annotations merge patch in `pkg/reconciler/instance/metadata.go`,
whose body never touches `spec`. The operator never writes
`Pod.Spec.ImagePullSecrets`, and the upstream ServiceAccount admission plugin
returns early on anything but CREATE
(`plugin/pkg/admission/serviceaccount/admission.go`, "we only mutate pods
during create requests"). So on stock Kubernetes that patch cannot be
rejected for `imagePullSecrets`: something else in those clusters mutates
Pods on UPDATE, most likely a policy engine rule such as Kyverno's
`add-imagepullsecrets` sample. The metadata patch also runs before the
rollout check and aborts the reconcile on error, so this PR would not
unblock that scenario either. I removed the "Fixes" line and kept #8370 as
related.

The change is still worth merging as a fix on its own: a Pod created before
the secret was added cannot restart a container from a now-private registry
until it is recreated, and the operator never recreated it.

The commit I added skips the check when `spec.serviceAccountName` is set.
The operator only validates a user-provided ServiceAccount and never adds
the Cluster pull secrets to it, so a recreated Pod would still miss the
secret and the checker would request a new rollout on every reconcile,
switching the primary over and over. It also adds the new trigger to the
rolling update docs.

Given that, I propose a change in scope for the PR:


## Title

```
fix(rollout): recreate Pods when a pull secret is added to the Cluster
```

## Body

```
The operator applies spec.imagePullSecrets to the instance ServiceAccount
only, and Kubernetes copies them into a Pod when the Pod is created. Pods
that were running before a secret was added to the Cluster never receive
it, so if the image later moves to a private registry a container restart
on those Pods fails to pull.

This adds a rollout checker that detects Pods missing one of the configured
pull secrets and recreates them through the normal replica-then-primary
rolling update. The check is skipped when spec.serviceAccountName is set,
because the operator does not manage the pull secrets of a user-provided
ServiceAccount and recreating the Pod would not change anything.

Removing a pull secret does not trigger a rollout, consistent with the
ServiceAccount update being append-only. The rolling update documentation
lists the new trigger.

Related to #8370. The stall reported there comes from a Pod metadata patch
being rejected because spec.imagePullSecrets changed on UPDATE. The operator
never writes that field and the ServiceAccount admission plugin only injects
pull secrets on CREATE, so that mutation comes from something else in those
clusters, and this change does not address it.

Unit tests cover no configured secrets, a Pod already carrying the secrets,
a Pod missing a secret, and a Cluster with a user-provided ServiceAccount.
```

mnencia:
On the reconcilePodSpec:disabled question: pull secrets should respect it. We only know the secret name is present on the pod, not that it works. That's a weaker reason to force a rollout against an explicit opt-out than PVC or image checks, which always fix something real.

Give checkPodImagePullSecretsOutdated its own applyCheckers call, placed after IsPodSpecReconciliationDisabled and before hasValidPodSpec. The legacy map won't work, it skips modern pods.

Atoms:
Thanks for digging into the actual root cause on #8370 — makes sense, and agreed the title/scope change is the right call.

@mnencia  re: reconcilePodSpec: disabled — agreed, and agreed on the placement. Pushing a commit now that gives checkPodImagePullSecretsOutdated its own applyCheckers call, right after IsPodSpecReconciliationDisabled and before hasValidPodSpec, with tests for both the disabled-annotation case and the "still fires when annotation isn't set" case.

PR #11551, opened 2026-09-22
docs: note the CloudNativePG defaults for the PgBouncer TLS sslmode options

no-hup:
The PgBouncer configuration options list in `docs/src/connection_pooling.md` says:

> Unless stated otherwise, the default values are the ones directly set by PgBouncer.

That is not true for the two TLS `sslmode` options. The operator overrides them in `pkg/management/pgbouncer/config/config.go` (the `defaultPgBouncerParameters` map, commented "parameters we want to have a default different from the default one"):

- `client_tls_sslmode` → `prefer`
- `server_tls_sslmode` → `verify-ca`

The `log_stats` and `auth_type` entries in the same list already call out their CloudNativePG-specific defaults; the two `sslmode` entries didn't, so a reader would assume TLS is off by default when the operator actually turns it on. This adds the same note to both, matching the source.

Docs-only, no code change.

Disclosure: I used an AI assistant to help find and write this change. I've reviewed and verified it against the source, and I'm happy to explain or revise anything.


no-hup:
The linked-issue check is red because this is a docs-only fix with no tracking issue. It just corrects the PgBouncer TLS `sslmode` defaults note in `connection_pooling.md` to match what the operator actually sets in `config.go`. I don't have permission to add the `no-issue` label myself — could a maintainer apply it, or let me know if you'd prefer I open a short issue to link instead? Happy either way.

